Issue #3, opened 2021-01-22
N_GPIOS was not declared in this scope, unless gpio.h is explicitly included


Gadgetoid:
I'm not sure this qualifies as an "issue" since I fixed it with: https://github.com/pimoroni/pimoroni-pico/commit/6a13dc5c4b8535df5c7e075cb20728b4be831f53#diff-ddd42d56faea5530bfb4a0200db4bbecaf574481ae9cfa7d66e8b296783ce838

However if don't explicitly:

```c++
#include "hardware/gpio.h"
```

before:

```c++
#include "hardware/pwm.h"
```

Then compiling will fail with:

```
[build] /home/phil/Development/pico/pico-sdk/src/rp2_common/hardware_pwm/include/hardware/pwm.h: In function 'uint pwm_gpio_to_channel(uint)':
[build] /home/phil/Development/pico/pico-sdk/src/rp2_common/hardware_pwm/include/hardware/pwm.h:86:33: error: 'N_GPIOS' was not declared in this scope
[build]    86 |     valid_params_if(PWM, gpio < N_GPIOS);
[build]       |                                 ^~~~~~~
[build] /home/phil/Development/pico/pico-sdk/src/rp2_common/hardware_pwm/include/hardware/pwm.h:86:5: note: in expansion of macro 'valid_params_if'
[build]    86 |     valid_params_if(PWM, gpio < N_GPIOS);
[build]       |     ^~~~~~~~~~~~~~~
[build] /home/phil/Development/pico/pico-sdk/src/rp2_common/hardware_pwm/include/hardware/pwm.h:86:33: note: the macro 'N_GPIOS' had not yet been defined
[build]    86 |     valid_params_if(PWM, gpio < N_GPIOS);
[build]       |                                 ^~~~~~~
[build] /home/phil/Development/pico/pico-sdk/src/rp2_common/hardware_pwm/include/hardware/pwm.h:86:5: note: in expansion of macro 'valid_params_if'
[build]    86 |     valid_params_if(PWM, gpio < N_GPIOS);
[build]       |     ^~~~~~~~~~~~~~~
[build] In file included from /home/phil/Development/pico/pico-sdk/src/rp2_common/hardware_adc/include/hardware/adc.h:12,
[build]                  from /home/phil/Development/pico/pimoroni-pico/libraries/pico_explorer/pico_explorer.cpp:5:
[...]

PR #7, opened 2021-01-23
N_GPIOS duplicates NUM_BANK0_GPIOS

kilograham:
Addresses #3 which could have been fixed by an include, however there is already a define for number of GPIOS in platform_defs.h


Wren6991 (approved):
We probably ought to deduplicate constants from the host libraries too at some point, but this change is a strict improvement over previous so no reason to block it

PR #9, opened 2021-01-23
fix #if in hardware/structs/dma.h

majbthrd:
I suspect the user's intent was to use "PARAM_ASSERTIONS_ENABLED_DMA" (default setting generated on lines 17-18).

### DIFF
--- a/src/rp2040/hardware_structs/include/hardware/structs/dma.h
+++ b/src/rp2040/hardware_structs/include/hardware/structs/dma.h
@@ -70,7 +70,7 @@ typedef struct {
 #define dma_debug_hw ((dma_debug_hw_t *const)(DMA_BASE + DMA_CH0_DBG_CTDREQ_OFFSET))
 
 static inline void check_dma_channel_param(uint channel) {
-#if PARAM_ASSERTIONS_ENABLED(DMA)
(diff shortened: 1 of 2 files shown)


kilograham (commented):
it is correct as is (as that is a macro that checks PARAM_ASSERTIONS_ENABLED_DMA amongst other stuff)

lurch:
> as that is a macro ...

Defined here: https://github.com/raspberrypi/pico-sdk/blob/master/src/common/pico_base/include/pico/assert.h#L32

majbthrd:
In that case, let me suggest a different approach.  I’ve updated the PR to move these inline functions from:

./src/rp2040/hardware_structs/include/hardware/structs/dma.h

which is a register definition include file where they seem out of place to:

./src/rp2_common/hardware_dma/include/hardware/dma.h

where they are right are home with the DMA driver code that uses them.

kilograham (commented):
Yes this is really where the belong, thanks

Not 100% sure why they were there to begin with - perhaps historical. Worth checking it doesn't break compilation of pico-examples, pico-extras, pico-playground (we're currently working on getting our CI re-setup after our move of everything to public github)

kilograham (approved):

Issue #6, opened 2021-01-23
Why does second stage bootloader not set VTOR and SP?


thejpster:
Having to make 0x100 in flash a function as opposed to a vector table complicates the use of other programming languages and SDKs, which assume flash starts with a vector table.

Is there a reason why the second stage bootloader doesn't set VTOR to 0x1000_0100, SP to [0x1000_0100] and then jump to [0x1000_0104]?

kilograham:
> Having to make 0x100 in flash a function as opposed to a vector table complicates the use of other programming languages and SDKs, which assume flash starts with a vector table.

I'm not sure that is an assumption they should make (unless flash is at 0x00000000), Still if they don't reference the stack pointer you could replace it with something like

0x47184b00, // ldr  r3, [pc, #0]; bx r3

to jump to the reset vector, and actually put a stack pointer in there (though all your bootstrap code need to be in the first 256 bytes of flash somewhere)

Wren6991:
> complicates the use of other programming languages and SDKs, which assume flash starts with a vector table.

Seems like the second stage bootloader already breaks this assumption, so I guess the other issue is the alignment requirement of VTOR, and not wanting to hoist the reset handler over the vector table to fill the alignment hole?

kilograham:
P.S. yay, rust i'm guessing - i wanted to give it a try myself, but i had zero time, and my Rust is not very idiomatic yet!

Wren6991:
> Is there a reason why the second stage bootloader doesn't set VTOR to 0x1000_0100, SP to [0x1000_0100] and then jump to [0x1000_0104]?

Purely a size consideration, in case we needed to squeeze more in there at a later time (as you can see it is already quite tight), and want to keep the interface stable. No strong objections to that mechanism but it looks like it would add a few bytes. Then again I guess now would be the time to change it, if ever.

Wren6991:
(I'm guessing one of the unstated priors here is that you would rather not ship your own second stages with those SDKs)

Wren6991:
So I understand, you're suggesting a patch like this?

```
--- a/src/rp2_common/boot_stage2/boot2_w25q080.S
+++ b/src/rp2_common/boot_stage2/boot2_w25q080.S
@@ -277,7 +277,10 @@ soft_reset:
     pop {r0}
     cmp r0, #0
     bne 1f
-    ldr r0, =(XIP_BASE + 0x101)
+    ldr r0, =(XIP_BASE + 0x100)
[...]

PR #10, opened 2021-01-23
Move vector table to beginning of flash image

Wren6991:
See #6.
- Vector table moved to start of user image for flash builds (immediately after boot2)
- boot2 vectors through this table, initialises VTOR and SP
- Move ELF entry point to start of `.reset` to simplify NO_FLASH builds (as there's no need to handle debugger entry specially there)

This is WIP, I haven't tested debugger flash entry, haven't modified the non-default second stages, haven't tried any non-default builds. We should probably take this chance to do something about the copy/paste between the boot2 files.

### DIFF
--- a/src/rp2_common/pico_standard_link/crt0.S
+++ b/src/rp2_common/pico_standard_link/crt0.S
@@ -209,11 +209,12 @@ _entry_point:
 .type _reset_handler,%function
 .thumb_func
 _reset_handler:
-    // Only core 0 should run this code (or indeed be able to reach it)
+    // Only core 0 should run the C runtime startup code; core 1 is normally
+    // sleeping in the bootrom at this point but check to be sure
     ldr r0, =(SIO_BASE + SIO_CPUID_OFFSET)
     ldr r0, [r0]
     cmp r0, #0
-    bne send_other_cores_to_bootrom
+    bne hold_non_core0_in_bootrom
 
     adr r4, data_cpy_table
 
@@ -301,7 +302,7 @@ runtime_init:
 // happen (it should sleep in the ROM until given an entry point via the
 // cross-core FIFOs) but it's good to be defensive.
 
-send_other_cores_to_bootrom:
+hold_non_core0_in_bootrom:
     ldr r0, = 'W' | ('V' << 8)
     bl rom_func_lookup
     bx r0
--- a/src/rp2_common/pico_standard_link/crt0.S
+++ b/src/rp2_common/pico_standard_link/crt0.S
@@ -144,35 +144,39 @@ __unhandled_user_irq:
 unhandled_user_irq_num_in_r0:
     bkpt #0
 
+// ----------------------------------------------------------------------------
+
 .section .reset, "ax"
 
-// This is the beginning of the image, which is entered from stage2 or bootrom USB MSD watchdog reboot
+// On flash builds, the vector table comes first in the image (conventional).
+// On NO_FLASH builds, the reset handler section comes first, as the entry
+// point is at offset 0 (fixed due to bootrom), and VTOR is highly-aligned.
+// Image is entered in various ways:
+//
+// - NO_FLASH builds are entered from beginning by UF2 bootloader
+//
+// - Flash builds vector through the table into _reset_handler from boot2
+//
+// - Either type can be entered via _entry_point by the debugger, and flash builds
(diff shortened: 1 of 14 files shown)


src/rp2_common/pico_standard_link/crt0.S
```diff
@@ -144,39 +144,53 @@ __unhandled_user_irq:
...
-    // When using flash, we install and use the ROM vector table to go thru regular bootrom/stage2 flash sequence
+    // Debugger tried to run code after loading, so SSI is in 03h-only mode.
+    // Go back through bootrom + boot2 to properly initialise flash.
     movs r0, #0
 #endif
-
-__vector_entry:
-    ldr r1, =(PPB_BASE + M0PLUS_CPUID_OFFSET)
-    str r0, [r1, #8]
+    ldr r1, =(PPB_BASE + M0PLUS_VTOR_OFFSET)
```
thejpster, line 178: Thank you!!

Wren6991: Yeah I had already fixed in passing, but I laughed when you mentioned it in the stream.

thejpster: Any chance you can not have both the processor model identification register and the core number identification register not both called CPUID whilst you're at it? :)

lurch: > the processor model identification register

That's `M0PLUS: CPUID Register` on page 80 of the RP2040 datasheet. 
https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_regs/include/hardware/regs/m0plus.h#L565
https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_structs/include/hardware/structs/scb.h#L14

> the core number identification register

That's `SIO: CPUID Register` on page 42 of the RP2040 datasheet.
https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_regs/include/hardware/regs/sio.h#L21
https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_structs/include/hardware/structs/sio.h#L15

> not both called CPUID whilst you're at it? :)

I suspect it's too late to change that at this stage. And I'm sure there's probably other registers which also have "duplicated names" between different peripheral blocks?

Wren6991: It's a bit late in the day to rename registers, but there is at least a documentation issue there, so I'll drop a couple of NOTE boxes in the relevant parts of the datasheet. I think you were lead down the garden path a little by this literal value appearing in the crt0 disassembly.

Wren6991: Note that the core number check in crt0 should be redundant anyway. It should be impossible to reach this point on core 1 (see https://github.com/raspberrypi/pico-bootrom/blob/00a4a19114195e20fb817bdfbca1165e157eef37/bootrom/bootrom_rt0.S#L222), there's just a paranoia check there in case you do something spectacularly broken

Wren6991:
Have briefly tried:

- blinky on flash, COPY_TO_RAM, NO_FLASH
- Stepping through the ELF entry point thing for catching debugger entries
- flash_program example, to make sure still callable as void
- blinky and flash_program with generic 03h second stage

should be most paths covered

src/rp2_common/pico_standard_link/crt0.S
```diff
@@ -284,11 +296,12 @@ runtime_init:
...
-// In case core 1's VTOR has already been moved into flash, we need to handle
-// core 1 reset. However, we do so by just jumping back into bootrom version of
-// wait_for_vector
+// If core 1 somehow gets into crt0 due to a spectacular VTOR mishap, we need to
+// catch it and send back to the sleep-and-launch code in the bootrom. Shouldn't
+// happen (it should sleep in the ROM until given an entry point via the
+// cross-core FIFOs) but it's good to be defensive.
 
-wait_for_vector:
+send_other_cores_to_bootrom:
```
kilograham, line 304: maybe "hold_non_core0_in_bootrom" or such

kilograham (changes requested):

src/rp2_common/pico_standard_link/crt0.S
Wren6991: :+1:

kilograham (approved):

PR #12, opened 2021-01-24
Add flash_get_unique_id to hardware_flash

Wren6991:
See https://github.com/raspberrypi/pico-examples/issues/6

This adds a function `flash_get_unique_id` which retrieves the unique identifier from external NOR flash. This is suitable for use as a unique board ID.

Also adds some lower-level flash functions for sending arbitrary commands to the SSI, but these are kept static.

### DIFF
--- a/src/rp2_common/hardware_flash/flash.c
+++ b/src/rp2_common/hardware_flash/flash.c
@@ -7,7 +7,16 @@
 #include "hardware/flash.h"
 #include "pico/bootrom.h"
 
+#include "hardware/structs/ssi.h"
+#include "hardware/structs/ioqspi.h"
+
 #define FLASH_BLOCK_ERASE_CMD 0xd8
+
+// Standard RUID instruction: 4Bh command prefix, 32 dummy bits, 64 data bits.
+#define FLASH_RUID_CMD 0x4b
+#define FLASH_RUID_DUMMY_BYTES 4
+#define FLASH_RUID_DATA_BYTES 8
+#define FLASH_RUID_TOTAL_BYTES (1 + FLASH_RUID_DUMMY_BYTES + FLASH_RUID_DATA_BYTES)
 
 #define __compiler_barrier() asm volatile("" ::: "memory")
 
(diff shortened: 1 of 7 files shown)


kilograham (commented):
I suggest we should also add a pico_ higher level library for this, which hooks in a _constructor_ to do this once at startup, then provides an accessor, so that people don't have to worry about IRQs and multi-core etc.

Wren6991:
Makes sense. I guess there is no overlap here with actual flash functionality. Will make that change.

tannewt:
The generic command API would be nice too. We'd be able to use it to get the size of the flash.

tannewt (approved):
Code looks good to me. Thanks for adding it!

src/rp2_common/hardware_flash/flash.c
```diff
@@ -99,3 +108,72 @@ void __no_inline_not_in_flash_func(flash_range_program)(uint32_t flash_offs, con
...
+
+static_assert(FLASH_UNIQUE_ID_SIZE_BYTES == FLASH_RUID_DATA_BYTES);
+
+void flash_get_unique_id(uint8_t *id_out) {
+#if PICO_NO_FLASH
+    panic_unsupported();
+#else
+    uint8_t txbuf[FLASH_RUID_TOTAL_BYTES] = {0};
+    uint8_t rxbuf[FLASH_RUID_TOTAL_BYTES] = {0};
+    txbuf[0] = 0x4b;
```
lurch, line 174: Would it make sense for the `0x4b` here to be `FLASH_RUID_CMD` instead?

Wren6991: Thanks, good spot

Wren6991:
Thanks @tannewt for squinting at the flash code, I want to get the `pico_` library in before merging so that we can update the example code to use that, but it's good that the flash part looks solid. The IRQ problem that @kilograham pointed out is a serious gotcha. Just trying to keep a lot of plates spinning right now :)

Wren6991:
Think this is ready but @kilograham  could you please look at the CMake changes in particular

kilograham (commented):
I'd be tempted to change the unique id (at least at the pico level) from an array of bytes to a structure containing an array of 8 bytes; easier to assign, safely pass pointers to etc.

src/rp2_common/pico_unique_id/unique_id.c
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2021 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "hardware/flash.h"
+#include "pico/unique_id.h"
+
+static_assert(PICO_UNIQUE_BOARD_ID_SIZE_BYTES == FLASH_UNIQUE_ID_SIZE_BYTES, "Board ID size must match flash ID");
```
lurch, line 10: "...flash ID size" ?

src/rp2_common/pico_unique_id/include/pico/unique_id.h
```diff
@@ -0,0 +1,64 @@
...
+
+#ifndef _PICO_UNIQUE_ID_H_
+#define _PICO_UNIQUE_ID_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** \file pico/unique_id.h
+ *  \defgroup unique_id unique_id
```
kilograham, line 15: should be pico_unique_id

kilograham: also should be added to docs/index.h (I notice that pico_stdio is missing from there, so if you could add that in passing!)

src/rp2_common/pico_unique_id/unique_id.c
Wren6991: Seems clear as it is?

src/rp2_common/pico_unique_id/include/pico/unique_id.h
Wren6991: pico_stdio seems to be listed fine under the runtime group?

kilograham (approved):

src/rp2_common/pico_unique_id/include/pico/unique_id.h
Wren6991: apparently I'm blind :)

Wren6991:
Thanks, both commits look good

kilograham (approved):

PR #14, opened 2021-01-24
elf2uf2: add the cache memory range described in Section 2.8.4.2

majbthrd:
In Section 2.8.4.2 (UF2 Format Details), the RP2040 datasheet indicates:

RAM only binary
▪ 0x20000000 - 0x20042000 Main RAM: Blocks can be positioned with byte alignment.
▪ 0x15000000 - 0x15004000 Flash Cache: (since flash is not being targeted, the Flash Cache is available for use as
RAM with same properties as Main RAM).

However, at present, the elf2uf2 utility doesn't support the second memory range.  This PR should address this.

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -74,7 +74,7 @@ const address_ranges rp2040_address_ranges_flash {
 const address_ranges rp2040_address_ranges_ram {
     address_range(MAIN_RAM_START, MAIN_RAM_END, address_range::type::CONTENTS),
     address_range(XIP_SRAM_START, XIP_SRAM_END, address_range::type::CONTENTS),
-    address_range(0x00000000u, 0x00002000u, address_range::type::IGNORE) // for now we ignore the bootrom if present
+    address_range(0x00000000u, 0x00004000u, address_range::type::IGNORE) // for now we ignore the bootrom if present
 };
 
 struct page_fragment {
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -215,6 +215,15 @@ static bool is_address_valid(const address_ranges& valid_ranges, uint32_t addr) 
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


kilograham (commented):
Hah, i thought it already did this....

Elsewhere CACHE_XXX is referred to as XIP_SRAM_XXX, so might be nice to change it to that

Also can you change the ROM size to 0x4000 - it grew!

FYI the bootrom will jump into the lowest address mapped preferring main RAM over XIP ram

majbthrd:
Can do.  I've done it as incremental commits for clarity.  There is a change to naming, the ROM region is enlarged, and I wasn't happy with the approach of the first commit, so I re-wrote the "ram_style" evaluation so that it is automatically in lock-step with the address ranges specified in "rp2040_address_ranges_ram".

lurch:
@kilograham 

> Elsewhere CACHE_XXX is referred to as XIP_SRAM_XXX, so might be nice to change it to that

I wonder if it might be worth using the same define-names (for all the addresses) as used in https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_regs/include/hardware/regs/addressmap.h ?

tools/elf2uf2/main.cpp
```diff
@@ -226,7 +238,7 @@ int elf2uf2(FILE *in, FILE *out) {
     bool ram_style = false;
     address_ranges valid_ranges = {};
     if (!rc) {
-        ram_style = 0x2 == eh.entry >> 28u;
+        ram_style = is_address_initialized(rp2040_address_ranges_ram, eh.entry);
```
lurch, line 241: @kilograham I guess this relies on there being no overlapping `address_range::type::CONTENTS` for any address covered by the ranges in `rp2040_address_ranges_flash` and `rp2040_address_ranges_ram`. Is it worth adding an assertion for that in `main()`, or would that just be overkill?

kilograham: yeah i think that is overkill

kilograham (approved):

PR #18, opened 2021-01-25
Pioasm Incorrectly Assembles Irq 'rel' Flag

monocasa:
By setting 5 instead of bit 4, pioasm sets the wait flag instead of the rel flag.

With this bug

```irq 0 rel```

is assembled to

```    0xc020, //  0: irq    wait 0```

After this patch, it is instead assembled to

```    0xc010, //  0: irq    nowait 0 rel```

### DIFF
--- a/tools/pioasm/pio_assembler.cpp
+++ b/tools/pioasm/pio_assembler.cpp
@@ -304,7 +304,7 @@ raw_encoding instr_wait::raw_encode(const program &program) {
 raw_encoding instr_irq::raw_encode(const program &program) {
     uint arg2 = num->resolve(program);
     if (arg2 > 7) throw syntax_error(num->location, "irq number must be must be >= 0 and <= 7");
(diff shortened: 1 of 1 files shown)


kilograham (commented):
Nice catch - that regressed at some point!

kilograham (approved):

kilograham (approved):

lurch:
> Nice catch - that regressed at some point!

As a regression-test, would it be technically feasible to check the "round trip", i.e. for each combination of command and options, check that the decompiled source matches the input source? Or would that be far too much work for far too little gain? :shrug: 
Although I guess that still wouldn't catch occurrences where you have the identical bug in both the assembler and the disassembler :laughing:

kilograham:
yes we should absolutely have regression tests that compare assembly with hand checked hex, however we had to settle for "run all ye programs" for release 1.

PR #26, opened 2021-01-27
Fix missing timer registers in DMA header

Wren6991:
See https://github.com/raspberrypi/pico-feedback/issues/22

Datasheet main text matched hardware, but DMA header and register listing generation was incorrectly parameterised (no single-source generation for DMA unlike other regblocks). The register listing is fixed in upstream datasheet, need to add the missing timers to the register headers too.


lurch (approved):
LGTM Luke! (matches what I get if I run through all the autogeneration steps)

kilograham (approved):

tannewt (commented):
Any chance we could get a #define for the number of pacing timers too? That way I could write a for loop to allocate one without have to take the size of `timer`.

PR #31, opened 2021-01-27
remove debugging left in by mistake

### LINKED ISSUE #28: pico_time/time.c using GPIO pins
I've been having a problem with calls to functions in time.c (eg. `sleep_ms()`) having side effects on GPIO pins.

I'm pretty sure this is related to the DEBUG_PINS macros; changing `#define PICO_DEBUG_PIN_BASE 19u` in src/rp2_common/hardware_gpio/include/hardware/gpio.h to something else changes the GPIO pins that are affected. (My project is using GPIO 21 as an output, which is how I discovered this.)

I don't immediately see how to turn this off!

kilograham:


src/common/pico_time/time.c
```diff
@@ -13,9 +13,6 @@
 #include "hardware/sync.h"
 #include "hardware/gpio.h"
 
-CU_REGISTER_DEBUG_PINS(core)
-CU_SELECT_DEBUG_PINS(core)
-
 const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(nil_time, 0);
 // use LONG_MAX not ULONG_MAX so we don't have sign overflow in time diffs
 const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(at_the_end_of_time, ULONG_MAX);
```
lurch, line 18: Hmm, the comment seems opposite to what the line below it is doing? :eyes: 
(yes, I realise this has nothing to do with the issue being fixed in this PR)

Wren6991 (approved):
Looks good to me other than the slightly dubious ULONG comment (which shouldn't block this PR -- looks like ULONG and LONG are just swapped in the comment)

Might be worth having a grep to see if there are more of these

dpgeorge:
I see this was merged to the develop branch.  Should we be tracking that one from MicroPython?  We are currently using the master branch and I'd prefer to stick to master if possible (if so, when does develop get merged to master?).

lurch:
> (if so, when does develop get merged to master?)

I believe we're planning to do that early next week (i.e. hopefully within the next few days).

PR #39, opened 2021-01-28
Missing pico stdio doxygen

kilograham:


docs/index.h
```diff
@@ -41,6 +41,7 @@
  * set of functionality above the basic hardware interfaces
  * @{
  * \defgroup pico_multicore pico_multicore
+ * \defgroup pico_stdlib pico_stdio
```
lurch, line 44: Should this have been `pico_stdio pico_stdio` ? :man_shrugging:

kilograham:
oops it was indeed there already

PR #41, opened 2021-01-28
Mutex owned and owner were not initialized by mutex_init. Combined ow…

kilograham:
…ned and owner as they had no separate value any more


Wren6991 (approved):
All seems sensible, doesn't seem to have broken anything

PR #49, opened 2021-01-29
Remove a dependecy on Pioasm to fix building with ninja

Daft-Freak:
This fixes the "'pioasm/pioasm', needed by '...', missing and no known rule to make it" error with ninja. I think the `add_custom_command` below adds the necessary dependency on the pioasm build.

I've successfully done a `ninja` and `make -j24` build on the examples with this (and a similar change to a couple of the examples).

(Clean build each time, 24-thread CPU)


kilograham:
does that work with a clean build?

Daft-Freak:
Yes, even doing something like:
```
cmake -GNinja -B build-ninja
ninja -C build-ninja pio_addition
```

to directly build an example using pioasm from nothing works fine.

kilograham (approved):
Yes the Pioasm dependency was completely wrong - CMake add_custom_target/add_custom_command are confusing!

PR #52, opened 2021-01-30
Update README.md

DhruvaG2000:
doxygen is not by default installed in most distributions. Hence needs to be added in ```sudo apt install ....``` part

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ instructions for other platforms, and just in general, we recommend you see [Pic
 
 1. Install CMake (at least version 3.12), and GCC cross compiler
    ```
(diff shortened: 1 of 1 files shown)


kilograham:
you only need doxygen if you are trying to build the docs which is not the default. we can perhaps mention that.

DhruvaG2000:
okay @kilograham do mention it then. closing this due to outdated pr

lurch:
Just a quick FYI that the (SDK) docs that `doxygen` _does_ build are hosted at https://raspberrypi.github.io/pico-sdk-doxygen/ so there seems to be little need for the user to install doxygen and build the docs themself?

PR #54, opened 2021-01-30
Improve spi write16/read16 documentation.

4ilo:
The write16/read16 functions only work as described after changing the amount of data_bits using spi_set_format.


Wren6991 (approved):
Thanks! Nice catch

scminvent:
Filed upstream after all, under the public skycmoon identity:

- Issue: raspberrypi/pico-sdk#2973
- PR (against `develop`): raspberrypi/pico-sdk#2974

The fix shape is the LWIP_PROVIDE_ERRNO=1 → LWIP_ERRNO_STDINCLUDE=1 swap on the FreeRTOS+lwIP variant only, equivalent to our existing `lwipopts.h` workaround. This issue stays open to track the upstream outcome; close once #2974 lands or is rejected.

PR #60, opened 2021-01-31
Add pio add program at offset

kilograham:


Wren6991 (commented):
Not the fault of this PR, but the doxygen comment says this:

```
/*! \brief Attempt to load the program at the specified instruction memory offset, panicking if not possible
 *  \ingroup hardware_pio
 *
 * \see pico_can_add_program_at_offset if you need to check whether the program can be loaded
 *
 * \param pio The PIO instance; either \ref pio0 or \ref pio1
 * \param program the program definition
 * \param offset the instruction memory offset wanted for the start of the program
 * \return the instruction memory offset the program is loaded at
 */
void pio_add_program_at_offset(PIO pio, const pio_program_t *program, uint offset);
```

And the return type is void. So this function doesn't do what is documented.

Wren6991 (approved):

PR #61, opened 2021-01-31
Bump cmake required

kilograham:


Wren6991 (approved):
elf2uf2 is still only on 3.12 but that should be fine as there's no `target_link_options` there

PR #63, opened 2021-01-31
Set thumb and executable attributes on padded boot2

Wren6991:
Nicer disassemblies


kilograham (approved):
works ok; i'll take the CODE segment being shown correctly

PR #70, opened 2021-02-01
enforce that resulting binaries dont have arm opcodes, fixes #50

### LINKED ISSUE #50: enforce that gcc toolchain only used thumb code
on my [distro](https://nixos.org) of choice, the gcc currently lacks a thumb version of `crtbegin.o`, so it silently inserted an arm32 version into all of my pico builds, resulting in nothing working
fixing that problem is outside of the scope of pico-sdk

but, detecting it, and warning/failing is probably within the scope of pico-sdk, and is very simple to do:
```
[nix-shell:~/apps/rpi/pico/pico-examples]$ readelf --arch-specific build/hello_world/usb/hello_usb.elf
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "ARM v6K"
  Tag_CPU_arch: v6K
  Tag_CPU_arch_profile: Microcontroller
  Tag_ARM_ISA_use: Yes
  Tag_THUMB_ISA_use: Thumb-1
  Tag_ABI_PCS_wchar_t: 4
  Tag_ABI_FP_denormal: Needed
  Tag_ABI_FP_exceptions: Needed
  Tag_ABI_FP_number_model: IEEE 754
  Tag_ABI_align_needed: 8-byte
  Tag_ABI_enum_size: small

[nix-shell:~/apps/rpi/pico/pico-examples]$ readelf --arch-specific build-ubuntu/hello_world/usb/hello_usb.elf
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "6S-M"
  Tag_CPU_arch: v6S-M
  Tag_CPU_arch_profile: Microcontroller
  Tag_THUMB_ISA_use: Thumb-1
  Tag_ABI_PCS_wchar_t: 4
  Tag_ABI_FP_denormal: Needed
  Tag_ABI_FP_exceptions: Needed
  Tag_ABI_FP_number_model: IEEE 754
  Tag_ABI_align_needed: 8-byte
  Tag_ABI_enum_size: small
```
when gcc did the wrong thing, it added a `Tag_ARM_ISA_use: Yes` into the arch specific flags on the resulting elf binary
it should be pretty simple to just check those flags after linking, and then fail if gcc somehow put arm32 code into the final elf

cleverca22:
the original fault, is that the `arm-none-gcc` on my host was built with multi-lib disabled
as a result, it could generate thumb binaries at compile time, but the `crtbegin.o` and the newlib `libc.a` had been pre-built in arm32 mode only

pico-sdk would silently link the arm32 variants, and then the code would just fault out at runtime, and only SWD could reveal why

this pr will cause the build to fail if such arm32 opcodes wind up in the binary, forcing you to fix the tooling before you can flash the pico

### DIFF
--- a/src/rp2_common.cmake
+++ b/src/rp2_common.cmake
@@ -17,10 +17,15 @@ function(pico_add_dis_output TARGET)
             )
 endfunction()
 
+function(pico_add_thumb_enforcement TARGET)
+    add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_READELF} -A ${TARGET}${CMAKE_EXECUTABLE_SUFFIX} | grep Tag_ARM_ISA_use && { echo error, the cross gcc and newlib must be built with thumb multilib support \; exit 1 \; } || exit 0)
+endfunction()
+
 function(pico_add_extra_outputs TARGET)
     pico_add_hex_output(${TARGET})
     pico_add_bin_output(${TARGET})
     pico_add_dis_output(${TARGET})
+    pico_add_thumb_enforcement(${TARGET})
 
     # PICO_CMAKE_CONFIG: PICO_NO_TARGET_NAME, Don't defined PICO_TARGET_NAME, type=bool, default=0, group=build
     # PICO_BUILD_DEFINE: PICO_TARGET_NAME, The name of the build target being compiled (unless PICO_NO_TARGET_NAME set in build), type=string, default=target name, group=build


src/rp2_common.cmake
```diff
@@ -17,10 +17,15 @@ function(pico_add_dis_output TARGET)
             )
 endfunction()
 
+function(pico_add_thumb_enforcement TARGET)
+    add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_READELF} -A ${TARGET}${CMAKE_EXECUTABLE_SUFFIX} | grep Tag_ARM_ISA_use && { echo error, the cross gcc and newlib must be built with thumb multilib support \; exit 1 \; } || exit 0)
```
lurch, line 21: Does there need to be an additional check somewhere that readelf is installed? Or does CMake do that "automatically"?

cleverca22: readelf and ld are both part of the binutils package, and you must have binutils installed to link anything, and pico-sdk needs both host and target linkers, since its compiling some host utils

kilograham (commented):
can't do that as it won't work on windows (all the shell script piping etc)

kilograham:
if we're going to do anything here, we should check in elf2uf2 (which should be able to tell the difference... it already checks a lot of things)...

cleverca22:
@kilograham should i just throw in some c code and have cmake compile it the same way it does the other host utils?

cleverca22:
ah, `elf2uf2` does sound like a good place to add the check

kilograham:
yup add something here https://github.com/raspberrypi/pico-sdk/blob/master/tools/elf2uf2/main.cpp#L92

kilograham:
closing because this belongs in elf2uf2

cleverca22:
@kilograham is that a new git repo?, or do you just mean a new PR against the elf2uf2 file?

lurch:
I suspect the latter. (as you can see above, elf2uf2 is maintained in this repo)

kilograham:
yes in this repo, but this PR seems a bit obsolete

Issue #22, opened 2021-01-26
SVD File -missing VTOR table under CPU


conkerkh:
From SVD file specification under CPU section, vtorPresent is missing. According to the datasheet of RP2040 VTOR relocation using VTOR register is supported. Please add vtorPresent field set to 1 under CPU section.

[https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_cpu.html](url)

Wren6991:
Note this is probably the reason it wasn't added:

> vtorPresent | Indicate whether the Vector Table Offset Register (VTOR) is implemented in Cortex-M0+ based devices. This tag is either set to true or false, 1 or 0. **If not specified, then VTOR is assumed to be present.**

Raised a PR to add this anyway as there's no need to be ambiguous

Wren6991:
Now merged

PR #79, opened 2021-02-03
Add vtorPresent=1 to rp2040.svd

Wren6991:
See https://github.com/raspberrypi/pico-sdk/issues/22

This is upstream in the SVD source too


lurch (approved):

lurch:
This fix is also included in #81 (because the SVD file is autogenerated).

Wren6991:
Good point, will close this one in favour of just #81

PR #80, opened 2021-02-03
Restore GPIO15 state after modifying it for USB enumeration fix

liamfraser:


Wren6991 (changes requested):
Would suggest also:
- Disabling pad output when switching to this function
- Enabling bus keep on GPIO15

to minimise disturbance to whatever's connected. E.g. rather than

```
gpio_set_function(dp, 8);
```

Perhaps

```
// Enable bus keep and force pin to tristate, so USB DP muxing doesn't affect
// pin state
gpio_set_pulls(15, true, true);
gpio_set_oeover(15, GPIO_GPIO_OVERRIDE_LOW);
// Select function 8 (USB debug muxing) without disturbing other controls
hw_write_masked(&iobank0_hw->io[15].ctrl,
	8 << IO_BANK0_GPIO15_CTRL_FUNCSEL_LSB,
	IO_BANK0_GPIO15_CTRL_FUNCSEL_BITS);
```

And the `pads` register would also need to be saved/restored. (Note this is already true because `gpio_set_function` modifies the pad OD/IE bits)

Wren6991:
Gave it a quick spin with `TUD_OPT_RP2040_USB_DEVICE_ENUMERATION_FIX=1` in my compile definitions. Confirmed that fix appears in the binary (i.e. it is reachable, does not get gc'd) and that the following gives me both good enumeration and a blinking LED on GPIO15:

```

#include <stdio.h>
#include "pico/stdlib.h"

int main() {
    // Setup GPIO before setting up USB CDC
    gpio_init(15);
    gpio_set_dir(15, GPIO_OUT);
    gpio_put(15, 1);

    stdio_init_all();
    while (true) {
        printf("Hello, world!\n");
        sleep_ms(1000);
        // Blink LED on GPIO15
        gpio_xor_mask(1u << 15);
    }
    return 0;
}
```
So think we are good to go

Wren6991 (approved):

PR #92, opened 2021-02-07
Change the quick-start instructions to include cstdlibc

ewpa:
Change the quick-start instructions to include the libstdc++-arm-none-eabi-newlib package.  This fixes the following error
encountered when building the examples.
```
[ 93%] Building CXX object [...]pico-sdk/src/rp2_common/
       pico_standard_link/new_delete.cpp.obj
[...]pico-sdk/src/rp2_common/pico_standard_link/new_delete.cpp:10:10:
fatal error: cstdlib: No such file or directory
   10 | #include <cstdlib>
      |          ^~~~~~~~~
compilation terminated.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
 
 1. Install CMake (at least version 3.13), and GCC cross compiler
    ```
-   sudo apt install cmake gcc-arm-none-eabi libnewlib-arm-none-eabi
+   sudo apt install cmake gcc-arm-none-eabi libnewlib-arm-none-eabi libstdc++-arm-none-eabi-newlib
    ```
 1. Set up your project to point to use the Raspberry Pi Pico SDK
 


bvgastel (dismissed):
Was also needed on my Ubuntu 20.04 installation.

aallan:
In the getting started instructions added `build-essentials` which seems cleaner.

```
$ sudo apt install cmake gcc-arm-none-eabi libnewlib-arm-none-eabi build-essential
```

README.md
```diff
@@ -38,7 +38,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
 
 1. Install CMake (at least version 3.13), and GCC cross compiler
    ```
-   sudo apt install cmake gcc-arm-none-eabi libnewlib-arm-none-eabi
+   sudo apt install cmake gcc-arm-none-eabi libnewlib-arm-none-eabi libstdc++-arm-none-eabi-newlib
```
aallan, line 41: Would prefer `build-essentials` to match the Getting Started book.

ewpa: I checked Ubuntu 20.04 and Debian 10 and `build-essential` does not depend on `libstdc++-arm-none-eabi-newlib`.  So I do not think that would solve this issue.

bvgastel: `libstdc++-arm-none-eabi-newlib` is a recommended package for `libnewlib-arm-none-eabi`, but was not installed on my rather default Debian Bullseye and Ubuntu 20.04 installations. It is probably better to fix the Getting Started book.

aallan: Instructions are for Raspberry Pi OS where `build-essentials` does include this package.

bvgastel: Maybe add to the regular text: '(Ubuntu and Debian users might need to also install `libstdc++-arm-none-eabi-newlib`)'? That would really be useful and prevents frustration.

ewpa: I ran the following on a _reasonably_ clean Raspbian 10 install on a 2B.

`sudo apt install cmake gcc-arm-none-eabi libnewlib-arm-none-eabi build-essential`

The result was this.

```
The following additional packages will be installed:
  binutils-arm-none-eabi cmake-data libarchive13 libjsoncpp1 libnewlib-dev
  librhash0 **libstdc++-arm-none-eabi-newlib** libuv1
```

Checking the package dependencies, `libnewlib-arm-none-eabi` recommends `libstdc++-arm-none-eabi-newlib`.  I also checked Debian 10 and Ubuntu 20.04 LTS and this is also the case.

For my Ubuntu workstation I did not have automatic installation of recommendations and had some of the ARM packages installed already, hence this package was missing for me.  Thus I think my PR might be better off as this.

`sudo apt install **--install-recommends** cmake gcc-arm-none-eabi libnewlib-arm-none-eabi build-essential`

However, re-running apt a second time with `--install-recommends` does not go back and reinstall the missed recommended packages.  So in my case where I already had `libnewlib-arm-none-eabi` installed, I did not get the `libstdc++-arm-none-eabi-newlib` installed running apt.  In that case, only including this package on the apt line explicitly results in its installation.

lurch:
Should this be moved to pico-feedback @aallan ?

aallan:
> Should this be moved to pico-feedback @aallan ?

It's a PR against the README file. You can't transfer PRs between repos, only Issues.

lurch:
Doh! I hadn't realised it was a PR and not an issue, sorry :laughing:

README.md
kilograham: I'm fine with adding this in the README... it didn't break my pi or ubuntu installations, so it seems like it does no harm

kilograham (changes requested):
needs to be rebased onto develop

ewpa:
I have rebased as requested.

kilograham (approved):

PR #95, opened 2021-02-07
Clean up headers in rp2_common to appease -Wstrict-prototypes

swetland:
In C, func() is a function taking an unspecified number of arguments,
vs func(void) a function taking no arguments. In C++ both forms indicate
"no arguments."
    
Update these headers to use the (void) form, which is correct in both
languages and avoids complaints when -Wstrict-prototypes is specified.

### DIFF
--- a/test/kitchen_sink/CMakeLists.txt
+++ b/test/kitchen_sink/CMakeLists.txt
@@ -59,6 +59,7 @@ target_compile_options(kitchen_sink_options INTERFACE
         -Wfloat-equal
         -Wmissing-format-attribute
         -Wno-long-long
+        $<$<COMPILE_LANGUAGE:C>:-Wstrict-prototypes>
 
         # todo not sure these are true, but investigate
         #-Wpacked
--- a/src/common/pico_base/include/pico/assert.h
+++ b/src/common/pico_base/include/pico/assert.h
@@ -36,7 +36,7 @@ extern "C" {
 #define hard_assert_if(x, test) ({if (PARAM_ASSERTIONS_ENABLED(x)) hard_assert(!(test));})
 
 #ifdef NDEBUG
-extern void hard_assertion_failure();
+extern void hard_assertion_failure(void);
 static inline void hard_assert(bool condition, ...) {
     if (!condition)
         hard_assertion_failure();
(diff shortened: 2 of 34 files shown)


kilograham (commented):
Can you add `-Wstrict-prototypes` to`test/kitchen_sink/CMakeLists.txt`and fixup the remaining issues from building that (it builds by default if you build from the SDK)
(Note worth doing both a release and debug build since some stuff is #ifdef-ed)

swetland:
Sure. Almost done, but it turns out that:
cc1plus: error: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++ [-Werror]

Do you know the Cmake incantation to add it just for C sources?

Also, how do you specify a debug vs release build?

kilograham:
you asked :-)

`$<$<COMPILE_LANGUAGE:CXX>:-Wstrict-prototypes>`

For debug, if not using an IDE which probably does this for you, you can pass `-DCMAKE_BUILD_TYPE=xxx` for e.g. Debug or Release, or the others..

Best to keep separate build directories for each config 

you can look at 

`cmake -LH` to see some more info

swetland:
Updated patch. Looks clean for me in Debug and Release build modes.

lurch:
image

test/kitchen_sink/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ target_compile_options(kitchen_sink_options INTERFACE
         -Wfloat-equal
         -Wmissing-format-attribute
         -Wno-long-long
+	$<$<COMPILE_LANGUAGE:C>:-Wstrict-prototypes>
```
kilograham, line 62: can you please fix the indent (8 spaces)

kilograham (changes requested):
otherwise great, thank you

kilograham (approved):
lgtm

test/kitchen_sink/CMakeLists.txt
swetland: Done.  Also re-based against origin/develop and fixed new instance of the warning in unqiue_iq.c

Pushed revised branch.

PR #107, opened 2021-02-09
Added version number of SDK to Doxygen navigation pane

JamesH65:


JamesH65:
@lurch @aallan @kilograham For review

aallan (approved):
LGTM

lurch:
I've built and viewed this version of the docs locally, and if I make my web-browser window super-narrow it then switches the layout into what I assume is the "mobile view", and the version number disappears. I dunno if we want to tweak this PR so that the SDK version-number is still visible for people browsing the API docs on their mobile phone? :shrug: 

As a side-issue, I noticed that in the "regular view" the top-left logo (which says "Raspberry Pi Pico SDK Documentation") links to `index.html` which takes me back to e.g. `file:///home/andrew.scheller/pico-sdk/build.docs/docs/doxygen/html/index.html` when clicked on; but in the "mobile view" the logo switches to just a RaspberryPi graphic which then links to `/` which takes me to e.g. `file:///` which isn't right.
In other words, if you're viewing https://raspberrypi.github.io/pico-sdk-doxygen/ in "mobile view" and you click on the RaspberryPi graphic at the top, it takes you to https://raspberrypi.github.io/ :confused:

JamesH65:
@lurch With regard to the mobile view, a lot of information disappears when you go narrow, not sure where to put a version number in that case. 

I'll take a look at the linking issue when time allows, just a bad reference somewhere.

PR #110, opened 2021-02-10
move PLL reset code from clocks driver to pll driver

majbthrd:
reset_block() and unreset_block_wait() calls on behalf of pll_init() are moved away from ./src/rp2_common/hardware_clocks/clocks.c and into ./src/rp2_common/hardware_pll/pll.c where they more rightly belong.  This follows the existing conventions already employed in the rtc, spi, uart, i2c, adc driver code.

The assert() calls to verify the validity of parameters are now done before such values are written to the PLL peripheral.

The peripheral state is checked before writing to the peripheral so that a PLL that already matches the provided configuration is not unnecessarily disrupted.

### DIFF
--- a/src/rp2_common/hardware_clocks/clocks.c
+++ b/src/rp2_common/hardware_clocks/clocks.c
@@ -7,7 +7,6 @@
 #include "pico.h"
 #include "hardware/regs/clocks.h"
 #include "hardware/platform_defs.h"
-#include "hardware/resets.h"
 #include "hardware/clocks.h"
 #include "hardware/watchdog.h"
 #include "hardware/pll.h"
@@ -148,9 +147,6 @@ void clocks_init(void) {
     // PLL SYS: 12 / 1 = 12MHz * 125 = 1500MHZ / 6 / 2 = 125MHz
     // PLL USB: 12 / 1 = 12MHz * 40  = 480 MHz / 5 / 2 =  48MHz
     /// \end::pll_settings[]
-
-    reset_block(RESETS_RESET_PLL_SYS_BITS | RESETS_RESET_PLL_USB_BITS);
-    unreset_block_wait(RESETS_RESET_PLL_SYS_BITS | RESETS_RESET_PLL_USB_BITS);
 
     /// \tag::pll_init[]
     pll_init(pll_sys, 1, 1500 * MHZ, 6, 2);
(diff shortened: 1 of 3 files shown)


Wren6991 (changes requested):
This makes a number of changes which are not listed in your description. For example, the USB controller is now not reset on the main boot path. That is a *significant* change that has potential to violate assumptions in the USB stack on a warm boot. If you have exhaustively followed up on the things this could break, it would be helpful to list that in your PR description.

Please also rebase your changes against `develop` and then force-push.

majbthrd:
The PR has been force rebased to the develop branch as requested.

In addition to what was described previously…

The USB controller reset already happens in the USB driver (rp2040_usb_init), just like rtc, spi, uart, i2c, adc driver code… as well as the PLL driver should this PR be adopted.

The SYSCFG reset does NOT happen as a result of the PR, but no pico-sdk code utilizes this peripheral.  Not resetting SYSCFG has strong benefits when the one core is debugging the other, as the reset disrupts the syscfg→dbgforce register.  Like all the above peripherals, the reset seems best located in the respective driver’s init code.

flit:
This would be really nice to have in the Pico SDK. Having a drag and drop _standard_ CMSIS-DAP debug probe with no extra hardware is one of the highly attractive features of the Pico, made possible by [pico-debug](/majbthrd/pico-debug). Unfortunately it requires a modified Pico SDK with this patch.

picoprobe is alright, but is slow, requires two Picos, and is non-standard, thus does not work with the majority of debuggers and IDEs. The only mainline release of a debug tool that supports picoprobe that I'm aware of is [pyocd](/pyocd/pyOCD) v0.30 and above.

There are also use cases beyond pico-debug for this PR.

frederikvs:
This PR enables debugging without requiring external hardware. A brilliant feature that would make the pico hardware **much** more convenient to use for slightly more advanced users. (i.e. those beyond the point of "drag and drop a firmware")

Personally, I feel this feature deserves a place in the getting started manual, right next to the picoprobe solution.

Are there any actions we can take to get this merged?

kilograham:
@Wren6991 can you take a look, I thought it was awaiting changes to be made

Wren6991 (approved):
I added a comment to call out the part I was confused about (removing USB/syscfg reset from runtime_init) and removed a couple of register writes that are unnecessary if PLL is in a clean state. Thank you for the patch

PR #118, opened 2021-02-11
Fix -Wsign-compare warnings

Wren6991:
We are clean with -Wall -Wextra except for a short list of disabled warnings. I tried turning these all back on, and they are mostly innocuous (insufficient gymnastics in assert-like macros mostly) but there are a few mixed-signed comparisons that are probably worth fixing.

This commit fixes those warnings and reenables `-Wsign-compare` on kitchen sink regression

### DIFF
--- a/src/rp2_common/hardware_pio/pio.c
+++ b/src/rp2_common/hardware_pio/pio.c
@@ -42,8 +42,9 @@ void pio_sm_unclaim(PIO pio, uint sm) {
 }
 
 int pio_claim_unused_sm(PIO pio, bool required) {
-    uint which = pio_get_index(pio);
-    uint base = which * NUM_PIO_STATE_MACHINES;
+    // PIO index is 0 or 1.
+    int which = (int)pio_get_index(pio);
+    int base = which * NUM_PIO_STATE_MACHINES;
     int index = hw_claim_unused_from_range((uint8_t*)&claimed, required, base,
                                       base + NUM_PIO_STATE_MACHINES - 1, "No PIO state machines are available");
     return index >= base ? index - base : -1;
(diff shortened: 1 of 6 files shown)


src/rp2_common/hardware_pio/pio.c
```diff
@@ -87,7 +88,7 @@ bool pio_can_add_program(PIO pio, const pio_program_t *program) {
 static bool _pio_can_add_program_at_offset(PIO pio, const pio_program_t *program, uint offset) {
     assert(offset < PIO_INSTRUCTION_COUNT);
     assert(offset + program->length <= PIO_INSTRUCTION_COUNT);
```
lurch, line 90: `pio_program_t.length` is a `uint8_t`, so I think that means the `assert(offset < PIO_INSTRUCTION_COUNT);` assertion is made redundant by the `assert(offset + program->length <= PIO_INSTRUCTION_COUNT);` assertion?

src/rp2_common/hardware_pio/pio.c
```diff
@@ -42,8 +42,9 @@ void pio_sm_unclaim(PIO pio, uint sm) {
...
 
 int pio_claim_unused_sm(PIO pio, bool required) {
-    uint which = pio_get_index(pio);
-    uint base = which * NUM_PIO_STATE_MACHINES;
+    // PIO index is 0 or 1.
+    int which = (int)pio_get_index(pio);
+    int base = which * NUM_PIO_STATE_MACHINES;
     int index = hw_claim_unused_from_range((uint8_t*)&claimed, required, base,
                                       base + NUM_PIO_STATE_MACHINES - 1, "No PIO state machines are available");
     return index >= base ? index - base : -1;
```
lurch, line 50: Experimentally, it seems you can leave `which` and `base` as `uint`, and then just change this last line to:
`return index >= (int)base ? index - (int)base : -1;`

(but perhaps that's actually not better than the version you've proposed :man_shrugging: )

src/rp2_common/hardware_pio/pio.c
Wren6991: Yes this is redundant, however it is also safe, clear, and not related to -Wsign-compare, so will leave it alone in this PR

Wren6991: actually this assert should fire if you pass in a value > UINT_MAX - PIO_INSTRUCTION_COUNT, whereas the second would not, so this should be left in
(e.g. if a casted -1 gets passed in by user code somehow)

lurch: LOL, I hadn't been considering PIO programs _that_ long :wink:

kilograham (approved):

PR #135, opened 2021-02-16
pioasm: Ada output format

JeremyGrosser:


tools/pioasm/ada_output.cpp
```diff
@@ -0,0 +1,148 @@
...
+        }
+    };
+
+    ada_output() : output_format("ada") {}
+
+    std::string get_description() override {
+        return "Ada specification";
+    }
+
+    FILE *open_spec_output(std::string package_name) {
```
kilograham, line 25: this doesn't seem to be used

tools/pioasm/ada_output.cpp
```diff
@@ -0,0 +1,148 @@
...
+                std::cerr << "error: too many options for Ada format" << std::endl;
+                return 1;
+        }
+
+        FILE *out = open_single_output(destination);
+        if (!out) return 1;
+
+        header(out, "This file is autogenerated by pioasm; do not edit!", 0);
+        fprintf(out, "pragma Style_Checks (Off);\n\n");
+        fprintf(out, "with RP.PIO;\n\n");
```
kilograham, line 104: where would users get this from?

JeremyGrosser: https://github.com/JeremyGrosser/rp2040_hal/blob/master/src/drivers/rp-pio.ads

PIO example project here: https://github.com/JeremyGrosser/pico_examples/tree/master/pio_blink

We're still working out the details of all the APIs and build process, so the project is light on documentation at the moment. I expect that situation to improve soon.

kilograham: cool; i guess you are the first there, you can claim "ada', and therefore what it does is up to you... we don't need to have the supporting files, though perhaps a link somewhere would be helpful

JeremyGrosser: I put links to the relevant github repositories in a comment at the top of ada_output.cpp... Is that where it belongs or is there a better place to document this dependency?

kilograham: i think that is fine

kilograham (dismissed):

kilograham:
can you pls rebase this onto develop

JeremyGrosser:
> can you pls rebase this onto develop

Done.

kilograham (approved):

PR #137, opened 2021-02-16
Add pico_bootsel_via_double_reset library, fix #87

### LINKED ISSUE #87: boot stage2: double reset to uf2
Hi hi, thank you for the pico. 

It is great to have uf2 built into the bootrom, as maintainer of similar [tinyuf2](https://github.com/adafruit/) (cross-platform uf2 bootloader), I kind of miss the double tap on reset button to enter uf2 feature. Which set a sram location to a magic number and wait for 0.5 second. If another reset occurs (double reset) during wait time then mcu will enter uf2 mode as following code does:

https://github.com/adafruit/tinyuf2/blob/master/src/main.c#L84 

- **pros**: it  requires only 1 button (RST) for small form factor PCB and in general easier to operate than press reset and hold BOOTSEL.
- **cons**: it does introduce additional 0.5s to boot process.

I wonder if RPI would be interested in adding this feature, my approach would be modify boot stage2 for double reset detection then use watchdog reset to jump to UF2. https://github.com/raspberrypi/pico-sdk/tree/master/src/rp2_common/boot_stage2

Thanks

Wren6991:
See #87. Pull in the double tap boot library from pico-examples, and add configs and doxygen.

I left the actual code unchanged, but wonder if it is worth adding a couple of compiler barriers.

### DIFF
--- a/src/rp2_common/CMakeLists.txt
+++ b/src/rp2_common/CMakeLists.txt
@@ -36,6 +36,7 @@ if (NOT PICO_BARE_METAL)
     # NOTE THE ORDERING HERE IS IMPORTANT AS SOME TARGETS CHECK ON EXISTENCE OF OTHER TARGETS
     pico_add_subdirectory(boot_stage2)
 
+    pico_add_subdirectory(pico_bootsel_via_double_reset)
     pico_add_subdirectory(pico_multicore)
     pico_add_subdirectory(pico_unique_id)
 
@@ -47,7 +48,6 @@ if (NOT PICO_BARE_METAL)
     pico_add_subdirectory(pico_mem_ops)
     pico_add_subdirectory(pico_malloc)
     pico_add_subdirectory(pico_printf)
-    pico_add_subdirectory(pico_bootsel_via_double_reset)
 
     pico_add_subdirectory(pico_stdio)
     pico_add_subdirectory(pico_stdio_semihosting)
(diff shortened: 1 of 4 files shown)


src/rp2_common/pico_bootsel_via_double_reset/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_library(pico_bootsel_via_double_reset INTERFACE)
+
+if (PICO_ON_DEVICE)
```
kilograham, line 3: this is in rp2_common; guard is not needed

src/rp2_common/CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ if (NOT PICO_BARE_METAL)
     pico_add_subdirectory(pico_mem_ops)
     pico_add_subdirectory(pico_malloc)
     pico_add_subdirectory(pico_printf)
+    pico_add_subdirectory(pico_bootsel_via_double_reset)
```
kilograham, line 50: I guess the ordering of things here isn't obvious, this really belongs before pico_multicore above

kilograham (changes requested):

kilograham (approved):
lgtm

lurch:
Once this has been merged into the `pico-sdk` repo, will the corresponding code in `pico-examples` be updated to match?

Wren6991:
> Once this has been merged into the `pico-sdk` repo, will the corresponding code in `pico-examples` be updated to match?

Yes, and that PR exists, and is linked to this PR. Scroll up

hathach:
Thank you for the PR, this will help small factor board with only 1 reset as button (BOOTSEL can be test pad for recover purpose) :)

PR #138, opened 2021-02-16
Added Pimoroni Tiny2040 and Keybow2040 board definitions

ZodiusInfuser:
Includes specific pin definitions for features of these boards.


kilograham:
oops; could you also rebase this onto develop

ZodiusInfuser:
Thanks @kilograham. You may need to point me to how to do that. I'm still learning my way around git

kilograham:
Let me do that then, i'm not quite sure how to explain that from this end for your end.

Can you verify that you are using B1s only, and that

```c
#ifndef PICO_FLOAT_SUPPORT_ROM_V1
#define PICO_FLOAT_SUPPORT_ROM_V1 0
#endif

#ifndef PICO_DOUBLE_SUPPORT_ROM_V1
#define PICO_DOUBLE_SUPPORT_ROM_V1 0
#endif
```

works for you (if so)

ZodiusInfuser:
I believe the launched products will be B1's but currently all we have are prototypes with B0, so cannot test.

Also, are #defines within the SDK already for the ADC pins? If so then we can remove our specific ADC ones on Tiny, as they'd be redundant

kilograham:
OK, gotcha; well will leave alone for now.

I do need to add common SDK defines for ADC etc, so will update yours accordingly - I'll leave your defines in place and just define the common ones using those.

kilograham:
oops

kilograham:
merged into board-headers branch

ZodiusInfuser:
Thanks @kilograham. 

> OK, gotcha; well will leave alone for now.

Just got confirmation that we will be B1, so although I cannot test myself, could you add those defines in for Tiny and Keybow please?
 
> I do need to add common SDK defines for ADC etc, so will update yours accordingly - I'll leave your defines in place and just define the common ones using those.

Thanks, that would be great!

lurch:
Slightly tangential to this @kilograham , in addition to the current "board definitions" concept, would there be any use in having an "addon definitions" concept? E.g. if using https://shop.pimoroni.com/products/pico-unicorn-pack with a Pico, I guess you might want to pick up the `#define PICO_DEFAULT_LED_PIN 25` from `boards/pico.h` but then also have e.g. `#define UNICORN_PACK_SW_A_PIN 16` in an `addons/unicorn_pack.h` ? Something similar might also be useful for Adafruit's Feather ecosystem? (i.e. separate pin-definition header files for the "base board" and different "addon boards")

If you think this is a good idea I'll spin it out into a separate issue, otherwise I'll just ignore it. :slightly_smiling_face:

ZodiusInfuser:
@lurch I think that's a great idea! A particular case I am dealing with at the moment is with our Breakout Garden range of I2C addon's. On most of our boards the default I2C pins we use are 4 and 5, yet the PicoExplorer has them wired to 20 and 21. If I could just include a `DEFAULT_I2C_SDA_PIN` and `DEFAULT_I2C_SCL_PIN` for instance, that would be nice.

I don't think we would add things like the Unicorn into it, but we could have a common "Pimoroni Packs" addon and then specific ones for our bases, as we are attempting to follow a standard to our pinout were possible.

It should be something independent from the boards I would say, or at least have an option of saying which set of RP2040 boards it applies to. No point building for the Unicorn addon for a Feather 2040 for instance.

A tangential thing to this. Some instructions in the documentation on how to link in the board/addon switching into VS code would be good. Currently if I want to build for the Tiny2040 I have to do it via command line.

hathach:
`PICO_DEFAULT_I2C_*_PIN` and `PICO_DEFAULT_SPI_*_PIN` would be nice, it will also makes the pico-examples run out of the box with most of the boards and their sensor, display add-on.

kilograham:
I just noticed
```
//////////////////////////
// brining over from pico.h  - need to verify...

// Drive high to force power supply into PWM mode (lower ripple on 3V3 at light loads)
#define PICO_SMPS_MODE_PIN 23
```
Do your boards actually have this hooked up?

ZodiusInfuser:
No we do not have that hooked up on our boards. Pin 23's used for the user switch instead. Going to double check the files now

ZodiusInfuser:
See https://github.com/raspberrypi/pico-sdk/pull/189 for tweaked board headers, and a new one for PicoSystem

lurch:
@ZodiusInfuser @hathach There's probably no need for my "addon boards" concept I outlined above... :thinking: 

The `PICO_DEFAULT_*` idea was added in #192 and https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/vgaboard.h#L77 shows how you would define an "addon board" header with the current SDK: define all the custom-pins you need to, and then just `#include "boards/pico.h"` to pick up all the default pin-values not already set. `boards/pico.h` is sprinkled with `#ifndef`s, so it won't override anything that's been set before you include it.

hathach:
Thanks for the update, yup, I think we are good for now with default i2c and spi pin :smiley:

PR #139, opened 2021-02-16
Fix typo in pwm.h?

tobias-simonsen:
While reviewing your code, I saw this possible inconsistency. Please apologize, if I should be mistaken.


src/rp2_common/hardware_pwm/include/hardware/pwm.h
```diff
@@ -343,7 +343,7 @@ static inline void pwm_retard_count(uint slice_num) {
  */
 static inline void pwm_set_clkdiv_int_frac(uint slice_num, uint8_t integer, uint8_t fract) {
     valid_params_if(PWM, slice_num >= 0 && slice_num < NUM_PWM_SLICES);
-    valid_params_if(PWM, fract >= 0 && slice_num <= 16);
+    valid_params_if(PWM, fract >= 0 && fract <= 16);
```
lurch, line 346: Nice catch. Should probably also be `< 16` though rather than `<= 16` ?
And (if you can) please switch this to PR against the `develop` branch rather than the `master` branch.

Wren6991:
Thank you, looks like this change was also made in #154

PR #140, opened 2021-02-16
elf2uf2: determine entry address selected by Boot ROM

majbthrd:
Presently, when confronted with an .elf with sections in both XIP_SRAM and MAIN_RAM, elf2uf2 erroneously insists that the lowest address (XIP_SRAM) must be where the entry address resides.

This PR fixes elf2uf2 by adopting the Boot ROM algorithm for the entry address, which can be summarized as:

If RAM boot, pick the lowest address in MAIN_RAM as the entry address

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -46,7 +46,8 @@ static int fail_write_error() {
 
 struct address_range {
     enum type {
-        CONTENTS,     // may have contents
+        CONTENTS_ENTRY,    // may have contents *and* allows an entry address
+        CONTENTS_NO_ENTRY, // may have contents but does NOT allow an entry address
         NO_CONTENTS,  // must be uninitialized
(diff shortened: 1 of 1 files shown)


kilograham (changes requested):
What the bootrom does is pick the lowest address in SRAM is present otherwise picks the the lowest in XIP_CACHE. There is a separate bootrom bug related to the fact that you cannot currently have a XIP_CACHE only binary.

That bug might get fixed in the future, so I'd rather leave the code roughly as it was, and rather adding a new contents type (which isn't able to specify the ordering conditions), just put that code for that in the ram-binary case only (and for now add an error if the result is in XIP_CACHE). (also it was confusing that flash was specified/specifiable as CONTENTS_NO_ENTRY, but that is really a side effect of the above).

majbthrd:
changes made as requested

kilograham (approved):

PR #145, opened 2021-02-17
fix the representation of at_the_end_of_time

### LINKED ISSUE #132: getchar() fails after 1.19 hours (Problem with time_reached / at_the_end_of_time)
I haven't reproduced this for myself, but user on the Adafruit Discord a user reported:
> Program runs without generating any junk; but after nearly an hour it suddenly starts seeing an endless string of:
0xff 0xff 0xff 0xff .. from the USB port!

and showed this loop, citing https://github.com/wa1tnr/camelforth-rp2040-aU/blob/main/17-camelforth-a.d/pico-examples/camelforth-a/forth/rp2040_pico_getkey_usb.inc:
```c++
uint8_t ch_read;
do {
    ch_read = getchar();
} while (( ch_read == '\0' ) || ( ch_read == 0xff ));
// valid character now in ch_read
```

I noticed that `getchar()` uses a wait until `at_the_end_of_time`.  However, at least superficially it looks like "the end of time" is just 1.19 hours after boot!

`absolute_time_t` is a 64-bit quantity, but `at_the_end_of_time` is initialized to `ULONG_MAX`, a 32-bit quantity (0xfffffffu):
> src/common/pico_time/time.c:17: const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(at_the_end_of_time, ULONG_MAX);

Instead, it seems it should be initialized to another value, possibly UINT64_MAX from *stdint.h*.

A second problem is likely to arise when waiting for absolute times after 1.19 hours:
```
PICO_WEAK_FUNCTION_DEF(time_reached)
bool PICO_WEAK_FUNCTION_IMPL_NAME(time_reached)(absolute_time_t t) {
    uint64_t target = to_us_since_boot(t);
    if (target > 0xffffffffu) return false;
    return time_us_64() >= target;
}
```
The `if` line should possibly read `if (target == at_the_end_of_time) return false;` and similarly for other sites.

kilograham:
at_the_end_of_time was erroneously ULONG_MAX which is only 32 bits; changed to INT64_MAX which is sufficiently far in the future, but allows for signed 64 bit arithmetic.

### DIFF
--- a/src/common/pico_base/include/pico/types.h
+++ b/src/common/pico_base/include/pico/types.h
@@ -27,8 +27,8 @@ typedef uint64_t absolute_time_t;
 
 /*! fn to_us_since_boot
  * \brief convert an absolute_time_t into a number of microseconds since boot.
- * \param t the number of microseconds since boot
- * \return an absolute_time_t value equivalent to t
+ * \param t the absolute time to convert
+ * \return a number of microseconds since boot, equivalent to t
  */
 static inline uint64_t to_us_since_boot(absolute_time_t t) {
     return t;
@@ -52,8 +52,8 @@ typedef struct {
 
 /*! fn to_us_since_boot
  * \brief convert an absolute_time_t into a number of microseconds since boot.
- * \param t the number of microseconds since boot
- * \return an absolute_time_t value equivalent to t
+ * \param t the absolute time to convert
+ * \return a number of microseconds since boot, equivalent to t
  */
 static inline uint64_t to_us_since_boot(absolute_time_t t) {
     return t._private_us_since_boot;
--- a/src/common/pico_base/include/pico/types.h
+++ b/src/common/pico_base/include/pico/types.h
@@ -7,13 +7,14 @@
 #ifndef _PICO_TYPES_H
 #define _PICO_TYPES_H
 
+#include "pico/assert.h"
+
 #include <stdint.h>
 #include <stdbool.h>
 #include <stddef.h>
 
 typedef unsigned int uint;
 
-#ifdef NDEBUG
 /*! \typedef absolute_time_t
     \brief An opaque 64 bit timestamp in microseconds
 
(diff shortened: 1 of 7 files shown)


src/common/pico_base/include/pico/types.h
```diff
@@ -49,11 +50,23 @@ typedef struct {
     uint64_t _private_us_since_boot;
 } absolute_time_t;
 
+/*! fn to_us_since_boot
+ * \brief convert an absolute_time_t into a number of microseconds since boot.
+ * \param t the number of microseconds since boot
+ * \return an absolute_time_t value equivalent to t
```
lurch, line 56: Are the param and return descriptions the wrong way around? :shrug:

kilograham:
fixes #132

lurch:
Looking at `types.h`, I wonder if it's also worth doing a small bit of refactoring in order to reduce the duplicated doxygen comments? (also, seems `assert.h` was missing)

```patch
diff --git a/src/common/pico_base/include/pico/types.h b/src/common/pico_base/include/pico/types.h
index 4b4f557..89ec210 100644
--- a/src/common/pico_base/include/pico/types.h
+++ b/src/common/pico_base/include/pico/types.h
@@ -10,10 +10,10 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <stddef.h>
+#include <assert.h>
 
 typedef unsigned int uint;
 
-#ifdef NDEBUG
 /*! \typedef absolute_time_t
     \brief An opaque 64 bit timestamp in microseconds
 
@@ -23,32 +23,13 @@ typedef unsigned int uint;
     \see to_us_since_boot
     \see update_us_since_boot
 */
+#ifdef NDEBUG
 typedef uint64_t absolute_time_t;
-
-/*! fn to_us_since_boot
- * \brief convert an absolute_time_t into a number of microseconds since boot.
- * \param t the absolute time to convert
- * \return a number of microseconds since boot, equivalent to t
- */
-static inline uint64_t to_us_since_boot(absolute_time_t t) {
-    return t;
-}
-
-/*! fn update_us_since_boot
- * \brief update an absolute_time_t value to represent a given number of microseconds since boot
- * \param t the absolute time value to update
- * \param us_since_boot the number of microseconds since boot to represent. Note this should be representable
- *                      as a signed 64 bit integer
- */
-static inline void update_us_since_boot(absolute_time_t *t, uint64_t us_since_boot) {
-    *t = us_since_boot;
-}
-
-#define ABSOLUTE_TIME_INITIALIZED_VAR(name, value) name = value
 #else
 typedef struct {
     uint64_t _private_us_since_boot;
 } absolute_time_t;
+#endif
 
 /*! fn to_us_since_boot
  * \brief convert an absolute_time_t into a number of microseconds since boot.
@@ -56,7 +37,11 @@ typedef struct {
  * \return a number of microseconds since boot, equivalent to t
  */
 static inline uint64_t to_us_since_boot(absolute_time_t t) {
+#ifdef NDEBUG
+    return t;
+#else
     return t._private_us_since_boot;
+#endif
 }
 
 /*! fn update_us_since_boot
@@ -67,8 +52,16 @@ static inline uint64_t to_us_since_boot(absolute_time_t t) {
  */
 static inline void update_us_since_boot(absolute_time_t *t, uint64_t us_since_boot) {
     assert(us_since_boot <= INT64_MAX);
+#ifdef NDEBUG
+    *t = us_since_boot;
+#else
     t->_private_us_since_boot = us_since_boot;
+#endif
 }
+
+#ifdef NDEBUG
+#define ABSOLUTE_TIME_INITIALIZED_VAR(name, value) name = value
+#else
 #define ABSOLUTE_TIME_INITIALIZED_VAR(name, value) name = {value}
 #endif
 
```

kilograham:
if you can figure out how to make doxygen work off one set of comments then great :-) - i'm not convinced it does the right things for these anyway - not much to refactor as they are different definitions of the same thing

kilograham:
oh let me look sorry, i didn't realize you had a patch

kilograham:
good idea

philipstarkey:
While I'm not the original reporter of #132, I've had a similar problem with the pico crashing after 1.19 hours. I've tested this PR today and can confirm it works (well other than the missing `assert.h` header that @lurch pointed out) and fixes the problem I was having with `getchar()`

liamfraser (dismissed):

liamfraser (approved):

PR #146, opened 2021-02-17
fixup not/reverse for pioasm python output (fixes #90)

### LINKED ISSUE #90: Python code generated by pioasm does is not correct
python code generated using the -o python  option in pioasm generates incorrect code

example mov isr ~x
generates the python code mov(isr, not x) which does not give the expected result

the correct code should be mov(irs, invert(x))

kilograham:


Wren6991 (approved):
This works, whilst testing it I also noticed that `wait()` is missing a comma, would be good to get some more systematic testing at some point

PR #147, opened 2021-02-17
Bunch of small fixes

kilograham:


src/common/pico_sync/lock_core.c
```diff
@@ -7,7 +7,7 @@
 #include "pico/lock_core.h"
 
 void lock_init(lock_core_t *core, uint lock_num) {
-    assert(lock_num >= 0 && lock_num < NUM_SPIN_LOCKS);
+    assert(lock_num < NUM_SPIN_LOCKS);
```
lurch, line 10: Should this be using the `valid_params_if` macro, as used elsewhere?

And on a related note, there's also a bunch of `valid_params_if` macros in `pwm.h` checking that a `uint` value is `>= 0`.

Wren6991: That would introduce a new config define (off by default), so would change the behaviour, not just fix a harmless warning. You can see this if you look at the definition of `valid_params_if`

Wren6991 (approved):
Looks good, doesn't seem to break anything

Issue #108, opened 2021-02-10
calling `sleep_ms` inside an interrupt handler causes lockup


Lard4:
Summary: when you call `sleep_ms` inside of an interrupt handler, some sort of infinite loop occurs inside of the hardware timer. Specifically, picoprobe/gdb is pointing to the code hanging somewhere in this function: https://github.com/raspberrypi/pico-sdk/blob/26653ea81e340cacee55025d110c3e014a252a87/src/rp2_common/hardware_timer/include/hardware/timer.h#L109

To reproduce: build/flash https://github.com/Lard4/pico-irq-test then when GP22 goes high (gets the interrupt), it turns off the LED then hangs.

lurch:
Looking at https://github.com/raspberrypi/pico-sdk/blob/master/src/common/pico_time/time.c#L332 it seems `sleep_ms` calls `sleep_us`, which in turn calls `sleep_until` which relies on an alarm-interrupt; so perhaps the problem is interrupts being fired from inside interrupts? :shrug: 
If you _must_ use `sleep_ms` inside an interrupt handler (which seems like an odd thing to do, because you typically want your interrupts to exit as quickly as possible?), it looks like you can set `PICO_TIME_DEFAULT_ALARM_POOL_DISABLED` to 1, and it'll then do the sleep using busy-waits instead of alarm IRQs?

Ahhh, and I've just noticed that https://raspberrypi.github.io/pico-sdk-doxygen/group__sleep.html says "These functions should not be called from an IRQ handler." :slightly_smiling_face:

Wren6991:
Yes, `sleep_` functions are interrupt-driven (to save power) and you are using them inside an interrupt. Sleeping in an IRQ is generally unsupported on *any* platform, and the function description tells you not to do this. Having long delays in an IRQ is also pretty unfriendly to the code being interrupted.

If you do want to insert a long delay in an IRQ, the function to use is `busy_wait_us_32()` which you will see is also listed as a "see also" under the sleep functions.

![image](https://user-images.githubusercontent.com/1298595/107514967-43205b00-6ba2-11eb-8711-e7280bc1467c.png)

![image](https://user-images.githubusercontent.com/1298595/107515000-4fa4b380-6ba2-11eb-81a5-39743e1f704c.png)

kilograham:
I think nonetheless we should downgrade a sleep in an IRQ handler to a busy_wait_us automatically (by default)

swetland:
[...]

PR #148, opened 2021-02-17
panic in debug mode if sleep methods are called in an exception hander

kilograham:
(unless PICO_ALLOW_SLEEP_IN_EXCEPTION is set)

### DIFF
--- a/src/common/pico_time/time.c
+++ b/src/common/pico_time/time.c
@@ -325,6 +325,11 @@ static int64_t sev_callback(__unused alarm_id_t id, __unused void *user_data) {
 #endif
 
 void sleep_until(absolute_time_t t) {
+#if PICO_ON_DEVICE && !PICO_ALLOW_SLEEP_IN_EXCEPTION && !defined(NDEBUG)
+    if (__get_current_exception()) {
+        panic("Attempted to sleep inside of an exception handler");
+    }
+#endif
 #if !PICO_TIME_DEFAULT_ALARM_POOL_DISABLED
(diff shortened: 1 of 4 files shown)


lurch:
Is exception handler the same "thing" (or a subset / superset) of interrupt handler? :confused: (apologies for my naivety)

src/common/pico_time/time.c
```diff
@@ -297,6 +297,11 @@ static int64_t sev_callback(__unused alarm_id_t id, __unused void *user_data) {
 #endif
 
 void sleep_until(absolute_time_t t) {
+#if PICO_ON_DEVICE && !PICO_ALLOW_SLEEP_IN_EXCEPTION && !defined(NDEBUG)
```
Wren6991, line 300: So `PICO_ALLOW_SLEEP_IN_EXCEPTION=1` just removes the `panic()`?  This only seems useful in combination with `PICO_TIME_DEFAULT_ALARM_POOL_DISABLED=1`, since otherwise you still die when you sleep in an exception, just quietly. Is it worth `#error`ing on invalid combinations of those flags?

kilograham:
fixes #108

src/common/pico_time/time.c
kilograham: ah yes, the intention was for sleep to work in that case (as a busy_wait) but in all fairness if the caller is going to go to that trouble they should call busy_wait themselves!

lurch:
Needs rebasing.

liamfraser (approved):

liamfraser (approved):

PR #151, opened 2021-02-18
Solve some file compilation failures in C++ environment

chenyang8094:
Solve the problem that the symbols of some files cannot be found in the C++ compilation environment.


kilograham:
this is already fixed in the develop branch

PR #152, opened 2021-02-18
Explicitly mark constants in autogenerated headers as unsigned values

lurch:
This should make it easier to enable more compiler-warnings in future.


kilograham:
ah, this doesn't seem to include the `_LSB` and `_MSB` constants!

kilograham:
Superceded by #182

PR #154, opened 2021-02-18
Bunch of small fixes

kilograham:


src/rp2_common/hardware_pwm/include/hardware/pwm.h
```diff
@@ -367,8 +371,8 @@ static inline void pwm_retard_count(uint slice_num) {
  * \param fract 4 bit fractional part of the clock divider
  */
 static inline void pwm_set_clkdiv_int_frac(uint slice_num, uint8_t integer, uint8_t fract) {
-    valid_params_if(PWM, slice_num >= 0 && slice_num < NUM_PWM_SLICES);
-    valid_params_if(PWM, fract >= 0 && slice_num <= 16);
+    check_slice_num_param(slice_num);
+    valid_params_if(PWM, fract <= 16);
```
lurch, line 375: See also #139
As I asked there, should this be `< 16` rather than `<= 16` ? Seems `DIV_FRAC` is only a 4-bit field? https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/pwm.h#L109

Wren6991 (approved):
All looks good, I did change the `<= 16` into a `< 16` in passing as that is a typo

PR #155, opened 2021-02-18
Add more memory barriers to avoid code re-ordering issues with DMA

kilograham:
added ` : : : "memory"` to the __dmb and __isb as it doesn't make much sense to have code motion across those (gcc seems to act like it does this anyway, but that just may be insufficient subtle test case)

add __dsb() and __compiler_memory_barrier()

added __compiler_memory_barrier( ) in `dma_wait_for_completion_blocking()`

### DIFF
--- a/src/rp2_common/hardware_sync/include/hardware/sync.h
+++ b/src/rp2_common/hardware_sync/include/hardware/sync.h
@@ -117,8 +117,9 @@ inline static void __dmb(void) {
 /*! \brief Insert a DSB instruction in to the code path.
  *  \ingroup hardware_sync
  *
- * The DSB (data synchronization barrier) acts as a special kind of. The DSB operation completes when all explicit
- * memory accesses before this instruction complete.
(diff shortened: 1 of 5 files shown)


src/rp2_common/hardware_sync/include/hardware/sync.h
```diff
@@ -111,7 +111,17 @@ inline static void __wfi(void) {
...
  */
 inline static void __dmb(void) {
-    __asm volatile ("dmb");
+    __asm volatile ("dmb" : : : "memory");
+}
+
+/*! \brief Insert a DSB instruction in to the code path.
+ *  \ingroup hardware_sync
+ *
+ * The DSB (data synchronization barrier) acts as a special kind of. The DSB operation completes when all explicit
```
lurch, line 120: special kind of...?

Wren6991 (approved):
All looks good, I fixed the comment typos. Also fixed up a redundant `__compiler_barrier` macro from the flash code, which should use the platform definition.

PR #159, opened 2021-02-19
Pico flash size bytes

### LINKED ISSUE #134: Build Compile Warning - PICO_FLASH_SIZE_BYTES redefined when setting this value in a board definition file.
Our rp2040 boards have 16MB on-board Flash, and we are defining this value in the board definition include file via the PICO_FLASH_SIZE_BYTES macro. 
i.e:
// board has 16M onboard flash
#define PICO_FLASH_SIZE_BYTES (16 * 1024 * 1024)

But when building a project, this board definition conflicts the value being set in:
../src/rp2040/hardware/regs/platform_defs.h

The pre-processor spits out a warning: "note: this is the location of the previous definition"

Which has the value set to:
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (2 * 1024 * 1024)
#endif

This seems like a bug, OR we're using this #define incorrectly in our board definition. 

Note - not a critical issue - everything still works, just an annoying message during project build.

kilograham:


lurch:
Looking at the other `#define` values in `platform_defs.h`, should `XOSC_MHZ` also be moved elsewhere? The RP2040 datasheet says "The RP2040 supports 1MHz to 15MHz crystals" so I guess XOSC_MHZ might be a board-level thing rather than a platform-level thing?

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -44,6 +33,23 @@
...
+
+#define FLASH_PAGE_SIZE (1u << 8)
+#define FLASH_SECTOR_SIZE (1u << 12)
+#define FLASH_BLOCK_SIZE (1u << 16)
+
+#define FLASH_UNIQUE_ID_SIZE_BYTES 8
+
+// PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes or -1 for unknown, type=int, default=-1, group=hardware_flash
+#ifndef PICO_FLASH_SIZE_BYTES
+#define PICO_FLASH_SIZE_BYTES -1
```
Wren6991, line 49: Is -1 the correct default value? E.g. see here in `hardware_flash.c`

```
#ifdef PICO_FLASH_SIZE_BYTES
    hard_assert(flash_offs + count <= PICO_FLASH_SIZE_BYTES);
#endif
```
Seems this would give a mixed sign comparison if the board header doesn't define flash size. Maybe it should be left undefined, or set to 16M?

Wren6991:
> Looking at the other `#define` values in `platform_defs.h`, should `XOSC_MHZ` also be moved elsewhere? The RP2040 datasheet says "The RP2040 supports 1MHz to 15MHz crystals" so I guess XOSC_MHZ might be a board-level thing rather than a platform-level thing?

Yeah the initial clock setup needs more thought in general, e.g. we should be able to support systems with no crystal at all, but currently that requires you to modify or roll your own `runtime_init`. It should also be hoisted up over the data init in crt0 to speed that up. I'll open an issue for it

kilograham:
fixes #134

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -44,6 +33,22 @@
...
+#define PARAM_ASSERTIONS_ENABLED_FLASH 0
+#endif
+
+#define FLASH_PAGE_SIZE (1u << 8)
+#define FLASH_SECTOR_SIZE (1u << 12)
+#define FLASH_BLOCK_SIZE (1u << 16)
+
+#define FLASH_UNIQUE_ID_SIZE_BYTES 8
+
+// PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes, type=int, default=-1, group=hardware_flash
```
lurch, line 47: Probably ought to remove the default from this line too?
(I've just checked, and all the scripts/programs dealing with `pico_configs.tsv` cope fine with `default` being absent)

lurch: #183 enforces this :slightly_smiling_face:

liamfraser (dismissed):

liamfraser (dismissed):

liamfraser (approved):

src/rp2_common/hardware_flash/include/hardware/flash.h
lurch: You still need to fix the PICO_CONFIG line to remove the `default`, i.e.
```
// PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes, type=int, group=hardware_flash
```

PR #170, opened 2021-02-20
systick cvr register is RW not RO

GregAC:
As specified in the RP2040 datasheet the SYST_CVR register is RW. Writes are required to clear the current count value.


lurch:
Already fixed on `develop` by #91 ?  (FYI I believe we're planning to do a pico-sdk release and bump `master` sometime during the next week)

GregAC:
Ah yes, hadn't realised there was a develop branch!

PR #171, opened 2021-02-20
b0 sincos shim was incorrect, causing crash

kilograham:
on B0 the sincos shim was written as a regular sincos method (i.e. expecting r2, r3 to be pointers for results) rather than returning values in r0-r3 as it should. This happened to be mostly harmless in the past, but got broken when fixes for out of range inputs were added (which meant shim was called without valid r2-r3)

### DIFF
--- a/src/rp2_common/pico_double/double_aeabi.S
+++ b/src/rp2_common/pico_double/double_aeabi.S
@@ -677,7 +677,7 @@ double_wrapper_section sincos
     bl sincostan_remainder
     pop {r2}
     mov lr, r2
-    b 1f
+    b 1f // continue with sincos
 
 wrapper_func sincos
     push {r2-r5, lr}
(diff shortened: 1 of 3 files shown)


src/rp2_common/pico_double/double_aeabi.S
```diff
@@ -688,25 +688,15 @@ wrapper_func sincos
...
-    bl 2f
+    bl 2f // call the shim
     pop {r4-r5}
     stmia r4!, {r0, r1}
     stmia r5!, {r2, r3}
     pop {r4, r5, pc}
 
 2:
     shimmable_table_tail_call SF_TABLE_V3_FSINCOS sincos_shim_bootstrap
-#if PICO_DOUBLE_PROPAGATE_NANS
```
kilograham, line 700: note this code was dead

kilograham:
fixes #172

liamfraser (approved):

PR #177, opened 2021-02-22
elf2uf2: fix address range bug

majbthrd:
This fixes a bug introduced in a recent PR.  The _END macros should not be added to _START.


lurch:
Relates to #140 and #168 

> The _END macros should not be added to _START.

Looks like you meant "The _START macros should not be added to _END."? :wink:

kilograham (approved):

PR #182, opened 2021-02-24
Fix compiler warnings2

kilograham:
generated headers now use _U() and conversion warnings have been fixed


liamfraser (approved):
Very long review but looks good to me!

PR #191, opened 2021-02-25
Updated and added new board headers for Pimoroni RP2040 boards

ZodiusInfuser:
Lets try this again


src/boards/include/boards/pimoroni_picosystem.h
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch, line 2: Should this be "Copyright (c) 2021 Pimoroni Ltd." ? :shrug:

ZodiusInfuser: Not got a clue! Both are valid, but it's on your repo, so...

kilograham:
Cool; we're getting there

1. Can you confirm that all boards with PICO_DEFAULT_LED_PIN have a 1/0 LED (it will now be valid to do without that define)
2. I added PICO_DEFAULT_WS2812_PIN which is currently unused by us, but I plan to added a pico_status_led in a future release which will use LED or WS2812
3. Can you let me know about default I2C and SPI pins as these will now have defaults too (again these can be omitted if no valid pins)

ZodiusInfuser:
Thanks for the merge @kilograham .

> 1\. Can you confirm that all boards with PICO_DEFAULT_LED_PIN have a 1/0 LED (it will now be valid to do without that define)

Yes, both Tiny and PicoSystem have a regular LED, although in our case it's active low rather than high.

> 2\. I added PICO_DEFAULT_WS2812_PIN which is currently unused by us, but I plan to added a pico_status_led in a future release which will use LED or WS2812

We don't have any products with a WS2812 as the LED at this time. Instead we have separate R, G and B LEDs, so for now I've set the G to be the `PICO_DEFAULT_LED_PIN`, and included separate defines for each LED element. E.g.
```
#define PICOSYSTEM_LED_G_PIN 13
#define PICOSYSTEM_LED_R_PIN 14
#define PICOSYSTEM_LED_B_PIN 15
```


> 3\. Can you let me know about default I2C and SPI pins as these will now have defaults too (again these can be omitted if no valid pins)

**Keybow 2040**
- I2C0 SDA = GP4
- I2C0 SCL = GP5
(no SPI used)

**PicoSystem**
- SPI0 CS = GP5
- SPI0 SCK = GP6
- SPI0 MOSI = GP7
(no I2C used)

**Tiny2040** is a weird case, as it's very much like the Pico in that any pins can be used, just from a smaller subset, so we haven't discussed specific pins to use as defaults. See pinout:
image
I suppose defaulting to SPI0 and I2C0 would be sufficient?

kilograham:
> I suppose defaulting to SPI0 and I2C0 would be sufficient?
Yes, we can do UART0, SPI0, and I2C0 all at the same time, so I'll default the pins for that

> Ah yes, I forgot about the R, G, B LEDs... I think that makes sense though I may add defines for the colors too - i think a future pico_status_led library might like to use different colors for different things if available.

kilograham:
Oh and yes active low makes sense as a setting too, although it won't be used for now

PR #193, opened 2021-02-26
Replace unavailable flag-preserving instruction

foopub:
It seems sub requires thumb2 instructions which are unavailable. This is in line with the rest of the sdk code base which uses subs.

### DIFF
--- a/src/rp2_common/hardware_clocks/clocks.c
+++ b/src/rp2_common/hardware_clocks/clocks.c
@@ -79,7 +79,7 @@ bool clock_configure(enum clock_index clk_index, uint32_t src, uint32_t auxsrc, 
             uint delay_cyc = configured_freq[clk_sys] / configured_freq[clk_index] + 1;
             asm volatile (
                 "1: \n\t"
-                "sub %0, #1 \n\t"
+                "subs %0, #1 \n\t"
                 "bne 1b"
                 : "+r" (delay_cyc)
             );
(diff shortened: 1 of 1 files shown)


kilograham:
some context needed here. why is this failing for you?  I'm guessing you are perhaps trying to build with clang?

foopub:
Yes, I'm trying to compile with clang. 
I don't know if this is of any use, but the disassembly of the gcc binary with `sub` has this:
```
157   62:   3001            adds    r0, #1
158   64:   3801            subs    r0, #1
159   66:   d1fd            bne.n   64 <clock_configure+0x64>
```
while the clang with `subs` it's:
```
 95   86:   1c40            adds    r0, r0, #1
 96   88:   3801            subs    r0, #1
 97   8a:   d1fd            bne.n   88 <clock_configure+0x88>
```
so it shouldn't make a difference to `bne`?

Also, I realised I was testing with the -masm-syntax-unified option on gcc, so I'll change the commit to include
`".syntax unified\n\t"`
which I've tested and it works with the normal cmake.

kilograham:
Yes the fact that it was converted to a `subs` was a good thing, since `sub` would never have worked.

I was going to suggest `.syntax unified` but wanted to make sure what you were doing first.

kilograham:
can you pls rebase

kilograham (approved):

Issue #112, opened 2021-02-10
pico_stdio_usb should (optionally) support a mechanism to reboot the RP2040 device


kilograham:
@MichaelBell has a fork which supports reboot when baud rate is set to 1200.

This is not ideal for picotool, because the OS is understandably not keen on you trying to talk CDC commands directly to a serial device it has claimed, and mapping from VID/PID to serial device is brittle/at-least-needing-of-an-external-library-for-cross-platform-support.

I suggest therefore, that we additionally define an additional interface for controlling the RP2040 (a sort of runtime companion to PICOBOOT but much simpler). picotool already needs a driver on windows anyways, so hopefully this isn't a huge issue for the user (although they may need to do this now for two interfaces?)

We can support both methods by #define in pico_stdio_usb, but we should document the extra interface so MicroPython, CircuitPython etc can support it in their REPL if they want

hathach:
Maybe it is worth looking at dfu runtime, it uses control transfer and doesn’t take any extra endpoint and work with dfu-util command`dfu-util -e`

https://github.com/hathach/tinyusb/blob/09868434cd9f53394350cce682333ada97f796c3/examples/device/dfu_runtime/src/main.c#L115

swetland:
This is probably a whole separate issue, but along the idea of "hands free" development just with picotool and debug over usb, it might be worth defining a default fault handler and protocol for passing minimal fault info (PC, FaultAddr, etc) across a soft reboot (and if sram can survive that, possibly park in picoboot or a postmortem interface where one could pull a ram dump, backtrace, etc)....

kilograham:
> This is probably a whole separate issue, but along the idea of "hands free" development just with picotool and debug over usb, it might be worth defining a default fault handler and protocol for passing minimal fault info (PC, FaultAddr, etc) across a soft reboot (and if sram can survive that, possibly park in picoboot or a postmortem interface where one could pull a ram dump, backtrace, etc)....

can you please open this as a separate issue; as the main issue is fixed

PR #197, opened 2021-02-28
Add support for resetting RP2040 via the USB connection when using pi…

kilograham:
…co_stdio_usb

	- setting baud rate to magic value (default=1200) will cause a reset to BOOTSEL mode
	- a VENDOR interface along side the CDC interface can be used to reset via refular flash boot, or into
	  BOOTSEL mode with control for the reset_usb_boot parameters for the latter
either method can be configured/enabled/disabled via #define

### DIFF
--- a/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
+++ b/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
@@ -19,6 +19,8 @@
  *  Note this library is a developer convenience. It is not applicable in all cases; for one it takes full control of the USB device precluding your
  *  use of the USB in device or host mode. For this reason, this library will automatically disengage if you try to using it alongside \ref tinyusb_device or
  *  \ref tinyusb_host. It also takes control of a lower level IRQ and sets up a periodic background task.
+ *
+ *  This library also includes (by default) functionality to enable the RP2040 to be reset over the USB interface.
  */
 
 // PICO_CONFIG: PICO_STDIO_USB_DEFAULT_CRLF, Default state of CR/LF translation for USB output, type=bool, default=PICO_STDIO_DEFAULT_CRLF, group=pico_stdio_usb
@@ -31,7 +33,7 @@
 #define PICO_STDIO_USB_STDOUT_TIMEOUT_US 500000
 #endif
 
-// todo perhaps unnecessarily high?
+// todo perhaps unnecessarily frequent?
 // PICO_CONFIG: PICO_STDIO_USB_TASK_INTERVAL_US, Period of microseconds between calling tud_task in the background, default=1000, advanced=true, group=pico_stdio_usb
 #ifndef PICO_STDIO_USB_TASK_INTERVAL_US
 #define PICO_STDIO_USB_TASK_INTERVAL_US 1000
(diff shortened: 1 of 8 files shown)


kilograham:
fixes #112

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -42,6 +44,36 @@
...
+#ifndef PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE
+#define PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE 1
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE, baud rate that if selected causes a reset into BOOTSEL mode (if PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE==1), default=1200, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE
+#define PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE 1200
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE, Enable/disable resetting into BOOTSEL mode via an additional VENDOR USB interface - enables picotool based reset, type=bool, default=1, group=pico_stdio_usb
```
lurch, line 57: `PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE` here should be `PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE`

liamfraser (approved):

liamfraser:
Looks good. I didn't realise you could implement your own vendor driver like that!

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -42,6 +44,36 @@
...
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_RESET_TO_FLASH_BOOT, If vendor reset interface is included allow rebooting with regular flash boot, type=bool, default=1, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_RESET_TO_FLASH_BOOT
+#define PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_RESET_TO_FLASH_BOOT 1
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_RESET_RESET_TO_FLASH_DELAY_MS, delays in ms before rebooting via regular flash boot, default=100, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_RESET_RESET_TO_FLASH_DELAY_MS
+#define PICO_STDIO_USB_RESET_RESET_TO_FLASH_DELAY_MS 100
```
lurch, line 74: Just out of curiosity, why would somebody ever need/want to change this?

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -42,6 +44,36 @@
 #define PICO_STDIO_USB_LOW_PRIORITY_IRQ 31
 #endif
 
+// PICO_CONFIG: PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE, Enable/disable resetting into BOOTSEL mode if the host sets the baud rate to a magic value (PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE), type=bool, default=1, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE
+#define PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE 1
```
lurch, line 49: Does it definitely make sense to enable-this-by-default, given your comments in #112 about "the OS is understandably not keen on you trying to talk CDC commands directly to a serial device it has claimed"? :man_shrugging:

src/rp2_common/pico_stdio_usb/reset_interface.c
```diff
@@ -0,0 +1,95 @@
...
+// Support for parameterized reset via vendor interface control request
+static bool resetd_control_request_cb(uint8_t __unused rhport, tusb_control_request_t const *request) {
+    if (request->wIndex == itf_num) {
+#if PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_RESET_TO_BOOTSEL
+        if (request->bRequest == RESET_REQUEST_BOOTSEL) {
+            uint gpio_mask = 0;
+            if (request->wValue & 0x100) {
+                gpio_mask = 1u << (request->wValue >> 9u);
+            }
+            reset_usb_boot(gpio_mask, request->wValue & 0x7f);
```
lurch, line 44: Does there need to be some comments about what `request->wValue` can be set to, or will that be covered by the picotool source-code?

kadamski:
Is there a related change to picotool that make use of this vendor interface? Or some other simple test script?

kilograham:
> Is there a related change to picotool that make use of this vendor interface? Or some other simple test script?

I didn't either until I found out the default one forced you to use endpoints (also this way pulls in much less code)

lurch:
I suspect @kilograham copied'n'pasted the wrong message into his reply there :slightly_smiling_face:

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
kilograham: the probably don't, but no harm in creating a define since I picked an arbitrary value

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
kilograham: that referred to trying to talk to a CDC device via libusb while the OS has it claime; you can use OS level things to set the baud rate just fne

kilograham:
> Looks good. I didn't realise you could implement your own vendor driver like that!

I didn't either until I found out the default one forced you to use endpoints (also this way pulls in much less code)

kilograham:
> Is there a related change to picotool that make use of this vendor interface? Or some other simple test script?

not public yet; i tested it locally within picotool, so i could release this, but the picotool changes will come soon separately

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
lurch: (this was fixed in #206 )

laf0rge:
The big question with this is now: How can the *actual* application (i.e. what I as a user of the SDK write) implement custom class drivers in my application?  It looks like the pico SDK is using this feature of tiny-usb for internal purposes?

PR #226, opened 2021-03-04
Add some extra defines to customise behaviour of STDIO_USB_RESET modes

lurch:
I _think_ I've got this right, but you'll obviously want to review it quite carefully @kilograham 

I've tested the defines in combination with VIA_BAUD_RATE and they all work fine, but I've been unable to test with VIA_VENDOR_INTERFACE.

### DIFF
--- a/src/rp2_common/pico_bootsel_via_double_reset/pico_bootsel_via_double_reset.c
+++ b/src/rp2_common/pico_bootsel_via_double_reset/pico_bootsel_via_double_reset.c
@@ -14,12 +14,12 @@
 #define PICO_BOOTSEL_VIA_DOUBLE_RESET_TIMEOUT_MS 200
 #endif
 
-// PICO_CONFIG: PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED, GPIO to use as bootloader activity LED when BOOTSEL mode is entered via reset double tap (or -1 for none), type=int, default=-1, group=pico_bootsel_via_double_reset
+// PICO_CONFIG: PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED, GPIO to use as bootloader activity LED when BOOTSEL mode is entered via reset double tap (or -1 for none), type=int, max=29, default=-1, group=pico_bootsel_via_double_reset
 #ifndef PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED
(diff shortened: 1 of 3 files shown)


src/rp2_common/pico_bootsel_via_double_reset/pico_bootsel_via_double_reset.c
```diff
@@ -14,12 +14,12 @@
 #define PICO_BOOTSEL_VIA_DOUBLE_RESET_TIMEOUT_MS 200
 #endif
 
-// PICO_CONFIG: PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED, GPIO to use as bootloader activity LED when BOOTSEL mode is entered via reset double tap (or -1 for none), type=int, default=-1, group=pico_bootsel_via_double_reset
+// PICO_CONFIG: PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED, GPIO to use as bootloader activity LED when BOOTSEL mode is entered via reset double tap (or -1 for none), type=int, max=29, default=-1, group=pico_bootsel_via_double_reset
```
kilograham, line 17: i think we should actually switch to using undefined rather than -1 now to match regular LED

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -54,6 +54,18 @@
 #define PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE 1200
 #endif
 
+// Setting this to something other than -1 will prevent picotool specifying an activity LED via VENDOR_INTERFACE. Set to -2 to specify none whilst also preventing picotool from specifying an activity LED.
```
kilograham, line 57: so again get rid of the -1 and use undefined, and split the -2 off into a separate define to disallow overriding the value via the reset interface

kilograham (changes requested):

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
lurch: I think I've done what you were asking for :slightly_smiling_face:

kilograham (approved):

kilograham (approved):
I would have added a static inline function for getting the pin mask, but not reason enough to change at this second

PR #229, opened 2021-03-05
Remove PICO_FLASH_SIZE_BYTES warning from header to allow -Werror

kilograham:
it breaks any build with -Werror; actual uses in C/CXX source that care should warn instead


lurch:
Doh! This PR reverts the latest commit on the tinyusb submodule. I expect you didn't mean to do that? :confused:

kilograham:
tsk that's what i get for giving myself temporary powers!

PR #236, opened 2021-03-06
pio: allow programs with 32 instructions

cmfcmf:
The PIO library currently errors when trying to find an offset for PIO programs with exactly 32 instructions.

I assume that this is a bug (?), because pioasm only errors (correctly) when a program has 33 instructions or more.


lurch:
Needs to be PRed against the `develop` branch rather than the `master` branch, but this LGTM :+1:

cmfcmf:
> Needs to be PRed against the `develop` branch rather than the `master` branch, but this LGTM

I see, done!

lurch (approved):

kilograham (approved):

PR #243, opened 2021-03-08
Timer race condition

### LINKED ISSUE #195: Repeated invocation of function "void sleep_us(uint64_t us)" can cause lockup
The attached code based on SDK C/C++ examples code test.c and built under Window10 demonstrates the problem:
[test.zip](https://github.com/raspberrypi/pico-sdk/files/6054267/test.zip)

#include <stdio.h>
#include "pico/stdlib.h"
#include "hardware/gpio.h"
#include "pico/binary_info.h"

const uint LED_PIN = 25;

int main() {

   stdio_init_all();

   gpio_init(LED_PIN);
   gpio_set_dir(LED_PIN, GPIO_OUT);
   while (1) {
      gpio_put(LED_PIN, 0);
      sleep_ms(250);
      gpio_put(LED_PIN, 1);
      puts("Hello World\n");
      sleep_ms(250);
	  
	  // invocation of function "void sleep_us(uint64_t us)"
	  // when the number of microseconds to sleep (parameter "us") value
      // is set to 9 program lockup occurs, values of 8 and 10 are OK
	  sleep_us(9);
   }
}

kilograham:
fixes #195

### DIFF
--- a/test/pico_time_test/pico_time_test.c
+++ b/test/pico_time_test/pico_time_test.c
@@ -64,12 +64,13 @@ static bool repeating_timer_callback(struct repeating_timer *t) {
 #define RESOLUTION_ALLOWANCE PICO_HARDWARE_TIMER_RESOLUTION_US
 #endif
 
+int issue_195_test(void);
+
 int main() {
     setup_default_uart();
     alarm_pool_init_default();
 
     PICOTEST_START();
-
     struct alarm_pool *pools[NUM_TIMERS];
     for(uint i=0; i<NUM_TIMERS; i++) {
         if (i == alarm_pool_hardware_alarm_num(alarm_pool_get_default())) {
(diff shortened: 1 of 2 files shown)


test/pico_time_test/pico_time_test.c
```diff
@@ -215,6 +216,35 @@ int main() {
     PICOTEST_CHECK(absolute_time_diff_us(near_the_end_of_time, at_the_end_of_time) > 0, "near the end of time should be before the end of time")
     PICOTEST_END_SECTION();
 
+    if (issue_195_test()) {
```
lurch, line 219: Apologies if this is a stupid question, but this seems to be checking the return from `issue_195_test()` but AFAICT that function will only ever `return 0` ? Was that function supposed to `return 1` if the delta between `expected_count` and `issue_195_counter` was too big?

kilograham: nope; there are macros

lurch: Perhaps I'm being _really_ stupid, but I can see that `issue_195_test()` has a `PICOTEST_START_SECTION` macro which [sets](https://github.com/raspberrypi/pico-sdk/blob/master/test/pico_test/include/pico/test.h#L22) `picotest_error_code = 0` and there's a `PICOTEST_END_SECTION` macro which [checks](https://github.com/raspberrypi/pico-sdk/blob/master/test/pico_test/include/pico/test.h#L24) `if (picotest_error_code != 0)`, but I can't tell where (in `issue_195_test()`) `picotest_error_code` actually gets modified :confused:

kilograham: the convention i am propagating is that tests return 0 on success or otherwise on failure which is what the macros do in general. Yes there is no actual check here atm (as it causes a hang in the failure case)... it _does not matter_

lurch: > (as it causes a hang in the failure case)... it does not matter

Oh, that's the bit I was missing. Sorry for pestering you!

Wren6991 (approved):
So the failure mode was an alarm firing at the point we were in the critical section setting an immediately-expiring timer on the same alarm, and getting dropped as a result?

kilograham:
note it also requires a mix of timers adding in IRQ (i.e. repeating) and non IRQ contexts (we already had a test which created a huge number of timers with random timeouts and makes sure they all happen and in the right order - just none of those repeated)

Issue #163, opened 2021-02-19
Missing GPIO out read function in SDK


Volksolive:
I haven't found in the SDK datasheet and hardware/gpio a function to read the GPIO output register.
Also in the hardware struct sio.h, the out register is mentioned as write only (io_wo_32 gpio_out;) but I think should be read write.

In the optic, the function gpio_get_pad() give the INFROMPAD info but there is no function to get the OUTTOPAD.

A side note on the RP2040 datasheet, the pads register are quite far down in the list of registers section, maybe it would be nice to add a reference to it (in the signets, or at the beginning of the section 2.19.6) so that it would be easier to access it instead of scrolling tens of pages.

lurch:
> I haven't found in the SDK datasheet and hardware/gpio a function to read the GPIO output register.

Note that https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/sio.h#L49 says "Reading back gives the last value written, NOT the input value from the pins." What's your use-case?

> Also in the hardware struct sio.h, the out register is mentioned as write only (io_wo_32 gpio_out;) but I think should be read write.

Yeah, https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/sio.h#L64 says `RW` so _perhaps_ https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_structs/include/hardware/structs/sio.h#L20 should be `io_rw_32` ? That's a question for @kilograham / @Wren6991 
(seems to be quite a few registers which are `RW` in hardware/regs/sio.h but `io_wo_32` in hardware/structs/sio.h)

> In the optic, the function gpio_get_pad() ...

@kilograham `gpio_get_pad` is declared at https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_gpio/gpio.c#L18 but doesn't have an entry (and therefore no doxygen) in https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_gpio/include/hardware/gpio.h - is that intentional?

> give the INFROMPAD info but there is no function to get the OUTTOPAD.

Again, an example use-case might be helpful?

Volksolive:
The use case is to be able to read what was written in the out register.
[...]

PR #247, opened 2021-03-11
Add gpio_get_out_level() accessor, and correct SIO GPIO_OUT struct ty…

Wren6991:
…pe from WO to RW

See #163

### DIFF
--- a/src/rp2_common/hardware_gpio/include/hardware/gpio.h
+++ b/src/rp2_common/hardware_gpio/include/hardware/gpio.h
@@ -432,7 +432,7 @@ static inline void gpio_put(uint gpio, bool value) {
  * \return true if the GPIO output level is high, false if low.
  */
 static inline bool gpio_get_out_level(uint gpio) {
-    return !!(sio_hw->gpio_out & 1u << gpio);
+    return !!(sio_hw->gpio_out & (1u << gpio));
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lurch:
https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/sio.h#L112 says the OE registers are RW too - should this PR also modify `gpio_oe` and `gpio_hi_oe` to be `io_rw_32`? (`gpio_oe` gets read from at https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_gpio/include/hardware/gpio.h#L454 and https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_gpio/include/hardware/gpio.h#L487 so I guess there must be nothing in the SDK to prevent you from reading from a `io_wo_32` ?)

And just for "debug use" completeness, is it worth adding accessor functions for reading the INOVER and OUTOVER registers?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -415,6 +415,26 @@ static inline void gpio_put(uint gpio, bool value) {
...
+ * To avoid races, this function must not be used for read-modify-write
+ * sequences when driving GPIOs -- instead functions like gpio_put() should be
+ * used to atomically update GPIOs. This accessor is intended for debug use
+ * only.
+ *
+ * \param gpio GPIO number
+ * \return true if the GPIO output level is high, false if low.
+ */
+static inline bool gpio_get_out_level(uint gpio) {
+    return !!(sio_hw->gpio_out & 1u << gpio);
```
kilograham, line 435: really trying to be consistent adding parenthesis to clarify things like & << precedence

kilograham (changes requested):

kilograham (approved):

PR #250, opened 2021-03-11
corrected clk_gpout3 to gpio pin 25 as per the rp2040 datasheet

twasiluk:
According to the RP2040 datasheet (https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf - 1.4.3. GPIO Functions table on page 14) the GPIO pin **25** is designated for clk_gpout3.

gpout3


lurch:
ping @liamfraser - is the code in the SDK wrong, or is the function table in the datasheet wrong?

lurch:
We've verified internally that the datasheet is correct and it's the SDK-code that is incorrect. Thanks @twasiluk

lurch (approved):

kilograham:
can you please rebase onto develop-1.1.1

PR #254, opened 2021-03-13
Tweaks to PIO clkdiv-setting code

lurch:
- refactor to reduce duplication
 - add extra param-validation

You get "unexpected results" if the divider is set too big :wink: https://github.com/micropython/micropython/issues/7025

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -196,29 +196,6 @@ static inline void sm_config_set_sideset(pio_sm_config *c, uint bit_count, bool 
                   (bool_to_bit(pindirs) << PIO_SM0_EXECCTRL_SIDE_PINDIR_LSB);
 }
 
-/*! \brief Set the state machine clock divider (from a floating point value) in a state machine configuration
- *  \ingroup sm_config
- *
- * The clock divider slows the state machine's execution by masking the
- * system clock on some cycles, in a repeating pattern, so that the state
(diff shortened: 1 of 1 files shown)


lurch:
The diff looks worse than it actually is because I swapped around the order of a couple of functions.

kilograham:
why?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -238,6 +215,38 @@ static inline void sm_config_set_clkdiv_int_frac(pio_sm_config *c, uint16_t div_
             (((uint)div_int) << PIO_SM0_CLKDIV_INT_LSB);
 }
 
+static inline void calculate_clkdiv_from_float(float div, uint16_t *div_int, uint8_t *div_frac) {
```
kilograham, line 218: this is a global namespace, you should call this something like pio_calculate_clkdiv_from_float

kilograham (changes requested):

src/rp2_common/hardware_pio/include/hardware/pio.h
lurch: Ahhh, this _did_ trigger my spidey-sense, and I wondered if it should have been named `_calculate_clkdiv_from_float` to mark it as "internal use only" as you'd typically do in Python. I've now renamed it to `pio_calculate_clkdiv_from_float` as you suggested. :+1:

lurch:
Manually "rebased" to `develop-1.1.1` and force-pushed.

kilograham (approved):

PR #269, opened 2021-03-19
Make flash_do_cmd public

tannewt:
This makes flash_do_cmd public. We're using it in CircuitPython to read the RDID register to determine the flash size. (We always link to 1MB and the rest is used for the user filesystem.)

### DIFF
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -80,6 +80,22 @@ void flash_range_program(uint32_t flash_offs, const uint8_t *data, size_t count)
  */
 void flash_get_unique_id(uint8_t *id_out);
 
+/*! \brief Execute bidirectional flash command
+ *  \ingroup hardware_flash
+ *
+ * Execute a bidirectional command to the flash. Bytes are simultaneously
+ * transmitted and received. Therefore, both buffers must be the same length,
+ * count, which is the lenght of the overall transaction. This is useful for
+ * reading metadata from the flash chip. Do *NOT* use it to mutate the data
+ * stored on the flash because it will not flush the cache.
+ *
+ *  \param txbuf Pointer to a byte buffer which will be transmitted to the flash
+ *  \param rxbuf Pointer to a byte buffer where data received from the flash will be written
+ *  \param count Length in bytes of txbuf and of rxbuf
+ */
+void flash_do_cmd(const uint8_t *txbuf, uint8_t *rxbuf, size_t count);
+
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -80,6 +80,22 @@ void flash_range_program(uint32_t flash_offs, const uint8_t *data, size_t count)
  */
 void flash_get_unique_id(uint8_t *id_out);
 
+/*! \brief Execute bidirectional flash command
+ *  \ingroup hardware_flash
+ *
+ * Execute a bidirectional command to the flash. Bytes are simultaneously
+ * transmitted and received. Therefore, both buffers must be the same length,
+ * count, which is the lenght of the overall transaction. This is useful for
```
lurch, line 88: lenght -> length

tannewt:
Typo fixed. Thanks @lurch

Wren6991 (dismissed):
Looks good. I have a fix for the comment (the cache is flushed) but pushes from maintainers are disabled on the PR branch, so will have to do that separately

```
commit c6f40d35603e15afa898af30a6f5beba3d92e341 (HEAD -> make_do_cmd_public)
Author: Luke Wren <wren6991@gmail.com>
Date:   Fri Apr 9 11:41:03 2021 +0100

    Clarify that cache is flushed, but that function is intended for low-level metadata access during startup

diff --git a/src/rp2_common/hardware_flash/include/hardware/flash.h b/src/rp2_common/hardware_flash/include/hardware/flash.h
index 51e780f..93532d2 100644
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -83,14 +83,24 @@ void flash_get_unique_id(uint8_t *id_out);
 /*! \brief Execute bidirectional flash command
  *  \ingroup hardware_flash
  *
- * Execute a bidirectional command to the flash. Bytes are simultaneously
- * transmitted and received. Therefore, both buffers must be the same length,
+ * Low-level function to execute a serial command on a flash device attached
+ * to the QSPI interface. Bytes are simultaneously transmitted and received
+ * from txbuf and to rxbuf. Therefore, both buffers must be the same length,
  * count, which is the length of the overall transaction. This is useful for
- * reading metadata from the flash chip. Do *NOT* use it to mutate the data
- * stored on the flash because it will not flush the cache.
+ * reading metadata from the flash chip, such as device ID or SFDP
+ * parameters.
+ *
+ * The XIP cache is flushed following each command, in case flash state
+ * has been modified. Like other hardware_flash functions, the flash is not
+ * accessible for execute-in-place transfers whilst the command is in
+ * progress, so entering a flash-resident interrupt handler or executing flash
+ * code on the second core concurrently will be fatal. To avoid these pitfalls
+ * is recommended that this function only be used to extract flash metadata
+ * during startup, before the main application begins to run: see the
+ * implementation of pico_get_unique_id() for an example of this.
  *
  *  \param txbuf Pointer to a byte buffer which will be transmitted to the flash
- *  \param rxbuf Pointer to a byte buffer where data received from the flash will be written
+ *  \param rxbuf Pointer to a byte buffer where data received from the flash will be written. txbuf and rxbuf may be the same buffer.
  *  \param count Length in bytes of txbuf and of rxbuf
  */
```

Wren6991 (approved):

tannewt:
Thank you!

PR #270, opened 2021-03-19
Boards typos

lurch:


src/boards/include/boards/vgaboard.h
```diff
@@ -74,4 +76,11 @@
 // vgaboard has a Pico on it, so default anything we haven't set above
 #include "boards/pico.h"
 
++// undefine UART if we're using 4-bit SD card (this needs to happen *after* including pico.h)
```
kilograham, line 79: The user may have overriden some of these themselves to make them not conflict (i do these sometimes  - e.g. i sometimes steal bit 0 of red for stdout). we should probably check for overlap - (also not you can undefine TX and RX pins independently, and don't need to worry about undefining the PICO_DEFAULT_UART)

kilograham (changes requested):

src/boards/include/boards/vgaboard.h
lurch: > we should probably check for overlap

So something like this?
```
#if (PICO_SD_DAT_PIN_COUNT == 4) && (VGABOARD_SD_DAT0_PIN == 19)
#if PICO_DEFAULT_UART_TX_PIN == 20
#undef PICO_DEFAULT_UART_TX_PIN
#endif
#if PICO_DEFAULT_UART_RX_PIN == 21
#undef PICO_DEFAULT_UART_RX_PIN
#endif
#endif
```
That feels a bit complicated, and there's obviously still corner-cases it'll miss out.
Alternatively I can remove the second commit from this PR if you'd prefer? (which is why I made them separate commits :wink: )

kilograham:
well it is very easy to forget that you lose your (default) UART pins with 4 pin SDIO. I would just leave it out for now though as I need to think some more about this - you can see this from picotool (that you have double-booked pins), and perhaps we should make it more obvious there.

lurch:
> I would just leave it out for now though

Done :+1:

Issue #252, opened 2021-03-12
I2C: Strange behaviour trying to read PA1010D GPS


Gadgetoid:
I've been testing the PA1010D i2c GPS (happy to supply one to anyone who's got a head for mysteries) with the Pico, starting with CircuitPython and reducing down to a minimum amount of C++ SDK code to reproduce.

Long story short, it will stall hard after about one i2c read using Pico's hardware I2C.

Using either CircuitPython `bitbangio` or the I2C PIO driver from the Pico Examples it will read perfectly happily and - indeed - in the same setup for testing this I've had valid GPS data pouring out using a CircuitPython example and `bitbangio` in lieu of hardware i2c.

Works:

```c++
#include <stdio.h>
#include "pico/stdlib.h"
#include "pio_i2c.h"

#define PIN_SDA 4
#define PIN_SCL 5

int main() {
    stdio_init_all();

    PIO pio = pio0;
    uint sm = 0;
    uint offset = pio_add_program(pio, &i2c_program);
    i2c_program_init(pio, sm, offset, PIN_SDA, PIN_SCL);

    uint addr = 0x10;

    while (1) {
        int ret;
        uint8_t rxdata[11];
        ret = pio_i2c_read_blocking(pio, sm, addr, rxdata, 10);
        rxdata[10] = '\n';
        printf(rxdata);
        sleep_ms(1000);
    }
    return 0;
}
```

Stalls:

```c++
#include <stdio.h>
#include "pico/stdlib.h"
#include "hardware/i2c.h"

int main() {
    stdio_init_all();

    i2c_init(i2c0, 100 * 1000);
    gpio_set_function(4, GPIO_FUNC_I2C);
    gpio_set_function(5, GPIO_FUNC_I2C);
    gpio_pull_up(4);
    gpio_pull_up(5);

    int addr = 0x10;

    while (1) {
        int ret;
        uint8_t rxdata[11];
        ret = i2c_read_blocking(i2c0, addr, rxdata, 10, false);
        rxdata[10] = '\n';
        printf(rxdata);
        sleep_ms(1000);
    }
    return 0;
}

```

I've played with the i2c baud rate to no avail. The hardware i2c seems to inevitably get stuck in this loop:

```c++
        do {
            abort_reason = i2c->hw->tx_abrt_source;
            abort = (bool) i2c->hw->clr_tx_abrt;
            if (timeout_check) {
                timeout = timeout_check(ts);
                abort |= timeout;
            }
[...]

PR #273, opened 2021-03-20
i2c: set hold time of SDA during transmit to an appropriate value

fivdi:
Fixes https://github.com/adafruit/circuitpython/issues/4082

~~Increases the set hold time of SDA during transmit from its default value of 1 to 2 enabling pico-sdk to function correctly with the TCS34725 color sensor.~~

On page 48 of the [I2C-bus specification and user manual](https://www.nxp.com/docs/en/user-guide/UM10204.pdf) Table 10 lists the "Characteristics of the SDA and SCL bus lines for Standard, Fast, and Fast-mode Plus I2C-bus devices". Below this table note [3] mentions the following:

> [3] A device must internally provide a hold time of at least 300 ns for the SDA signal (with respect to the V IH(min) of the SCL signal) to bridge the undefined region of the falling edge of SCL.

This PR sets the IC_SDA_TX_HOLD field in the IC_SDA_HOLD register to value required for a hold time of 300 ns in Standard-mode and Fast-mode and 120 ns  in Fast-mode Plus.

The following tests were performed with this PR at 10000, 100000 and 400000 baud with two devices ([MCP9808](https://www.adafruit.com/product/1782) and [TCS34725](https://www.adafruit.com/product/1334)) on the I2C bus and all tests ran successfully.

- Scan the I2C bus with [bus_scan](https://github.com/raspberrypi/pico-examples/tree/master/i2c/bus_scan) from pico-examples. Without this this PR the TCS34725 is not detected. With this PR the TCS34725 is detected.
- Read the Device ID register form the TCS34725. Without this PR there is an error. With this PR the Device ID register can be successfully read and is 0x44.
- Read the temperature from the MCP9808 temperature sensor in an infinite loop (works with or without this PR.)

Tests were also successfully performed with a BME280 and TCS34725 at 1000000 baud.

~~Unfortunately, I can't really explain why the modification made by this PR is needed to to get pico-sdk to function correctly with the TCS34725. According to figure 11 on page 10 of the [TCS34725 datasheet](https://www.mouser.de/datasheet/2/588/TCS3472_DS000390_3-00-1844543.pdf) the minimum data hold time, t(HDDAT), is 0μs and the maximum is 0.9μs. However, the notes below figure 11 do say the following:~~

~~Note(s):~~
~~1. Specified by design and characterization; not production tested.~~

### DIFF
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -76,11 +76,11 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     invalid_params_if(I2C, lcnt < 8);
 
     // The SDA hold time should be at least 300 ns, per the I2C standard.
-    // sda_hold_count [cycles] = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
+    // sda_tx_hold_count [cycles] = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
     // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
     // Add 1 to avoid division truncation.
-    uint sda_hold_count = ((freq_in * 3) / 10000000) + 1;
-    invalid_params_if(I2C, sda_hold_count > lcnt - 2);
+    uint sda_tx_hold_count = ((freq_in * 3) / 10000000) + 1;
+    invalid_params_if(I2C, sda_tx_hold_count > lcnt - 2);
 
     i2c->hw->enable = 0;
     // Always use "fast" mode (<= 400 kHz, works fine for standard mode too)
@@ -92,7 +92,7 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     i2c->hw->fs_scl_lcnt = lcnt;
     i2c->hw->fs_spklen = lcnt < 16 ? 1 : lcnt / 16;
     hw_write_masked(&i2c->hw->sda_hold,
-                    sda_hold_count,
+                    sda_tx_hold_count,
                     I2C_IC_SDA_HOLD_IC_SDA_TX_HOLD_BITS);
 
     i2c->hw->enable = 1;
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -87,9 +87,9 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
         sda_tx_hold_count = ((freq_in * 3) / 10000000) + 1;
     } else {
         // sda_tx_hold_count = freq_in [cycles/s] * 120ns * (1s / 1e9ns)
-        // Reduce 120/1e9 to 12/1e8 to avoid numbers that don't fit in uint.
+        // Reduce 120/1e9 to 3/25e6 to avoid numbers that don't fit in uint.
         // Add 1 to avoid division truncation.
-        sda_tx_hold_count = ((freq_in * 12) / 100000000) + 1;
+        sda_tx_hold_count = ((freq_in * 3) / 25000000) + 1;
     }
     assert(sda_tx_hold_count <= lcnt - 2);
 
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -75,11 +75,22 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     invalid_params_if(I2C, hcnt < 8);
     invalid_params_if(I2C, lcnt < 8);
 
-    // The SDA hold time should be at least 300 ns, per the I2C standard.
-    // sda_tx_hold_count [cycles] = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
-    // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
-    // Add 1 to avoid division truncation.
-    uint sda_tx_hold_count = ((freq_in * 3) / 10000000) + 1;
+    // Per I2C-bus specification a device in standard or fast mode must
+    // internally provide a hold time of at least 300ns for the SDA signal to
+    // bridge the undefined region of the falling edge of SCL. A smaller hold
+    // time of 120ns is used for fast mode plus.
+    uint sda_tx_hold_count;
+    if (baudrate < 1000000) {
+        // sda_tx_hold_count = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
+        // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
+        // Add 1 to avoid division truncation.
+        sda_tx_hold_count = ((freq_in * 3) / 10000000) + 1;
+    } else {
+        // sda_tx_hold_count = freq_in [cycles/s] * 120ns * (1s / 1e9ns)
+        // Reduce 120/1e9 to 12/1e8 to avoid numbers that don't fit in uint.
+        // Add 1 to avoid division truncation.
+        sda_tx_hold_count = ((freq_in * 12) / 100000000) + 1;
+    }
     assert(sda_tx_hold_count <= lcnt - 2);
 
     i2c->hw->enable = 0;
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -75,6 +75,13 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     invalid_params_if(I2C, hcnt < 8);
     invalid_params_if(I2C, lcnt < 8);
 
+    // The SDA hold time should be at least 300 ns, per the I2C standard.
+    // sda_hold_count [cycles] = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
+    // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
+    // Add 1 to avoid division truncation.
+    uint sda_hold_count = ((freq_in * 3) / 10000000) + 1;
+    invalid_params_if(I2C, sda_hold_count > lcnt - 2);
+
     i2c->hw->enable = 0;
     // Always use "fast" mode (<= 400 kHz, works fine for standard mode too)
     hw_write_masked(&i2c->hw->con,
@@ -84,10 +91,9 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     i2c->hw->fs_scl_hcnt = hcnt;
     i2c->hw->fs_scl_lcnt = lcnt;
     i2c->hw->fs_spklen = lcnt < 16 ? 1 : lcnt / 16;
-    // Set hold time of SDA during transmit to 2 for TCS34725 color sensor
-    i2c->hw->sda_hold =
-            I2C_IC_SDA_HOLD_IC_SDA_RX_HOLD_RESET << I2C_IC_SDA_HOLD_IC_SDA_RX_HOLD_LSB |
-            2 << I2C_IC_SDA_HOLD_IC_SDA_TX_HOLD_LSB;
+    hw_write_masked(&i2c->hw->sda_hold,
+                    sda_hold_count,
+                    I2C_IC_SDA_HOLD_IC_SDA_TX_HOLD_BITS);
 
     i2c->hw->enable = 1;
     return freq_in / period;
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -103,7 +103,7 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     i2c->hw->fs_scl_lcnt = lcnt;
     i2c->hw->fs_spklen = lcnt < 16 ? 1 : lcnt / 16;
     hw_write_masked(&i2c->hw->sda_hold,
-                    sda_tx_hold_count,
+                    sda_tx_hold_count << I2C_IC_SDA_HOLD_IC_SDA_TX_HOLD_LSB,
                     I2C_IC_SDA_HOLD_IC_SDA_TX_HOLD_BITS);
 
     i2c->hw->enable = 1;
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -84,6 +84,10 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     i2c->hw->fs_scl_hcnt = hcnt;
     i2c->hw->fs_scl_lcnt = lcnt;
     i2c->hw->fs_spklen = lcnt < 16 ? 1 : lcnt / 16;
+    // Set hold time of SDA during transmit to 2 for TCS34725 color sensor
```
lurch, line 87: I dunno enough about I2C to comment on the validity of this PR, but the comment here seems a bit misleading: it implies that this hold time is set _only_ for the "TCS34725 color sensor", whereas this PR actually changes it for all I2C devices.

fivdi: Yes, I agree, the comment is a bit misleading. Would the following be better?

The IC_SDA_TX_HOLD field in the IC_SDA_HOLD register has a default value of 1 which is known to be too low for the TCS34725 color sensor so set IC_SDA_TX_HOLD to 2.

lurch: Like I said earlier, I don't really know much about I2C, and I've not (yet) pored through the datasheet in enough detail to know what changing the `SDA_TX_HOLD` does. But with that disclaimer out of the way, it feels a "bit blunt" to change it for **all** devices when there's only _one_ device that's causing a problem? Perhaps it would be "cleaner" to leave the defaults as they are, but then provide an accessor-function for changing the hold times, which any driver-code for the TCS34725 could then use? :shrug:

dhalbert: The TCS34725 does work properly on other microcontrollers with their I2C hardware without special settings. So it may be that the default for `IC_SDA_TX_HOLD` is a bit low. It would be interesting to look at the HAL-level I2C drivers for other chips that use the licensed Synopsys DW_apb_i2c peripheral that the RP2040 uses.

As for the comment, it might be phrased as: "The TCS34725 and possibly other sensors require ..."

fivdi: > it feels a "bit blunt" to change it for all devices when there's only one device that's causing a problem?

Well, I've only used three I2C devices on a Pico so far and the TCS34725 is one of them. I don't know if there are other devices causing the same problem. The exact same TCS34725 breakout board that I'm using on the Pico functions flawlessly on a Raspberry Pi 4 from user space without the need to do anything special. It would be great if nothing special needed to be done to use the TCS34725 with the pico-sdk.

> Perhaps it would be "cleaner" to leave the defaults as they are, but then provide an accessor-function for changing the hold times, which any driver-code for the TCS34725 could then use?

Introducing a new function for changing hold times would certainly be less of a risk. On the other hand, it would also introduce new complexity. SDA hold times are really low level details and not exactly easy to understand. I think it's good that the pico-sdk hides the details of SDA hold times from users. I think the current pico-sdk API for I2C is really good. You just need to call ```i2c_init(i2c_inst_t *i2c, uint baudrate)``` and you're ready to go.

dhalbert:
There is a thorough implementation for the Synopsys DW_apb_i2c in the Linux kernel:
https://github.com/torvalds/linux/blob/master/drivers/i2c/busses/, files `i2c-designware-*.c`.

kilograham:
cc @Wren6991

dhalbert:
I have verified this fix works for the TC34725 and does not seem to interfere with the other I2C sensors I tried. I took Saleae traces and cannot see any timing differences, but without the fix the TCS34725 does not ACK several writes sent to it, and it works fine with the fix. We plan to incorporate this fix into CircuitPython's fork of pico-sdk for now.

dhalbert:
We (CircuitPython) have also seen [some flakiness](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_SH1107/issues/4) with SH1107-based I2C OLED displays. This fix also seems to fix the SH1107 problems.

dhalbert:
Increasing `IC_SDA_TX_HOLD` to `5` instead of `2` will support the PA1010D GPS as well (https://github.com/raspberrypi/pico-sdk/issues/252#issuecomment-808803326).

fivdi:
On page 48 of the [I2C-bus specification and user manual](https://www.nxp.com/docs/en/user-guide/UM10204.pdf) Table 10 lists the "Characteristics of the SDA and SCL bus lines for Standard, Fast, and Fast-mode Plus I2C-bus devices". Below this table note [3] mentions the following:

> [3] A device must internally provide a hold time of at least 300 ns for the SDA signal (with respect to the V IH(min) of the SCL signal) to bridge the undefined region of the falling edge of SCL.

On page 511 of the RP2040 Datasheet the documentation for the IC_SDA_HOLD register mentions the following:

> The bits [15:0] of this register are used to control the hold time of SDA during transmit in both slave and master mode
(after SCL goes from HIGH to LOW).

It would also be interesting to know what the DW_apb_i2c documentation has to say, but I don't have access to the documentation.

Should the value programmed in IC_SDA_TX_HOLD be what every is necessary to result in a 300 ns hold time?

dhalbert:
> It would also be interesting to know what the DW_apb_i2c documentation has to say, but I don't have access to the documentation.

I think what is in the RP2040 datasheet is just copied from the DW_apb_i2c documentation; it says so at the top.


> Should the value programmed in IC_SDA_TX_HOLD be what every is necessary to result in a 300 ns hold time?

That makes perfect sense; the value can be calculated instead of this empirical testing. Would you like to revise the PR to do that?

The Linux driver mentions 300ns:
https://github.com/torvalds/linux/blob/a5e13c6df0e41702d2b2c77c8ad41677ebb065b3/drivers/i2c/busses/i2c-designware-master.c#L54-L56:
```c
	/* Set standard and fast speed dividers for high/low periods */
	sda_falling_time = t->sda_fall_ns ?: 300; /* ns */
	scl_falling_time = t->scl_fall_ns ?: 300; /* ns */
```

dhalbert:
>Would you like to revise the PR to do that?

I made a PR  for the Adafruit fork of pico-sdk: https://github.com/adafruit/pico-sdk/pull/3. It calculates the SDA hold time. There's a Saleae screenshot included showing the difference.

fivdi:
> Would you like to revise the PR to do that?

I'll revise it today.

src/rp2_common/hardware_i2c/i2c.c
fivdi: The comment has been changed. Hopefully it's more appropriate now.

fivdi:
The PR has been revised.

src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -75,6 +75,13 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     invalid_params_if(I2C, hcnt < 8);
     invalid_params_if(I2C, lcnt < 8);
 
+    // The SDA hold time should be at least 300 ns, per the I2C standard.
+    // sda_hold_count [cycles] = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
+    // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
+    // Add 1 to avoid division truncation.
+    uint sda_hold_count = ((freq_in * 3) / 10000000) + 1;
+    invalid_params_if(I2C, sda_hold_count > lcnt - 2);
```
lurch, line 83: Is the usage of the `invalid_params_if` macro appropriate here, given that both `sda_hold_count` and `lcnt` are calculated values? :man_shrugging:  Perhaps a regular `assert` would be better?

src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -75,6 +75,13 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
     invalid_params_if(I2C, hcnt < 8);
     invalid_params_if(I2C, lcnt < 8);
 
+    // The SDA hold time should be at least 300 ns, per the I2C standard.
+    // sda_hold_count [cycles] = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
+    // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
+    // Add 1 to avoid division truncation.
+    uint sda_hold_count = ((freq_in * 3) / 10000000) + 1;
```
lurch, line 82: Perhaps this should be explicitly named `sda_tx_hold_count`, to match up with `I2C_IC_SDA_HOLD_IC_SDA_TX_HOLD_BITS` ?

fivdi: Good idea. `sda_hold_count` has been renamed to `sda_tx_hold_count`.

src/rp2_common/hardware_i2c/i2c.c
fivdi: `invalid_params_if` was used on line 83 because it's also used [a few lines up on line 76](https://github.com/raspberrypi/pico-sdk/blob/2ab09b28feb9fdedd784a16269b69646deea1562/src/rp2_common/hardware_i2c/i2c.c#L76):
```c
    invalid_params_if(I2C, lcnt < 8);
```
 If `invalid_params_if` is appropriate for `lcnt < 8` I assumed it's appropriate for `sda_hold_count > lcnt - 2`.

Is this assumption incorrect?

That being said, I'm not super familiar with the pico-sdk codebase so I've no objections to making this change.

lurch: I'm not sure either, which is why I shrugged. It's really a question for @kilograham

dhalbert:
@tannewt pointed out in our own PR that for "Fast Mode +", which is a 1MHz SCL clock rate, the specified fall time is only 120ns, so `IC_SDA_TX_HOLD` need only be 120ns or just over rather than 300ns. We did not hold up our own PR for this, but it could be added to this PR. The way the pico-sdk driver is written, it uses the "fast mode" setting of the hardware for all possible clock speeds. However, I don't know that the other calculations are applicable to 1MHz either.

I have never actually seen anyone use a 1MHz clock with CircuitPython, so if this is a limitation, we can live with it for now.

src/rp2_common/hardware_i2c/i2c.c
kilograham: invalid_params_if to catch the result of out of range parameters (unless the function returns an error code which most don't) - can be for a knock on calculation from the original 

assert() should be used to check internal consistency/assumptions

fivdi: Ok, I think I get it now.

Switched from `invalid_params_if` to `assert` for checking the consistency of the following constraint mentioned at the top of page 512 in the RP2040 datasheet:

> The programmed SDA hold time during transmit (IC_SDA_TX_HOLD) cannot exceed at any time the duration of the low part of scl. Therefore the programmed value cannot be larger than N_SCL_LOW-2, where N_SCL_LOW is the duration of the low part of the scl period measured in ic_clk cycles.

fivdi:
> @tannewt pointed out in our own PR that for "Fast Mode +", which is a 1MHz SCL clock rate, the specified fall time is only 120ns, so `IC_SDA_TX_HOLD` need only be 120ns or just over rather than 300ns. We did not hold up our own PR for this, but it could be added to this PR. The way the pico-sdk driver is written, it uses the "fast mode" setting of the hardware for all possible clock speeds. However, I don't know that the other calculations are applicable to 1MHz either.
> 
> I have never actually seen anyone use a 1MHz clock with CircuitPython, so if this is a limitation, we can live with it for now.

@dhalbert Thank you for the information. I'll look into it.

fivdi:
I did some tests with a TCS34725 and a BME280 using the 300 ns SDA TX hold time in Fast-mode Plus and there were no issues. The wording in the specification doesn't explicitly mention what the SDA TX hold time should be in Fast-mode Plus. However, given that the fall time of both SDA and SCL signals in Fast-mode Plus is 120 ns a hold time of 120ns is more appropriate. [This application note](https://www.cypress.com/file/364871/download) is also good. The PR has been updates to reflect this.

src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -75,11 +75,22 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
...
+    if (baudrate < 1000000) {
+        // sda_tx_hold_count = freq_in [cycles/s] * 300ns * (1s / 1e9ns)
+        // Reduce 300/1e9 to 3/1e7 to avoid numbers that don't fit in uint.
+        // Add 1 to avoid division truncation.
+        sda_tx_hold_count = ((freq_in * 3) / 10000000) + 1;
+    } else {
+        // sda_tx_hold_count = freq_in [cycles/s] * 120ns * (1s / 1e9ns)
+        // Reduce 120/1e9 to 12/1e8 to avoid numbers that don't fit in uint.
+        // Add 1 to avoid division truncation.
+        sda_tx_hold_count = ((freq_in * 12) / 100000000) + 1;
```
dhalbert, line 92: You reduce this to the smallest possible integer numerator and denominator here.  `3` and `10000000` for the non-1MHz case, and `3` and `25000000` for 1MHz.

dhalbert (changes requested):

src/rp2_common/hardware_i2c/i2c.c
fivdi:
```c
uint sda_tx_hold_count = ((freq_in * 3) / (baudrate < 1000000 ? 10000000 : 25000000)) + 1;
```

fivdi: Good idea. 120/1e9 reduced to 3/25e6. I choose readability over compactness here.

lurch:
This looks reasonable to me, and it's encouraging to see that it fixes some of the reported problems with I2C, but I don't know enough about I2C to be able to give it a "formal approval" :slightly_smiling_face:

Wren6991 (approved):
Thank you, this is a good change as per spec, and calculations look correct to me. I did to a double take on 25e6 vs 2.5e6, but seems correct.

dhalbert:
@fividi Thank you for your initial idea about this and your detailed spec reading to find the 300ns/120ns values.

fivdi:
> @fividi Thank you for your initial idea about this and your detailed spec reading to find the 300ns/120ns values.

@dhalbert You're welcome 😄

PR #274, opened 2021-03-21
rtc_get_datetime: read the rtc registers just once each

jepler:
In order for the returned value to accurately reflect a single moment in time, ensure the registers are read just once and in the datasheet order. Before this change, the RTC registers would each be read multiple times, leading (infrequently) to the returned fields not all reflecting the same moment in time.

The rp2040 datasheet has what I believe is an incorrect example (embedding the source of this function); will the datasheet be updated if this function is fixed?

This problem is only a speculative one; I did not actually observe it in the wild.

### DIFF
--- a/src/rp2_common/hardware_rtc/rtc.c
+++ b/src/rp2_common/hardware_rtc/rtc.c
@@ -92,13 +92,16 @@ bool rtc_get_datetime(datetime_t *t) {
     }
 
     // Note: RTC_0 should be read before RTC_1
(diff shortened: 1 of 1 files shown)


lurch:
> The rp2040 datasheet has what I believe is an incorrect example (embedding the source of this function); will the datasheet be updated if this function is fixed?

Yes, if the code in the SDK gets updated then the next release of the PDF documentation (whenever that happens) will automatically include the updated code :slightly_smiling_face: 

Could you rebase this PR against `develop-1.1.1` please?

jepler:
@lurch Rebased & base branch changed to develop-1.1.1.

kilograham (approved):
yuk; thank you...

Issue #275, opened 2021-03-21
Make boot_stage2 choice a function of the board header by default + make generic flash the default


kilograham:
The "default bootstage2" mechanism should include a meta file which (effectively) includes the other bootstage 2's with `#ifdef`, thus allowing the board header to choose a bootstage2 without having to move to move up to using the cmake level board config.

The default if unspecified should be for generic slow flash, so ppl don't get bitten by non working flash.

kilograham:
moving to 1.1.1 as it blocks simple fixes for some boards

kilograham:
merged

PR #279, opened 2021-03-23
Adafruit Feather RP2040 fixes

sandeepmistry:
The following changes are based on information from the Adafruit forum: https://forums.adafruit.com/viewtopic.php?f=57&t=176745&p=860819&hilit=rp2040#p860819

The pre-built CircuitPython .uf2 was working on my board, but none of the `pico-examples` were without the following SDK changes.

1) The board has 8 MB of flash according to the product page: https://www.adafruit.com/product/4884

2) CircultPython sets `PICO_FLASH_SPI_CLKDIV` to `4`: https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/Makefile#L237

3) CircultPython uses `boot2_generic_03h.S` for boot2: https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/Makefile#L236 

There's probably a better have a board specific `PICO_DEFAULT_BOOT_STAGE2_FILE` value instead of editing `src/rp2_common/boot_stage2/CMakeLists.txt` - but I wanted to get the discussion going :)

### DIFF
--- a/src/boards/include/boards/adafruit_feather_rp2040.h
+++ b/src/boards/include/boards/adafruit_feather_rp2040.h
@@ -69,11 +69,11 @@
 
 //------------- FLASH -------------//
 #ifndef PICO_FLASH_SPI_CLKDIV
-#define PICO_FLASH_SPI_CLKDIV 2
+#define PICO_FLASH_SPI_CLKDIV 4
 #endif
 
 #ifndef PICO_FLASH_SIZE_BYTES
-#define PICO_FLASH_SIZE_BYTES (4 * 1024 * 1024)
+#define PICO_FLASH_SIZE_BYTES (8 * 1024 * 1024)
 #endif
 
 // All boards have B1 RP2040


hathach:
I encounter the issue with `PICO_DEFAULT_BOOT_STAGE2_FILE` as well, but not sure how to fix it. sdk probably need to have a way for board to specify its bootstage2. Currently it is done at application CMakeLists.txt

lurch:
See #275 :wink: 
Also, it's (apparently) possible to have a board-specific cmake file in addition to / instead of a board-specific header file.

kilograham:
Yup, for `PICO_BOARD=foo` you can create a foo.cmake in boards/ or in another path set via `PICO_BOARD_CMAKE_DIRS` an that will be used to configure the build (including adding the board header to `PICO_CONFIG_HEADER_FILES`)... not too bad really, but the issue @lurch linked above will mean that you can just do it via your board header which is much easier

sandeepmistry:
@lurch @kilograham thanks for the link to #275 (and apologies for not taking a look at the issue list prior to opening this PR).

Would you prefer I drop commit https://github.com/raspberrypi/pico-sdk/pull/279/commits/57e51b46762a2c9a44705a17d392938c992c1eb5 from this pull request? Alternatively, as an interim solution, would it make sense for the default in `src/rp2_common/boot_stage2/CMakeLists.txt` to be `boot2_generic_03h.S` as this seems to be the plan in #275.

brentru:
@sandeepmistry  Thanks for the PR, I was able to get blink/ from the examples working properly.

tannewt:
Thanks for opening this PR! I was just about to look at this myself.

> 3. CircultPython uses `boot2_generic_03h.S` for boot2: https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/Makefile#L236

I [*just* changed this](https://github.com/adafruit/circuitpython/pull/4446). We now use our own [templated C stage2 code](https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/stage2.c.jinja) that pulls flash info from our [nvm.toml repo](https://github.com/adafruit/nvm.toml). This separates out the init code from specific flash skus. (It also uses 0xe7 quad word read which is 2 dummy cycles instead of 0xeb's four cycles.)

One reason the `boot2_w25q080.S` stage2 doesn't work because [GigaDevice flash](https://github.com/adafruit/nvm.raw/blob/main/flash/gigadevice/GD25Q64C.pdf) on the feather doesn't support two byte status register writes with the 0x01 command. Instead it supports 0x31 which writes the second status register byte directly. The Winbond chips support this too and it seems to be the trend. (0x01 on the winbond says it's backwards compatible to support two byte writes.)

I'd expect the 2 divisor to work with the quad setup after QE has been set. (The datasheet says commands work up to 80mhz.) However, it seems to be unreliable and the 4 divisor is a much safer bet.

kilograham:
> @lurch @kilograham thanks for the link to #275 (and apologies for not taking a look at the issue list prior to opening this PR).
> 
> Would you prefer I drop commit [57e51b4](https://github.com/raspberrypi/pico-sdk/commit/57e51b46762a2c9a44705a17d392938c992c1eb5) from this pull request? Alternatively, as an interim solution, would it make sense for the default in `src/rp2_common/boot_stage2/CMakeLists.txt` to be `boot2_generic_03h.S` as this seems to be the plan in #275.

I've move #275 into our next bug fix release, so nothing going to happen b4 then - yes please remove 57e4 tho (and rebase to develop-1.1.1)

sandeepmistry:
> I've move #275 into our next bug fix release, so nothing going to happen b4 then - yes please remove 57e4 tho (and rebase to develop-1.1.1)

@kilograham done, please let me know if anything needs to be changed in this PR.

hathach:
> See #275 
> Also, it's (apparently) possible to have a board-specific cmake file in addition to / instead of a board-specific header file.

Thanks @lurch That is  great to know that we can add board-specific cmake file. while waiting for #275 should I add cmake for Feather rp2040 to dictate the boot stage2, or just wait for it since header only is nicer I guess.

kilograham:
subsumed by #293

kilograham:
note 1.1.1 has been released which should fix feather issues

PR #280, opened 2021-03-24
Use pico_unique_id as the USB device serial number

earlephilhower:
Use the pico_unique_id to generate a unique serial number for the
STDIO_USB device, allowing for individual boards to be identified.


lurch:
I thought this was a duplicate, but I did a search and found https://github.com/raspberrypi/picoprobe/pull/8 which is actually slightly different :laughing: 

Can you rebase this against `develop-1.1.1` please?

earlephilhower:
@lurch  By inspection it looks like we get the same results.  `strcat` is heavier weight, but this generation is called exactly 1 time and if you've got `stdio` running then `strcat` is along for the ride anyway (i.e. no add'l code space increase).

lurch:
I wonder if it's worth adding a "common" `pico_unique_id_to_string(char* str, uint strlen)` function to the `pico_unique_id` lib? :shrug: (For bonus points maybe you could copy only a subset of the unique_id if strlen is too short, and start repeating the unique_id if strlen is too long? :laughing: )

earlephilhower:
As a lowly end user I was hesitant to suggest a new API call, but your suggestion does make sense (modulo the repeating copies if the buffer is too long...that seems like not the right thing to do).  Let me adjust this PR accordingly.  Then, if merged, picoprobe can be modified to use this new API and only 1 copy of the code will exist (yay!).

src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c
```diff
@@ -121,6 +124,18 @@ const uint16_t *tud_descriptor_string_cb(uint8_t index, uint16_t langid) {
     #define DESC_STR_MAX (20)
     static uint16_t desc_str[DESC_STR_MAX];
 
+    // lazy update the serial number if not set yet
+    if (!usbd_serial_str[0]) {
+        pico_unique_board_id_t id;
+        pico_get_unique_board_id(&id);
+
+        for (int i = 0; i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES; i++) {
```
kilograham, line 132: yeah, we should pull in printf etc. just to have the serial number. please replace with a trivial loop to fill in the hex values directly

kilograham (changes requested):

kilograham:
ah i see this is concurrent editing... i'm ok with adding a helper function to do the trivial loop since it seems useful.

kilograham:
also please rebase this onto develop not develop-1.1.1 @lurch steered you wrongly!

earlephilhower:
Closing due to being stuck in a weird git rebase heck.

New PR #281 supersedes with the suggested new API and low-code implementation.

PR #281, opened 2021-03-25
Add pico_get_unique_board_id_string API

earlephilhower:
Add a new API to pico_unique which will turn the unique ID into a
canonical text string.

Use this API to update the USB device serial number in stdio_usb.

Supersedes #280

### DIFF
--- a/src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c
@@ -30,6 +30,7 @@
 
 #include "tusb.h"
 #include "pico/stdio_usb/reset_interface.h"
+#include "pico/unique_id.h"
 
 #define USBD_VID (0x2E8A) // Raspberry Pi
 #define USBD_PID (0x000a) // Raspberry Pi Pico SDK CDC
@@ -98,10 +99,12 @@ static const uint8_t usbd_desc_cfg[USBD_DESC_LEN] = {
 #endif
 };
 
+static char usbd_serial_str[PICO_UNIQUE_BOARD_ID_SIZE_BYTES * 2 + 1];
+
 static const char *const usbd_desc_str[] = {
     [USBD_STR_MANUF] = "Raspberry Pi",
     [USBD_STR_PRODUCT] = "Pico",
-    [USBD_STR_SERIAL] = "000000000000", // TODO
+    [USBD_STR_SERIAL] = usbd_serial_str,
     [USBD_STR_CDC] = "Board CDC",
 #if PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE
     [USBD_STR_RPI_RESET] = "Reset",
@@ -120,6 +123,10 @@ const uint8_t *tud_descriptor_configuration_cb(uint8_t index) {
 const uint16_t *tud_descriptor_string_cb(uint8_t index, uint16_t langid) {
     #define DESC_STR_MAX (20)
     static uint16_t desc_str[DESC_STR_MAX];
+
+    // Assign the SN using the unique flash id
+    if (!usbd_serial_str[0])
+        pico_get_unique_board_id_string(usbd_serial_str, sizeof(usbd_serial_str));
 
     uint8_t len;
     if (index == 0) {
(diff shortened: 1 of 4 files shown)


src/rp2_common/pico_unique_id/unique_id.c
```diff
@@ -26,3 +26,16 @@ static void __attribute__((constructor)) _retrieve_unique_id_on_boot(void) {
...
 }
+
+void pico_get_unique_board_id_string(char *id_out, size_t len) {
+    pico_unique_board_id_t temp = retrieved_id;
+    size_t i;
+    // Generate hex one nibble at a time
+    for (i = 0; (i < len - 1) && (i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES * 2); i++) {
+        size_t bi = i / 2;
+        int nibble = (temp.id[bi] >> 4) & 0x0F;
+        temp.id[bi] = (uint8_t)(temp.id[bi] << 4);
```
lurch, line 37: Apologies if this is a stupid question (I'm not a low-level C programmer), but is this modifying the bytes in the underlying `retrieved_id` ?

earlephilhower: No, this works on the bytes in`temp` which are copied in line 31.

src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c
```diff
@@ -121,6 +124,10 @@ const uint16_t *tud_descriptor_string_cb(uint8_t index, uint16_t langid) {
     #define DESC_STR_MAX (20)
     static uint16_t desc_str[DESC_STR_MAX];
 
+    // Assign the SN using the unique flash id
+    if (!usbd_serial_str[0])
```
kilograham, line 128: can you please use braces here. I guess we should check in a style guide

src/rp2_common/pico_unique_id/unique_id.c
kilograham: this does feel a bit weird though, i'd be tempted to just do
```
nibble = (temp.id[i/2] >> (4 * (i&1))) & 0xf;
```

src/rp2_common/pico_unique_id/unique_id.c
```diff
@@ -26,3 +26,16 @@ static void __attribute__((constructor)) _retrieve_unique_id_on_boot(void) {
 void pico_get_unique_board_id(pico_unique_board_id_t *id_out) {
     *id_out = retrieved_id;
 }
+
+void pico_get_unique_board_id_string(char *id_out, size_t len) {
```
kilograham, line 30: also for better or worse we've tended to use `uint` rather than `size_t` for sizes in APIs

kilograham (changes requested):

src/rp2_common/pico_unique_id/unique_id.c
lurch: (I guess @kilograham 's suggestion doesn't modify `temp`, in which case perhaps there's no need for the copy and you can just access `retrieved_id`?)

earlephilhower: That's a much better way and it doesn't need the copy, thanks.

src/rp2_common/pico_unique_id/unique_id.c
earlephilhower: All done.  To preserve hex ordering, we actually need to shift `(4 - 4 * (i&1))` or the hex nibbles end up backwards.

src/rp2_common/pico_unique_id/unique_id.c
```diff
@@ -26,3 +26,14 @@ static void __attribute__((constructor)) _retrieve_unique_id_on_boot(void) {
 void pico_get_unique_board_id(pico_unique_board_id_t *id_out) {
     *id_out = retrieved_id;
 }
+
+void pico_get_unique_board_id_string(char *id_out, uint len) {
+    assert(len > 0);
```
lurch, line 31: Isn't a uint _always_ > 0 ? :wink:

kilograham (approved):

src/rp2_common/pico_unique_id/unique_id.c
lurch: It's a good thing that `*` has a higher precedence than `-` :wink:

PR #289, opened 2021-03-30
Don't advertise REMOTE_WAKEUP for USB stdio as it is is not currently…

liamfraser:
… implemented. Longer term need to implement REMOTE_WAEKUP properly in TinyUSB. Similar fix needed in MicroPython to fix MicroPython issue #6866. The symptom is that USB serial will stop working if there is a delay of > 2 seconds ish between messages. We have seen this issue on Linux laptops with USB autosuspend enabled.


lurch (approved):

kilograham:
pls rebase onto develop-1.1.1

liamfraser:
Rebased

kilograham (approved):

PR #293, opened 2021-03-31
build: allow board config header to specify bootstage 2 without using custom per board cmake

### LINKED ISSUE #275: Make boot_stage2 choice a function of the board header by default + make generic flash the default

kilograham:
also changes default if unspecified to boot2_generic_03h
fixes #275 
encompasses #279


src/rp2_common/boot_stage2/CMakeLists.txt
```diff
@@ -49,17 +66,22 @@ function(pico_define_boot_stage2 NAME SOURCES)
             COMMAND ${Python3_EXECUTABLE} ${PICO_BOOT_STAGE2_DIR}/pad_checksum -s 0xffffffff ${ORIGINAL_BIN} ${PADDED_CHECKSUMMED_ASM}
             )
 
-
     add_library(${NAME}_library INTERFACE)
     add_dependencies(${NAME}_library ${NAME}_padded_checksummed_asm)
     # not strictly (or indeed actually) a link library, but this avoids dependency cycle
     target_link_libraries(${NAME}_library INTERFACE ${PADDED_CHECKSUMMED_ASM})
+    target_link_libraries(${NAME}_library INTERFACE boot_stage2_headers)
```
kilograham, line 73: wahhh? how does this work? wasn't supposed to remove the PADDED_CHECKSUMMED_ASM!

Wren6991 (approved):

PR #296, opened 2021-04-01
make spi_init return baud rate set

kilograham:


liamfraser (approved):

lurch:
@kilograham I didn't review this when I first saw it, because I wanted to do full debug and release builds of pico-sdk and pico-examples, to see if this change of prototype introduced any warnings, and I'm afraid I was too busy to do that at the time.
But I've now done that, and no new warnings were seen :+1:

lurch (approved):

Reneg973:
Nice, I already changed this locally some weeks ago

Issue #306, opened 2021-04-05
The parameters check in spin_lock_get_num() from hardware/sync.h trips up -WError


Lyle-Alloy:
Here's my (trimmed) user code that caused the issue:
```
#include "pico/sync.h"

critical_section_t critical_section;

void a(){
    critical_section_init(&critical_section);
    critical_section_enter_blocking(&critical_section);
    critical_section_exit(&critical_section);
}

```

And the error:

```
In file included from /Applications/ARM/arm-none-eabi/include/c++/10.2.1/cassert:44,
                 from /Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/common/pico_base/include/pico/assert.h:14,
                 from /Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/common/pico_base/include/pico/types.h:10,
                 from /Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/common/pico_base/include/pico.h:16,
                 from /Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/common/pico_sync/include/pico/lock_core.h:10,
                 from /Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/common/pico_sync/include/pico/sem.h:10,
                 from /Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/common/pico_sync/include/pico/sync.h:15,
                 from /Users/work/Documents/Workspace/controls/Projects/application/src/user_file.h:10,
                 from /Users/work/Documents/Workspace/controls/Projects/application/src/user_file.cpp:6:
/Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/rp2_common/hardware_sync/include/hardware/sync.h: In function 'uint spin_lock_get_num(spin_lock_t*)':
/Users/work/Documents/Workspace/controls/Projects/application/cmake-build-debug/_deps/pico_sdk-src/src/rp2_common/hardware_sync/include/hardware/sync.h:208:54: error: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Werror=sign-compare]
[...]

PR #307, opened 2021-04-05
Added a typecast to int before doing the parameter check for spin locks

Lyle-Alloy:
This prevents a comparison between a signed and an unsigned number which
will create a warning tripping -Werror.

This is my fix to #306

### DIFF
--- a/src/rp2_common/hardware_sync/include/hardware/sync.h
+++ b/src/rp2_common/hardware_sync/include/hardware/sync.h
@@ -204,9 +204,10 @@ inline static spin_lock_t *spin_lock_instance(uint lock_num) {
  * \return The Spinlock ID
  */
 inline static uint spin_lock_get_num(spin_lock_t *lock) {
-    int lock_num = lock - (spin_lock_t *) (SIO_BASE + SIO_SPINLOCK0_OFFSET);
-    invalid_params_if(SYNC, lock_num < 0 || lock_num >= NUM_SPIN_LOCKS);
-    return (uint) lock_num;
+    invalid_params_if(SYNC, (uint) lock < SIO_BASE + SIO_SPINLOCK0_OFFSET ||
+                            (uint) lock >= NUM_SPIN_LOCKS * sizeof(spin_lock_t) + SIO_BASE + SIO_SPINLOCK0_OFFSET ||
+                            ((uint) lock - SIO_BASE + SIO_SPINLOCK0_OFFSET) % sizeof(spin_lock_t) != 0);
+    return (uint) (lock - (spin_lock_t *) (SIO_BASE + SIO_SPINLOCK0_OFFSET));
 }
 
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_sync/include/hardware/sync.h
```diff
@@ -205,7 +205,7 @@ inline static spin_lock_t *spin_lock_instance(uint lock_num) {
  */
 inline static uint spin_lock_get_num(spin_lock_t *lock) {
     int lock_num = lock - (spin_lock_t *) (SIO_BASE + SIO_SPINLOCK0_OFFSET);
-    invalid_params_if(SYNC, lock_num < 0 || lock_num >= NUM_SPIN_LOCKS);
+    invalid_params_if(SYNC, lock_num < 0 || lock_num >= (int) NUM_SPIN_LOCKS);
```
kilograham, line 208: would make more sense to make `lock_num` a `uint` from the start an skip the < 0 check.

kilograham (changes requested):

src/rp2_common/hardware_sync/include/hardware/sync.h
Lyle-Alloy: I think I actually like it how it is right now, because I think the intent was to check if lock was either pointing below the `spin_lock` block or pointing above it, which is reflected in the two conditionals.

Making `lock_num` a `uint` would achieve the same thing, but it would be because if it was below the `spin_lock` block `lock_num` would roll over on subtraction, which is kinda opaque to the casual reader in my opinion.

Anyways, you're the maintainer and thanks for the work! Don't want to add more mental load than needed so I'll just make the change.

lurch: I agree that we ought to try and avoid under/over flows! Perhaps it's worth adding a separate check for `lock >= (spin_lock_t *) (SIO_BASE + SIO_SPINLOCK0_OFFSET)` ? :man_shrugging:

Lyle-Alloy:
@kilograham @lurch How's this? I just moved all the checks into the `invalid_params_if` and made them very human readable.

All those constants should be evaluated at compile time and merged together so it shouldn't be much work at runtime.

I also added a check for alignment because that would be undefined behaviour as well!

kilograham:
cool thanks, yeah i noticed that our `-Werror` checks aren't run with `PARAM_ASSERTIONS_ENABLE_ALL=1` so I will fix the others

kilograham:
can you please rebase onto develop

Lyle-Alloy:
@kilograham should be all set

kilograham:
thank you... sorry to mess you around, but do you mind rebasing onto develop-1.1.2 now; i want to squeeze it in there actually!

Lyle-Alloy:
Haha you're making me remember my `git rebase`-foo. Should be all set now

kilograham (approved):

PR #311, opened 2021-04-06
Fix param-validation for PIO sideset encoding

lurch:
Same change as #282 , but re-opened against the `develop` branch.

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
@@ -80,19 +80,20 @@ inline static uint _pio_encode_instr_and_src_dest(enum pio_instr_bits instr_bits
 }
 
 inline static uint pio_encode_delay(uint cycles) {
+    // note that the maximum cycles will be smaller if sideset_bit_count > 0
     valid_params_if(PIO_INSTRUCTIONS, cycles <= 0x1f);
     return cycles << 8u;
 }
 
 inline static uint pio_encode_sideset(uint sideset_bit_count, uint value) {
     valid_params_if(PIO_INSTRUCTIONS, sideset_bit_count >= 1 && sideset_bit_count <= 5);
-    valid_params_if(PIO_INSTRUCTIONS, value <= (0x1fu >> sideset_bit_count));
+    valid_params_if(PIO_INSTRUCTIONS, value <= ((1u << sideset_bit_count) - 1));
     return value << (13u - sideset_bit_count);
 }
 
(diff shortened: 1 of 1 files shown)


kilograham (commented):
This one has been lurking because we have a bit of confusion regarding number of side set bits;

The RP2040 data sheet refers to `SIDESET_COUNT` which is the number of bits set aside from the delay field for _side set_, whereas in `pioasm` we refer to `.side_set N` or `side_set N opt` in which cases the `SIDESET_COUNT` is 3 or 4 respectively. We think of N as the number of bits you want to side set with

Documenting these functions would help - I think the only other place you'd notice is in the implementation of pioasm, but I think perhaps we need a better name for `sideset_bit_count` here

lurch:
Just FYI I only wrote this patch by cross-referencing `pio_instructions.h` against what the RP2040 datasheet says, I didn't look at any of the pioasm stuff. But of course that's still no guarantee that I've not made a mistake in my understanding!

Any chance you could take a look @Wren6991 ?

Wren6991:
To add to the confusion, `pio.h` uses the same convention as the hardware, i.e. inclusive of the `EN` bit, which is different from `pio_encode_sideset_opt()`. Bit of a shame, too late to fix this now. It's a good thing the `pio.h` function is usually called from code generated by `pioasm`, which uses the exclusive count convention

```
/*! \brief Set the 'sideset' options in a state machine configuration
 *  \ingroup sm_config
 *
 * \param c Pointer to the configuration structure to modify
 * \param bit_count Number of bits to steal from delay field in the instruction for use of side set (max 5)
 * \param optional True if the topmost side set bit is used as a flag for whether to apply side set on that instruction
 * \param pindirs True if the side set affects pin directions rather than values
 */
static inline void sm_config_set_sideset(pio_sm_config *c, uint bit_count, bool optional, bool pindirs) {
    valid_params_if(PIO, bit_count <= 5);
    valid_params_if(PIO, !optional || bit_count >= 1);
    c->pinctrl = (c->pinctrl & ~PIO_SM0_PINCTRL_SIDESET_COUNT_BITS) |
                 (bit_count << PIO_SM0_PINCTRL_SIDESET_COUNT_LSB);
```

Think we should merge this patch as is and then go through and add doxygen to the encode functions, making sure the inconsistency is highlighted.

Wren6991 (approved):

lurch:
Ahhh, I hadn't noticed that inconsistency because I was _only_ looking at `pio_instructions.h` and not `pio.h`. But yes that does explain the confusion!

PR #312, opened 2021-04-06
boot_stage2: cleanup and correctness fixes

kilograham:


lurch (approved):
Only a very small change from #310 , nice :+1:

PR #314, opened 2021-04-06
i2c: set high and low times to values that conform to the i2c specification

### LINKED ISSUE #305: i2c: values for tlow lower than the minimum values specified in the I2C specification
Per table 10 on page 48 of the [I<sup>2</sup>C-bus specification and user manual](https://www.nxp.com/docs/en/user-guide/UM10204.pdf) the minimum values for t<sub>LOW</sub> and t<sub>HIGH</sub> are as follows:

Symbol | Parameter | Standard-mode | Fast-mode | Fast-mode Plus | Unit
:---: | :---: | :---: | :---: | :---: | :---
t<sub>LOW</sub> | LOW period of the SCL clock | 4.7 | 1.3 | 0.5 | μs
t<sub>HIGH</sub> | HIGH period of the SCL clock | 4.0 | 0.6 | 0.26 | μs

With pico-sdk v1.1.0, the actual values used for t<sub>LOW</sub> are lower than the minimum values specified in the specification (<sup>1</sup>):

Symbol | Parameter | Standard-mode | Fast-mode | Fast-mode Plus | Unit
:---: | :---: | :---: | :---: | :---: | :---
t<sub>LOW</sub> | LOW period of the SCL clock | 4.0<sup>1</sup> | 1.008<sup>1</sup> | 0.4<sup>1</sup> | μs
t<sub>HIGH</sub> | HIGH period of the SCL clock | 6.0 | 1.496 | 0.6 | μs

The corresponding values written to the `IC_FS_SCL_LCNT` and `IC_FS_SCL_HCNT` registers are:

Symbol | Register | Standard-mode | Fast-mode | Fast-mode Plus
:---: | :---: | :---: | :---: | :---:
lcnt | IC_FS_SCL_LCNT | 500 | 126 | 50
hcnt | IC_FS_SCL_HCNT | 750 | 187 | 75

This isn't resulting in any issues with the I<sup>2</sup>C devices that I'm using with pico-sdk but it would probably be worth changing. Perhaps the values written to `lcnt` should have been written to `hcnt` and vice versa? See [here](https://github.com/raspberrypi/pico-sdk/blob/1.1.0/src/rp2_common/hardware_i2c/i2c.c#L69-L71).

fivdi:
Fixes #305.

### DIFF
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -67,8 +67,8 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
 
     // TODO there are some subtleties to I2C timing which we are completely ignoring here
     uint period = (freq_in + baudrate / 2) / baudrate;
-    uint hcnt = period * 3 / 5; // oof this one hurts
-    uint lcnt = period - hcnt;
+    uint lcnt = period * 3 / 5; // oof this one hurts
+    uint hcnt = period - lcnt;
     // Check for out-of-range divisors:
     invalid_params_if(I2C, hcnt > I2C_IC_FS_SCL_HCNT_IC_FS_SCL_HCNT_BITS);
     invalid_params_if(I2C, lcnt > I2C_IC_FS_SCL_LCNT_IC_FS_SCL_LCNT_BITS);


src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -67,8 +67,8 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
 
     // TODO there are some subtleties to I2C timing which we are completely ignoring here
     uint period = (freq_in + baudrate / 2) / baudrate;
```
dhalbert, line 69: Thanks for this. Do you also understand why `period = (freq_in + baudrate / 2) / baudrate;` rather than simply `period = freq_in / baudrate;` ? I was puzzled by this.

dhalbert: maybe it's just rounding.

fivdi: Yes, it's rounding for int/uint: (dividend + divisor / 2) / divisor

Wren6991 (approved):
Nice clean fix, thank you

fivdi:
@Wren6991 Thank you for merging. Does this PR need to be added to a milestone so that it ends up in a release at some point? I was just about to ask the same question about [this PR](https://github.com/raspberrypi/pico-sdk/pull/273).

lurch:
@fivdi I'm not a mind-reader, but my assumption is that @kilograham 's workflow is that anything merged into `develop` will get included in the next "big release" (which would be a minor release in semver terms e.g. 1.X.0) and it's only specifically small bugfixes that get selectively pulled into the next "bugfix release" (which would be a patch release in semver terms e.g. 1.1.Y).
I think / hope that makes sense?

fivdi:
@lurch yes, that makes sense, thanks.

kilograham:
yes, although it should have the milestone so it is easier to see what is happening - i have addedd

Issue #161, opened 2021-02-19
bump TinyUSB to latest upstream release


kilograham:
We will at least try to get this in for 1.0.2 unless there are issues in which case we'll move to 1.0.3

lurch:
Links to https://github.com/raspberrypi/pico-sdk/issues/85#issuecomment-778049788

kilograham:
we have decided not to do this for 1.0.2 release, because we will need some upstream changes

lurch:
Links to #73 also

kilograham:
Fixed by #462

kilograham:
merged

Issue #85, opened 2021-02-04
Usb assertion "hw_data_offset(next_buffer_ptr) <= USB_DPRAM_MAX"


ndinsmore:
This is a duplicate of the issue I created at please see the issue there for sull documentation:
hathach/tinyusb/issues/628

In the issue I created in the tinyusb repo I mention that I am using the new code coming directly from that repo, but the code at that is at issue is in the SDK tiny USB repo as well.

lurch:
Sounds like something @liamfraser needs to look at?

ndinsmore:
This issue is fixed by hathach/tinyusb#642 @liamfraser it is one real line of code so it should be easy for you to take a look at.

liamfraser:
Looks like a plausible fix to me. I think the audio example may be the first one that repeatedly opens and closes an ep which is why we didn't see this?

liamfraser:
(I don't own the upstream tinyusb repo so I can't merge the PR)

hathach:
@liamfraser please accept my contributor invitation, and feel free to review, approval and merge any PRs that make sense to you. I still haven't got into the usb specs of rp2040 yet. Worse I am on vacation just now :wink:

liamfraser:
@ndinsmore I have merged this into tinyusb upstream

liamfraser:
Will keep this open until either a) our sdk tracks upstream or b) we pick this commit into our fork. We need to review what changes have been made in tinyusb upstream before switching the SDK to it.

kilograham:
closing in favor of #161

kilograham:
changed my mind, will get closed with #321

kilograham:
actually closed by #462

kilograham:
released

PR #321, opened 2021-04-08
Move to tinyusb 0.9.0

kilograham:
tinyusb: add cmake functions to add example from tinyusb itself; fixup for sdtio_usb for tinyusb API changes

### DIFF
--- a/src/rp2_common/hardware_uart/include/hardware/uart.h
+++ b/src/rp2_common/hardware_uart/include/hardware/uart.h
@@ -101,6 +101,12 @@ static inline uint uart_get_index(uart_inst_t *uart) {
     return uart == uart1 ? 1 : 0;
 }
 
+static inline uart_inst_t *uart_get_instance(uint instance) {
+    static_assert(NUM_UARTS == 2, "");
+    invalid_params_if(UART, instance >= NUM_UARTS);
+    return instance ? uart1 : uart0;
+}
+
 static inline uart_hw_t *uart_get_hw(uart_inst_t *uart) {
     uart_get_index(uart); // check it is a hw uart
     return (uart_hw_t *)uart;
--- a/src/rp2_common/pico_stdio_usb/reset_interface.c
+++ b/src/rp2_common/pico_stdio_usb/reset_interface.c
@@ -38,7 +38,10 @@ static uint16_t resetd_open(uint8_t __unused rhport, tusb_desc_interface_t const
 }
 
 // Support for parameterized reset via vendor interface control request
-static bool resetd_control_request_cb(uint8_t __unused rhport, tusb_control_request_t const *request) {
+static bool resetd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t const * request) {
+    // nothing to do with DATA & ACK stage
+    if (stage != CONTROL_STAGE_SETUP) return true;
+
     if (request->wIndex == itf_num) {
 
 #if PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_RESET_TO_BOOTSEL
@@ -69,10 +72,6 @@ static bool resetd_control_request_cb(uint8_t __unused rhport, tusb_control_requ
     return false;
 }
 
-static bool resetd_control_complete_cb(uint8_t __unused rhport, tusb_control_request_t __unused const *request) {
-    return true;
-}
-
 static bool resetd_xfer_cb(uint8_t __unused rhport, uint8_t __unused ep_addr, xfer_result_t __unused result, uint32_t __unused xferred_bytes) {
     return true;
 }
@@ -85,8 +84,7 @@ static usbd_class_driver_t const _resetd_driver =
     .init             = resetd_init,
     .reset            = resetd_reset,
     .open             = resetd_open,
-    .control_request  = resetd_control_request_cb,
-    .control_complete = resetd_control_complete_cb,
+    .control_xfer_cb  = resetd_control_xfer_cb,
     .xfer_cb          = resetd_xfer_cb,
     .sof              = NULL
 };
--- a/src/rp2_common/tinyusb/CMakeLists.txt
+++ b/src/rp2_common/tinyusb/CMakeLists.txt
@@ -104,8 +104,46 @@ if (EXISTS ${PICO_TINYUSB_PATH}/${TINYUSB_TEST_PATH})
 
     add_library(tinyusb_board INTERFACE)
     target_sources(tinyusb_board INTERFACE
-            ${PICO_TINYUSB_PATH}/hw/bsp/raspberry_pi_pico/board_raspberry_pi_pico.c
+            ${PICO_TINYUSB_PATH}/hw/bsp/rp2040/family.c
     )
+    target_include_directories(tinyusb_board INTERFACE ${PICO_TINYUSB_PATH}/hw/bsp/rp2040/boards/sdk_selected)
 
+    set(PICO_TINYUSB_PATH_INTERNAL ${PICO_TINYUSB_PATH} CACHE INTERNAL "")
+    add_library(tinyusb_example_support INTERFACE)
+    target_compile_definitions(tinyusb_example_support INTERFACE
+            CFG_TUSB_OS=OPT_OS_PICO
+            BOARD=pants
+    )
+    target_link_libraries(tinyusb_example_support INTERFACE tinyusb_board)
 endif()
+
+# SHORT_TYPE is 'dev' or 'host'; LONG_TYPE is 'device' or 'host'
+macro(internal_add_tinyusb_example SHORT_TYPE LONG_TYPE NAME )
+    if (TARGET tinyusb_${LONG_TYPE})
+        set(TARGET_NAME "usb_${SHORT_TYPE}_${NAME}")
+        # todo we ought to be able to include the examples, but we can't because they are very standalone
+        # todo maybe allow passing file names if necessary - doesn't seem to be right now
+        set(EXAMPLE_SOURCE_PATH ${PICO_TINYUSB_PATH_INTERNAL}/examples/${LONG_TYPE}/${NAME}/src)
+        file(GLOB EXAMPLE_SOURCE ${EXAMPLE_SOURCE_PATH}/*.c)
+        add_executable(${TARGET_NAME}
+                ${EXAMPLE_SOURCE}
+        )
+
+        target_include_directories(${TARGET_NAME} PRIVATE ${EXAMPLE_SOURCE_PATH})
+        target_link_libraries(${TARGET_NAME} PRIVATE pico_stdlib tinyusb_${LONG_TYPE} tinyusb_example_support)
+        pico_add_extra_outputs(${TARGET_NAME})
+
+    # add url via pico_set_program_url
+    # todo lets point at the right place
+    #example_auto_set_url(${NAME})
+    endif()
+endmacro()
+
+function(add_tinyusb_device_example NAME)
+    internal_add_tinyusb_example(dev device ${NAME})
+endfunction()
+
+function(add_tinyusb_host_example NAME)
+    internal_add_tinyusb_example(host host ${NAME})
+endfunction()
 


kilograham:
Note right now, this requires a patched tinyusb 0.9.0 which is in the raspberry pi repo here

https://github.com/raspberrypi/tinyusb/tree/pico-0.9.0-wip

This fixes some bugs in the family implementation, and provides a somewhat sub-optimal workaround with some impedence mismatches with the tinyusb examples proper

1. Use their own board abstraction distinct from that of the SDK
2. Use cmake, but as standalone apps with the SDK as a depedency rather than as apps that use TinyUSB via the tinyusb wrapper libraries in the SDK.

Both of these probably require some rationalization

kilograham:
fixes #85 
fixes #161

src/rp2_common/pico_stdio_usb/reset_interface.c
```diff
@@ -38,7 +38,10 @@ static uint16_t resetd_open(uint8_t __unused rhport, tusb_desc_interface_t const
 }
 
 // Support for parameterized reset via vendor interface control request
-static bool resetd_control_request_cb(uint8_t __unused rhport, tusb_control_request_t const *request) {
+static bool resetd_control_xfer_cb(uint8_t rhport, uint8_t stage, tusb_control_request_t const * request) {
+    // nothing to do with DATA & ACK stage
+    if (stage != CONTROL_STAGE_SETUP) return true;
```
hathach, line 43: Since you are reset mcu, it should be better to reset when `stage == ACK`  (ACK is complete), this will prevent host OS to complain about the failure of control transfer. For example,  touch1200 is common way for Arduino to do reset to DFU. 

https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.c#L354
```C
    case CDC_REQUEST_SET_LINE_CODING:
      if (stage == CONTROL_STAGE_SETUP)
      {
        TU_LOG2("  Set Line Coding\r\n");
        tud_control_xfer(rhport, request, &p_cdc->line_coding, sizeof(cdc_line_coding_t));
      }
      else if ( stage == CONTROL_STAGE_ACK)
      {
        if ( tud_cdc_line_coding_cb ) tud_cdc_line_coding_cb(itf, &p_cdc->line_coding);
      }
    break;
```

src/rp2_common/hardware_uart/include/hardware/uart.h
```diff
@@ -101,6 +101,12 @@ static inline uint uart_get_index(uart_inst_t *uart) {
     return uart == uart1 ? 1 : 0;
 }
 
+static inline uart_inst_t *uart_get_instance(uint instance) {
```
lurch, line 104: Is this change needed by this PR, or was it just pulled in coincidentally?

src/rp2_common/tinyusb/CMakeLists.txt
```diff
@@ -104,8 +104,46 @@ if (EXISTS ${PICO_TINYUSB_PATH}/${TINYUSB_TEST_PATH})
...
-            ${PICO_TINYUSB_PATH}/hw/bsp/raspberry_pi_pico/board_raspberry_pi_pico.c
+            ${PICO_TINYUSB_PATH}/hw/bsp/rp2040/family.c
     )
+    target_include_directories(tinyusb_board INTERFACE ${PICO_TINYUSB_PATH}/hw/bsp/rp2040/boards/sdk_selected)
 
+    set(PICO_TINYUSB_PATH_INTERNAL ${PICO_TINYUSB_PATH} CACHE INTERNAL "")
+    add_library(tinyusb_example_support INTERFACE)
+    target_compile_definitions(tinyusb_example_support INTERFACE
+            CFG_TUSB_OS=OPT_OS_PICO
+            BOARD=pants
```
lurch, line 115: :jeans:

kilograham: oops - i was checking something by deliberately using an invalid board ;-) that ain't gonna compile

src/rp2_common/pico_stdio_usb/reset_interface.c
kilograham: ah; good point thanks

kilograham:
@hathach wanted to have a discussion somewhere, and this is probably as good a place as any.

It might be nice to unify the tinyusb and sdk notions of a board, and possible share CMakeLists.txt between the two. I don't have a good mental picture of what the intentions are in TinyUSB to help me decide how to proceed. Here's some random points/questions.

1) Part of the confusion comes from the fact that TinyUSB has TinyUSB->pico-sdk and we have pico-sdk->TinyUSB ;-)
2) I assume/believe that part of the reason for the common FAMILY/BOARD setup within the examples is for CI and automated builds. Do people also expect to be able to build run the examples from within the TinyUSB tree (note there were some bugs in the new family.c which suggest they might not have). Is it a requirement of CI to be able to specify BOARD explicitly here rather than via PICO_BOARD when using the SDK
3) I'm a bit confused by the CMakeLists.txt in the examples. They are IFed based on the FAMILY, but the esp32 (or whatever I'm not looking atm) code paths in the CMakeLists.txt don't actually add any targets or set any sources. The SDK based GLOBing for examples works for now (to create our own build targets within pico-examples later), but if the example CMakeLists.txt provided some information on source files in a common way, maybe we could utilize those in some way. The way this was(/would have been :-) ) handled across multiple families would have helped here.

hathach:
> @hathach I wanted to have a discussion somewhere, and this is probably as good a place as any.
> 
> It might be nice to unify the tinyusb and sdk notions of a board, and possible share CMakeLists.txt between the two. I don't have a good mental picture of what the intentions are in TinyUSB to help me decide how to proceed. Here's some random points/questions.

yeah, it would be nice to have tinyusb example to work out of the box with pico-sdk.

> 1. Part of the confusion comes from the fact that TinyUSB has TinyUSB->pico-sdk and we have pico-sdk->TinyUSB ;-)

Yeah, I feel the pain as well, like all other mcus, the sdk/low level driver (clock, gpio ) are needed only for running tinyusb stock examples and mostly for me to develop and test out new feature (basically maintaining the repo). It is actually not needed when tinyusb is part of another project or in this case pico-sdk. I am not sure what is best to ease this out. Typically it shouldn't be an issue if user don't submodule init with `--recursive`. For this very reason, I have recently created an source-only mirror https://github.com/hathach/tinyusb_src that can be used to be included as part of comprehensive project. The src is exactly the same to the tinyusb/src, it is pushed automatically by ci whenever there is changes within the src folder of tinyusb's master branch https://github.com/hathach/tinyusb/blob/master/.github/workflows/trigger.yml#L45. Maybe it is a good alternative.

> 2. I assume/believe that part of the reason for the common FAMILY/BOARD setup within the examples is for CI and automated builds. Do people also expect to be able to build run the examples from within the TinyUSB tree (note there were some bugs in the new family.c which suggest they might not have). Is it a requirement of CI to be able to specify BOARD explicitly here rather than via PICO_BOARD when using the SDK

The `family/board` is way to ease adding new board within a same mcu family. `make BOARD=` is actually what I used to develop and maintain the repo. And yes, it is what user expect to run examples to testing out the stack. TinyUSB only use a thin board api that only use 1 LED, 1 Button and Uart (and tick timer mostly for blinky). https://github.com/hathach/tinyusb/blob/master/docs/boards.md  https://github.com/hathach/tinyusb/blob/master/docs/getting_started.md#examples . In general, I will also ask user to try out this way when opening an issue. 

> 3. I'm a bit confused by the CMakeLists.txt in the examples. They are IFed based on the FAMILY, but the esp32 (or whatever I'm not looking atm) code paths in the CMakeLists.txt don't actually add any targets or set any sources. The SDK based GLOBing for examples works for now (to create our own build targets within pico-examples later), but if the example CMakeLists.txt provided some information on source files in a common way, maybe we could utilize those in some way. The way this was(/would have been :-) ) handled across multiple families would have helped here.

esp32s2 requires to run with FreeRTOS config, you would see it in the _freertos example https://github.com/hathach/tinyusb/blob/master/examples/device/cdc_msc_freertos/CMakeLists.txt#L13 . Actually, I don't use CMakeLists.txt as top level, I typically type `make BOARD=pico`  which include the [make.mk
](https://github.com/hathach/tinyusb/blob/master/examples/make.mk) which will include and run the `all` target by [family.mk](https://github.com/hathach/tinyusb/blob/master/hw/bsp/rp2040/family.mk#L11). It is a bit complicated, but I did to keep it consistent with all other mcu ports. 

Actually, I am not too familiar with CMake, and I only try enough to get it running with existing `make BOARD=`( have no time to play with build system :D ). If you have any suggestion or better an PR, I am more than happy to collaborate with.

kilograham:
yup, for the initial SDK release we copied some of the examples into the sdk because we weren't sure about changing some settings (like `CFG_TUSB_OS` being unconditionally set to `OPT_OS_NONE` in the the configs - which have now mostly been changed upstream to be guarded by `#ifndef`)

This https://github.com/raspberrypi/pico-examples/pull/99 PR in in pico-examples changes us to use a curated list of your examples, but by globbing the the source files using the cmake helper function in this PR.

Given that you already have some support for saying which examples belong with which families, I am hoping that maybe long term we could just import a CMakeLists.txt from your example root into our pico-examples tree to add all your current examples so there is non manual work. This should be possible, but I wanted to understand the current usage of the CMakeLists.txt in tinyusb examples first because figuring out how best to refactor them.

Note also there is also this https://github.com/raspberrypi/tinyusb/tree/pico-0.9.0-wip branch, which has a few tweaks, and allows for specifying a BOARD on the TinyUSB side which just uses the PICO_BOARD settings from the pico-sdk

lurch:
> Part of the confusion comes from the fact that TinyUSB has TinyUSB->pico-sdk and we have pico-sdk->TinyUSB ;-)

And there's also MicroPython which [submodules](https://github.com/micropython/micropython/blob/master/.gitmodules) both `lib/tinyusb` and `lib/pico-sdk` ;-)

hathach:
> yup, for the initial SDK release we copied some of the examples into the sdk because we weren't sure about changing some settings (like `CFG_TUSB_OS` being unconditionally set to `OPT_OS_NONE` in the the configs - which have now mostly been changed upstream to be guarded by `#ifndef`)
> 
> This [raspberrypi/pico-examples#99](https://github.com/raspberrypi/pico-examples/pull/99) PR in in pico-examples changes us to use a curated list of your examples, but by globbing the the source files using the cmake helper function in this PR.

Ah yeah, naturally it is default no OS configuration previously. I made some changes when adding rp2040 support. There maybe a few left, feel free to make PR to fix those.  

> Given that you already have some support for saying which examples belong with which families, I am hoping that maybe long term we could just import a CMakeLists.txt from your example root into our pico-examples tree to add all your current examples so there is non manual work. This should be possible, but I wanted to understand the current usage of the CMakeLists.txt in tinyusb examples first because figuring out how best to refactor them.

Yeah, ready to use CMakeLists.txt from tinyusb example for pico-sdk is what I really want as well. That will help people testing out examples easier. I will try to tackle the build system later on when having more time.
 
> Note also there is also this https://github.com/raspberrypi/tinyusb/tree/pico-0.9.0-wip branch, which has a few tweaks, and allows for specifying a BOARD on the TinyUSB side which just uses the PICO_BOARD settings from the pico-sdk

Thanks for the link, I will check it out next time I rework the cmake build for rp2040.

hathach:
> > Part of the confusion comes from the fact that TinyUSB has TinyUSB->pico-sdk and we have pico-sdk->TinyUSB ;-)
> 
> And there's also MicroPython which [submodules](https://github.com/micropython/micropython/blob/master/.gitmodules) both `lib/tinyusb` and `lib/pico-sdk` ;-)

Yeah, lots of project depends on each others. pico-sdk and tinyusb is kind of weird since they include each other, therefore submodule init with recursive flag shouldn't be run :sweat_smile: . Maybe I will drop the pico-sdk submodule in tinyusb, and have an instruction to tell people to install pico-sdk themself. After all, tinyusb is supposed to be included as submodules by others. Will do it later on.

hathach:
update I have removed pico-sdk as part of tinyusb submodules, it would be much better for rp2040 user. They definitely need to pico-sdk somewhere already. https://github.com/hathach/tinyusb/pull/782

src/rp2_common/hardware_uart/include/hardware/uart.h
kilograham: fixed

src/rp2_common/tinyusb/CMakeLists.txt
kilograham: fixed

kilograham:
obsoleted by new #462

Issue #284, opened 2021-03-26
Unable to disable FIFO_IRQ_PROC0


alexandruradovici:
I am trying to port [Tock](https://github.com/tock/tock) to the Raspberry Pi Pico and I am facing a strange issue regarding the FIFO interrupt.

As Tock is using only one core, I do not need the FIFO. My understanding is that the second core is stopped by the bootloader. The problems that I am facing are the following:
  1.  the FIFO_IRQ_PROC) (15) interrupt keeps firing continuously, I tried disabling it from the NVIC, but it seems to ignore it
  2.  when trying to to write to the NVIC::ISER register the value 1<<15 it hard faults, any other value works

I am using `openocd` via a Raspberry Pi to load and debug the software. I used the same init steps as in the SDK.

I think I might be missing something, but I can't figure out what. Any advice would be great.

lurch:
I've not tried this myself, but from a quick look at the datasheet it seems you need to write to NVIC ICER to disable the interrupt? (ISER is the enable register; although I dunno why re-enabling an interrupt should cause a hard fault?)

alexandruradovici:
> I've not tried this myself, but from a quick look at the datasheet it seems you need to write to NVIC ICER to disable the interrupt? (ISER is the enable register; although I dunno why re-enabling an interrupt should cause a hard fault?)
> 
I did try disabling it by writing the ICER register. I cleared the pending register, wrote 1 << 15 to ICER and then read the pending register again. It was still signaling the interrupt.

> 
> 
> EDIT: And section 2.3.1.4. of the datasheet says you can read FIFO_ST to see the reason the interrupt was fired.

I did this, the register returns 0b11, meaging that the RDY and VLD are set. Reading the FIFO_RD register still does not clear the interrupt. I tried reding it 100 times, once every time the interrupt fired, and always got 0.

lurch:
> I did try disabling it by writing the ICER register.

Sorry, from your original message it sounded like you were trying to disable the interrupt by writing to NVIC::ISER, apologies if I misunderstood.

Well I'm afraid that's as far as my knowledge goes, so you'll have to wait for @kilograham or @Wren6991 to provide further assistance :slightly_smiling_face:

kilograham:
You need to clear the status register by writing say 0xF to the register; this will clear the IRQ.

[...]

PR #323, opened 2021-04-09
Clarify multicore_fifo doxygen

lurch:
Based on my observations in #284

### DIFF
--- a/src/rp2_common/pico_multicore/include/pico/multicore.h
+++ b/src/rp2_common/pico_multicore/include/pico/multicore.h
@@ -119,7 +119,7 @@ uint32_t multicore_fifo_pop_blocking(void);
 
 bool multicore_fifo_pop_timeout_us(uint64_t timeout_us, uint32_t *out);
 
-/*! \brief Flush any data in the outgoing FIFO
+/*! \brief Flush any data in the incoming FIFO
  *  \ingroup multicore_fifo
  *
  */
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_multicore/include/pico/multicore.h
```diff
@@ -130,9 +130,12 @@ static inline void multicore_fifo_drain(void) {
 
 /*! \brief Clear FIFO interrupt
  *  \ingroup multicore_fifo
+ *
+ * Note that this only clears an interrupt that was caused by the ROE or WOF flags.
+ * To clear the VLD flag you need to use one of the 'pop' or 'drain' functions.
 */
```
kilograham, line 136: yeah; i guess we should rename this function as it doesn't really do what it says; perhaps `multicore_clear_sticky_errors`

Note we need a little documentation update here (this file) anyway to discourage use of the FIFOs

lurch: > i guess we should rename this function

That would be a backwards-incompatible change, which according to semver conventions would necessitate a version-bump to 2.0.0

> Note we need a little documentation update here (this file) anyway to discourage use of the FIFOs

Sounds like a Graham-job to me! :wink:  I know very little about the FIFOs (and even less about multicore in general).

kilograham (approved):

PR #324, opened 2021-04-09
Fix implementation config listing in structs/i2c.h

Wren6991:


kilograham (approved):

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
```diff
@@ -62,80 +62,73 @@ typedef struct {
...
 // IC_UFM_SCL_HIGH_COUNT ............. 0x0006
 // IC_TX_TL .......................... 0x0
-// IC_STOP_DET_IF_MASTER_ACTIVE ...... 0x0
-// IC_SS_SCL_LOW_COUNT ............... 0x01d6
-// IC_HAS_DMA ........................ 0x1
-// IC_RX_FULL_GEN_NACK ............... 0x0
-// IC_CLOCK_PERIOD ................... 100
-// IC_EMPTYFIFO_HOLD_MASTER_EN ....... 1
+// IC_TX_CMD_BLOCK ................... 0x1
+// IC_HAS_DMA ........................ 0x0
```
fivdi, line 71: If IC_HAS_DMA is 0x0, does this mean that the RP2040 I2C peripherals don't support DMA?
If so, this would mean that section 4.3.15. of the RP2040 Datasheet which says that the I2C peripherals have a DMA controller interface is incorrect.

Wren6991: ...I see what happened, good catch thank you, that field should be 0x1, all the others match

```
8c8
< #define IC_HAS_DMA 0x0
---
> #define IC_HAS_DMA 0x1
```

Missing automation link here as you can see!

Wren6991: (and yes the hardware certainly does support the DMA handshake)

PR #331, opened 2021-04-11
i2c: improve communication with i2c devices in i2c_write_blocking

### LINKED ISSUE #278: i2c: i2c_write_blocking not returning PICO_ERROR_GENERIC when expected
The program below is a variant of [bus_scan](https://github.com/raspberrypi/pico-examples/blob/sdk-1.1.0/i2c/bus_scan/bus_scan.c) from pico-examples. It calls ```i2c_write_blocking``` rather than ```i2c_read_blocking```.

```c
#include <stdio.h>
#include "pico/stdlib.h"
#include "hardware/i2c.h"

static bool reserved_addr(uint8_t addr) {
    return (addr & 0x78) == 0 || (addr & 0x78) == 0x78;
}

int main() {
    stdio_init_all();

    i2c_init(i2c_default, 100 * 1000);
    gpio_set_function(PICO_DEFAULT_I2C_SDA_PIN, GPIO_FUNC_I2C);
    gpio_set_function(PICO_DEFAULT_I2C_SCL_PIN, GPIO_FUNC_I2C);
    gpio_pull_up(PICO_DEFAULT_I2C_SDA_PIN);
    gpio_pull_up(PICO_DEFAULT_I2C_SCL_PIN);

    printf("\ni2c_write_blocking test\n");
    printf("   0  1  2  3  4  5  6  7  8  9  A  B  C  D  E  F\n");

    for (int addr = 0; addr < (1 << 7); ++addr) {
        if (addr % 16 == 0) {
            printf("%02x ", addr);
        }

        int ret;
        uint8_t txdata = 0;
        if (reserved_addr(addr)) {
            ret = PICO_ERROR_GENERIC;
        } else {
            ret = i2c_write_blocking(i2c_default, addr, &txdata, 1, false);
            if (ret != 1) {
               printf("never printed as ret is always 1\n");
            }
        }

        printf(ret < 0 ? "." : "@");
        printf(addr % 16 == 15 ? "\n" : "  ");
    }
}
```

Although there is only one device on the I2C bus, an MCP9808 temperature sensor at address 0x18, the output of this program is:

```
i2c_write_blocking test
   0  1  2  3  4  5  6  7  8  9  A  B  C  D  E  F
00 .  .  .  .  .  .  .  .  @  @  @  @  @  @  @  @
10 @  @  @  @  @  @  @  @  @  @  @  @  @  @  @  @
20 @  @  @  @  @  @  @  @  @  @  @  @  @  @  @  @
30 @  @  @  @  @  @  @  @  @  @  @  @  @  @  @  @
40 @  @  @  @  @  @  @  @  @  @  @  @  @  @  @  @
50 @  @  @  @  @  @  @  @  @  @  @  @  @  @  @  @
60 @  @  @  @  @  @  @  @  @  @  @  @  @  @  @  @
70 @  @  @  @  @  @  @  @  .  .  .  .  .  .  .  .
```

```i2c_write_blocking``` returns 1 every time it's called although it should only return 1 when called for `addr` 0x18. ```i2c_write_blocking``` should return `PICO_ERROR_GENERIC` every time it's called except when called for `addr` 0x18.

**Edit**: Tested with pico-sdk v1.1.0

### PR DESCRIPTION
**Pull request by fivdi**
This PR should fix #278.

For testing, the following three test programs were used:

- [bus_scan_issue_278](https://github.com/fivdi/rp2040-tests/tree/master/i2c/bus_scan_issue_278)
- [scan_at_many_baudrates](https://github.com/fivdi/rp2040-tests/tree/master/i2c/scan_at_many_baudrates)
- [read_and_write_devices](https://github.com/fivdi/rp2040-tests/tree/master/i2c/read_and_write_devices)

Each test program has a readme with a short explanation of their purpose.

~~(While working on this issue I noticed a problem with baudrate 1144. All other baud rates from 1000 to 2048000 look good but 1144 doesn't work. I'm not sure why but I'll create a separate issue for that.)~~

**Edit:** There isn't actually an issue with baudrate 1144. I went a step too far with underclocking. The lowest baudrate supported is 1145. At baudrates lower than 1145, the value of `lcnt` is greater than 2^16-1 (65535) and no longer fits in the 16-bit IC_FS_SCL_LCNT field of the IC_FS_SCL_LCNT register. Should there be `asserts` in `i2c_set_baudrate` to check if `lcnt` or `hcnt` is > 65535?

### DIFF
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -43,7 +43,8 @@ uint i2c_init(i2c_inst_t *i2c, uint baudrate) {
             I2C_IC_CON_SPEED_VALUE_FAST << I2C_IC_CON_SPEED_LSB |
             I2C_IC_CON_MASTER_MODE_BITS |
             I2C_IC_CON_IC_SLAVE_DISABLE_BITS |
-            I2C_IC_CON_IC_RESTART_EN_BITS;
+            I2C_IC_CON_IC_RESTART_EN_BITS |
+            I2C_IC_CON_TX_EMPTY_CTRL_BITS;
 
     // Set FIFO watermarks to 1 to make things simpler. This is encoded by a register value of 0.
     i2c->hw->tx_tl = 0;
@@ -145,7 +146,7 @@ static int i2c_write_blocking_internal(i2c_inst_t *i2c, uint8_t addr, const uint
     bool abort = false;
     bool timeout = false;
 
-    uint32_t abort_reason;
+    uint32_t abort_reason = 0;
     int byte_ctr;
 
     int ilen = (int)len;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Wren6991** [comment]
Thank you for looking so deeply into this. This patch causes the existing `i2c_bus_scan` example (with `i2c_read_blocking()`) to fail to find any devices. I haven't looked into the reason yet, but the `I2C_IC_CON_TX_EMPTY_CTRL_BITS` in `i2c_init()` seems like the only change that could have affected this.

**Comment by fivdi** [comment]
There are some issues in `i2c_read_blocking_internal` too and my assumption that they can be fixed later may be incorrect. I'll take a look. The change with `I2C_IC_CON_TX_EMPTY_CTRL_BITS` may be relevant for `i2c_read_blocking_internal`.

**Comment by lurch** [comment]
> Should there be asserts in i2c_set_baudrate to check if lcnt or hcnt is > 65535?

It already does? Or is there an off-by-one error? https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_i2c/i2c.c#L72

**Comment by fivdi** [comment]
>>Should there be asserts in i2c_set_baudrate to check if lcnt or hcnt is > 65535?
>
>It already does? Or is there an off-by-one error? https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_i2c/i2c.c#L72

Sorry, my mistake, you are right, the checks are there already and look good.

**Comment by fivdi** [comment]
> Thank you for looking so deeply into this.

You're welcome 😄 

>This patch causes the existing `i2c_bus_scan` example (with `i2c_read_blocking()`) to fail to find any devices. I haven't looked into the reason yet, but the `I2C_IC_CON_TX_EMPTY_CTRL_BITS` in `i2c_init()` seems like the only change that could have affected this. (edit; actually am now struggling to reproduce this)
> 
> edit2: yes, loose wiring, sorry. Odd because I could swap the branches back and forth and it seemed repeatable until I commented!

Just to be sure, I ran the existing `i2c_bus_scan` example (with `i2c_read_blocking()`) about 100 times and it ran successfully each time only detecting the three devices on the bus.

To be extra sure, I also ran [this test program](https://github.com/fivdi/rp2040-tests/tree/master/i2c/scan_with_read_at_100k) which continuously scans the I2C bus with `i2c_read_blocking`. It's running at the moment and has done over 165000 scans and each scan only detected the three devices on the bus.

**Comment by fivdi** [comment]
> There are some issues in `i2c_read_blocking_internal` too and my assumption that they can be fixed later may be incorrect. I'll take a look. The change with `I2C_IC_CON_TX_EMPTY_CTRL_BITS` may be relevant for `i2c_read_blocking_internal`.

I don't think the change with `I2C_IC_CON_TX_EMPTY_CTRL_BITS` has an effect on `i2c_read_blocking_internal`. For the potential issues in `i2c_read_blocking_internal` see #336

**Comment by Wren6991** [comment]
Reading through the patch again in detail:

```
// TODO Could there be an abort while waiting for the STOP
// here? If so, additional code would be needed here
// to take care of the abort.
```

If I understand correctly, the outcome when this happens is that we leave the I2C in an aborted state when returning from the function. However this already happens when we see an abort type other than TX_ABRT (and it seems you would need to re-call i2c_init() to clear those aborts), so this only causes additional trouble in the case where

- A TX_ABRT occurs
- The abort is cleared
- Whilst waiting for the STOP following the TX_ABRT, another abort takes place

Which I don't think is reachable, so I think this is ok as is.

Wren6991:
It's a bit hard to figure out exactly *when* the aborts take place from the docs, so probably worth leaving that comment there and merging as is. Seems like this change is a strict improvement

Wren6991 (approved):
Thank you for this!

PR #334, opened 2021-04-12
Add networking stack from TinyUSB.

luigifcruz:
I'm not sure if this is a path the maintainers want to pursue. But I added a `tinyusb_net` target on CMakeLists to import the necessary headers and files to make it possible to use the Pico as a network device based on the RNDIS [example](https://github.com/hathach/tinyusb/tree/master/examples/device/net_lwip_webserver) from TinyUSB.

As of now, I use these patches to build my PiccoloSDR [project](https://github.com/luigifcruz/pico-stuff) without known problems.

### DIFF
--- a/src/rp2_common/tinyusb/CMakeLists.txt
+++ b/src/rp2_common/tinyusb/CMakeLists.txt
@@ -46,6 +46,7 @@ if (EXISTS ${PICO_TINYUSB_PATH}/${TINYUSB_TEST_PATH})
             ${PICO_TINYUSB_PATH}/src
             ${PICO_TINYUSB_PATH}/src/common
             ${PICO_TINYUSB_PATH}/hw
+            ${PICO_TINYUSB_PATH}/lib/networking
     )
 
     add_library(tinyusb_device_unmarked INTERFACE)
@@ -91,6 +92,13 @@ if (EXISTS ${PICO_TINYUSB_PATH}/${TINYUSB_TEST_PATH})
             ${PICO_TINYUSB_PATH}/src/class/hid/hid_host.c
             ${PICO_TINYUSB_PATH}/src/class/msc/msc_host.c
             ${PICO_TINYUSB_PATH}/src/class/vendor/vendor_host.c
+            )
+
+    add_library(tinyusb_net INTERFACE)
+    target_sources(tinyusb_host INTERFACE
+            ${PICO_TINYUSB_PATH}/lib/networking/dhserver.c
+            ${PICO_TINYUSB_PATH}/lib/networking/dnserver.c
+            ${PICO_TINYUSB_PATH}/lib/networking/rndis_reports.c
             )
 
     # Sometimes have to do host specific actions in mostly


kilograham (changes requested):
might be useful; not sure why some things are on tinyusb_common, and some on tinyusb_host and none on tinyusb_net!

luigifcruz:
Rebased to the latest branch. :)

kilograham (changes requested):

peterharperuk:
I've been playing around with the rndis stuff in tiny usb. I can't really see the point of running a dhcp server on the device. What would you use this for?

It would be nicer if dhcp could somehow obtain an address for the pico and use that as a network to get access to the internet. But I can't really get network sharing to work reliably. It looks like it sort of works on Windows if you're lucky. I'm not sure I can get it to work at all on Ubuntu.

Unless this is easy to use, I can't really see this being worth it?

luigifcruz:
I mean, this is not the easiest way to get data out of the Pico. But, unfortunately, it's the only way to do it with a decent speed without extra hardware. I can get a stable 5 Mbps running an `iperf3` server on the Pico using the TinyUSB interface. This enables projects like my [PiccoloSDR](https://hackaday.com/2021/03/16/the-raspberry-pi-pico-as-an-sdr-receiver/) that require large amounts of data to be transferred to the computer. I also can think about a few projects that would require the bandwidth offered by this stack. For example, digital signal analyzer using the PIO, digital oscilloscope, high-sampling rate sensors, etc.

Regarding compatibility, the current version of the PiccoloSDR, which uses this PR, works very well with Linux and macOS (including Apple Silicon). You plug it in, an IP is automatically assigned, and you can access it with zero configuration. I never tested this on Windows, but looks like [Microsoft added support](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/supported-usb-classes) to USB NCM on Windows 11. This theoretically solves support for Windows.

While developing PiccoloSDR, a major hurdle was patching the SDK to give me the right headers and binaries. The actual development of the network application was trivial. This PR comes at almost zero cost for those who don't use this feature while significantly facilitating the development process of network apps for those who need it.

kilograham:
i'm just confused by the PR in general; why are existing libraries being modified; can you just not add a new "tinyusb_net library"?

peterharperuk:
> can you just not add a new "tinyusb_net library"?

I guess you could, but tiny usb already seems to pile everything into tinyusb_device_base? What's more confusing is that this doesn't live in the pico-sdk anymore.

We're missing ${PICO_TINYUSB_PATH}/lib/networking/rndis_reports.c and the include folder ${PICO_TINYUSB_PATH}/lib/networking, which are easy enough to add to your own project.

I'm more interested in knowing if we can use this for real.

peterharperuk:
@luigifcruz Sorry, I missed your comment. So your use case is just to connect two computers and communicate with IP stacks. The IP address comes from the Pico - effectively a small private network.

mordae:
I am trying to build a subsampling receiver, learning as I go, and to my surprise the excellent demo from luigifcruz wouldn't work with stock SDK. :sweat_smile:

Now seriously, using UDP to stream data over USB is a decent idea and with DHCP server it works out-of-box. I believe that USB tethering on Android works like that as well. It saves you the hassle of configuring network interfaces manually or writing specific device drivers.

peterharperuk:
@luigifcruz @mordae Did you test by connecting to Windows or a Linux PC? I still don't seem to get an IP address when connecting to Linux.

Issue #337, opened 2021-04-13
Arithmetic underflow in SPI read routines, leading to deadlock


michaelstoops:
I've identified an arithmetic underflow in the SPI read routines which seems to lead a deadlock, at least in my application.

The repeatability code is https://github.com/raspberrypi/pico-examples/pull/101, in the case that the SPI slave starts while the SPI master is in the middle of a transmission. The slave correctly transmits 8 bytes and then deadlocks.

The flaw is in https://github.com/raspberrypi/pico-sdk/blob/fc10a97c386f65c1a44c68684fe52a56aaf50df0/src/rp2_common/hardware_spi/spi.c#L79 and similar lines below. In the error condition, `tx_remaining` is 248 while `rx_remaining` is 247. The operation underflows and gives a result of `0xFF FF FF FF FF FF FF FF`.

I'm not 100% clear on what happens next, and it probably depends on the timing of signals coming in anyway. Either way, I think the arithmetic underflow is not intended.

kilograham:
presumably you mean _hang_ not _deadlock_

michaelstoops:
> presumably you mean _hang_ not _deadlock_

Fair. :)

michaelstoops:
Submitted proposed fix at https://github.com/raspberrypi/pico-sdk/pull/338.

Wren6991:
I have merged #338, since the unsigned wrapping in the comparison was not intentional!

AFAIK it is an invariant for master mode that `rx_remaining >= tx_remaining` at any point in time, and it seems like slave mode breaks this invariant in some or all cases.

> in the case that the SPI slave starts while the SPI master is in the middle of a transmission.

Could you expand on that please? Is the timing of the start of the slave transfer vs the start of the master transfer significant?

michaelstoops:
Yes. Are you able to read Salae Logic data files?

michaelstoops:
> Yes. Are you able to read Saleae Logic data files?

I guess you can download the application here, whether or not you have a Saleae device. https://www.saleae.com/downloads/

kilograham:
i don't have enough context, but it would seem sensible that a slave might receive before transmitting? However whatever the outcome in that case, I worry about an off by one error on the other end of things (i.e. buffer being full)

michaelstoops:
Here is the happy path. Slave starts first and waits for clock pulses from the master.
[...]

PR #338, opened 2021-04-13
Fixing arithmetic underflow https://github.com/raspberrypi/pico-sdk/i…

michaelstoops:
I think this fixes https://github.com/raspberrypi/pico-sdk/issues/337.


Wren6991 (dismissed):
Good catch, thank you. I'm a little confused as to the exact situation where `rx_remaining < tx_remaining` but the unsigned wrapping was certainly not intentional, and this change seems to just fix that.

Wren6991 (approved):

michaelstoops:
> Good catch, thank you. I'm a little confused as to the exact situation where `rx_remaining < tx_remaining` but the unsigned wrapping was certainly not intentional, and this change seems to just fix that.

Thanks! 🙂

Could it have something to do with being in SPI slave mode?

PR #343, opened 2021-04-16
Updated existing Pimoroni board headers to match latest style, and added a new board

ZodiusInfuser:


src/boards/include/boards/pimoroni_picolipo_16mb.h
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
+//       SHOULD ONLY CONSIST OF PREPROCESSOR DIRECTIVES
+// -----------------------------------------------------
+
+#ifndef _BOARDS_PIMORONI_PICOLIPO_4MB_H
```
lurch, line 12: Says `4MB` here, but `16mb` in the filename?

src/boards/include/boards/pimoroni_picolipo_16mb.h
```diff
@@ -0,0 +1,96 @@
...
+// For board detection
+#define PIMORONI_PICOLIPO_4MB
+
+// --- BOARD SPECIFIC ---
+#ifndef PICOLIPO_USER_SW_PIN
+#define PICOLIPO_USER_SW_PIN 23
+#endif
+
+#ifndef PICOLIPO_VBUS_DETECT_PIN
+#define PICOLIPO_VBUS_DETECT 24
```
lurch, line 24: Missing '_PIN` ?

src/boards/include/boards/pimoroni_picolipo_4mb.h
```diff
@@ -0,0 +1,96 @@
...
+// For board detection
+#define PIMORONI_PICOLIPO_4MB
+
+// --- BOARD SPECIFIC ---
+#ifndef PICOLIPO_USER_SW_PIN
+#define PICOLIPO_USER_SW_PIN 23
+#endif
+
+#ifndef PICOLIPO_VBUS_DETECT_PIN
+#define PICOLIPO_VBUS_DETECT 24
```
lurch, line 24: `_PIN` again

src/boards/include/boards/pimoroni_picosystem.h
```diff
@@ -115,15 +117,39 @@
...
+//no PICO_DEFAULT_I2C
+//no PICO_DEFAULT_I2C_SDA_PIN
+//no PICO_DEFAULT_I2C_SCL_PIN
+
+// --- SPI ---
+#ifndef PICO_DEFAULT_SPI
+#define PICO_DEFAULT_SPI 0
+#endif
+#ifndef PICO_DEFAULT_SPI_SCK_PIN
+#define PICO_DEFAULT_SPI_SCK_PIN PICOSYSTEM_LCD_SCLK_PIN
```
lurch, line 141: Does it make sense to define these as `PICO_DEFAULT_SPI_SCK_PIN` if they're hardwired to the LCD?

src/boards/include/boards/pimoroni_picolipo_16mb.h
ZodiusInfuser: Thanks for spotting this! Fixed.

src/boards/include/boards/pimoroni_picolipo_16mb.h
ZodiusInfuser: Thanks. Fixed

src/boards/include/boards/pimoroni_picolipo_4mb.h
ZodiusInfuser: Thanks. Fixed

src/boards/include/boards/pimoroni_picosystem.h
ZodiusInfuser: You are right. In this case the SPI is not exposed onto any headers, so the specific LCD defines should be used. I have added a `PICOSYSTEM_LCD_SPI` define in place of `PICO_DEFAULT_SPI`

ZodiusInfuser:
Thanks for the review @lurch. I have corrected the issues you identified.

lurch:
Can you rebase this against the `develop` branch please?

lurch:
Given the large overlap between `pimoroni_picolipo_4mb.h` and `pimoroni_picolipo_16mb.h`, I wonder if it might be worth factoring the common parts out into e.g. a `pimoroni_picolipo_generic.h` file, and then `pimoroni_picolipo_16mb.h` could define the flash-size and just include `pimoroni_picolipo_generic.h` ? :shrug:

ZodiusInfuser:
> Can you rebase this against the `develop` branch please?
Okay. Done.

> Given the large overlap between `pimoroni_picolipo_4mb.h` and `pimoroni_picolipo_16mb.h`, I wonder if it might be worth factoring the common parts out into e.g. a `pimoroni_picolipo_generic.h` file, and then `pimoroni_picolipo_16mb.h` could define the flash-size and just include `pimoroni_picolipo_generic.h` ? 🤷

I thought about that, but that gives people the ability to enter that generic header as a board option. Don't know if that's a problem. Tested my end and it compiles pico-examples, which is odd, as I thought it would fall over with no flash size defined, unless a default gets defined somewhere later?

ZodiusInfuser:
A hacky way around it would be to put the generic/common header into a sub directory of boards, then this fails (as we want):

`cmake -D"PICO_BOARD=pimoroni_picolipo_common" ..`

but the common can still be selected like this:

`cmake -D"PICO_BOARD=common/pimoroni_picolipo_common" ..`

Not a great solution though. Perhaps you have a better idea?

lurch:
A similarly hacky solution might be to put something like:
```
#if !defined(PIMORONI_PICOLIPO_4MB) && !defined(PIMORONI_PICOLIPO_16MB)
#error Please include pimoroni_picolipo_4mb.h or pimoroni_picolipo_16mb.h rather than directly including pimoroni_picolipo_generic.h
#endif 
```
into `pimoroni_picolipo_generic.h` ? :shrug: 

Or @kilograham might have a better suggestion.

ZodiusInfuser:
That's a decent idea, but will wait for graham before committing the change.

By the way, would you prefer _generic or _common for such header names? The latter seems more appropriate to me. And yes or no to putting it into a subfolder like I showed?

lurch:
Those are Graham-questions :wink:

kilograham (approved):

PR #348, opened 2021-04-19
correct adafruit flash size for itsybitsy and qtpy rp2040 from 4 MB to 8 MB

hathach:
This PR correct the flash size for [itsybitsy](https://www.adafruit.com/product/4888) and [qtpy](https://www.adafruit.com/product/4900) rp2040. Both use 8MB W25Q64JVxQ flash device


kilograham (approved):

hathach:
Thank you for the quick merge

PR #365, opened 2021-04-27
Change various (confusing to user) message to be DEBUG only

kilograham:


liamfraser (approved):

lurch:
Perhaps the `message("Using PICO_EXAMPLES_PATH from environment ('${PICO_EXAMPLES_PATH}')")` chunk in `docs/CMakeLists.txt` should be moved inside the `if(PICO_BUILD_DOCS)` clause?

jonathangjertsen:
The DEBUG level was added in CMake 3.15, so with 3.14 it looks even more confusing:

```
DEBUGPICO_GCC_TRIPLE defaulted to arm-none-eabi
...
DEBUGIncluding custom CMake file [redacted]/pico-sdk/src/common/pico_base/generate_config_header.cmake
DEBUGELF2UF2 will need to be built
```

Should the minimum required version be bumped?

kilograham:
ah good catch can you please open an issue

PR #370, opened 2021-05-01
Feature - RTC alarm on every second

Reneg973:
The PR includes 2 changes

1. better const correctness
2. the mentioned feature, every second the ISR is triggered to update the clock output


src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -13,7 +13,7 @@
 
 // Set this when setting an alarm
 static rtc_callback_t _callback = NULL;
-static bool _alarm_repeats = false;
+static int8_t _alarm_repeats = 0;
```
kilograham, line 16: prefer to use an enum for 3 repeat types

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +144,51 @@ static void rtc_irq_handler(void) {
...
+       s1 |= RTC_IRQ_SETUP_1_HOUR_ENA_BITS   | (((uint)t->hour)  << RTC_IRQ_SETUP_1_HOUR_LSB);
+    if (t->min   >= 0)
+        s1 |= RTC_IRQ_SETUP_1_MIN_ENA_BITS   | (((uint)t->min)   << RTC_IRQ_SETUP_1_MIN_LSB);
+    if (t->sec   >= 0)
+        s1 |= RTC_IRQ_SETUP_1_SEC_ENA_BITS   | (((uint)t->sec)   << RTC_IRQ_SETUP_1_SEC_LSB);
+    else if (_alarm_repeats == -1) // repeatable every second! All entries are -1
+    {
+        datetime_t tNew;
+        rtc_get_datetime(&tNew);
+        s1 = RTC_IRQ_SETUP_1_SEC_ENA_BITS | ((((uint)tNew.sec + 1) % 60) << RTC_IRQ_SETUP_1_SEC_LSB);
```
kilograham, line 186: This seems race prone; is using the RTC clock for an every second interrupt useful?

kilograham (commented):
1) Please rebase onto develop
2) We don't have a CONTRIBUTING.md but please match the existing code style:
    - 4 spaces, not tabs
    - { on same line as if
    - if's without braces on same line as if only
    - variable names are not Hungarian

src/rp2_common/hardware_rtc/rtc.c
Reneg973: at least it works fine - updating my display every second while pushing the current time on UART.
But there is another more critical thing missing. ATM an attached exclusive interrupt can't be removed.

src/rp2_common/hardware_rtc/rtc.c
Reneg973: I'll upate it

src/rp2_common/hardware_rtc/rtc.c
Reneg973: the current ISR still enables rtc_alarm before user callback has returned. Maybe we should change this too?

Reneg973: Another question, usually RTC uses 32.768kHz clock, here 46.875kHz is programmed. What is the cause for this?

lurch:
You need to click the "Edit" button at the top of this page and change the base branch :wink:  (which will clean up the messy commit history that GH is currently displaying)

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -13,7 +13,14 @@
 
 // Set this when setting an alarm
 static rtc_callback_t _callback = NULL;
-static bool _alarm_repeats = false;
+
+typedef enum {
+    NO_REPEAT                =  0,
+    CONTINUOUS_REPEAT        =  1,
+    CONTINUOUS_REPEAT_ON_SEC =  2,
```
lurch, line 20: This looks like it might be a typo for `CONTINUOUS_REPEAT_ONE_SEC` :wink: Perhaps `CONTINUOUS_REPEAT_EVERY_SECOND` might be less ambiguous?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +151,46 @@ static void rtc_irq_handler(void) {
...
-    if (t->dotw  < 0) return true;
-    if (t->hour  < 0) return true;
-    if (t->min   < 0) return true;
-    if (t->sec   < 0) return true;
-    return false;
+// return != 0 on repeat, -1 on repeat every second
+static int8_t rtc_alarm_repeats(const datetime_t *t) {
+  // If any value is set to -1 then we don't match on that value
+  // hence the alarm will eventually repeat
+    if((t->year & t->month & t->day & t->dotw & t->hour & t->min & t->sec) < 0) return CONTINUOUS_REPEAT_ON_SEC;
```
lurch, line 158: Graham might feel differently to me, but I wonder if explicitly doing
```
if (t->year < 0 && t->month < 0 && t->day < 0 && t->dotw < 0 && t->hour < 0 && t->min < 0 && t->sec < 0)
```
might be more readable than using bitwise-AND ?
(your version is also missing the space between the `if` and the `(` :slightly_smiling_face: )

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +151,46 @@ static void rtc_irq_handler(void) {
...
-    // hence the alarm will eventually repeat
-    if (t->year  < 0) return true;
-    if (t->month < 0) return true;
-    if (t->day   < 0) return true;
-    if (t->dotw  < 0) return true;
-    if (t->hour  < 0) return true;
-    if (t->min   < 0) return true;
-    if (t->sec   < 0) return true;
-    return false;
+// return != 0 on repeat, -1 on repeat every second
```
lurch, line 154: This comment should probably refer to the enum-values rather than -1 ?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +151,46 @@ static void rtc_irq_handler(void) {
...
+static int8_t rtc_alarm_repeats(const datetime_t *t) {
```
lurch, line 155: Should this be declared to return `repeat_type` rather than `int8_t`?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +151,46 @@ static void rtc_irq_handler(void) {
...
-    if (t->dotw  >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_DOTW_ENA_BITS);
-    if (t->hour  >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_HOUR_ENA_BITS);
-    if (t->min   >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_MIN_ENA_BITS);
-    if (t->sec   >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_SEC_ENA_BITS);
+    uint32_t s0 = 0, s1 = 0;
 
     // Does it repeat? I.e. do we not match on any of the bits
     _alarm_repeats = rtc_alarm_repeats(t);
 
+    if( (!valid_datetime(t) && !_alarm_repeats) || !user_callback)	// none of the parameters is valid
```
lurch, line 173: Would an alarm be considered "valid" if it e.g. had a month of `14` and a day of `-1` ?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +151,46 @@ static void rtc_irq_handler(void) {
...
+        return false;
+
+    // Set the match enable bits for things we care about
+    if (t->year  >= 0) s0 |= RTC_IRQ_SETUP_0_YEAR_ENA_BITS  | (((uint)t->year)  << RTC_IRQ_SETUP_0_YEAR_LSB );
+    if (t->month >= 1) s0 |= RTC_IRQ_SETUP_0_MONTH_ENA_BITS | (((uint)t->month) << RTC_IRQ_SETUP_0_MONTH_LSB);
+    if (t->day   >= 1) s0 |= RTC_IRQ_SETUP_0_DAY_ENA_BITS   | (((uint)t->day)   << RTC_IRQ_SETUP_0_DAY_LSB  );
```
lurch, line 179: I'm not sure if changing the month and day checks to `1` instead of `0` is wanted here? :shrug:

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -131,43 +151,46 @@ static void rtc_irq_handler(void) {
...
+    if (t->dotw  >= 0) s1 |= RTC_IRQ_SETUP_1_DOTW_ENA_BITS  | (((uint)t->dotw)  << RTC_IRQ_SETUP_1_DOTW_LSB);
+    if (t->hour  >= 0) s1 |= RTC_IRQ_SETUP_1_HOUR_ENA_BITS  | (((uint)t->hour)  << RTC_IRQ_SETUP_1_HOUR_LSB);
+    if (t->min   >= 0) s1 |= RTC_IRQ_SETUP_1_MIN_ENA_BITS   | (((uint)t->min)   << RTC_IRQ_SETUP_1_MIN_LSB);
+    if (t->sec   >= 0) s1 |= RTC_IRQ_SETUP_1_SEC_ENA_BITS   | (((uint)t->sec)   << RTC_IRQ_SETUP_1_SEC_LSB);
+    else if (_alarm_repeats == CONTINUOUS_REPEAT_ON_SEC) {
```
lurch, line 184: Should this be the first test in this chain-of-ifs, rather than only appearing in the `else` clause?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -122,6 +137,11 @@ static void rtc_irq_handler(void) {
 
     if (_alarm_repeats) {
         // If it is a repeatable alarm, re enable the alarm.
+        if(_alarm_repeats == CONTINUOUS_REPEAT_ON_SEC) {
+          datetime_t t;
+          rtc_get_datetime(&t);
+          rtc_hw->irq_setup_1 = RTC_IRQ_SETUP_1_SEC_ENA_BITS | ((((uint)t.sec + 1) % 60) << RTC_IRQ_SETUP_1_SEC_LSB);
```
lurch, line 143: Is this change needed because otherwise the RTC alarm would just fire continuously?

src/rp2_common/hardware_rtc/rtc.c
lurch, line 173: Might be better to check `_alarm_repeats != NO_REPEAT` rather than `!_alarm_repeats` ?

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Not a typo, just a different point of view. An alarm is triggered ON change of x, where x may be SEC or MIN or HOUR, but all other possible alarms/triggers are already included in CONTINUOUS_REPEAT.
From SW pov you are right, I'll modify it

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Right, I already changed it, just not pushed

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Overseen to remove this comment, changed also return type of function

src/rp2_common/hardware_rtc/rtc.c
Reneg973: very good objection, needs a better breakdown

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Yes, that is a strange inconsistency in the original code. The RTC reports months from 1-12, but the input allowed a month of 0. If I assume the RTC follows the kiss principle, input and output will have the same ranges. Anyway I am not 100% sure

src/rp2_common/hardware_rtc/rtc.c
Reneg973: > Is this change needed because otherwise the RTC alarm would just fire continuously?

IRQ has been triggered on second x. This code is to renable it for second (x+1)%60. If not incremented, it would delay until it reaches second x again

src/rp2_common/hardware_rtc/rtc.c
Reneg973: oh yes, it should be first case

src/rp2_common/hardware_rtc/rtc.c
Reneg973: right and done

src/rp2_common/hardware_rtc/rtc.c
Reneg973: done

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -11,9 +11,28 @@
...
+#define RANGE_CHECK_DOTW(t)  (t->dotw >= 0 && t->dotw <= 6)
+#define RANGE_CHECK_HOUR(t)  (t->hour >= 0 && t->hour <= 23)
+#define RANGE_CHECK_MIN(t)   (t->min >= 0 && t->min <= 59)
+#define RANGE_CHECK_SEC(t)   (t->sec >= 0 && t->sec <= 59)
+
+
 // Set this when setting an alarm
 static rtc_callback_t _callback = NULL;
-static bool _alarm_repeats = false;
+static uint8_t _seconds_increment = 1;
```
kilograham, line 25: not sure why this is state; it seems like it should be a macro parameter as the usage on line 161 should always be 1 (not the last value)

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -114,60 +139,88 @@ void rtc_enable_alarm(void) {
...
+    valid_params_if(RTC, check_params);
+    if(!check_params)	// none of the parameters is valid
+        return false;
+
+    // Set the match enable bits for things we care about
+    if(_alarm_repeats == CONTINUOUS_REPEAT_EVERY_SEC) {
+        // repeatable every second! All entries are -1
+        datetime_t new_dt;
+        _rtc_get_datetime(&new_dt);
+        _seconds_increment = (-t->sec) % 60;
```
kilograham, line 206: even more confused, because t->sec is -1 here always

kilograham (commented):
also white space/formatting issues still

src/rp2_common/hardware_rtc/rtc.c
Reneg973: maybe best described with an example:
if sec input parameter for rtc_set_alarm() is in range 0<=s<=59, than we have absolute behavior. The alarm always appears when internal comparison of s with the RTC seconds is true.

If sec value is negative, it is interpreted as a relative value. Means if it is -1, _seconds_increment is set to 1, and the alarm appears whenever the seconds changes. If now input sec is -2, the increment will be 2 and therefore the ISR is called every 2 seconds. As this input value is just local within rtc_set_alarm(), it must be remembered as a state, so that ISR can access it.

src/rp2_common/hardware_rtc/include/hardware/rtc.h
```diff
@@ -32,6 +32,10 @@
 extern "C" {
 #endif
 
+#ifndef PARAM_ASSERTIONS_ENABLED_RTC
```
lurch, line 35: Could you add a PICO_CONFIG line as well please? See https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_adc/include/hardware/adc.h#L47 for an example.
(This allows the value to be set by https://github.com/raspberrypi/pico-project-generator )

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -11,9 +11,28 @@
 #include "hardware/resets.h"
 #include "hardware/clocks.h"
 
+#define RANGE_CHECK_YEAR(t)  (t->year >= 0 && t->year <= 4095)
+#define RANGE_CHECK_MONTH(t) (t->month >= 1 && t->month <= 12)
+#define RANGE_CHECK_DAY(t)   (t->day >= 1 && t->day <= 32)
```
lurch, line 16: Which months have 32 days? :wink:

src/rp2_common/hardware_rtc/rtc.c
Reneg973: E.g. like snooze function on a clock. Set alarm to -30sec repeat, alarm appears, you press button to switch off the alarm, and after 30s it appears again.

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Oh there are months that won't end at all :)
I was not sure whether <= or < has better readability, and so this slipped through.

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -114,60 +139,88 @@ void rtc_enable_alarm(void) {
...
-    if (t->dotw  >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_DOTW_ENA_BITS);
-    if (t->hour  >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_HOUR_ENA_BITS);
-    if (t->min   >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_MIN_ENA_BITS);
-    if (t->sec   >= 0) hw_set_bits(&rtc_hw->irq_setup_1, RTC_IRQ_SETUP_1_SEC_ENA_BITS);
+    uint32_t s0 = 0, s1 = 0;
 
     // Does it repeat? I.e. do we not match on any of the bits
     _alarm_repeats = rtc_alarm_repeats(t);
 
+    bool check_params = (is_valid_datetime(t) || _alarm_repeats != NO_REPEAT) && user_callback;
```
lurch, line 196: @kilograham Do we definitely want to just `return false` here (and not set the alarm) if a `user_callback` wasn't supplied?

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Oh maybe you're confused because _seconds_increment is processed within the macro ADD_AND_ENABLE_REPEATABLE_SECOND?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -180,12 +233,24 @@ void rtc_set_alarm(datetime_t *t, rtc_callback_t user_callback) {
...
-void rtc_disable_alarm(void) {
-    // Disable matching and wait for it to stop being active
-    hw_clear_bits(&rtc_hw->irq_setup_0, RTC_IRQ_SETUP_0_MATCH_ENA_BITS);
-    while(rtc_hw->irq_setup_0 & RTC_IRQ_SETUP_0_MATCH_ACTIVE_BITS) {
-        tight_loop_contents();
+void rtc_delete_alarm(void)
+{
+    if (_callback == NULL) {
+        return; // rtc_set_alarm was not called or not successful
     }
```
lurch, line 243: Feels like this function would be more robust if this check was done near the end rather than near the start? (or simply remove the check altogether, since `_callback` later gets set to NULL anyway)

src/rp2_common/hardware_rtc/rtc.c
lurch: I think @Reneg973 was intending that you could set e.g. "repeat every 5 seconds" by setting `t->sec` to -5 ?
But I'm not sure if that's desirable behaviour, since e.g. there's no similar functionality to set "repeat every 5 minutes" (and we're not trying to re-implement `cron` :wink: ) so I think it might be more consistent to revert the "repeat every 5 seconds" functionality, and just stick to the "setting everything to -1 causes the alarm to repeat every second" behaviour. For more sophisticated behaviour, the user can always add their own custom logic inside their callback function.

kilograham: no i was confused because I didn't notice you could now set this to negative values other than -1. I'm not sure that this functionality belongs here; there are plenty of other ways of getting repeating timers. `hardware_foo` libraries are generally intended for exposing hardware functionality in a clean API, not using the hardware to provide additional functionality.
I need to take a closer look at the RTC h/w, so I'll get back to you

src/rp2_common/hardware_rtc/rtc.c
Reneg973: ok this one is wrong - I fix it

Datasheet 4.8.3. says this event is also used to wakeup the pico

Reneg973: Anyway I think the HAL should differ between a user event and a wakeup from dormant mode, because if it is in dormant mode, the user callback may not be executed as clk_sys may be 0

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -114,60 +139,88 @@ void rtc_enable_alarm(void) {
...
+    else {
+        if (RANGE_CHECK_YEAR(t))  s0 |= RTC_IRQ_SETUP_0_YEAR_ENA_BITS  | (((uint)t->year)  << RTC_IRQ_SETUP_0_YEAR_LSB);
+        if (RANGE_CHECK_MONTH(t)) s0 |= RTC_IRQ_SETUP_0_MONTH_ENA_BITS | (((uint)t->month) << RTC_IRQ_SETUP_0_MONTH_LSB);
+        if (RANGE_CHECK_DAY(t))   s0 |= RTC_IRQ_SETUP_0_DAY_ENA_BITS   | (((uint)t->day)   << RTC_IRQ_SETUP_0_DAY_LSB);
+        if (RANGE_CHECK_DOTW(t))  s1 |= RTC_IRQ_SETUP_1_DOTW_ENA_BITS  | (((uint)t->dotw)  << RTC_IRQ_SETUP_1_DOTW_LSB);
+        if (RANGE_CHECK_HOUR(t))  s1 |= RTC_IRQ_SETUP_1_HOUR_ENA_BITS  | (((uint)t->hour)  << RTC_IRQ_SETUP_1_HOUR_LSB);
+        if (RANGE_CHECK_MIN(t))   s1 |= RTC_IRQ_SETUP_1_MIN_ENA_BITS   | (((uint)t->min)   << RTC_IRQ_SETUP_1_MIN_LSB);
+        if (RANGE_CHECK_SEC(t))   s1 |= RTC_IRQ_SETUP_1_SEC_ENA_BITS   | (((uint)t->sec)   << RTC_IRQ_SETUP_1_SEC_LSB);
+
+        if(!s0 && !s1) return false; // out of range datetime_t input
```
lurch, line 218: Is this a condition that can actually happen? Won't it be caught by the earlier checks?

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -114,60 +139,88 @@ void rtc_enable_alarm(void) {
...
+    }
+}
+
+static void __no_inline_not_in_flash_func(rtc_irq_handler)(void) {
     // Always disable the alarm to clear the current IRQ.
     // Even if it is a repeatable alarm, we don't want it to keep firing.
     // If it matches on a second it can keep firing for that second.
     rtc_disable_alarm();
 
     if (_alarm_repeats) {
```
lurch, line 156: Should all the `if (_alarm_repeats)` checks actually be `if (_alarm_repeats != NO_REPEAT)` ? :shrug:

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -114,60 +139,88 @@ void rtc_enable_alarm(void) {
...
-    if (t->dotw  < 0) return true;
-    if (t->hour  < 0) return true;
-    if (t->min   < 0) return true;
-    if (t->sec   < 0) return true;
-    return false;
+static repeat_type rtc_alarm_repeats(const datetime_t *t) {
+  // If any value is set to -1 then we don't match on that value
+  // hence the alarm will eventually repeat
+    if (t->year < 0 && t->month < 0 && t->day < 0 && t->dotw < 0
+        && t->hour < 0 && t->min < 0 && t->sec < 0) return CONTINUOUS_REPEAT_EVERY_SEC;
```
lurch, line 178: Probably worth adding a comment here saying that if all values are -1 then we repeat every second?

src/rp2_common/hardware_rtc/rtc.c
Reneg973: it can happen if one parameter is set to a value > range and all others are -1.

src/rp2_common/hardware_rtc/rtc.c
lurch: > it needs nearly twice the operations and therefore cycles to check

But with RP2040 running at 125MHz, and the RTC alarm only firing at most once per second, I don't think that should be a huge problem? :wink:

src/rp2_common/hardware_rtc/rtc.c
Reneg973: yes, I removed the user_callback check and therefore it must be removed here too. But this leads to the next issue, see
irq_remove_handler()/irq_set_exclusive_handler(). Seems like a NULL handler is handled as a valid function ptr there.

src/rp2_common/hardware_rtc/rtc.c
Reneg973: a question of coding guidelines. Technically it's the same

lurch: Fair enough, that's a Graham-decision then :slightly_smiling_face:

src/rp2_common/hardware_rtc/rtc.c
lurch: Perhaps it's worth adding an `is_valid_alarmtime` function, and using that as part of the `valid_params_if` checks?
```
static bool is_valid_alarmtime(const datetime_t *t) {
    return (RANGE_CHECK_YEAR(t) || t->year < 0)
        && (RANGE_CHECK_MONTH(t) || t->month < 0)
        && (RANGE_CHECK_DAY(t) || t->day < 0)
// etc.
```

src/rp2_common/hardware_rtc/rtc.c
lurch: The RP2040 datasheet says: "XOSC @ 12MHz / 256 = 46875Hz. To get a 1Hz reference CLKDIV_M1 should be set to 46874."
As long as `CLKDIV_M1` is set to an appropriate value, I don't think it matters what frequency the RTC clock runs at? (Looks like https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_clocks/clocks.c#L187 is where clk_rtc is actually configured)

src/rp2_common/hardware_rtc/rtc.c
```diff
@@ -180,12 +234,20 @@ void rtc_set_alarm(datetime_t *t, rtc_callback_t user_callback) {
...
+{
+    // first disable the alarm
+    rtc_disable_alarm();
+
+    // don't receive interrupts anymore
+    rtc_hw->inte = RTC_INTE_RESET;
+
+    // disable IRQ and remove handler
+    irq_set_enabled(RTC_IRQ, false);
+    irq_remove_handler(RTC_IRQ, _callback);
```
lurch, line 250: Should this be `irq_remove_handler(RTC_IRQ, rtc_irq_handler)` ? :shrug: (IRQ-related code isn't my strong point!)

src/rp2_common/hardware_rtc/rtc.c
Reneg973: Exactly,  12MHz / 256...  I found it. Would be nice if the comment you linked would say the same

src/rp2_common/hardware_rtc/rtc.c
Reneg973: nice, so I had in mind that I made too many changes with 1 commit ... and prompt forgot to fix this.

Reneg973: done

src/rp2_common/hardware_rtc/rtc.c
Reneg973: I don't think it's worth the effort as the function already returns false then. Just an assertion could appear - if they are enable for rtc.

I think the more critical question is, what should happen with a given user_callback, if pico is woken up from RTC event?

lurch: > > Perhaps it's worth adding an `is_valid_alarmtime` function

> I don't think it's worth the effort

But I _think_ this would successfully reject the "a month of 14 and a day of -1" that I mentioned earlier? Meh :shrug: 

>  what should happen with a given user_callback, if pico is woken up from RTC event?

Looking at https://github.com/raspberrypi/pico-extras/blob/master/src/rp2_common/pico_sleep/sleep.c#L106 (linked to from the RP2040 datasheet) it seems that whatever callback is passed to `sleep_goto_sleep_until` (which of course might be NULL) overwrites any existing RTC user_callback.

BTW I hope you don't feel that I'm being too picky with all my comments, I'm only trying to offer constructive feedback :slightly_smiling_face:   And if any of my comments conflict with what @kilograham says, please ignore my comments and listen to Graham :grinning:

Reneg973: It's fine as your comments make sense. And now, good find! Yes, at least `sleep_goto_sleep_until` tries to overwrite the user callback. But the user callback is set with `irq_set_exclusive_handler`. And now we find ourselves within an `hard_assert(current == __unhandled_user_irq || current == handler);`

So my freshly implemented rtc_delete_alarm() is directly required for  `sleep_goto_sleep_until` :-)

lurch: > And now we find ourselves within an hard_assert(current == __unhandled_user_irq || current == handler);

Are you sure? I think you might be mixing up the irq_handler and the user_callback again?  AFAICT (and I could easily be entirely wrong!) the `sleep_goto_sleep_until` only changes the user_callback (which is called by the irq_handler), and doesn't actually change the irq_handler? :shrug:

Reneg973: ok, second condition then is true (current==handler)

src/rp2_common/hardware_rtc/rtc.c
Reneg973: From SW side you are right, this functionality could be implemented in a level above.
There are 2 solutions then
1. set an alarm "repeat every second" -> count the number of IRQs arrived :: fast ISR but highly increased number of IRQs
2. set an alarm at current_time + x -> get the IRQ -> query datetime_t -> increment by x -> set new alarm via rtc_set_alarm() :: slow ISR but not more IRQs than required.

So both methods produce more overhead than really required.
Anyway I don't know which HAL strategy you want to follow for Pico.

lurch: 3rd approach: repeat every second -> query datetime_t -> if `t->sec % 5 == 0` do something, else ignore :wink: 

As I mentioned before, with RP2040 running with a CPU clock of 125MHz I don't think you really need to worry about "slow ISR"?

kilograham: You do need to worry!

Reneg973: 3rd approach = 1st :) - I meant the same! And t->sec % 5 is not correct, you need include the start seconds.
And 125MHz is just the default. I have improved my SPI performance by 22% (increase SPI freq 31.25MHz -> 40MHz), but for this I had to sacrifice 36% of clk_sys by setting it to 80MHz!

src/rp2_common/hardware_rtc/rtc.c
Reneg973: ok no user callback in sleep because it's overwritten

src/rp2_common/hardware_rtc/include/hardware/rtc.h
Reneg973: done

src/rp2_common/hardware_rtc/rtc.c
Reneg973: I am simply no friend of increasing the number of execution cycles just to increase readability. If it's not clear, it could be documented.

lurch: Another Graham-decision I think :wink:

Reneg973: It could, I would be happy to change it if accepted. But of course it's a little bit bit "hacking" at a place where it is not really time critical. Just a few clock cycles and a few bytes binary code size reduction.

src/rp2_common/hardware_rtc/rtc.c
Reneg973: > no i was confused because I didn't notice you could now set this to negative values other than -1. I'm not sure that this functionality belongs here; there are plenty of other ways of getting repeating timers. `hardware_foo` libraries are generally intended for exposing hardware functionality in a clean API, not using the hardware to provide additional functionality.
> I need to take a closer look at the RTC h/w, so I'll get back to you

@kilograham:
Repeating timers are not repeating clock ticks. Anyway here is my proposal for a modification to be able to do this outside:
1. generate a new interface function rtc_modify_alarm() which reduces the overhead of rtc_set_alarm() and therefore can be called within the ISR.
2. in almost all use cases when the user ISR is called, the RTC time will be queried. Give it as a parameter (tell don't ask principle).

What do you think?

PR #372, opened 2021-05-03
add PICO_DIVIDER_DISABLE_INTERRUPTS flag 

kilograham:
…which makes PICO_DIVIDER disable interrupts around division rather than using co-operative guards to protect nested use (i.e. within IRQ/exception). Use of this flag can simplify porting of RTOSes but with a different performance profile

### DIFF
--- a/src/rp2_common/pico_divider/divider.S
+++ b/src/rp2_common/pico_divider/divider.S
@@ -11,16 +11,21 @@
 .cpu cortex-m0plus
 .thumb
 
+// PICO_CONFIG: PICO_DIVIDER_DISABLE_INTERRUPTS, Disable interrupts around division such that divider state need not be saved/restored in exception handlers, default=0, group=pico_divider
+
 #include "pico/asm_helper.S"
 
+// PICO_CONFIG: PICO_DIVIDER_CALL_IDIV0, Whether 32 bit division by zero should call __aeabi_idiv0, default=1, group=pico_divider
 #ifndef PICO_DIVIDER_CALL_IDIV0
 #define PICO_DIVIDER_CALL_IDIV0 1
 #endif
 
+// PICO_CONFIG: PICO_DIVIDER_CALL_IDIV0, Whether 64 bit division by zero should call __aeabi_ldiv0, default=1, group=pico_divider
 #ifndef PICO_DIVIDER_CALL_LDIV0
 #define PICO_DIVIDER_CALL_LDIV0 1
 #endif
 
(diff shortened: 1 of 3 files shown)


src/rp2_common/pico_divider/divider.S
```diff
@@ -11,16 +11,21 @@
 .cpu cortex-m0plus
 .thumb
 
+// PICO_CONFIG: PICO_DIVIDER_DISABLE_INTERRUPTS, Disable interrupts around division such that divider state need not be saved/restored in exception handlers, default=0, group=pico_divider
```
lurch, line 14: There's no "#ifndef ... #define ... #endif" default-value defined for PICO_DIVIDER_DISABLE_INTERRUPTS ?

kilograham:
nope, no point as it is only used in that source file

lurch:
The reason I raised this is that I thought the https://github.com/raspberrypi/pico-sdk/blob/master/tools/extract_configs.py script would get upset if it finds a `PICO_CONFIG` entry without a matching default `#define` value. But I just ran the script against this PR and it didn't get upset! But that's because the script [currently](https://github.com/raspberrypi/pico-sdk/blob/master/tools/extract_configs.py#L118) only checks `.h` and `.c` files and not `.S` files.
This means that the `PICO_DIVIDER_DISABLE_INTERRUPTS` currently won't appear in the `pico_configs.tsv`, and so also won't appear in https://github.com/raspberrypi/pico-project-generator  (but this may or may not be a bad thing? :shrug: )

liamfraser (approved):
LGTM if the config thing Andrew mentioned is okay

kilograham:
> The reason I raised this is that I thought the https://github.com/raspberrypi/pico-sdk/blob/master/tools/extract_configs.py script would get upset if it finds a `PICO_CONFIG` entry without a matching default `#define` value. But I just ran the script against this PR and it didn't get upset! But that's because the script [currently](https://github.com/raspberrypi/pico-sdk/blob/master/tools/extract_configs.py#L118) only checks `.h` and `.c` files and not `.S` files.
> This means that the `PICO_DIVIDER_DISABLE_INTERRUPTS` currently won't appear in the `pico_configs.tsv`, and so also won't appear in https://github.com/raspberrypi/pico-project-generator (but this may or may not be a bad thing? 🤷 )

yeah in all fairness you wouldn't want this in project generator - would be nice if it still showed up in the documentation. perhaps we need two classes of #define anyway then.

lurch:
> would be nice if it still showed up in the documentation. perhaps we need two classes of #define anyway then.

I'll hopefully get back to working on `extract_configs.py` again at some point, as I still need to modify it to pull out all the `PICO_BUILD_DEFINE` settings which we want to automatically include in the documentation...

PR #375, opened 2021-05-03
Disable core 0 SIO FIFO IRQ during core1 launch sequence

kilograham:
This way the user can pre-install an IRQ handler to receive FIFO requests once the core starts, without it receiving any during the init


liamfraser (approved):

lurch:
Apologies if this is a stupid question, but is this a non-symmetrical process (i.e. core0 can only launch tasks on core1), or can core1 also launch tasks on core0 ?

kilograham:
> Apologies if this is a stupid question, but is this a non-symmetrical process (i.e. core0 can only launch tasks on core1), or can core1 also launch tasks on core0 ?

only one "task" on each core; core0 is already running, so yes this is just for getting code running on core 1

lurch:
> only one "task" on each core;

Sorry, bad terminology on my part - I meant "process" (singular) rather than "tasks" (plural).

> so yes this is just for getting code running on core 1

Yeah, I understood that the bootrom starts core0 and puts core1 into "sleep" (again please excuse my bad terminology!), with the "launch sequence" required for core0 to "wake up" core1; but I wasn't sure if core1 (once it's running) can also put core0 into this  same "sleep" state?

PR #380, opened 2021-05-05
Add hardware_exception for setting exception handlers at runtime

### LINKED ISSUE #127: Mechanism for setting exception handlers
I could be wrong, but there doesn't appear to be any way of setting exception handlers from code other than to create specially-named functions such as `isr_svcall`, which the documentation suggestsnot to do.

The implementation numbers interrupts/exceptions from 0 for IRQ0. As these appear immediately after the vector table, it ought to be possible to use negative numbers to refer to the exceptions --- these are actually documented in the ARM documentation at https://developer.arm.com/documentation/dui0497/a/the-cortex-m0-processor/exception-model/exception-types, so `SVCall` is interrupt -5. However, calling `irq_set_exclusive_handler()` doesn't work because you can only use it on vectors which are already pointing at `__unhandled_user_irq`.

Given how clean the rest of the SDK's design is, it'd be nice to have a blessed-by-the-API way of doing this.

kilograham:
fixes #127


liamfraser (approved):

src/rp2_common/hardware_exception/exception.c
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "hardware/exception.h"
+#include "hardware/regs/m0plus.h"
```
lurch, line 8: Only a minor nit-pick, but this is already included in hardware/exception.h

src/rp2_common/hardware_exception/include/hardware/exception.h
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * \ingroup hardware_exception
+ */
+enum exception_number {
+    NMI_EXCEPTION        = -14,     /* Non Maskable Interrupt */
+    HARDFAULT_EXCEPTION  = -13,     /* HardFault Interrupt */
+    SVCALL_EXCEPTION     =  -5,     /* SV Call Interrupt */
+    PENDSV_EXCEPTION     =  -2,     /* Pend SV Interrupt */
+    SYSTICK_EXCEPTION    =  -1,     /* System Tick Interrupt */
+};
```
lurch, line 57: What happens for numbers outside of this list, e.g. -3 ?

src/rp2_common/hardware_exception/include/hardware/exception.h
```diff
@@ -0,0 +1,106 @@
...
+
+/*! \brief Exception handler function type
+ *  \ingroup hardware_exception
+ *
+ * All exceptions handlers should be of this type, and follow normal ARM EABI register saving conventions
```
lurch, line 62: "exceptions handlers" -> "exception handlers"

src/rp2_common/pico_standard_link/crt0.S
```diff
@@ -74,6 +74,11 @@ __vectors:
 .word isr_irq30
 .word isr_irq31
 
+// all default exception handlers do nothing, and we can check for them being set to our
+// default values by them pointing to between __defaults_isrs_start and __default_isrs_end
```
lurch, line 78: "by them pointing" -> "by pointing them" ?

lurch, line 78: I initially thought "by them pointing to" was a typo for "by pointing them to", but then I reread it a couple of times and figured out what it actually meant :wink: 
Perhaps "by seeing if they point to somewhere" would be clearer?

src/rp2_common/hardware_exception/include/hardware/exception.h
kilograham: they continue to exist

you can set handlers for them, but there is no point

PR #385, opened 2021-05-06
allow setting exception handler to the same value

kilograham:
Thiis might happen if exceptions are set dynamically on both cores, but VTOR is the same... this should indeed cause an error if the values differ, but not otherwise

### DIFF
--- a/src/rp2_common/hardware_exception/exception.c
+++ b/src/rp2_common/hardware_exception/exception.c
@@ -46,7 +46,7 @@ exception_handler_t exception_set_exclusive_handler(enum exception_number num, e
 #if !PICO_NO_RAM_VECTOR_TABLE
     uint32_t save = save_and_disable_interrupts();
     exception_handler_t current = exception_get_vtable_handler(num);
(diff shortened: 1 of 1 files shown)


lurch (approved):

src/rp2_common/hardware_exception/exception.c
```diff
@@ -46,7 +46,7 @@ exception_handler_t exception_set_exclusive_handler(enum exception_number num, e
 #if !PICO_NO_RAM_VECTOR_TABLE
     uint32_t save = save_and_disable_interrupts();
     exception_handler_t current = exception_get_vtable_handler(num);
-    hard_assert(exception_is_compile_time_default(current));
+    hard_assert(handler == current || exception_is_compile_time_default(current));
```
lurch, line 49: If `handler == current` I guess this whole function _effectively_ becomes a no-op?

kilograham: yes this is a nop

liamfraser (commented):
LGTM

PR #400, opened 2021-05-12
REDACTED: Adafruit qtpy rp2040 target starts CDC/ACM USB only once - suspect flash timing

wa1tnr:
Changes:

    PICO_FLASH_SPI_CLKDIV from 2 to 4
    (to slow the clock, by halving it)    and

    PICO_FLASH_SIZE_BYTES from 4 to 8
    (from branch 'develop' which already has the flash sized to 8MB)

Reference system Linux host PC will not enumerate /dev/ttyACM0
(or related device names) except upon UF2 firmware upload.

UF2 upload is fine and program will run once, until power is removed.
Then, never again.

This patch is meant to allow the (end-user authored) firmware/program
to run repeatedly, including cycling power to the target board.

First (SPI clock divisor) value tried was '4' (from '2').

No other experiments done, in search of the optimal clock divisor.

Not at all sure what the clock divisor 'does'.

Simple guess:

Going from 2, to 4 would 'halve' the clock frequency (divide it
by 2).

Halving the frequency would cause the subsystem to evolve
all events more slowly, which apparently helps in enumeration
of the CDC/ACM device to the host PC.


lurch:
@wa1tnr Please change this PR to target `develop` instead of `master`.

ping @tannewt and @dhalbert

wa1tnr:
@lurch I think I've complied - created a parallel branch and PR.

Don't know what to do with this one #400.

Thanks for your help!

PR #406, opened 2021-05-13
Small API additions and minor fixes

kilograham:

PR #407, opened 2021-05-13
Hw claim bugfix

Reneg973:
modified hw_is_claimed

Note a lock is not required as it is the access to a single memory cell which is locked automatically by HW


kilograham:
Thank you, but i had actually already fixed locally in #406, although you spotted a typo I hadn't... please feel free to review my change there

PR #414, opened 2021-05-16
add comment about using clk_gpout0 enable bit (Fixes #413)

### LINKED ISSUE #413: It is not obvious what unsetting and setting CLOCKS_CLK_GPOUT0_CTRL_ENABLE_BITS in clock_configure() achieves
(Likely reason being that I don't understand what's going on, but would still appreciate if someone who is knowledgeable enough, would do the explaining.)

The [`clock_configure()`](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_clocks/clocks.c#L43) function can be used to set up any of the clock. In that function there's a [pair](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_clocks/clocks.c#L74) [of](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_clocks/clocks.c#L105) `hw_{clear,set}_bits()` calls that write `CLOCKS_CLK_GPOUT0_CTRL_ENABLE_BITS` to the clock's `CTRL` register.

But the value of `CLOCKS_CLK_GPOUT0_CTRL_ENABLE_BITS`, `0x00000800`, does not make sense with `clk_sys` for example, because in that clock's `CTRL` register the bits 31:8 are marked as "Reserved" ([2.15.7. List of Registers](https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf)).

kahara:


lurch:
We don't accept PRs directly into `master`, can you target this against the `develop` branch instead please?

Wren6991 (approved):

PR #421, opened 2021-05-19
Only warn about Doxygen missing when Pico SDK is the top-level project

jonathangjertsen:
This silences the following warning for projects that include the Pico SDK:

```
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
```

I don't think projects that include the pico-sdk would want to generate the
Pico SDK documentation (perhaps with the sole exception of the example repo?),
so this warning is not all that relevant outside of the top-level project.


lurch:
Was this already fixed by #365 ?

jonathangjertsen:
Yes; I didn't notice there was a develop branch. Thanks!

Issue #424, opened 2021-05-21
add a board header for the Arduino Nano RP2040


2bndy5:
I could throw this together myself based on the available schematic, but I'm unsure about some things like the Flash SPI CLK divider to use... I don't know who on the @arduino team to ping about this (@facchinm ?). 

I don't want to overstep here. If this request is explicitly undesired, then I'll back off and close this issue.

My reason for wanting a _arduino_nano_rp2040.h_ in _/src/boards/include/boards/_ folder is because these board header files have proven extremely useful for 3rd-party libraries (that have incorporated native Pico SDK support) to choose a default set of pins in a rather agnostic way - leaving the actual choice of board to build for up to the user via `-DPICO_BOARD=<board_name>`.

lurch:
> these board header files have proven extremely useful

That's nice to hear :smiley: Well done @kilograham ! :clap:

PR #425, opened 2021-05-21
Add support for Arduino Nano RP2040 Connect

facchinm:
Beside the pin definitions for the board, the PR also includes a specially crafted second stage that allows setting the QE bit for the flash we are using.
Since that bit is sticky and it gets set during the factory test, all boards should "just work" using the second stage for W25Q080-like devices.

To avoid any maintenance burden, we can safely drop the second commit and the boards will keep working unless the QE bit is unset (very unlikely?)

Fixes https://github.com/raspberrypi/pico-sdk/issues/424
@sandeepmistry

### DIFF
--- a/src/boards/include/boards/arduino_nano_rp2040_connect.h
+++ b/src/boards/include/boards/arduino_nano_rp2040_connect.h
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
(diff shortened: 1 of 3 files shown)


kilograham:
@Wren6991 can you comment on this - is this the same bit that needs setting in factory test on Pico/B0? If it is otherwise identical to the W25Q080 then it might be worth dropping the special boot stage 2

facchinm:
@kilograham the two bytes write to SR registers doesn't work with this flash indeed, so this seconds stage just uses 0x31 (CMD_WRITE_STATUS2) to write SR2. After setting it, all the code is 100% the same as W25Q080

Wren6991:
Wow, yes, very small diff between this and `boot2_w25q080.S`. Just the different SR write and the different continuation bits. I think the copy paste is ok for now, it's a general problem that existed before this commit.

Wren6991 (approved):

PR #428, opened 2021-05-21
- interp_ fixes

Reneg973:
- fix interp_claim_lane (in case of interp1 and lane==1 "bit" was 0b11 instead of 0b1000)

- added missing function interp_unclaim_lane_mask()

### DIFF
--- a/src/rp2_common/hardware_interp/include/hardware/interp.h
+++ b/src/rp2_common/hardware_interp/include/hardware/interp.h
@@ -285,9 +285,9 @@ static inline void interp_set_force_bits(interp_hw_t *interp, uint lane, uint bi
 }
 
 typedef struct {
-    io_rw_32 accum[2];
-    io_rw_32 base[3];
-    io_rw_32 ctrl[2];
+    uint32_t accum[2];
+    uint32_t base[3];
+    uint32_t ctrl[2];
 } interp_hw_save_t;
 
 /*! \brief Save the specified interpolator state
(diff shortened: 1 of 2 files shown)


kilograham (changes requested):
we've already discussed not adding const to hardware pointers in another issue; remove these as the rest looks more useful.

Reneg973:
> we've already discussed not adding const to hardware pointers in another issue; remove these as the rest looks more useful.

Well, it was worth a try - you may have seen that I've splitted the const correct patch from the real bug.
It just that unlovely because my emerging const correct C++ classes need to spam a lot of const_cast<>s

kilograham:
> Well, it was worth a try - you may have seen that I've splitted the const correct patch from the real bug.
Are you referring to the previous PR?

> it's just that unlovely because my emerging const correct C++ classes need to spam a lot of const_cast<>s
The last I checked, your previous _const_ changes caused a bunch of new warnings. perhaps post a sample of your C++ code, so I can see where the pain is, and see if I am moved, I'm guessing this is const methods. Still philosophically i'm not convinced that what you call "correct" is actually philosophically so, since many thing (particularly interp) have side effects on read.

Reneg973:
> Are you referring to the previous PR?

No, I split this PR into 2 commits, first to fix the bug and second to add some costs.

> The last I checked, your previous _const_ changes caused a bunch of new warnings. perhaps post a sample of your C++ code, so I can see where the pain is, and see if I am moved, I'm guessing this is const methods. Still philosophically ...

I know and I am sorry that I just compiled my code and not all references to see where build problems exist. 
May it be possible to add some auto checks for PRs?

I don't want to discuss const. It may be used or may be not. It is just if you worked many years with it you don't want to miss it anymore (like me). And it is still not clear which strategy picosdk is following, as const is already used, but more or less sporadically.

kilograham (approved):

Issue #304, opened 2021-04-03
new_delete - Wsized-deallocation


Wiz-IO:
pico_standard_link\new_delete.cpp: warning: the program should also define 'void operator delete [](void*, unsigned int)' [-Wsized-deallocation]

workaround
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wsized-deallocation"
// the code
#pragma GCC diagnostic pop
```

kilograham:
compiler? version?

Wiz-IO:
gcc version 7.2.1 20170904 (release) 
[ARM/embedded-7-branch revision 255204] (GNU Tools for Arm Embedded Processors 7-2017-q4-major)

Wiz-IO:
4 delete - no warn
https://github.com/Wiz-IO/framework-wizio-pico/blob/main/SDK/pico/pico_standard_link/new_delete.cpp#L22

lurch:
Fixed by #430 ?

kilograham:
merged into develop

PR #430, opened 2021-05-21
Define `void operator delete[](void *p, std::size_t n)` in new_delete.cpp

jonathangjertsen:
This prevents the following warning when compiling with `-Wextra`:

```C
[redacted]/pico-sdk/src/rp2_common/pico_standard_link/new_delete.cpp:24:6: error: the program should also define 'void operator delete [](void*, unsigned int)' [-Werror=sized-deallocation]
   24 | void operator delete[](void *p) noexcept { std::free(p); }
      |      ^~~~~~~~
```

This function signature was introduced in C++14. I don't _think_ that should cause any issues, but I'm not sure.

### DIFF
--- a/src/rp2_common/pico_standard_link/new_delete.cpp
+++ b/src/rp2_common/pico_standard_link/new_delete.cpp
@@ -23,4 +23,6 @@ void operator delete(void *p) { std::free(p); }
 
 void operator delete[](void *p) noexcept { std::free(p); }
 
+void operator delete[](void *p, __unused std::size_t n) noexcept { std::free(p); }
+
 #endif
--- a/src/rp2_common/pico_standard_link/new_delete.cpp
+++ b/src/rp2_common/pico_standard_link/new_delete.cpp
@@ -17,12 +17,16 @@ void *operator new[](std::size_t n) {
     return std::malloc(n);
 }
 
-void operator delete(void *p, __unused std::size_t n) noexcept { std::free(p); }
-
 void operator delete(void *p) { std::free(p); }
 
 void operator delete[](void *p) noexcept { std::free(p); }
 
+#if __cpp_sized_deallocation
+
+void operator delete(void *p, __unused std::size_t n) noexcept { std::free(p); }
+
 void operator delete[](void *p, __unused std::size_t n) noexcept { std::free(p); }
+
+#endif
 
 #endif


Reneg973:
see https://github.com/raspberrypi/pico-sdk/issues/304

src/rp2_common/pico_standard_link/new_delete.cpp
```diff
@@ -23,4 +23,6 @@ void operator delete(void *p) { std::free(p); }
 
 void operator delete[](void *p) noexcept { std::free(p); }
 
+void operator delete[](void *p, __unused std::size_t n) noexcept { std::free(p); }
```
fivdi, line 26: The C++ new header only adds the declarations for the sized deallocation functions if `__cpp_sized_deallocation` is true.

So, should the sized deallocation function that is being added with this PR (and the already existing size deallocation function on [line 20](https://github.com/raspberrypi/pico-sdk/blob/5446fa7ae922dd1aced35028799c4f615fd0dad9/src/rp2_common/pico_standard_link/new_delete.cpp#L20) be conditionally compiled using `__cpp_sized_deallocation` (see [here](https://en.cppreference.com/w/cpp/feature_test))?

Here is what the new header has:

```c++
#if __cpp_sized_deallocation
void operator delete(void*, std::size_t) _GLIBCXX_USE_NOEXCEPT
  __attribute__((__externally_visible__));
void operator delete[](void*, std::size_t) _GLIBCXX_USE_NOEXCEPT
  __attribute__((__externally_visible__));
#endif
``´

jonathangjertsen: I think it will work either way. I looked up some other implementations of this fix and found one case where the conditional is included ([here](https://hg.mozilla.org/mozilla-central/rev/d346efc8449b)), and two where it was not ([here](https://hg.openjdk.java.net/jdk/jdk/rev/4ec9fc2b2f0d) and [here](https://github.com/cpputest/cpputest/issues/957)).

fivdi: I'm sure it will work too, however, I'm not sure if it's a good idea to make functions intended for C++14 and higher available in C++11.

jonathangjertsen: Updated

kilograham (approved):

PR #431, opened 2021-05-21
Prevent warnings about some unused parameters in pico_stdio_usb when building with -Wextra

jonathangjertsen:


Reneg973:
maybe a macro for better readability would be preferable?

like

`#define UNUSED_PARAMETER(x) (void)x`

fivdi:
The pico-sdk uses the `__unused` macro to prevent the compiler from warning about variables that are declared but never used. For example, see [resetd_xfer_cb](https://github.com/raspberrypi/pico-sdk/blob/1.1.2/src/rp2_common/pico_stdio_usb/reset_interface.c#L76-L78) in `reset_interface.c`.

`__unused` is defined [here](https://github.com/raspberrypi/pico-sdk/blob/1.1.2/src/host/pico_platform/include/pico/platform.h#L44-L46) and [here](https://github.com/raspberrypi/pico-sdk/blob/1.1.2/src/host/pico_platform/include/pico/platform.h#L64-L66) in `platform.h`.

jonathangjertsen:
I have amended the commit to use `__unused` instead

src/rp2_common/pico_stdio_usb/reset_interface.c
```diff
@@ -6,6 +6,7 @@
 #include "tusb.h"
 
 #include "pico/bootrom.h"
+#include "pico/platform.h"
```
kilograham, line 9: we don't generally include `pico/platform.h` directly unless this is a non SDK tool (usually `pico.h` is included because it also includes all user/board cinfig), and that should be included by all other library headers (as it is by `pico/bootrom.h`) so, i'd rather not include this

kilograham (changes requested):
also worth removing the `(void)` cast of an unused parameter elsewhere in the file

kilograham (approved):

Issue #422, opened 2021-05-19
DEBUG messages are printed by CMake versions < 3.15


jonathangjertsen:
The minimum CMake version is 3.13, but the DEBUG level for `message()` was added in CMake 3.15 - with earlier versions, the message is simply concatenated with `DEBUG` resulting in output like:

```
DEBUGPICO_GCC_TRIPLE defaulted to arm-none-eabi
...
DEBUGIncluding custom CMake file [redacted]/pico-sdk/src/common/pico_base/generate_config_header.cmake
DEBUGELF2UF2 will need to be built
```

(from PR #365)

PR #433, opened 2021-05-22
Prevent the literal string DEBUG from being prepended to some messages in CMake < 3.15

jonathangjertsen:
Fixes issue #422


kilograham (changes requested):
as a rule we've been prefixing our methods with `pico_`

jonathangjertsen:
> as a rule we've been prefixing our methods with `pico_`

Fixed

kilograham (approved):

PR #437, opened 2021-05-23
Fix wrong format string in alarm_pool_dump_key

jonathangjertsen:
Fixes the following warning when building for host

```
[...]/pico-sdk/src/common/pico_time/time.c: In function 'alarm_pool_dump_key':
[...]/pico-sdk/src/common/pico_time/time.c:282:15: warning: format '%ld' expects argument of type 'long int', but argument 2 has type 'uint64_t' {aka 'long long unsigned int'} [-Wformat=]
     printf("%ld", to_us_since_boot(get_entry(pool, id)->target));
             ~~^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
             %I64d
```

`PRIu64` and `PRId64` are in the C99 standard, so maybe all uses of the nonstandard `%llu` and `%lld` specifiers should be replaced with those?

### DIFF
--- a/src/common/pico_time/time.c
+++ b/src/common/pico_time/time.c
@@ -5,6 +5,7 @@
  */
 
 #include <limits.h>
+#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include "pico.h"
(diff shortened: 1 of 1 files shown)


lurch:
Doing a quick `git grep` of the source code for `to_us_since_boot`, it seems that the SDK variously uses:
* "%lld" https://github.com/raspberrypi/pico-sdk/blob/master/src/common/pico_time/time.c#L277
* PRIu64 https://github.com/raspberrypi/pico-sdk/blob/master/test/pico_stdlib_test/pico_stdlib_test.c#L34
* PRIi64 https://github.com/raspberrypi/pico-sdk/blob/master/test/pico_time_test/pico_time_test.c#L179

kilograham (approved):

PR #438, opened 2021-05-23
Implement `uart_write_blocking` and `uart_read_blocking` for host

jonathangjertsen:


src/host/hardware_uart/uart.c
```diff
@@ -88,10 +88,22 @@ size_t uart_is_readable(uart_inst_t *uart) {
 }
 
 // Write len bytes directly from src to the UART
-//void uart_write_blocking(uart_inst_t uart, const uint8_t *src, size_t len);
+void uart_write_blocking(uart_inst_t *uart, const uint8_t *src, size_t len)
+{
```
kilograham, line 92: please match code formatting

kilograham (changes requested):
Nice to have someone using `host` mode - i thought i was the only one! It definitely needs some fleshing out in places.

If you are not aware of `pico_host_sdl` it is worth checking that out too, though everything that can reasonably be implemented in pico_sdk should be.

kilograham (approved):

PR #443, opened 2021-05-24
Fix function-name misspelling

lurch:
If "init tty" is `_inittty` then "reset tty" ought to be `_resettty`


Reneg973:
it may be static too?

kilograham (approved):

PR #444, opened 2021-05-24
Optional feature to get the max level that has ever been held by a queue

jonathangjertsen:
I'm using this in my application to get a better idea of how many entries I need to allocate to avoid blocking or losing data, and to debug performance issues. It adds some overhead, hence the opt-in.


kilograham (approved):
seems like it might be useful... thanks for the addition

PR #451, opened 2021-05-27
Add function to get the currently selected ADC channel

jonathangjertsen:
This is most useful in combination with round-robin sampling to avoid having to track it manually.

### DIFF
--- a/src/rp2_common/hardware_adc/include/hardware/adc.h
+++ b/src/rp2_common/hardware_adc/include/hardware/adc.h
@@ -88,6 +88,15 @@ static inline void adc_select_input(uint input) {
     hw_write_masked(&adc_hw->cs, input << ADC_CS_AINSEL_LSB, ADC_CS_AINSEL_BITS);
 }
 
+/*! \brief  Get the currently selected ADC input channel
+ *  \ingroup hardware_adc
+ *
+ * \return The currently selected input channel. 0...3 are GPIOs 26...29 respectively. Input 4 is the onboard temperature sensor.
+ */
+static inline uint adc_get_selected_input(void) {
+    return (adc_hw->cs & ADC_CS_AINSEL_BITS) >> ADC_CS_AINSEL_LSB;
+}
+
 /*! \brief  Round Robin sampling selector
  *  \ingroup hardware_adc
  *


src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -88,6 +88,16 @@ static inline void adc_select_input(uint input) {
     hw_write_masked(&adc_hw->cs, input << ADC_CS_AINSEL_LSB, ADC_CS_AINSEL_BITS);
 }
 
+/*! \brief  ADC get input
```
lurch, line 91: "ADC get input" is probably a bit ambiguous? It could either mean "get the input channel" or "get the input value" ? :wink:

src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -88,6 +88,16 @@ static inline void adc_select_input(uint input) {
...
+ *  \ingroup hardware_adc
+ *
+ * Returns the currently selected input. 0...3 are GPIOs 26...29 respectively.
```
lurch, line 94: Could you make this a `\return` statement please? See e.g. line 132

src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -88,6 +88,16 @@ static inline void adc_select_input(uint input) {
...
+ * Input 4 is the onboard temperature sensor.
+ */
+static inline uint adc_selected_input(void) {
```
lurch, line 97: @kilograham obviously has the final decision, but I wonder if a function name of `adc_get_selected_input` would be slightly clearer? (In hindsight I guess it's a shame that `adc_select_input` wasn't named `adc_select_input_channel`? Because that would allow this function to be named `adc_get_input_channel`.)

jonathangjertsen:
Good suggestions @lurch , I've updated accordingly

src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -88,6 +88,15 @@ static inline void adc_select_input(uint input) {
     hw_write_masked(&adc_hw->cs, input << ADC_CS_AINSEL_LSB, ADC_CS_AINSEL_BITS);
 }
 
+/*! \brief  Get the currently selected ADC input channel
+ *  \ingroup hardware_adc
+ *
+ * \return Returns the currently selected input channel. 0...3 are GPIOs 26...29 respectively. Input 4 is the onboard temperature sensor.
```
lurch, line 94: Don't need the "Returns". So just ` * \return The currently selected ...`

lurch (approved):
LGTM :+1:

src/rp2_common/hardware_adc/include/hardware/adc.h
kilograham: yes, thank adc_get_selected_input is the better name... i did go thru and tried to make all names logical and consistent but I may have missed a couple.

kilograham (approved):

PR #457, opened 2021-05-28
Allow lengthening xosc startup delay with a compile option

dhalbert:
This PR allows you to specify a multiplier, at compile time to lengthen the default XOSC startup time. The multiplier, `PICO_XOSC_STARTUP_DELAY_MULTIPLIER`, is `1` by default.

Motivation: On a few samples of some Adafruit RP2040 boards, the XOSC takes much longer to stabilize than would be expected. This appears to be due to some mismatch between the chosen XOSC crystal and capacitors. Sometimes the problem is intermittent, and may depend on handling the board while restarting, humidity, etc. We are going to fix future boards, but this PR is a software fix that accommodates all boards.

The Adafruit board include files set `PICO_XOSC_STARTUP_DELAY_MULTIPLIER` to be 64. This was chosen empirically. 16 can be too low sometimes. 32 seems to work, and so 64 includes a safety margin.

This issue has been discussed at length in #401. The default XOSC startup time is used in the boot ROM when using USB, but the XOSC startup problem does not seem to be an issue: the boards all can connect to USB successfully. Also relevant: https://github.com/raspberrypi/pico-feedback/issues/158/.

I've tested this on an Adafruit QT Py RP2040 that exhibits the problem, with the `pico-examples` `blink` example.

One minor thing: `xosc.c` did not end in a newline, and my editor fixed that automatically.

### DIFF
--- a/src/rp2_common/hardware_xosc/xosc.c
+++ b/src/rp2_common/hardware_xosc/xosc.c
@@ -13,6 +13,11 @@
 #include "hardware/regs/xosc.h"
 #include "hardware/structs/xosc.h"
 
+// Allow lengthening startup delay to accommodate slow-starting oscillators
+#ifndef PICO_XOSC_STARTUP_DELAY_MULTIPLIER
+#define PICO_XOSC_STARTUP_DELAY_MULTIPLIER 1
+#endif
+
 void xosc_init(void) {
     // Assumes 1-15 MHz input
     assert(XOSC_MHZ <= 15);
@@ -20,7 +25,9 @@ void xosc_init(void) {
 
     // Set xosc startup delay
     uint32_t startup_delay = (((12 * MHZ) / 1000) + 128) / 256;
-    xosc_hw->startup = startup_delay;
+
+    // Allow lengthening startup delay to accommodate slow-starting oscillators
+    xosc_hw->startup = startup_delay * PICO_XOSC_STARTUP_DELAY_MULTIPLIER;
 
     // Set the enable bit now that we have set freq range and startup delay
     hw_set_bits(&xosc_hw->ctrl, XOSC_CTRL_ENABLE_VALUE_ENABLE << XOSC_CTRL_ENABLE_LSB);
--- a/src/rp2_common/hardware_xosc/xosc.c
+++ b/src/rp2_common/hardware_xosc/xosc.c
@@ -13,7 +13,7 @@
 #include "hardware/regs/xosc.h"
 #include "hardware/structs/xosc.h"
 
-// Allow lengthening startup delay to accommodate slow-starting oscillators
+// PICO_CONFIG: PICO_XOSC_STARTUP_DELAY_MULTIPLIER, Multiplier to lengthen xosc startup delay to accommodate slow-starting oscillators, type=int, min=1, default=1, group=hardware_xosc
 #ifndef PICO_XOSC_STARTUP_DELAY_MULTIPLIER
 #define PICO_XOSC_STARTUP_DELAY_MULTIPLIER 1
 #endif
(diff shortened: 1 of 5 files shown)


src/rp2_common/hardware_xosc/xosc.c
```diff
@@ -13,14 +13,21 @@
 #include "hardware/regs/xosc.h"
 #include "hardware/structs/xosc.h"
 
+// Allow lengthening startup delay to accommodate slow-starting oscillators
```
kilograham, line 16: Can you change this to a `// PICO_CONFIG: ` so that it shows up in the docs (see other places)

kilograham (changes requested):

src/rp2_common/hardware_xosc/xosc.c
dhalbert: Sure. Should I also do some kind of checking on the value to make sure the resultant multiplication doesn't overflow the register field of 14 bits? The `struct` definition for the register does not take that into account.

dhalbert: image

https://github.com/raspberrypi/pico-sdk/blob/afc10f3599c27147a6f34781b7102d86f58aa5f6/src/rp2040/hardware_structs/include/hardware/structs/xosc.h#L15-L22

dhalbert:
I pushed a `PICO_CONFIG:` fix, though I'm a little confused as to where I should see it. I built the doxygen docs and `PICO_XOSC_STARTUP_DELAY_MULTIPLIER` shows up in the `doxygen/xml/` but not the `doxygen/xml`. However, there are many other `PICO_CONFIG`d definitions that similarly are only in the xml. They don't show up in the search, for instance.

lurch:
The `PICO_CONFIG` entries are extracted by `tools/extract_configs.py` to `pico_configs.tsv` which is then used by e.g. https://github.com/raspberrypi/pico-project-generator
They're also included into Appendix B of https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-c-sdk.pdf by internal build-tools.

src/rp2_common/hardware_xosc/xosc.c
kilograham: yes makes sense - please follow invalid_params_if() style... (which will require you to add some macros to xosc.h - just copy/paste from somewhere else)

dhalbert:
I ended up not using `invalid_params_if()`, which seems more meant for run-time checking. Instead I did some cpp-compile-time checking, which is done similarly in other places.

Most recent changes:
- Check that `XOSC_MHZ` is in range (was done with an `assert()`).
- Check that the delay calculation fits in 14 bits.
- Use `XOSC_MHZ` instead of the literal `12` in the delay calculation.

kilograham (approved):

src/rp2_common/hardware_xosc/xosc.c
```diff
@@ -13,14 +13,30 @@
...
+
+// PICO_CONFIG: PICO_XOSC_STARTUP_DELAY_MULTIPLIER, Multiplier to lengthen xosc startup delay to accommodate slow-starting oscillators, type=int, min=1, default=1, group=hardware_xosc
```
kilograham, line 18: stylistically (going forward, there may be a few stragglers) we are trying to put these  defaulted values in the header in case any other code (source files) cares about the actual value. can you move this to xsoc.h

kilograham (changes requested):

kilograham (approved):

src/rp2_common/hardware_xosc/xosc.c
dhalbert: Sure. Moved PICO_CONFIG and default setting to `src/rp2_common/hardware_xosc/include/hardware/xosc.h`, and changed `#include` in xosc.c to accommodate this.

PR #460, opened 2021-05-30
Fix flash size for Adafruit RP2040 boards

eightycc:
Adafruit QT Py RP2040 and Itsybitsy RP2040 ship with 8M of flash.


lurch:
ping @hathach @dhalbert

dhalbert (approved):
This is right, thanks!

hathach:
this already fixed in develop, we should find a way to have develop as main branch target of a new PR.

eightycc:
Doh, I didn't realize there was a develop branch. Should I base future pulls off of develop?

hathach:
That is very common mistake people make here. I made a few one myself as well

PR #465, opened 2021-06-02
Add usb device dpram to svd file. Fixes #351

### LINKED ISSUE #351: Missing memory layout in svd
I'm trying to implement USB driver in rust and I cannot find endpoint and buffer controll registers. It looks like there is nothing from base 0x5010000 to 0x5011000.
I confirmed that there should be controller structures in datasheet and are used in c implementation.

Am I missing something or there is some issue with svd?

liamfraser:


Wren6991 (approved):
Nice, matches layout in handwritten struct header and in datasheet

PR #466, opened 2021-06-02
fix qtpy rp2040 uart rx rev B

hathach:
This Pr address the changes in production of Adafruit qtpy rp2040 revB with uart rx pin from 9 to 5. https://cdn-learn.adafruit.com/assets/assets/000/101/678/original/adafruit_products_QTRP_sch.png?1618955879

also add static check for USB DPRAM (a bit randomly, but it could be useful in the future).

### DIFF
--- a/src/boards/include/boards/adafruit_qtpy_rp2040.h
+++ b/src/boards/include/boards/adafruit_qtpy_rp2040.h
@@ -30,7 +30,7 @@
 #endif
 
 #ifndef PICO_DEFAULT_UART_RX_PIN
-#define PICO_DEFAULT_UART_RX_PIN 9
+#define PICO_DEFAULT_UART_RX_PIN 5
 #endif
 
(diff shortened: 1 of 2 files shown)


kilograham (approved):

lurch:
Just out of curiosity, how many qtpy rp2040s were made before this change? (Could this PR lead to some example code not working if the user has an "early" qtpy rp2040?)

hathach:
> Just out of curiosity, how many qtpy rp2040s were made before this change? (Could this PR lead to some example code not working if the user has an "early" qtpy rp2040?)

To be honest, I am not entirely sure, one of the early prototype is revA. I think most of production board is revB. 3 of my production boards are revB.

dhalbert:
@ladyada I'm not sure we sold any Rev A?

ladyada:
correct - that was just my testing protos

PR #470, opened 2021-06-03
Add gpio_set_irqover to match inover/outover/oeover (fixes #265)

### LINKED ISSUE #265: No function for gpio irq over
There is functions to invert the signal for gpio out, in and out enable (oe) but there is no function for the gpio irq signal invert, something like gpio_set_irqover to stay consistent with the other functions.
Maybe it could also be nice in the gpio_set_inover to add a third parameter like bool irqover which would invert both in and irq if true or only in if false.
I can see cases where you would want both to be inverted, for example a debounce/filter where you want the IT to start a delay and after this delay, check the state of the pin.

Wren6991:
fixes #265

### DIFF
--- a/src/host/hardware_gpio/gpio.c
+++ b/src/host/hardware_gpio/gpio.c
@@ -27,6 +27,10 @@ void gpio_set_pulls(uint gpio, bool up, bool down) {
 
 }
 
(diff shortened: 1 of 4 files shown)


kilograham (commented):
cool

PR #471, opened 2021-06-03
Added board definition for upcoming Pimoroni product

ZodiusInfuser:


kilograham:
Are you happy with this; if so, i will squeeze it into 1.2.0 release which is moments away!

ZodiusInfuser:
I am, yes. It's a pretty basic config :)

PR #475, opened 2021-06-03
include pico.h so that uint is defined

fivdi:
Fixes #474

PR #476, opened 2021-06-04
portability issue on RPI2 -> uint to uin32_t

jbinko:
Compiler on RPI2 was failing on type uint. I changed code from uint to uin32_t


fivdi:
@jbinko This appears to be a duplicate of #475 which was discussed at #474

kilograham:
duplicate

PR #479, opened 2021-06-07
fix pioasm python output

urish:
While working on the [online pioasm](https://wokwi.com/tools/pioasm), I found several PIO instructions that result in invalid python code. Here is a small program that demonstrates the issue:

```
.program python_issue

    push block
    wait 0 irq 1 rel
    irq clear 1 rel
```

And the resulting Python program:

```python
# -------------------------------------------------- #
# This file is autogenerated by pioasm; do not edit! #
# -------------------------------------------------- #

import rp2
from machine import Pin
# ----------- #
# python_test #
# ----------- #

@rp2.asm_pio()
def python_test():
    wrap_target()
    push(, block)                         # 0
    wait(0, irq, 1 rel)                   # 1
    irq(clear 1 rel)                      # 2
    wrap()
```

After this fix, the above program compiles to a valid python syntax:

```python
# -------------------------------------------------- #
# This file is autogenerated by pioasm; do not edit! #
# -------------------------------------------------- #

import rp2
from machine import Pin
# ----------- #
# python_test #
# ----------- #

@rp2.asm_pio()
def python_test():
    wrap_target()
    push(block)                           # 0
    wait(0, irq, rel(1))                  # 1
    irq(clear, rel(1))                    # 2
    wrap()
```

### DIFF
--- a/tools/pioasm/python_output.cpp
+++ b/tools/pioasm/python_output.cpp
@@ -195,9 +195,12 @@ struct python_output : public output_format {
                         if (arg2 & 0x8u) {
                             invalid = true;
                         } else {
-                            guts = "irq, " + std::to_string(arg2 & 7u);
+                            guts = "irq, ";
+                            auto irq = std::to_string(arg2 & 7u);
                             if (arg2 & 0x10u) {
-                                guts += " rel";
+                                guts += "rel(" + irq + ")";
+                            } else {
+                                guts += irq;
                             }
                         }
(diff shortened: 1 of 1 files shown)


kilograham (approved):
Ha, yeah, thought there might be some of these... didn't get a lot of testing as it isn't really used in any actual micropython workflow - just added for completeness/porting

thanks for fixing.

PR #480, opened 2021-06-08
Exclude unused shared IRQ handler code when  PICO_DISABLE_SHARED_IRQ_HANDLERS=1

lurk101:
Exclude unusable code

### DIFF
--- a/src/rp2_common/hardware_irq/irq.c
+++ b/src/rp2_common/hardware_irq/irq.c
@@ -63,7 +63,7 @@ void irq_set_pending(uint num) {
     *((io_rw_32 *) (PPB_BASE + M0PLUS_NVIC_ISPR_OFFSET)) = 1u << num;
 }
 
-#if PICO_MAX_SHARED_IRQ_HANDLERS
+#if !PICO_DISABLE_SHARED_IRQ_HANDLERS
 // limited by 8 bit relative links (and reality)
 static_assert(PICO_MAX_SHARED_IRQ_HANDLERS >= 1 && PICO_MAX_SHARED_IRQ_HANDLERS < 0x7f, "");
 
@@ -133,6 +133,7 @@ irq_handler_t irq_get_exclusive_handler(uint num) {
 }
 
 
+#if !PICO_DISABLE_SHARED_IRQ_HANDLERS
 static uint16_t make_branch(uint16_t *from, void *to) {
     uint32_t ui_from = (uint32_t)from;
     uint32_t ui_to = (uint32_t)to;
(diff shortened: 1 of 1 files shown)


kilograham (changes requested):
well obviously it was an oversight/unfinished thought (as evidenced by the empty #ifdef). can you

1) perhaps reword your PR so it is a description not a question. note also that this stuff is likely GCed by the linker anyway (i haven't actually double checked), so is more of a warning nuisance than anything else
2)  factor  out a #define in this file based on the other two/(three/) for a combination of them disables a bunch of things

lurk101:
Yes, obviously.
1 - These functions are indeed GC'd if you tell the compiler and the linker via the section options, but presently you don't by default it seems.
2 - PICO_DISABLE_SHARED_IRQ_HANDLERS=1 and PICO_MAX_SHARED_IRQ_HANDLERS=0 look like they were applied as meaning the same. I think it would have made more sense to disallow PICO_MAX_SHARED_IRQ_HANDLERS=0 if PICO_DISABLE_SHARED_IRQ_HANDLERS=1, and only use it for chain slot array sizing.

I've reworked and rebaed to a single commit.

kilograham:
> Yes, obviously.
> 1 - These functions are indeed GC'd if you tell the compiler and the linker via the section options, but presently you don't by default it seems.

That should be the default, unless you've made your own frankenbuild.

lurk101:
When I dump CMAKE_C_FLAGS for the default Release build I get:
`-- CFlags -mcpu=cortex-m0plus -mthumb`
I normally add
```
if ("${CMAKE_BUILD_TYPE}" MATCHES "Release")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ffunction-sections -fdata-sections -Wl,–gc-sections -O3")
endif()
```
to get
`-- CFlags -mcpu=cortex-m0plus -mthumb -ffunction-sections -fdata-sections -Wl,–gc-sections -O3`

kilograham:
They aren't necessarily in `CMAKE_C_FLAGS`, that is just the basic global compiler flags.

If you do a `make VERBOSE=1` or whatever you'll see they've been added (via `pico_standard_link` as it happens)

lurk101:
Ah, ok. Good to know. Save me that CMakeList edit with every project.

lurch:
> 2 - PICO_DISABLE_SHARED_IRQ_HANDLERS=1 and PICO_MAX_SHARED_IRQ_HANDLERS=0 look like they were applied as meaning the same. I think it would have made more sense to disallow PICO_MAX_SHARED_IRQ_HANDLERS=0 if PICO_DISABLE_SHARED_IRQ_HANDLERS=0, and only use it for chain slot array sizing.

Perhaps the min and max values checked here https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_irq/irq.c#L68
should also be mentioned in the PICO_CONFIG line here https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_irq/include/hardware/irq.h#L11
? :man_shrugging:

kilograham:
subsumed by #496

PR #481, opened 2021-06-08
remove -Winline

kilograham:
Had been vaguely wondering why we get some rather dubiously useful warnings about inlining with some compilers. on looking at `make VERBOSE=1` output for something else I noticed we were passing `-Winline`

Looking at history this was added as temporary for a debug check during development (as you'd expect since it is not the correct way to add compiler options)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,8 +33,6 @@ if (NOT TARGET _pico_sdk_inclusion_marker)
     add_subdirectory(tools)
     add_subdirectory(src)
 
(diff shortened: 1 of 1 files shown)


lurch:
Ahh. Removal of dubiously-useful-warnings is good!

Does this PR make this line redundant too? https://github.com/raspberrypi/pico-sdk/blob/master/test/kitchen_sink/CMakeLists.txt#L73
:man_shrugging:

liamfraser (approved):

PR #484, opened 2021-06-16
Using ' inline __always_inline' also for GNUC 7.

zapta:
Per the discussion at https://www.raspberrypi.org/forums/viewtopic.php?f=145&t=313961

GNU7 also expects the 'inline' keyword and generates warnings otherwise.

### DIFF
--- a/src/rp2_common/pico_platform/include/pico/platform.h
+++ b/src/rp2_common/pico_platform/include/pico/platform.h
@@ -49,7 +49,7 @@ extern "C" {
 
 #define __packed_aligned __packed __aligned(4)
 
-#if defined(__GNUC__) && __GNUC__ < 7
+#if defined(__GNUC__) && __GNUC__ <= 7
 #define __force_inline inline __always_inline
 #else
 #define __force_inline __always_inline


kilograham:
can you confirm the actual compiler version you saw this on (i.e. `gcc --version`)

zapta:
I don't have gcc in my path but when I build in platformio I get this

 - toolchain-gccarmnoneeabi 1.70201.0 (7.2.1)

When I run the gcc from the platformio cache I get this

~/.platformio/packages/toolchain-gccarmnoneeabi/bin$ ./arm-none-eabi-gcc
--version
arm-none-eabi-gcc.exe (GNU Tools for Arm Embedded Processors
7-2017-q4-major) 7.2.1 20170904 (release) [ARM/embedded-7-branch revision
255204]

Does it answer your question?

kilograham:
Ok this thoroughly confused me because it worked in 7 for me (but i misread q4 as later than q2 when they are just quarters not absolute)... this was not a GCC change, but rather a newlib change:

https://github.com/bminor/newlib/commit/82339fcd717b4365caf65abf67072840342b4730

And got picked up by arm-none-eabi-gcc somewhere between 7.2.1 and 7.3.1... in any case, it seems that <= 7 although not strictly correct, seems to not cause any complaints

kilograham (dismissed):

kilograham (approved):

zapta:
Thanks @kilograham. If it helps, I can send a second pull request that puts it in a #ifndef condition, in case somebody needs an escape route and override it manually with gcc -D.

#ifndef __force_inline
<the existing set up of __force_inline>
#endif

PR #485, opened 2021-06-17
Allow one of float/double to have 'none' impl but not the other - pre…

kilograham:
…viously caused a link error


lurch:
Is the tinyusb bump in this PR accidental?

Wren6991 (approved):
Had to look up the semantics of multiple weak symbols :) fix looks good, makes sense.

PR #493, opened 2021-06-27
U64 divide save restore issue

kilograham:
The 64 bit divide code path made repeated use of the divider with the same DIVISOR without reloading it, which fails if an interrupt which uses the divider happens *between* the actual hardware divisions. In this case the hardware divider is not _dirty_, so the nested division does not bother to save/restore the divider state.

Fixed by reloading the divisor each time (it was in a CPU register anyway)

### DIFF
--- a/test/pico_divider_test/pico_divider_nesting_test.c
+++ b/test/pico_divider_test/pico_divider_nesting_test.c
@@ -158,6 +158,23 @@ void test_nesting() {
             if (0xffffffffu != a * 11 + b) {
                 FAILED();
             }
+            static uint64_t z64;
+            z64 -= 47;
+            uint64_t a64 = z64 / -13635;
+            uint64_t b64 = z64 % -13635;
+            if (z64 != a64 * -13635 + b64) {
+                FAILED();
+            }
+            // specifically check 64/32 divide
+            static uint64_t c64 = 0x13ffffffffull;
+            static uint32_t cd = 1;
+            a64 = c64 / cd;
+            b64 = c64 % cd;
+            if (c64 != a64 * cd + b64) {
+                FAILED();
+            }
+            cd++;
+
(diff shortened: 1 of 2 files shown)


kilograham:
p.s. fortunately the boot rom doesn't do this ever (i checked)

test/pico_divider_test/pico_divider_nesting_test.c
```diff
@@ -158,6 +158,23 @@ void test_nesting() {
...
+            static uint64_t z64;
+            z64 -= 47;
+            uint64_t a64 = z64 / -13635;
+            uint64_t b64 = z64 % -13635;
+            if (z64 != a64 * -13635 + b64) {
+                FAILED();
+            }
+            // specifically check 64/32 divide
+            static uint64_t c64 = 0x13ffffffffull;
+            static uint32_t cd = 1;
```
lurch, line 170: I dunno if it'd be "nicer" (more readable?) to declare these outside the `for` loop (like you did with `uint z = 0;`), rather than declaring them as `static` inside the `for` loop? :shrug: 

The first time I saw
```
static uint32_t cd = 1;
a64 = c64 / cd;
```
I thought "Huh, what's the point of testing divide-by-one?" :wink:  (it was only later that I spotted the `cd++;` further down)

(but then I'm not a regular C programmer, so perhaps this is just a common idiom that I'm not familiar with?)

kilograham: yeah i thought about that when i wrote it, but didn't feel motivated enough then, and so probably less so now... these tests are a bit opaque anyway

Wren6991 (approved):
> which fails if an interrupt which uses the divider happens between the actual hardware divisions

Ow. Fix looks good

PR #495, opened 2021-06-28
Split recursive mutex into their own functions (was Reduce performance hit of recursive mutex)

kilograham:
~~The recursive mutex code seemed to negatively effect some time critical code, so the code was reworked slightly
for the regular mutex methods, but additional explciit mutex_funx_r() and mutex_func_nr() vairants were added for
recursive mutexes and non recursive mutexes respectively.~~

~~It might have been nice to swtich to different mutex types for recursive/non recursive (which i had considered
when adding recursive mutexes), but that is a bit too fraught with backwards compatibility issues.~~

~~I imagine not _that_ many people care about the performance of this so critically that they much care (which is why the _r and _nr names are a bit oblique - the naming was a bit confusing as they will silently fail in release mode for the wrong type of mutex)~~

As mentioned in the title update,`mutex_t` and `mutex_` are reverted to non recursive versions (pre SDK1.2.0) and new `recursive_mutex_t` and `recursive_mutex_` functions have been added

`PICO_MUTEX_ENABLE_SDK120_COMPATIBILITY` flag has been added to allow old SDK1.2.0 compatibility (i.e. mutex_t can be used recursively or not) but this is slower (and is will be removed in a future version)

NOTE: incorrect spelling of `mutex_is_initialzed` has been changed to `mutex_is_initialized` as this is not a function people have likely use externally.

### DIFF
--- a/src/common/pico_sync/include/pico/mutex.h
+++ b/src/common/pico_sync/include/pico/mutex.h
@@ -19,20 +19,20 @@ extern "C" {
  * \brief Mutex API for non IRQ mutual exclusion between cores
  *
  * Mutexes are application level locks usually used protecting data structures that might be used by
- * multiple threads of executions. Unlike critical sections, the mutex protected code is not necessarily
+ * multiple threads of execution. Unlike critical sections, the mutex protected code is not necessarily
  * required/expected to complete quickly, as no other sytem wide locks are held on account of an acquired mutex.
  *
  * When acquired, the mutex has an owner (see \ref lock_get_caller_owner_id) which with the plain SDK is just
  * the acquiring core, but in an RTOS it could be a task, or an IRQ handler context.
  *
  * Two variants of mutex are provided; \ref mutex_t (and associated mutex_ functions) is a regular mutex that cannot
- * be acquired recursively by the same owner (a deadlock will occur if you try). \ref recursive_mutex
+ * be acquired recursively by the same owner (a deadlock will occur if you try). \ref recursive_mutex_t
  * (and associated recursive_mutex_ functions) is a recursive mutex that can be recursively obtained by
  * the same caller, at the expense of some more overhead when acquiring and releasing.
  *
- * It is generally a bad idea to call blocking mutex_ or recursive_mutex_ functions from within an IRQ handler
- . It is valid to call \ref mutex_try_enter or \ref recursive_mutex_try_enter from within an IRQ handler, if the operation
- * that would be conducted under lock can be skipped if the mutex is locked (at least by the same core).
+ * It is generally a bad idea to call blocking mutex_ or recursive_mutex_ functions from within an IRQ handler.
+ * It is valid to call \ref mutex_try_enter or \ref recursive_mutex_try_enter from within an IRQ handler, if the operation
+ * that would be conducted under lock can be skipped if the mutex is locked (at least by the same owner).
  *
  * NOTE: For backwards compatibility with version 1.2.0 of the SDK, if the define
  * PICO_MUTEX_ENABLE_SDK120_COMPATIBILITY is set to 1, then the the regular mutex_ functions
@@ -121,7 +121,7 @@ bool mutex_try_enter(mutex_t *mtx, uint32_t *owner_out);
  * caller will *NOT* own the mutex.
  *
  * \param mtx Pointer to recursive mutex structure
- * \param owner_out If mutex was already owned, and this pointer is non-zero, it will be filled in with the core number of the current owner of the mutex
+ * \param owner_out If mutex was already owned, and this pointer is non-zero, it will be filled in with the owner id of the current owner of the mutex
  */
 bool recursive_mutex_try_enter(recursive_mutex_t *mtx, uint32_t *owner_out);
 
@@ -130,8 +130,7 @@ bool recursive_mutex_try_enter(recursive_mutex_t *mtx, uint32_t *owner_out);
  *
  * Wait for up to the specific time to take ownership of the mutex. If the caller
  * can be granted ownership of the mutex before the timeout expires, then true will be returned
- * and the caller will own the mutex, otherwise false will be returned and the calling
- * core will *NOT* own the mutex.
+ * and the caller will own the mutex, otherwise false will be returned and the caller will *NOT* own the mutex.
  *
  * \param mtx Pointer to mutex structure
  * \param timeout_ms The timeout in milliseconds.
@@ -144,8 +143,8 @@ bool mutex_enter_timeout_ms(mutex_t *mtx, uint32_t timeout_ms);
  *
  * Wait for up to the specific time to take ownership of the recursive mutex. If the caller
  * already has ownership of the mutex or can be granted ownership of the mutex before the timeout expires,
- * then true will be returned and the caller will own the mutex, otherwise false will be returned and the calling
- * core will *NOT* own the mutex.
+ * then true will be returned and the caller will own the mutex, otherwise false will be returned and the caller
+ * will *NOT* own the mutex.
  *
  * \param mtx Pointer to recursive mutex structure
  * \param timeout_ms The timeout in milliseconds.
@@ -185,7 +184,7 @@ bool recursive_mutex_enter_timeout_us(recursive_mutex_t *mtx, uint32_t timeout_u
  *  \ingroup mutex
  *
  * Wait until the specific time to take ownership of the mutex. If the caller
- * core can be granted ownership of the mutex before the timeout expires, then true will be returned
+ * can be granted ownership of the mutex before the timeout expires, then true will be returned
  * and the calling core will own the mutex, otherwise false will be returned and the calling
  * core will *NOT* own the mutex.
  *
@@ -204,7 +203,7 @@ bool mutex_enter_block_until(mutex_t *mtx, absolute_time_t until);
  * core will *NOT* own the mutex.
  *
  * \param mtx Pointer to recursive mutex structure
- * \param until The time after which to return if the core cannot take ownership of the mutex
+ * \param until The time after which to return if the caller cannot be granted ownership of the mutex
  * \return true if mutex now owned, false if timeout occurred before mutex became available
  */
 bool recursive_mutex_enter_block_until(recursive_mutex_t *mtx, absolute_time_t until);
@@ -227,7 +226,7 @@ void recursive_mutex_exit(recursive_mutex_t *mtx);
  *  \ingroup mutex
  *
  * \param mtx Pointer to mutex structure
- * \return true if the mutex is initialised, false otherwise
+ * \return true if the mutex is initialized, false otherwise
  */
 static inline bool mutex_is_initialized(mutex_t *mtx) {
     return mtx->core.spin_lock != 0;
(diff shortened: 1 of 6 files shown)


src/common/pico_sync/mutex.c
```diff
@@ -73,41 +129,56 @@ bool __time_critical_func(mutex_enter_timeout_us)(mutex_t *mtx, uint32_t timeout
...
+    assert(lock_is_owner_id_valid(mtx->owner));
+    mtx->owner = LOCK_INVALID_OWNER_ID;
+    lock_internal_spin_unlock_with_notify(&mtx->core, save);
+}
+
+static void __time_critical_func(mutex_exit_r_guts)(mutex_t *mtx) {
+    uint32_t save = spin_lock_blocking(mtx->core.spin_lock);
+    assert(mtx->enter_count);
+    if (!--mtx->enter_count) {
         lock_internal_spin_unlock_with_notify(&mtx->core, save);
```
lurch, line 162: Trying to compare the old-code with the new-code, is an `mtx->owner = LOCK_INVALID_OWNER_ID;` also needed before this line? Or is that not necessary now? :shrug:

lurch:
This is probably a silly suggestion, but could it be worth adding `mutex_init_nr` as an alias of `mutex_init` and `mutex_init_r` as an alias of `recursive_mutex_init` ?

lurch:
`auto_init_recursive_mutex` still references `.recursion_state` ? And once that's "fixed" (if it is indeed a bug) I think you can/could also then get rid of `MAX_RECURSION_STATE`?

lurch:
Just for future reference, it might be worth adding a note to the commit-comment that the spelling of `mutex_is_initialized` has been "fixed" too? (or is that function internal-use-only?)

src/common/pico_sync/mutex.c
kilograham: ha good catch, i was rather saddened to see there was no test currently (I did test it originally; i guess I'll add one)

lurch:
`mutex_try_enter_r_guts` still checks ` && mtx->recursive` which I think isn't necessary?

Should there also be `mutex_enter_block_until_nr` and `mutex_enter_block_until_r` specialisations, or is that function inherently less timing-critical?

liamfraser (approved):
I have reviewed to the best of my ability, but that doesn't mean there is not a bug in here!

kilograham:
i've decided that this is a case of trying to correct a wrong with another wrong. I am going to make a backwards compatible change to split mutex back to being non recursive and use a separate rmutex for recursive. I will attempt to add a #define to allow 1.2.0 code using recursive mutex to work

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -51,38 +77,60 @@ void mutex_init(mutex_t *mtx);
...
 
+/*! \brief Attempt to take ownership of a recursive mutex
+ *  \ingroup mutex
+ *
+ * If the mutex wasn't owned or was owner by the caller, this will claim the mutex and return true.
+ * Otherwise (if the mutex was already owned) this will return false and the
+ * caller will *NOT* own the mutex.
+ *
+ * \param mtx Pointer to recursive mutex structure
+ * \param owner_out If mutex was already owned, and this pointer is non-zero, it will be filled in with the core number of the current owner of the mutex
```
lurch, line 124: "core number" -> "owner id" ? (like you did on line 112)

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -51,38 +77,60 @@ void mutex_init(mutex_t *mtx);
...
 /*! \brief Wait for mutex with timeout
  *  \ingroup mutex
  *
- * Wait for up to the specific time to take ownership of the mutex. If the calling
- * core can take ownership of the mutex before the timeout expires, then true will be returned
- * and the calling core will own the mutex, otherwise false will be returned and the calling
+ * Wait for up to the specific time to take ownership of the mutex. If the caller
+ * can be granted ownership of the mutex before the timeout expires, then true will be returned
+ * and the caller will own the mutex, otherwise false will be returned and the calling
  * core will *NOT* own the mutex.
```
lurch, line 134: "calling core" -> "caller"?

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -105,11 +167,25 @@ bool mutex_enter_timeout_ms(mutex_t *mtx, uint32_t timeout_ms);
...
+ */
+bool recursive_mutex_enter_timeout_us(recursive_mutex_t *mtx, uint32_t timeout_us);
+
 /*! \brief Wait for mutex until a specific time
  *  \ingroup mutex
  *
- * Wait until the specific time to take ownership of the mutex. If the calling
- * core can take ownership of the mutex before the timeout expires, then true will be returned
+ * Wait until the specific time to take ownership of the mutex. If the caller
+ * core can be granted ownership of the mutex before the timeout expires, then true will be returned
```
lurch, line 188: "caller core" -> "caller"?

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -119,20 +195,51 @@ bool mutex_enter_timeout_us(mutex_t *mtx, uint32_t timeout_us);
...
+/*! \brief Wait for mutex until a specific time
+ *  \ingroup mutex
+ *
+ * Wait until the specific time to take ownership of the mutex. If the caller
+ * already has ownership of the mutex or can be granted ownership of the mutex before the timeout expires,
+ * then true will be returned and the caller will own the mutex, otherwise false will be returned and the calling
+ * core will *NOT* own the mutex.
+ *
+ * \param mtx Pointer to recursive mutex structure
+ * \param until The time after which to return if the core cannot take ownership of the mutex
```
lurch, line 207: core -> caller ?

src/common/pico_sync/mutex.c
```diff
@@ -67,47 +101,83 @@ bool __time_critical_func(mutex_enter_timeout_ms)(mutex_t *mtx, uint32_t timeout
...
 }
 
+bool __time_critical_func(recursive_mutex_enter_timeout_us)(recursive_mutex_t *mtx, uint32_t timeout_us) {
+    return recursive_mutex_enter_block_until(mtx, make_timeout_time_us(timeout_us));
+}
+
 bool __time_critical_func(mutex_enter_block_until)(mutex_t *mtx, absolute_time_t until) {
+#if PICO_MUTEX_ENABLE_SDK120_COMPATIBILITY
+    if (mtx->recursive) {
+        return mutex_enter_block_until(mtx, until);
```
lurch, line 119: Was this supposed to be `recursive_mutex_enter_block_until` ?

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -19,25 +19,51 @@ extern "C" {
...
+ * (and associated recursive_mutex_ functions) is a recursive mutex that can be recursively obtained by
+ * the same caller, at the expense of some more overhead when acquiring and releasing.
+ *
+ * It is generally a bad idea to call blocking mutex_ or recursive_mutex_ functions from within an IRQ handler
+ . It is valid to call \ref mutex_try_enter or \ref recursive_mutex_try_enter from within an IRQ handler, if the operation
  * that would be conducted under lock can be skipped if the mutex is locked (at least by the same core).
  *
+ * NOTE: For backwards compatibility with version 1.2.0 of the SDK, if the define
+ * PICO_MUTEX_ENABLE_SDK120_COMPATIBILITY is set to 1, then the the regular mutex_ functions
+ * may also be used for recursive mutexes. This flag will be removed in a future version of the SDK.
```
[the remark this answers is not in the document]
kilograham: I don't think there is any benefit; they still work

src/common/pico_sync/mutex.c
kilograham: yes, thank you, good catch - more tests required!

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -119,20 +194,51 @@ bool mutex_enter_timeout_us(mutex_t *mtx, uint32_t timeout_us);
  */
 bool mutex_enter_block_until(mutex_t *mtx, absolute_time_t until);
 
+/*! \brief Wait for mutex until a specific time
+ *  \ingroup mutex
+ *
+ * Wait until the specific time to take ownership of the mutex. If the caller
+ * already has ownership of the mutex or can be granted ownership of the mutex before the timeout expires,
+ * then true will be returned and the caller will own the mutex, otherwise false will be returned and the calling
+ * core will *NOT* own the mutex.
```
lurch, line 203: calling core -> caller

src/common/pico_sync/mutex.c
```diff
@@ -67,47 +101,83 @@ bool __time_critical_func(mutex_enter_timeout_ms)(mutex_t *mtx, uint32_t timeout
...
+#if PICO_MUTEX_ENABLE_SDK120_COMPATIBILITY
+    if (mtx->recursive) {
+        recursive_mutex_exit(mtx);
+        return;
+    }
+#endif
     uint32_t save = spin_lock_blocking(mtx->core.spin_lock);
-    assert(mtx->owner != LOCK_INVALID_OWNER_ID);
-    if (!mtx->recursion_state) {
+    assert(lock_is_owner_id_valid(mtx->owner));
```
lurch, line 169: Should this `assert` also be in `recursive_mutex_exit`?

src/common/pico_sync/mutex.c
```diff
@@ -7,53 +7,87 @@
...
             mtx->owner = caller;
-            if (mtx->recursion_state) {
-                assert(mtx->recursion_state == MAX_RECURSION_STATE);
-                mtx->recursion_state--;
-            }
-        } else if (mtx->owner == caller && mtx->recursion_state > 1) {
-            mtx->recursion_state--;
+            uint __unused total = ++mtx->enter_count;
+            spin_unlock(mtx->core.spin_lock, save);
+            assert(total); // check for overflow
```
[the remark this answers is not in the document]
kilograham: i think it is an unlikely eventuality other than a bug

src/common/pico_sync/include/pico/mutex.h
```diff
@@ -51,88 +77,168 @@ void mutex_init(mutex_t *mtx);
...
  * \param mtx Pointer to mutex structure
- * \param owner_out If mutex was already owned, and this pointer is non-zero, it will be filled in with the core number of the current owner of the mutex
+ * \param owner_out If mutex was already owned, and this pointer is non-zero, it will be filled in with the owner id of the current owner of the mutex
  */
 bool mutex_try_enter(mutex_t *mtx, uint32_t *owner_out);
 
+/*! \brief Attempt to take ownership of a recursive mutex
+ *  \ingroup mutex
+ *
+ * If the mutex wasn't owned or was owner by the caller, this will claim the mutex and return true.
```
lurch, line 119: owner -> owned ?

lurch (approved):
I haven't tested any of this code, but LGTM.

liamfraser (approved):
LGTM. Also very good documentation in mutex.h which I like!

PR #497, opened 2021-06-28
fix operatore precedence of */ vs +- in pioasm

### LINKED ISSUE #416: pioasm expressions ignore order of operations
When using an expression in the PIO assembly format, then the order of expressions is ignored. Addition is evaluated **before** multiplication.

Take the following example program:

```
.program example

set X (20/2-1)
in X 8
push
```

My expectation is, that this will write `9` into register X, but the current behaviour is that `20` will be written into X.

Changing the expression to read `20/2-2` will result in an floating point exception while compiling:

```
[ 30%] Generating controller.pio.h
make[2]: *** [CMakeFiles/example_controller_pio_h.dir/build.make:73: controller.pio.h] Floating point exception (core dumped)
make[1]: *** [CMakeFiles/Makefile2:1277: CMakeFiles/example_controller_pio_h.dir/all] Error 2
```

I am using the pico C SDK, commit `afc10f3599c27147a6f34781b7102d86f58aa5f6`.

kilograham:
fix #416


Wren6991 (approved):
Built it, tried it, seems to fix the issue

PR #502, opened 2021-07-04
bug in sleep_until on host mode for macOS

kilograham:
if sleep target was in the past, it would hang


Wren6991 (approved):
Don't have a mac host to try it on but fix makes sense and looks good.

PR #509, opened 2021-07-14
Improve wording

matiasilva:
Just two minor changes to improve the wording on IRQ functions.


Wren6991 (commented):
First one is a good change, the second comment refers to the pending exception state, as described here in the architecture reference:

image

(from section B1.3.2 of Armv6-M Architecture Reference)

so is probably fine either way

Wren6991 (approved):

lurch:
This fixes the second-half of https://github.com/raspberrypi/pico-feedback/issues/175
(the first-half of that issue is still in backlog)

PR #514, opened 2021-07-14
automatically calculate crc32 for flash based application

GorgonMeducer:
The new revision will calculate crc32 for the stage2 boot block and store at 0x100000FC. 
This operation only works for flash-based applications.

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -24,6 +24,8 @@ typedef unsigned int uint;
 static char error_msg[512];
 static bool verbose;
 
+static uint32_t crc32_calculator (uint8_t *pbuf, uint_fast32_t size, uint32_t init);
+
 static int fail(int code, const char *format, ...) {
     va_list args;
     va_start(args, format);
@@ -67,6 +69,7 @@ typedef std::vector<address_range> address_ranges;
 #define XIP_SRAM_END          0x15004000u
 #define MAIN_RAM_BANKED_START 0x21000000u
 #define MAIN_RAM_BANKED_END   0x21040000u
+#define BOOT2_SIZE_BYTES      0x00000100u
 
 const address_ranges rp2040_address_ranges_flash {
     address_range(FLASH_START, FLASH_END, address_range::type::CONTENTS),
@@ -293,7 +296,8 @@ int elf2uf2(FILE *in, FILE *out) {
                         MAIN_RAM_START, sp);
         }
 #endif
-    }
+    } 
+    
     uf2_block block;
     block.magic_start0 = UF2_MAGIC_START0;
     block.magic_start1 = UF2_MAGIC_START1;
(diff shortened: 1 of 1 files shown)


tools/elf2uf2/main.cpp
```diff
@@ -311,6 +315,13 @@ int elf2uf2(FILE *in, FILE *out) {
         memset(block.data, 0, sizeof(block.data));
         rc = realize_page(in, page_entry.second, block.data, sizeof(block.data));
         if (rc) return rc;
+
+        if (!ram_style) {
+            if (block.target_addr == FLASH_START && block.payload_size >= BOOT2_SIZE_BYTES - 4) {
```
lurch, line 320: I may be totally misunderstanding what this code is doing, but was the `>=` here supposed to be `<=` ?

GorgonMeducer: It should be “>=”. It is guarantee that the target block contains the first 252 bytes (maybe more). 
For the condition where block.target_address == FLASH_START and the block.payload_size smaller than 252, that means the stage2 boot block is not padded to 252 as required, so we simply leave it untouched.

lurch: > I may be totally misunderstanding what this code is doing

(My naive concern was that e.g. if `block.payload_size == 254`, then the last two bytes of the payload will be overwritten with the first two bytes of the CRC?)
However I'm probably way out of my depth here, so apologies for the irrelevant comment and I'll let Graham deal with this :confused:

GorgonMeducer: No, the stage2 boot block request that the valid payload in the first 256 bytes is less than or equals to 252 bytes. It is written in the documents. And it is required by the bootroom.

https://github.com/raspberrypi/pico-bootrom/blob/00a4a19114195e20fb817bdfbca1165e157eef37/bootrom/bootrom_main.c#L74


https://github.com/raspberrypi/pico-sdk/blob/bfcbefafc5d2a210551a4d9d80b4303d4ae0adf7/src/rp2_common/boot_stage2/boot2_w25q080.S#L26

GorgonMeducer: The reason why I use "block.payload_size >= BOOT2_SIZE_BYTES - 4" is because:

- The elf might already has a CRC32 checksum (or just a dummy place holder), it is just like what I have done in a linker script: https://github.com/GorgonMeducer/Pico_Template/blob/b941fb754820580c22c32f3cf4b7641220f0a185/project/mdk/RP2040.sct#L24
- The block might be bigger than just 256 bytes (contains the vector table and etc.). There is no limitation to say that there could not be a big section that contains the stage2 boot (256bytes) and the following code, vector table and data. 
- Since pico-sdk explicitly requires that the stage2 boot block should be padded to 252 bytes, so I add this check to make sure it is a, literally speaking, valid stage2 boot block.

GorgonMeducer:
I was thinking these days, maybe I should also handle the condition when people have a stage2 block less than 252?
I have got a solution. If you think this is a good idea. Please reject this one. and I will push another merge request with a new solution.

Thank you.

PR #520, opened 2021-07-19
Clarify 7-bit address use for I2C

matiasilva:
This PR clarifies the length of the I2C address parameter in the SDK comments.

## Motivation

A mistake I ran into while writing examples that use the I2C API was thinking that the address parameter in any I2C function was 8-bit and thus (wrongly) assumed that I had to left shift the 7-bit slave address and add the read/write bit myself. Little did I know that the SDK actually takes care of this for us and expects a 7-bit address. The comments added here hopefully clarify that for future use.


Wren6991 (approved):
Nice, thanks

Issue #204, opened 2021-03-01
Check that #define-d PIN combinations make sense at compile-time


lurch:
(as initially suggested in #192)

It might be nice to have a compile-time #error if you give an invalid combination of UART_DEFAULT settings? e.g. setting {PICO_DEFAULT_UART=0, PICO_DEFAULT_UART_TX_PIN=13} (invalid because GPIO13 is UART0 **RX**) or {PICO_DEFAULT_UART=1, PICO_DEFAULT_UART_TX_PIN=0} (invalid because GPIO0 is **UART0** TX).
(with similar things being done for I2C and SPI pins too)

kilograham:
#1724 has been merged into develop

PR #526, opened 2021-07-23
adafruit_feather_rp2040: Fix default I2C instance

mwick83:
Use i2c1 as default, because the default pins can't be muxed to i2c0.

### DIFF
--- a/src/boards/include/boards/adafruit_feather_rp2040.h
+++ b/src/boards/include/boards/adafruit_feather_rp2040.h
@@ -44,7 +44,7 @@
 
 //------------- I2C -------------//
 #ifndef PICO_DEFAULT_I2C
-#define PICO_DEFAULT_I2C 0
+#define PICO_DEFAULT_I2C 1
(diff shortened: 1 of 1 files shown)


lurch (dismissed):
LGTM :+1:   I wonder why nobody spotted this before? :laughing:

mwick83:
> LGTM  I wonder why nobody spotted this before?

I have no idea, but it took me some hours debugging, why the blockng functions wouldn't return and SCL and SDA didn't toggle at all until I cross checked the MUXing options. The default pins are indeed correct.

lurch:
I once suggested that we could add a check that the pins "match up correctly" (which would have e.g. caught the error you spotted here, and also prevent you assigning `#define PICO_DEFAULT_I2C_SDA_PIN 3`) - but unfortunately I've been far far too busy with other stuff.
Ah yes, #204 :slightly_smiling_face:

kilograham (approved):

PR #533, opened 2021-07-29
Documentation fix for SIE_STATUS.CONNECTED

richardanaya:
This PR contains a documentation fix for the issue referred to in https://github.com/raspberrypi/pico-feedback/issues/110


lurch:
Thanks for the PR, but unfortunately it's not quite this simple. These files are autogenerated, so there's a (private) "upstream" file that we need to edit.

PR #534, opened 2021-07-29
Documentation fix, SPINLOCK0-31 should be RW

richardanaya:
This PR fixes documentation related to the issue at https://github.com/raspberrypi/pico-feedback/issues/162


lurch:
Same as #533 - these files are autogenerated, so there's a (private) "upstream" file that we need to edit.

PR #535, opened 2021-08-05
Added definitions for some upcoming Pimoroni LED driver boards

ZodiusInfuser:


src/boards/include/boards/pimoroni_interstate75.h
```diff
@@ -0,0 +1,185 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch, line 2: Should this be 2021 now? :man_shrugging:

src/boards/include/boards/pimoroni_interstate75.h
```diff
@@ -0,0 +1,185 @@
...
+#ifndef PICO_DEFAULT_LED_PIN_INVERTED
+#define PICO_DEFAULT_LED_PIN_INVERTED 1
+#endif
+
+// --- I2C ---
+#ifndef PICO_DEFAULT_I2C
+#define PICO_DEFAULT_I2C INTERSTATE75_I2C
+#endif
+#ifndef PICO_DEFAULT_I2C_SDA_PIN
+#define PICO_DEFAULT_I2C_SDA_PIN INTERSTATE75_SDA_PIN
```
lurch, line 152: Is having both `PICO_DEFAULT_I2C_SDA_PIN` and `INTERSTATE75_SDA_PIN` genuinely useful? :shrug:

src/boards/include/boards/pimoroni_interstate75.h
ZodiusInfuser: Probably, but that seems like a wider change than this specific PR, as most (if not all) of the SDK says 2020...

src/boards/include/boards/pimoroni_interstate75.h
ZodiusInfuser: Not sure. I added it because we have an INT pin exposed alongside SDA and SCL, following our breakout format, and because a user may not intuitively think to look for a Pico define when using a board not called Pico. This is also what I did for the Keybow2040 header (although I notice now that I don't have a KEYBOW2040_I2C define...) https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/pimoroni_keybow2040.h
Could drop the PICO_DEFAULT... but I suspect they are used in various parts of the SDK and examples, so make sense to keep for compatibility.

Honestly, I'm not sure how much these headers are used. From my experience it's not trivial to set a board type in VSCode or another editor that would pull from this list, so you have to build from command line to take advantage of them. Unless there's an easy way I completely missed?

src/boards/include/boards/pimoroni_interstate75.h
lurch: I think the general convention is that the copyright year on a file reflects the year in which it was created? :man_shrugging:

src/boards/include/boards/pimoroni_interstate75.h
lurch: We _definitely_ want to keep the `PICO_DEFAULT...` defines, because as you say they're used throughout the SDK and examples (and IIRC is why these board header files were first added?)

The other bits sound like questions for @kilograham :wink: 

@JamesH65 @matiasilva I wonder if it makes sense to add support for setting the `PICO_BOARD` from the project-generator, e.g. via a drop-down list or something?

@kilograham On a side note I see that page 20 of https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-c-sdk.pdf says "The default PICO_BOARD if not specified is pico." but then on page 287 it says the default value is "rp2040"? :confused:

src/boards/include/boards/pimoroni_interstate75.h
ZodiusInfuser: If that's the case, then whoops on behalf of all the board commits that we, Adafruit, and Sparkfun have pushed this year 😅 

In any case, let me know if you want it updating for these two board. The others would be best to do by someone as a separate PR I think.

lurch: I'm gonna back out and let @kilograham decide - apologies for causing any confusion!

kilograham:
> From my experience it's not trivial to set a board type in VSCode or another editor that would pull from this list,

I think it is trivial in any IDE to store cmake settings (including PICO_BOARD) but not there is no dropdown.

src/boards/include/boards/pimoroni_interstate75.h
lurch: > On a side note I see that page 20 of https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-c-sdk.pdf says "The default PICO_BOARD if not specified is pico." but then on page 287 it says the default value is "rp2040"?

I think #536 should fix that.

kilograham (approved):

PR #537, opened 2021-08-10
Use W25Q080 second stage loader for Nano RP2040 Connect

facchinm:
Due to the well known electronic market situation, we were forced to mount an alternative part number in a batch of Arduino Nano RP2040 Connect.
These flash chips, from ISSI, need yet another way to configure the QE sticky bit :disappointed: 
At the moment, the safest way to handle the dual sourcing is to fallback using W25Q080 loader, and requiring that the sticky bit has already been programmed during production.

@pnndra @alranel @iabdalkader

### DIFF
--- a/src/boards/include/boards/arduino_nano_rp2040_connect.h
+++ b/src/boards/include/boards/arduino_nano_rp2040_connect.h
@@ -66,7 +66,7 @@
 
 //------------- FLASH -------------//
 
-#define PICO_BOOT_STAGE2_CHOOSE_AT25SF128A 1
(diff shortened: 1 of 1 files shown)


dhalbert:
What is the part number you are substituting? We have seen Nano's with AT25SF128A and IS25LP128F. Thanks.

pnndra:
AT25SF128A is the chip we'll be using for most of the production. for now we just had a small batch with IS25LP128F.

kilograham (approved):

PR #542, opened 2021-08-23
Add _REG_ macro for new headers

matiasilva:


lurch:
Does this need to be folded into #543 to fix some of the build errors in that PR?

matiasilva:
Yes, it does. The build errors are due to a lack of this macro.

matiasilva:
I'll combine this with the other one.

PR #544, opened 2021-08-24
Fix access type for "WO" registers in SVD

lurch:


Wren6991 (approved):

Wren6991:
Looks good, all of these registers are read-all-zeroes in hardware

PR #550, opened 2021-08-30
Update i2c.h

k1r0s:
typo


lurch:
AFAIK this _isn't_ a typo - see e.g. Figure 87 in the RP2040 datasheet.

Wren6991:
To clarify, PRESETn/presetn is the active-low reset for the I2C block. Text is correct as it stands, but maybe a bit misleading

Wren6991:
I hadn't realised this was an edit to a register header. Those headers are generated from the same source as the SVD and the datasheet listings (and it's important they match) so we can't merge PRs to those files.

Here is the register table in the datasheet:

image

Other relevant parts of the I2C section of the datasheet:

image

image

k1r0s:
lol, I thought it was a typo. Im dumb

k1r0s:
@Wren6991 To clarify I would recommend that the existing text is displayed like so:

```
means once the 'PRESETn' signal is applied
```

Issue #548, opened 2021-08-26
UART only works for 8N1; API exposes functions that mishandle UARTLCR


MustBeArt:
The documented Hardware UART API exposes useful-sounding functions such as `uart_set_baudrate()` and `uart_set_format()` (and others) that write to the UARTLCR registers. This can only be done safely if the UART is not enabled, and then only if the UART is not currently busy transmitting or receiving a character, as documented in ARM's _PrimeCell UART (PL011) Technical Reference Manual_, in the section on the UARTLCR_H register. These functions make no attempt to ensure this is the case, and there is no warning in the API documentation.

In fact there is no safe way to call these functions through the API, because the API does not expose the UART enable separately, but only through uart_init() and uart_deinit(). The programmer can test the enable with uart_is_enabled(), but cannot control it. If the programmer tries to comply with the restriction on writing UARTLCR by first calling uart_deinit(), then calling (say) uart_set_format(), and finally calling uart_init(), their work is immediately undone by uart_init(). **The uart_init() functions HARD CODES the format value to 8 bits, 1 stop bit, and no parity! The result is that any attempt to use word size, stop bits, or parity values other than 8, 1, and none is doomed to failure.**

Changing baud rate on the fly (by just calling uart_set_baudrate()) does seem to work empirically, but it still violates the requirements set out in the PL011 manual, so there may be cases I didn't try where it fails.

Changing the stop bits, word width, or parity on the fly (by just calling uart_set_format()) seems to work sometimes and fail sometimes. My current project requires changing all those values quite frequently (to exercise another device's serial port implementation). Sometimes it transmits with the wrong parameters, and other times it fails to transmit at all after a parameter change.

I did not investigate whether other settings in UARTLCR work or not. FIFO enable, for instance, is in UARTLCR_H, and uart_init() has a hard-coded value for that one, too. However, uart_init() erroneously sets the FIFO enable bit AFTER enabling the UART, so it is always in violation of the rule from the PL011 manual. The API function for this, uart_set_fifo_enabled(), writes UARTLCR_H without any checks, just as uart_set_format() does.

[...]

PR #556, opened 2021-09-08
(WIP) Disable UART when writing to LCR

Wren6991:
See #548.

Details are quite messy. Our API supposedly lets you poke the UART settings whilst it's enabled, and doesn't impose restrictions on what state it can be in when you do this (e.g. it might be mid-character).

First problem is the UART requires you to disable it when you change the settings, slightly bigger problem is there is seemingly no way to cleanly halt and restart the UART when it may be mid character, because it doesn't have the flags you would need to poll to do this. This PR attempts to address this by asserting a slightly longer than one character busy wait after disabling the UART.

Another approach would be to allow the UART to be left disabled after `uart_init()` is called, so the user can poke settings and then enable it. To change settings they would need to reinitialise the UART, which under the hood would just toggle the reset like it does today. This would be a breaking API change, but as the current implementation is quite broken, it doesn't seem like many people are using the uncommon UART formats.

### DIFF
--- a/src/rp2_common/hardware_uart/uart.c
+++ b/src/rp2_common/hardware_uart/uart.c
@@ -53,10 +53,10 @@ uint uart_init(uart_inst_t *uart, uint baudrate) {
     uint baud = uart_set_baudrate(uart, baudrate);
     uart_set_format(uart, 8, 1, UART_PARITY_NONE);
 
+    // Enable FIFOs (must be before setting UARTEN, as this is an LCR access)
+    hw_set_bits(&uart_get_hw(uart)->lcr_h, UART_UARTLCR_H_FEN_BITS);
     // Enable the UART, both TX and RX
     uart_get_hw(uart)->cr = UART_UARTCR_UARTEN_BITS | UART_UARTCR_TXE_BITS | UART_UARTCR_RXE_BITS;
-    // Enable FIFOs
-    hw_set_bits(&uart_get_hw(uart)->lcr_h, UART_UARTLCR_H_FEN_BITS);
     // Always enable DREQ signals -- no harm in this if DMA is not listening
     uart_get_hw(uart)->dmacr = UART_UARTDMACR_TXDMAE_BITS | UART_UARTDMACR_RXDMAE_BITS;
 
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_uart/uart.c
```diff
@@ -86,19 +122,80 @@ uint uart_set_baudrate(uart_inst_t *uart, uint baudrate) {
         baud_fbrd = ((baud_rate_div & 0x7f) + 1) / 2;
     }
 
-    // Load PL011's baud divisor registers
+    // Need to cleanly disable UART before touching LCR
+    bool was_enabled = uart_is_enabled(uart);
+    uint32_t cr_save;
+    if (was_enabled)
```
kilograham, line 128: please match code style. IF you are going to make a non braced if, then please put it on same line

daveythacher:
Two options:
https://github.com/raspberrypi/pico-sdk/issues/548#issuecomment-915210409

Note would need struct initialization function. Documentation update would be required. However documentation is already needed for delay.

Another option is to hide these functions since they are not really usable. Full hardware module is not supported. Force user to init and deinit to change format and baud rate. Add format to init would be minimal API change.

Deprecate/remove hardware_uart?

kilograham:
> Deprecate/remove hardware_uart?

Why on earth would we do that? It provides useful functionality. there is always register access for what it doesn't provide

kilograham:
yeah, i mean in general i think i would have preferred if the UART had used a builder pattern as per some of the other h/w. This was something i wanted to look at before launch (the uart code came from an earlier incarnation).

note that all this talk of CMake still makes no sense. i think your complaint is that the hardware_uart packaged is somewhat focused on simple 8 bit transmission/receive, but the fact that pico_stdio_uart uses it is neither here nor there, and that there are some CMake convenience functions for choosing amongst the stdio implementations seems even less germaine.

we should certainly make the existing uart functions behave correctly (as @Wren6991 is looking at) such that you can configure things correctly. it may well be that adding a new init function which is builder based alongside the rest is desirable (and maybe leaving the old way broken if it can't cleanly be fixed).

As for more advanced uart functionality, there doesn't seem to have been a lot of requests for this thus far, so i'm not particularly in favor of adding a whole abstraction. I don't know enough about UART use cases to know what this would look like.

Wren6991:
> Only the CMAKE logic could change the format

Yes, due to a bug in the implementation. This isn't an API problem.

To be clear, this PR is to make the current API work as advertised. Feel free to open another issue to make the case for API changes.

src/rp2_common/hardware_uart/uart.c
```diff
@@ -86,19 +122,80 @@ uint uart_set_baudrate(uart_inst_t *uart, uint baudrate) {
...
+                   (bool_to_bit(parity != UART_PARITY_NONE) << UART_UARTLCR_H_PEN_LSB) |
+                   (bool_to_bit(parity == UART_PARITY_EVEN) << UART_UARTLCR_H_EPS_LSB),
+                   UART_UARTLCR_H_WLEN_BITS |
+                   UART_UARTLCR_H_STP2_BITS |
+                   UART_UARTLCR_H_PEN_BITS |
+                   UART_UARTLCR_H_EPS_BITS);
+
+    if (was_enabled)
+        uart_get_hw(uart)->cr = cr_save;
+}
```
lurch, line 168: I wonder if it's worth abstracting this out further into e.g. a `uart_write_lcr_bits(uart_inst_t *uart, uint32_t values, uint32_t write_mask)` function, so that `uart_set_format` would just be:
```c
void uart_set_format(uart_inst_t *uart, uint data_bits, uint stop_bits, uart_parity_t parity) {
    invalid_params_if(UART, data_bits < 5 || data_bits > 8);
    invalid_params_if(UART, stop_bits != 1 && stop_bits != 2);
    invalid_params_if(UART, parity != UART_PARITY_NONE && parity != UART_PARITY_EVEN && parity != UART_PARITY_ODD);

    uart_write_lcr_bits(uart,
                       ((data_bits - 5u) << UART_UARTLCR_H_WLEN_LSB) |
                       ((stop_bits - 1u) << UART_UARTLCR_H_STP2_LSB) |
                       (bool_to_bit(parity != UART_PARITY_NONE) << UART_UARTLCR_H_PEN_LSB) |
                       (bool_to_bit(parity == UART_PARITY_EVEN) << UART_UARTLCR_H_EPS_LSB),
                       UART_UARTLCR_H_WLEN_BITS |
                       UART_UARTLCR_H_STP2_BITS |
                       UART_UARTLCR_H_PEN_BITS |
                       UART_UARTLCR_H_EPS_BITS);
}
```
and `uart_set_fifo_enabled` would just be:
```c
void uart_set_fifo_enabled(uart_inst_t *uart, bool enabled) {
    uart_write_lcr_bits(uart,
                       (bool_to_bit(enabled) << UART_UARTLCR_H_FEN_LSB),
                       UART_UARTLCR_H_FEN_BITS);
}
```
:man_shrugging: 

(or perhaps this is one of those cases where too much abstraction is confusing rather than helpful)

kilograham:
i believe this is superseded by #1347, right?

andygpz11:
> i believe this is superseded by #1347, right?

Yes, it is.

PR #561, opened 2021-09-16
Change _watchdog_enable to trigger immediate reboot when no delay set

rjp5th:
This fixes an issue where where calling watchdog_reboot will appear as a TIMER reset event in the WATCHDOG_REASON register rather than a FORCE event, making it difficult to tell whether the microcontroller was reset from a watchdog expiring or a software reset. This makes it where if the delay_ms is 0 it will trigger a watchdog reset immediately rather than defaulting to a value of 50ms.


Wren6991 (commented):
Seems like a sensible change. The 50 ms default is not documented as far as I can see, and seems a bit unhelpful.

liamfraser (approved):
Looks good to me. In case anyone is curious WATCHDOG_CTRL_TRIGGER doesn't need WATCHDOG_CTRL_ENABLE to be set. Enable just starts the decrementing timer.

PR #564, opened 2021-09-17
Add board definition for Melopero Shake RP2040

leoli51:
We created the board header files for the upcoming Melopero Shake RP2040, a development board based on the RP2040 microcontroller.


leoli51:
Sorry we fixed the name of the header

PR #565, opened 2021-09-17
Add Melopero Shake RP2040 Board header file

leoli51:
Hello, we created a board definition for the upcoming Melopero Shake RP2040, a development board based on the RP2040 microcontroller.


src/boards/include/boards/melopero_shake_rp2040.h
```diff
@@ -0,0 +1,95 @@
...
+#ifndef PICO_DEFAULT_SPI_RX_PIN
+#define PICO_DEFAULT_SPI_RX_PIN 20
+#endif
+
+#ifndef PICO_DEFAULT_SPI_SCK_PIN
+#define PICO_DEFAULT_SPI_SCK_PIN 18
+#endif
+
+#ifndef PICO_DEFAULT_SPI_CSN_PIN
+#define PICO_DEFAULT_SPI_CSN_PIN 1
```
lurch, line 71: This is the same value as `PICO_DEFAULT_UART_RX_PIN` ?

lurch:
Can you change the base branch to `develop` please?

src/boards/include/boards/melopero_shake_rp2040.h
leoli51: Yes, we confirm that it was done on purpose. Many thanks for the advice.

lurch: This might lead to "unexpected results" for any example-code that tries to use both the DEFAULT_UART and the DEFAULT_SPI at the same time? (e.g. if https://github.com/raspberrypi/pico-examples/tree/master/spi/spi_flash was configured to use UART rather than USB-CDC for printf)
@kilograham can probably offer more detailed advice on any ramifications that might have...

leoli51: We updated the default UART pins. Now it should be ok, let me know if something is still off. Thank you very much for your support.

PR #575, opened 2021-09-27
pico_stdin: Increase sleeping time for stdin read

pingiun:
This solve an issue where USB serial seems to hang when using a function like getchar()

### DIFF
--- a/src/rp2_common/pico_stdio/stdio.c
+++ b/src/rp2_common/pico_stdio/stdio.c
@@ -127,7 +127,7 @@ static int stdio_get_until(char *buf, int len, absolute_time_t until) {
         }
         // we sleep here in case the in_chars methods acquire mutexes or disable IRQs and
         // potentially starve out what they are waiting on (have seen this with USB)
-        busy_wait_us(1);
+        busy_wait_us(10);
     } while (!time_reached(until));
     return PICO_ERROR_TIMEOUT;
 }


lurch:
Can you rebase against the `develop` branch please? (we don't accept PRs to `master`)

kilograham:
can you provide an example where this hangs - how did you pick 10 rather than 1?

pingiun:
> can you provide an example where this hangs - how did you pick 10 rather than 1?

I was running this code:
```C
#include <stdio.h>
#include "pico/stdlib.h"

int main() {
    stdio_init_all();
    while (true) {
        char c = getchar();
        if (c == 'a') {
            printf("You sent a!\n");
        } else {
            printf("You did not send a!\n");
        }
    }
    return 0;
}
```

When I run this code on my pico with no other changes, it will not respond to any usb serial input. I changed it to 10, but maybe a lower value will also work, I didn't do much testing on different values

kilograham:
moving to 1.2.1 if i can reproduce

kilograham:
moving out of 1.3.0 as this isn't really reproducable (Seems to be the host stack getting into a bad state?)

see however https://forums.raspberrypi.com/viewtopic.php?f=145&t=321326

lurch:
@pingiun see that latest comment in #623 . Would you like to test if #918 fixes things for you?

PR #580, opened 2021-10-02
Fix syntax error in cmake when using custom stage2

deltabeard:
This fixes a bug in the CMakeLists for boot_stage2 whereby specifying a custom boot_stage2 file would cause a syntax error due to `elif` being used instead of `elseif`. With this patch, the SDK now compiles when a custom boot_stage2 file is specified.

Signed-off-by: Mahyar Koshkouei <mk@deltabeard.com>


kilograham (approved):

kilograham:
thanks

Issue #582, opened 2021-10-03
Possible typo in structs/dma.h


lurch:
https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_structs/include/hardware/structs/dma.h#L55 (which is hand-edited) says `ctrdeq`, but looking at https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/dma.h#L5049 (which is auto-generated) it seems like it _should_ have been named `ctdreq` ?

Although this looks like a simple typo, perhaps the fact that this has ended up in a "public header" means that it's now too late to fix this? :confused:  (If this is a "wontfix", feel free to close this issue)

kilograham:
yeah too late to fix this really

PR #478, opened 2021-06-04
Header info string updates for PIO DBG_PADx and XOSC STARTUP

Wren6991:
Related to https://github.com/raspberrypi/pico-feedback/issues/158

PR #583, opened 2021-10-04
New PR for autogenerated struct headers

kilograham:
Reviewers please scan the diffs as there are some members that have changed to match what was in the SVD. Everything (SDK, examples) compiles OK (with one change I made), however since we found some other SVD bugs it is worth checking that the SVD wasn't what was wrong in these cases


src/rp2040/hardware_regs/rp2040.svd
```diff
@@ -22331,7 +22331,7 @@
             </field>
           </fields>
           <name>STARTUP</name>
-          <resetValue>0x00000000</resetValue>
+          <resetValue>0x000000c4</resetValue>
```
kilograham, line 22334: note i saw this in passing, which was I think missed from the previous change #478

lurch: This SVD file is auto-generated, so shouldn't be hand-edited as any changes will just get overwritten. I'll do my best to propagate the changes you've made here into the "upstream source files".

src/rp2040/hardware_structs/include/hardware/structs/adc.h
```diff
@@ -1,27 +1,88 @@
...
 #ifndef _HARDWARE_STRUCTS_ADC_H
 #define _HARDWARE_STRUCTS_ADC_H
 
 #include "hardware/address_mapped.h"
 #include "hardware/regs/adc.h"
 
+// reference to datasheet: https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf#tab-registerlist_adc
+
+// BITMASK [BITRANGE]: FIELDNAME (RESETVALUE): DESCRIPTION
+// The REG macro is intended to help make the register navigable in your IDE (for example, using the "Go to Definition" feature)
```
lurch, line 18: `The REG macro` -> `The _REG_ macro` ? (just to avoid any ambiguity)

src/rp2040/hardware_structs/include/hardware/structs/dma.h
```diff
@@ -8,46 +10,173 @@
...
     io_rw_32 ints1;
+
+    _REG_(DMA_TIMER0_OFFSET)
     io_rw_32 timer[4];
-    io_wo_32 multi_channel_trigger;
+
+    _REG_(DMA_MULTI_CHAN_TRIGGER_OFFSET)
+    // Trigger one or more channels simultaneously
+    // 0x0000ffff [0-15]  : MULTI_CHAN_TRIGGER (0): Each bit in this register corresponds to a DMA channel
+    io_rw_32 multi_channel_trigger;
```
lurch, line 151: The datasheet PDF says that this is SC (self-clearing) so arguably `io_wo_32` was more correct than `io_rw_32`?
And I checked and the SVD has `<modifiedWriteValues>clear</modifiedWriteValues>` (see https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_register) for `MULTI_CHAN_TRIGGER`, so I guess theoretically this _could_ still be `io_wo_32`? Entirely up to you @kilograham if you feel that the distinction and effort are worth it though!

src/rp2040/hardware_structs/include/hardware/structs/dma.h
```diff
@@ -8,46 +10,173 @@
...
+    // 0x00ff0000 [16-23] : RAF_LVL (0): Current Read-Address-FIFO fill level
+    // 0x0000ff00 [8-15]  : WAF_LVL (0): Current Write-Address-FIFO fill level
+    // 0x000000ff [0-7]   : TDF_LVL (0): Current Transfer-Data-FIFO fill level
     io_ro_32 fifo_levels;
-    io_wo_32 abort;
+
+    _REG_(DMA_CHAN_ABORT_OFFSET)
+    // Abort an in-progress transfer sequence on one or more channels
+    // 0x0000ffff [0-15]  : CHAN_ABORT (0): Each bit corresponds to a channel
+    io_rw_32 abort;
```
lurch, line 179: Hmmm, in contrast to what I wrote above (for `MULTI_CHAN_TRIGGER`, this `CHAN_ABORT` probably **should** be `io_rw_32` because the datasheet says "The bit will remain high until any inflight transfers have been flushed through the address and data FIFOs. After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel.". However the datasheet also labels the field as SC, whereas Appendix A in the datasheet describes SC as "This is a single bit that is written to by the processor and then **cleared on the next clock cycle**."

Hmmm, tricky... :confused:

src/rp2040/hardware_structs/include/hardware/structs/rtc.h
```diff
@@ -8,22 +10,101 @@
...
 typedef struct {
+    _REG_(RTC_CLKDIV_M1_OFFSET)
+    // Divider minus 1 for the 1 second counter
+    // 0x0000ffff [0-15]  : CLKDIV_M1 (0)
     io_rw_32 clkdiv_m1;
+
+    _REG_(RTC_SETUP_0_OFFSET)
+    // RTC setup register 0
+    // 0x00fff000 [12-23] : YEAR (0): Year
+    // 0x00000f00 [8-11]  : MONTH (0): Month (1
```
lurch, line 30: Any idea why this line has been truncated so prematurely, rather than saying `Month (1..12)` ?

src/rp2040/hardware_structs/include/hardware/structs/sio.h
```diff
@@ -11,51 +13,162 @@
...
-    io_wo_32 gpio_togl;
 
-    io_wo_32 gpio_oe;
-    io_wo_32 gpio_oe_set;
-    io_wo_32 gpio_oe_clr;
-    io_wo_32 gpio_oe_togl;
+    _REG_(SIO_GPIO_OUT_SET_OFFSET)
+    // GPIO output value set
+    // 0x3fffffff [0-29]  : GPIO_OUT_SET (0): Perform an atomic bit-set on GPIO_OUT, i
+    io_rw_32 gpio_set;
```
lurch, line 47: I wonder why this was previously `io_wo_32 gpio_set;` when the datasheet says `GPIO_OUT_SET` is RW? Perhaps the datasheet "should" list the all atomic bit-manipulation GPIO registers in the SIO block (`GPIO_xxxx_SET`, `GPIO_xxxx_CLR` and `GPIO_xxxx_XOR`) as WO? :man_shrugging:

src/rp2040/hardware_structs/include/hardware/structs/spi.h
```diff
@@ -10,17 +12,92 @@
...
+    // 0x00000001 [0]     : RORIC (0): Clears the SSPRORINTR interrupt
     io_rw_32 icr;
+
+    _REG_(SPI_SSPDMACR_OFFSET)
+    // DMA control register, SSPDMACR on page 3-12
+    // 0x00000002 [1]     : TXDMAE (0): Transmit DMA Enable
+    // 0x00000001 [0]     : RXDMAE (0): Receive DMA Enable
     io_rw_32 dmacr;
+
+    uint32_t _pad0[1006];
```
lurch, line 94: Wow :wink:

src/rp2040/hardware_structs/include/hardware/structs/spi.h
```diff
@@ -10,17 +12,92 @@
...
+
+    _REG_(SPI_SSPPERIPHID0_OFFSET)
+    io_ro_32 periphid[4];
+
+    _REG_(SPI_SSPPCELLID0_OFFSET)
+    io_ro_32 pcellid[4];
```
lurch, line 100: Do we "want" to include these, given that they weren't previously?

src/rp2040/hardware_structs/include/hardware/structs/uart.h
```diff
@@ -10,23 +12,169 @@
...
+    // 0x00000001 [0]     : RXDMAE (0): Receive DMA enable
     io_rw_32 dmacr;
+
+    uint32_t _pad2[997];
+
+    _REG_(UART_UARTPERIPHID0_OFFSET)
+    io_ro_32 periphid[4];
+
+    _REG_(UART_UARTPCELLID0_OFFSET)
+    io_ro_32 pcellid[4];
```
lurch, line 177: See previous comment about these not being included before.

src/rp2040/hardware_structs/include/hardware/structs/xip_ctrl.h
```diff
@@ -1,29 +1,70 @@
...
+    _REG_(XIP_STREAM_FIFO_OFFSET)
+    // FIFO stream data
+    io_ro_32 stream_fifo;
+} xip_ctrl_hw_t;
 
 #define xip_ctrl_hw ((xip_ctrl_hw_t *const)XIP_CTRL_BASE)
 
+#define XIP_STAT_FIFO_FULL XIP_STAT_FIFO_FULL_BITS
+#define XIP_STAT_FIFO_EMPTY XIP_STAT_FIFO_EMPTY_BITS
+#define XIP_STAT_FLUSH_RDY XIP_STAT_FLUSH_READY_BITS
```
lurch, line 68: I wonder if it would make more sense for these `#define`s to be in a non-autogenerated header file?

src/rp2040/hardware_structs/include/hardware/structs/usb.h
```diff
@@ -112,36 +120,446 @@ static_assert(sizeof(usb_host_dpram_t) == USB_DPRAM_MAX, "");
...
+    // 0x80000000 [31]    : SIM_TIMING (0): Reduced timings for simulation
+    // 0x00000002 [1]     : HOST_NDEVICE (0): Device mode = 0, Host mode = 1
+    // 0x00000001 [0]     : CONTROLLER_EN (0): Enable controller
     io_rw_32 main_ctrl;
-    io_rw_32 sof_rw;
+
+    _REG_(USB_SOF_WR_OFFSET)
+    // Set the SOF (Start of Frame) frame number in the host controller
+    // 0x000007ff [0-10]  : COUNT (0)
+    io_wo_32 sof_wr;
```
lurch, line 142: This has changed from `io_rw_32 sof_rw;` to `io_wo_32 sof_wr;` - is this like #582 where it's a typo that we actually need to preserve for backwards-compatibility?

src/rp2040/hardware_structs/include/hardware/structs/usb.h
```diff
@@ -112,36 +120,446 @@ static_assert(sizeof(usb_host_dpram_t) == USB_DPRAM_MAX, "");
...
+    // 0x00000002 [1]     : DM_PULLUP_HISEL_OVERRIDE_EN (0)
+    // 0x00000001 [0]     : DP_PULLUP_HISEL_OVERRIDE_EN (0)
     io_rw_32 phy_direct_override;
+
+    _REG_(USB_USBPHY_TRIM_OFFSET)
+    // Used to adjust trim values of USB phy pull down resistors
+    // 0x00001f00 [8-12]  : DM_PULLDN_TRIM (0x1f): Value to drive to USB PHY
+    // 0x0000001f [0-4]   : DP_PULLDN_TRIM (0x1f): Value to drive to USB PHY
     io_rw_32 phy_trim;
-    io_rw_32 linestate_tuning;
```
lurch, line 135: This seems to have disappeared, but I can't see any reference to it in the datasheet either?

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
```diff
@@ -10,125 +12,320 @@
...
+
+    _REG_(I2C_IC_COMP_VERSION_OFFSET)
+    // I2C Component Version Register
+    // 0xffffffff [0-31]  : IC_COMP_VERSION (0x3230312a)
+    io_ro_32 comp_version;
+
+    _REG_(I2C_IC_COMP_TYPE_OFFSET)
+    // I2C Component Type Register
+    // 0xffffffff [0-31]  : IC_COMP_TYPE (0x44570140): Designware Component Type number = 0x44_57_01_40
+    io_ro_32 comp_type;
```
lurch, line 325: Again, new fields that we didn't include before.

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
```diff
@@ -10,125 +12,320 @@
...
 #define i2c0_hw ((i2c_hw_t *const)I2C0_BASE)
 #define i2c1_hw ((i2c_hw_t *const)I2C1_BASE)
 
-// List of configuration constants for the Synopsys I2C hardware (you may see
-// references to these in I2C register header; these are *fixed* values,
-// set at hardware design time):
-
-// IC_ULTRA_FAST_MODE ................ 0x0
-// IC_UFM_TBUF_CNT_DEFAULT ........... 0x8
-// IC_UFM_SCL_LOW_COUNT .............. 0x0008
```
lurch, line 67: Eeek, see https://github.com/raspberrypi/pico-feedback/issues/191 - this  definitely needs to be moved "somewhere else".

src/rp2040/hardware_regs/rp2040.svd
kilograham: it looks like this must have been hand edited b4 otherwise it wouldn't be out of step

lurch: Where is it "out of step"? It says `<resetValue>0x00000000</resetValue>` here because https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/xosc.h#L142 says `#define XOSC_STARTUP_DELAY_RESET  "-"` (i.e. unknown / undefined).
https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/include/hardware/regs/xosc.h and https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_regs/rp2040.svd are both generated from the same "upstream source" :slightly_smiling_face:

src/rp2040/hardware_structs/include/hardware/structs/xip_ctrl.h
kilograham: not sure where these came from really, we don't really have anywhere better to put them

src/rp2040/hardware_structs/include/hardware/structs/spi.h
kilograham: yeah i remove them.. no one has ever complained about them not being there

src/rp2040/hardware_structs/include/hardware/structs/dma.h
kilograham: it sounds like it is read/write but always reads as zero, no? in which case I guess `io_rw_32` is fine... if the SVD set it was wrtie only then we'd use wo

src/rp2040/hardware_structs/include/hardware/structs/dma.h
kilograham: yeah this looks like it it is intended to be read

src/rp2040/hardware_structs/include/hardware/structs/usb.h
lurch: I.e. any external code that was previously using `sof_rw` (if such code exists :wink: ) will now break, because it's been renamed to `sof_wr`.

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
kilograham: ugh - yes this should not have been here, perhaps it was _hard_ to put it in the regs header where it really belongs ... can we do that now (presumably upstream even of the regs header)

src/rp2040/hardware_structs/include/hardware/structs/rtc.h
kilograham: i bet it is something to do with elipsis

src/rp2040/hardware_structs/include/hardware/structs/uart.h
kilograham: fixed

src/rp2040/hardware_structs/include/hardware/structs/usb.h
kilograham: oops didn't see this, will fix

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
lurch: The regs header is also autogenerated, but I'll have a think about how to include this comment-text.

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
```diff
@@ -10,125 +12,322 @@
...
+
+    _REG_(I2C_IC_COMP_VERSION_OFFSET) // I2C_IC_COMP_VERSION
+    // I2C Component Version Register
+    // 0xffffffff [0-31]  : IC_COMP_VERSION (0x3230312a)
+    io_ro_32 comp_version;
+
+    _REG_(I2C_IC_COMP_TYPE_OFFSET) // I2C_IC_COMP_TYPE
+    // I2C Component Type Register
+    // 0xffffffff [0-31]  : IC_COMP_TYPE (0x44570140): Designware Component Type number = 0x44_57_01_40
+    io_ro_32 comp_type;
```
lurch, line 327: Should we get rid of these additions, the same way you got rid of the SPI and UART version fields?

liamfraser: I don't see a problem with adding these. They don't have to be used by software

src/rp2040/hardware_structs/include/hardware/structs/rtc.h
liamfraser: Created an issue for this https://github.com/raspberrypi/svd2struct/issues/2

src/rp2040/hardware_regs/rp2040.svd
Wren6991: Yeah the correct reset value is 0xc4 but there is currently no way to express the actual as-taped-out reset value of an `undef` field in VRB, so I added it to the info string so that it at least appears in the datasheet and header. I'll go see if there is an internal issue for this already.

src/rp2040/hardware_structs/include/hardware/structs/rtc.h
```diff
@@ -8,22 +10,103 @@
...
+    // 0x20000000 [29]    : MIN_ENA (0): Enable minute matching
+    // 0x10000000 [28]    : SEC_ENA (0): Enable second matching
+    // 0x07000000 [24-26] : DOTW (0): Day of the week
+    // 0x001f0000 [16-20] : HOUR (0): Hours
+    // 0x00003f00 [8-13]  : MIN (0): Minutes
+    // 0x0000003f [0-5]   : SEC (0): Seconds
     io_rw_32 irq_setup_1;
-    io_rw_32 rtc_1;
-    io_rw_32 rtc_0;
-    io_rw_32 intr;
```
Wren6991, line 23: Little confused by this one but it's because looks like `intr` is read-only because all the IRQs are level-sensitive

src/rp2040/hardware_structs/include/hardware/structs/sio.h
Wren6991: I mean arguably they are write-only -- you get garbage when you read, and all-zeroes just happens to be the cheapest possible garbage to generate. I had a quick look at the hardware source and there is a comment saying `// write-only`. So I guess this is more of a philosophical question.

src/rp2040/hardware_structs/include/hardware/structs/usb.h
```diff
@@ -112,38 +122,450 @@ static_assert(sizeof(usb_host_dpram_t) == USB_DPRAM_MAX, "");
...
+    _REG_(USB_ADDR_ENDP1_OFFSET) // USB_ADDR_ENDP1
+    io_rw_32 int_ep_addr_ctrl[USB_HOST_INTERRUPT_ENDPOINTS]; // 15
+
+    _REG_(USB_MAIN_CTRL_OFFSET) // USB_MAIN_CTRL
+    // Main control register
+    // 0x80000000 [31]    : SIM_TIMING (0): Reduced timings for simulation
+    // 0x00000002 [1]     : HOST_NDEVICE (0): Device mode = 0, Host mode = 1
+    // 0x00000001 [0]     : CONTROLLER_EN (0): Enable controller
     io_rw_32 main_ctrl;
-    io_rw_32 sof_rw;
```
Wren6991, line 118: Is it SOF_WR or SOF_RW? Is this just a typo we need to preserve?

src/rp2040/hardware_structs/include/hardware/structs/rtc.h
lurch: Does that mean if you set the RTC alarm to e.g. match on everything except the seconds field, that the RTC IRQ will remain active for the entire minute?

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
liamfraser: Why not have it in a platform def type file. Could be in the main platform_defs.h or in a specific, i2c_defs.h. Then we don't need to hack the autogenerated files and it can be included by the i2c driver

src/rp2040/hardware_regs/rp2040.svd
lurch: Fixed in #601

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
lurch: Also now fixed in #601 :grin:

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
Wren6991: Agree with Liam

src/rp2040/hardware_structs/include/hardware/structs/i2c.h
Wren6991: Yep these are fine, they aren't very useful but no reason not to include them

src/rp2040/hardware_structs/include/hardware/structs/sio.h
Wren6991: Thanks @lurch

Issue #545, opened 2021-08-24
Firmware update from firmware not possible


geurtv:
I'm trying to update the RP2040 firmware from firmware running on that same RP2040. I download new firmware (via USB) into an unused region in flash and when done copy that data to the start of the flash and then reboot. 

So the function that copies from flash to flash and then waits for reboot must run entirely from memory and never ever from flash. This, however, is currently not possible because the function `rom_func_lookup`, which both `flash_range_erase` and `flash_range_program` call, is still in flash.

When I wrap `rom_func_lookup` with a `__not_in_flash_func` (and make sure the compiler doesn't use `memcpy` in flash-to-flash) it all works fine.

kilograham:
Ah yes, that is a weird oversight

lurch:
Is any of the discussion in #410 applicable here?

geurtv:
Hmm, hadn't noticed that issue. Yes, I'd say this is a duplicate.

Wren6991:
> When I wrap rom_func_lookup with a __not_in_flash_func (and make sure the compiler doesn't use memcpy in flash-to-flash) it all works fine.

It might be worth adding that decorator, though the user in #410 also saw functions like this one fail to inline, which was surprising:

```c
static inline uint32_t rom_table_code(uint8_t c1, uint8_t c2) {
    return (((uint)c2) << 8u) | (uint)c1;
}
```

Erasing the image you are currently XIPing from is pretty fraught, for the reasons mentioned in #410. To solve it fully we would need to pull all of the hardware_flash code into RAM, which is a waste of RAM for people who are using it to write to flash regions that don't contain the current program image :) This would *still* be dangerous, because there is a period of several milliseconds where you have *no* bootable image in flash, because it just erased itself, so you have bricked your board until you can access the USB boot and reflash. Because of this it doesn't seem sensible to support this use case for the hardware_flash library.

Slightly cleaner approach is to have a self-contained stub program in e.g. the first 4k of flash which:

- Runs on every flash boot, because it's entered by the second stage
- Checks watchdog registers to see if it should do a flash->flash copy, to install a new program image
- Checks for a valid image immediately following it in flash, and jumps in in the same way that boot2 jumps in
- If there is no valid image, go to a rescue/bootloader mode so that a new image can be installed

[...]

PR #586, opened 2021-10-04
Adding/propage macros/signature typedefs for ROM functions. Make rom_func_lookup non-flash safe for flash functions

### LINKED ISSUE #410: __not_in_flash_func() requested around rom_func_lookup() and others.
Hi guys,
    I think I'm pushing the boundaries a little here which is why I have created the issue I have.

Here's the background;

- I've ported an app from an Adafruit Arduino M0 board with an SD card connected to it.
- The Arduino had the facility to put a UF2 file on the SD card and it would reflash itself in the same way that on the Pico you can drop the UF2 file on the drive that appears when you plug in the USB.
- I wanted to emulate what the Adafruit board did, so I have written some code which goes into the startup of my app and does the following;

1. Initialises an SD card interface using SPI.
2. Looks for a UF2 file
3. If found, processes the contents of the UF2 file to a base address half way up the flash (ignoring the addresses specified in the UF2 file itself)
4. Now running from code based in RAM, it moves the processed data back down to the Flash base address and then resets the device.

So long as all of my apps contain the same code, I can reprogram the device just by putting the UF2 image on the SD card and inserting it.

The issue I seem to have caused myself is that calling the flash routines in the code below causes the app to crash because there are certain dependent functions that are in the flash I'm writing to. 

`
	for (size_t PageCount = 0; PageCount < NewNumBlocks; PageCount++)
	{
		uint8_t DataBlock[FLASH_PAGE_SIZE];
									
		// Sectors are 4096 bytes, so erase sectors at a time...
		if((PageCount % 16) == 0)
			flash_range_erase((uint32_t)(PageCount * FLASH_PAGE_SIZE), FLASH_SECTOR_SIZE);
						
		// Pages are 256 bytes...
		// Copy from the destination FLASH address into the DataBlock[] in RAM.
		for (int i = 0 ; i < FLASH_PAGE_SIZE ; i++)
			DataBlock[i] = *pNewFW++;
						
		// Now program from DataBlock[] in RAM to new address in FLASH.
		flash_range_program(OriginalFWOffset, &DataBlock[0], FLASH_PAGE_SIZE);
						
		OriginalFWOffset += FLASH_PAGE_SIZE;
	}
`
It appears that flash_range_erase() and flash_range_program() are declared **__no_inline_not_in_flash_func**, but the dependencies within them are not.

I can fix the problem by changing the declarations of the functions below to be **__not_in_flash_func()**

- rom_func_lookup (bootrom.c)
- rom_funcs_lookup (bootrom.c)
- __compiler_memory_barrier (platform.h)
- rom_table_code (bootrom.h)

My questions are;
1. Is that the right way to fix the problem?
2. I'm not entirely sure what __compiler_memory_barrier does, so do I risk "breaking" that?
3. should these anciliary functions already be declared to be not in flash anyway?

P.S. I do have this all working, but I'm after clarification as to whether or not what I am doing to delcare these additional functions as **not in flash** is correct and should be ported into the standard SDK or not.

### LINKED ISSUE #562: add #define to pico/bootrom.h with `rom_table_code`s of the functions in the bootrom

kilograham:
fixes #562 
fixed #410

### DIFF
--- a/src/rp2_common/pico_bootrom/include/pico/bootrom.h
+++ b/src/rp2_common/pico_bootrom/include/pico/bootrom.h
@@ -22,7 +22,7 @@ extern "C" {
  * \ingroup pico_bootrom
  *
  * These codes are uses to lookup data or function addresses in the bootrom
- * 
+ *
  * \param c1 the first character
  * \param c2 the second character
  * \return the 'code' to use in rom_func_lookup() or rom_data_lookup()
@@ -30,6 +30,7 @@ extern "C" {
 static inline uint32_t rom_table_code(uint8_t c1, uint8_t c2) {
     return (((uint)c2) << 8u) | (uint)c1;
 }
+
 
 /*!
  * \brief Lookup a bootrom function by code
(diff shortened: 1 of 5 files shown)


kilograham:
fixes #545

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -60,7 +61,72 @@ void *rom_data_lookup(uint32_t code);
...
+
+/*!
+ * \brief Lookup a bootrom function by code. This method is forceably inlined into the caller for FLASH/RAM sensitive code usage
+ * \ingroup pico_bootrom
+ * \param code the code
+ * \return a pointer to the function, or NULL if the code does not match any bootrom function
+ */
+static __force_inline void *rom_func_lookup_inline(uint32_t code) {
+    rom_table_lookup_fn rom_table_lookup = (rom_table_lookup_fn) rom_hword_as_ptr(0x18);
+    uint16_t *func_table = (uint16_t *) rom_hword_as_ptr(0x14);
```
lurch, line 127: Where does the magic `0x14` come from? Is it worth making it a `#define` ? :shrug: 
(Doesn't matter if not, but hopefully it doesn't hurt to ask)

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -60,7 +61,72 @@ void *rom_data_lookup(uint32_t code);
  */
 bool rom_funcs_lookup(uint32_t *table, unsigned int count);
 
-typedef void __attribute__((noreturn)) (*reset_usb_boot_fn)(uint32_t, uint32_t);
+// The following is a list of bootrom functions available via rom_func_lookup and their signature
+
+typedef uint32_t (*rom_popcount32_fn)(uint32_t);
+#define ROM_FUNC_POPCOUNT32             rom_table_code('P','3')
```
lurch, line 67: Could / should these `#define`s also be used in https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_bit_ops/bit_ops_aeabi.S ?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -60,7 +61,72 @@ void *rom_data_lookup(uint32_t code);
  */
 bool rom_funcs_lookup(uint32_t *table, unsigned int count);
 
-typedef void __attribute__((noreturn)) (*reset_usb_boot_fn)(uint32_t, uint32_t);
+// The following is a list of bootrom functions available via rom_func_lookup and their signature
```
lurch, line 64: Should this list also include `rom_table_code('S', 'D')` and `rom_table_code('S', 'F')` from https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_double/double_init_rom.c ? (and `src/rp2_common/pico_float/float_init_rom.c`)

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -60,7 +61,72 @@ void *rom_data_lookup(uint32_t code);
...
+
+typedef uint32_t *(*rom_memcpy_fn)(uint8_t *, const uint8_t *, uint32_t);
+#define ROM_FUNC_MEMCPY                 rom_table_code('M','C')
+
+typedef uint32_t *(*rom_memcpy44_fn)(uint32_t *, const uint32_t *, uint32_t);
+#define ROM_FUNC_MEMCPY44               rom_table_code('C','4')
+
+typedef void __attribute__((noreturn)) (*rom_reset_usb_boot_fn)(uint32_t, uint32_t);
+typedef rom_reset_usb_boot_fn reset_usb_boot_fn; // kept for backwards compatibility
+#define ROM_FUNC_RESET_USB_BOOT         rom_table_code('U','B')
```
lurch, line 92: Should there be a space after the comma, or is the missing-space deliberate?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
kilograham: it is in the datasheet - dont care to add a define

src/rp2_common/pico_bootrom/include/pico/bootrom.h
kilograham: can't be because that is assembly

src/rp2_common/pico_bootrom/include/pico/bootrom.h
kilograham: nah, those aren't for public consumption

PR #592, opened 2021-10-06
respect OUTPUT_NAME target property when generating supplemental files

kilograham:


src/rp2_common.cmake
```diff
@@ -36,8 +36,8 @@ function(pico_add_extra_outputs TARGET)
 
         add_custom_command(TARGET ${TARGET}_symlinked POST_BUILD
                 COMMAND rm -f "${PICO_SYMLINK_ELF_AS_FILENAME}"
-                COMMAND ln -s -r ${TARGET}${CMAKE_EXECUTABLE_SUFFIX} "${PICO_SYMLINK_ELF_AS_FILENAME}"
-                COMMENT "Symlinking from ${PICO_SYMLINK_ELF_AS_FILENAME} to ${TARGET}${CMAKE_EXECUTABLE_SUFFIX}"
+                COMMAND ln -s -r $<TARGET_FILE:${TARGET}> "${PICO_SYMLINK_ELF_AS_FILENAME}"
+                COMMENT "Symlinking from ${PICO_SYMLINK_ELF_AS_FILENAME} to $<TARGET_FILE:${TARGET}>"
```
lurch, line 40: Should the COMMENT be before the COMMAND ? :shrug: (Or perhaps the COMMENT should be changed to say "Created symlink ..." instead of "Symlinking ..."?)

lurch: Oh yes, mea culpa https://cmake.org/cmake/help/latest/command/add_custom_command.html
Whoops.

PR #594, opened 2021-10-07
add watchdog_enable_caused_reboot (minor compatibility issue)

kilograham:
Note there is a small backwards incompatibility in so far as watchdog_enable sets scratch[4] to a magic value not 0 now, however this behavior was not documented before.

### DIFF
--- a/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
+++ b/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
@@ -90,9 +90,15 @@ bool watchdog_caused_reboot(void);
  * \brief Did watchdog_enable cause the last reboot?
  * \ingroup hardware_watchdog
  *
- * This method checks watchdog scratch register 4 for a magic number set by \ref watchdog_enable. This will not be present
- * if the watchdog reset is initiated by \ref watchdog_reboot or by the RP2040 bootrom (e.g. dragging a UF2 onto the RPI-RP2 drive)
- * @return true if the watchdog timer or a watchdog force caused the last reboot and the watchdog was initiated by \ref watchdog_enable
+ * Perform additional checking along with \ref watchdog_cuased_reboot to determin if a watchdog time out initiated by
+ * \ref watchdog_enable caused the last reboot.
+ *
+ * This method checks for a special value in watchdog scratch register 4 placed there by \ref watchdog_enable.
+ * This would not be present if a watchdog reset is initiated by \ref watchdog_reboot or by the RP2040 bootrom
+ * (e.g. dragging a UF2 onto the RPI-RP2 drive).
+ *
+ * @return true if the watchdog timer or a watchdog force caused (see \reg watchdog_caused_reboot) the last reboot
+ *              and the watchdog reboot happened after \ref watchdog_enable was called
  * @return false there has been no watchdog reboot since run has been
  */
 bool watchdog_enable_caused_reboot(void);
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
```diff
@@ -80,6 +86,17 @@ void watchdog_enable(uint32_t delay_ms, bool pause_on_debug);
...
 bool watchdog_caused_reboot(void);
 
+/**
+ * \brief Did watchdog_enable cause the last reboot?
+ * \ingroup hardware_watchdog
+ *
+ * This method checks watchdog scratch register 4 for a magic number set by \ref watchdog_enable. This will not be present
+ * if the watchdog reset is initiated by \ref watchdog_reboot or by the RP2040 bootrom (e.g. dragging a UF2 onto the RPI-RP2 drive)
+ * @return true if the watchdog timer or a watchdog force caused the last reboot and the watchdog was initiated by \ref watchdog_enable
+ * @return false there has been no watchdog reboot since run has been
```
lurch, line 96: "since run has been" ??

lurch:
Is it perhaps worth adding a note about this to "2.8.1.1. Watchdog Boot" or "4.7.4. Scratch Registers" in the RP2040 Datasheet?
Or are the Watchdog Scratch registers implicitly "for use by the bootrom and SDK only"? :man_shrugging:

src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
lurch: Still not clear what "since run has been" is supposed to mean, and I'm having a hard time interpreting the `@return true` part too. I'll have another look in the morning when I'm less tired.

kilograham: @Wren6991 @liamfraser  "watchdog run has been" was in the comments at release - perhaps a good sentence which described whatever resets the watchdog reboot flag

src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
```diff
@@ -80,6 +86,23 @@ void watchdog_enable(uint32_t delay_ms, bool pause_on_debug);
...
+ * \ingroup hardware_watchdog
+ *
+ * Perform additional checking along with \ref watchdog_caused_reboot to determine if a watchdog timeout initiated by
+ * \ref watchdog_enable caused the last reboot.
+ *
+ * This method checks for a special value in watchdog scratch register 4 placed there by \ref watchdog_enable.
+ * This would not be present if a watchdog reset is initiated by \ref watchdog_reboot or by the RP2040 bootrom
+ * (e.g. dragging a UF2 onto the RPI-RP2 drive).
+ *
+ * @return true if the watchdog timer or a watchdog force caused (see \reg watchdog_caused_reboot) the last reboot
```
lurch, line 100: Does #561 mean that this "watchdog force" part is no longer true? Which I guess would mean this could be shortened to
`@return true if the watchdog timer caused a reboot after \ref watchdog_enable was called` which IMHO seems much more readable? :shrug:

src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
Wren6991: New wording makes more sense to me

Wren6991 (approved):

PR #596, opened 2021-10-07
add template PR - very polite

kilograham:


andrum99:
Note that another way of (almost completely) solving the problem of prospective contributors trying to submit against the `master` branch is to change the default branch from `master` to `develop`. Obviously this has other implications as well, so may not be desirable. (That's how aalan has the documentation repo set up).

PR #602, opened 2021-10-08
move inline assembly constant out into C code so it doesn't get out of range in a large function

kilograham:


Wren6991 (approved):
Ouch, makes sense

PR #604, opened 2021-10-08
add dma_timer related methods

kilograham:


penguin42:
Yep that looks good (especially with the compiler warning; (I think it had a point!)

PR #609, opened 2021-10-11
add irq_get_priority method - improve efficiency of irq_init_priorities

### LINKED ISSUE #606: core1 takes over 100us to start executing code using multicore_launch_core1()
Testing the 'multicore_runner_queue' example, I found out that calling multicore_launch_core1(), it seems to take over 100µs for core1 to actually start executing the code in the entry function passed as an argument.

I'm not saying this is a bug or an issue - but I wanted to know if that was "normal", as I could not find this documented. I think this could be interesting to know.

kilograham:
fixes #606 (well speeds it up from 100us to 40us anyway)

### DIFF
--- a/src/rp2_common/hardware_irq/irq.c
+++ b/src/rp2_common/hardware_irq/irq.c
@@ -407,7 +407,7 @@ void irq_init_priorities() {
     static_assert(!(NUM_IRQS & 3), "");
     uint32_t prio4 = (PICO_DEFAULT_IRQ_PRIORITY & 0xff) * 0x1010101u;
     io_rw_32 * p = (io_rw_32 *)(PPB_BASE + M0PLUS_NVIC_IPR0_OFFSET);
-    for (int i = 0; i < NUM_IRQS / 4; i++) {
+    for (uint i = 0; i < NUM_IRQS / 4; i++) {
         *p++ = prio4;
     }
 #endif


src/rp2_common/hardware_irq/irq.c
```diff
@@ -400,8 +404,11 @@ void irq_add_tail_to_free_list(struct irq_handler_chain_slot *slot) {
 
 void irq_init_priorities() {
 #if PICO_DEFAULT_IRQ_PRIORITY != 0
-    for (uint irq = 0; irq < NUM_IRQS; irq++) {
-        irq_set_priority(irq, PICO_DEFAULT_IRQ_PRIORITY);
+    static_assert(!(NUM_IRQS & 3), "");
+    uint32_t prio4 = (PICO_DEFAULT_IRQ_PRIORITY & 0xff) * 0x1010101u;
+    io_rw_32 * p = (io_rw_32 *)(PPB_BASE + M0PLUS_NVIC_IPR0_OFFSET);
+    for (uint i = 0; i < NUM_IRQS / 4; i++) {
+        *p++ = prio4;
```
lurch, line 411: So it took a bit of head-scratching, but I guess this "speeds things up" by writing 4 priorities at a time, instead of writing the priorities individually, which means that `M0PLUS_NVIC_IPR0` only gets written to 8 times instead of 32 times?
(perhaps a little one-line comment would prevent other people getting confused in future? :wink: )

Wren6991 (approved):
Nice, surprised that seemingly innocuous loop took so many cycles but it makes sense

src/rp2_common/hardware_irq/irq.c
lurch: And so I guess this also implies that setting `PICO_DEFAULT_IRQ_PRIORITY` to 0 would speed up the issue in #606 even further? (assuming of course that the user isn't bothered about IRQs having different priorities)

PR #610, opened 2021-10-12
Use reg[0] description for hardware structs register arrays

kilograham:


src/rp2040/hardware_structs/include/hardware/structs/dma.h
```diff
@@ -145,6 +145,11 @@ typedef struct {
     io_rw_32 ints1;
 
     _REG_(DMA_TIMER0_OFFSET) // DMA_TIMER0
+    // (Description copied from DMA_TIMER0 applies similarly all array members)
```
lurch, line 148: I'm not sure we need the "Description copied from" part? Would "(DMA_TIMER0 applies similarly all array members)" be sufficient?

src/rp2040/hardware_structs/include/hardware/structs/interp.h
```diff
@@ -22,21 +22,52 @@
...
+    // (Description copied from SIO_INTERP0_BASE0 applies similarly all array members)
+    //
+    // Read/write access to BASE0 register
     io_rw_32 base[3];
 
     _REG_(SIO_INTERP0_POP_LANE0_OFFSET) // SIO_INTERP0_POP_LANE0
+    // (Description copied from SIO_INTERP0_POP_LANE0 applies similarly all array members)
+    //
+    // Read LANE0 result, and simultaneously write lane results to both accumulators (POP)
     io_ro_32 pop[3];
```
lurch, line 40: Trying to match this up to the RP2040 datasheet, it's slightly confusing here that `pop[0]` refers to `INTERP0_POP_LANE0`, `pop[1]` refers to `INTERP0_POP_LANE1` but `pop[2]` refers to `INTERP0_POP_FULL`.
However I guess there's not really much that can be done about that in an automated manner.

src/rp2040/hardware_structs/include/hardware/structs/iobank0.h
```diff
@@ -45,12 +45,117 @@ typedef struct {
 
 typedef struct {
     _REG_(IO_BANK0_PROC0_INTE0_OFFSET) // IO_BANK0_PROC0_INTE0
+    // (Description copied from IO_BANK0_PROC0_INTE0 applies similarly all array members)
+    //
+    // Interrupt Enable for proc0
+    // 0x80000000 [31]    : GPIO7_EDGE_HIGH (0)
```
lurch, line 51: For `inte[0]` this is `GPIO7_EDGE_HIGH`, for `inte[1]` this is `GPIO15_EDGE_HIGH`, for `inte[2]` it's `GPIO23_EDGE_HIGH` and for `inte[3]` it's `Reserved.`.
I wonder if it's worth adding a special-case to the svd2struct comment-generator-code for these `IO_BANK0` array-regs and instead just say something like `// Please refer to the RP2040 datasheet for full details` ? Which is a bit of a cop-out but I can't think of anything better off the top of my head. :confused:

kilograham: yeah you can just follow the _REG_ link to see the real registers

PR #612, opened 2021-10-14
Updated reg_headers and SVD

lurch:


lurch:
I didn't update the struct_headers because that created a conflict with #610

PR #616, opened 2021-10-15
Added some waveshare boards definition

EngineerWill:
Added some waveshare boards definition
[RP2040-LCD-0.96](https://www.waveshare.net/shop/RP2040-LCD-0.96.htm)
The other boards are not on the shop yet, but I've got them.
Do you need me to provide corresponding schematic diagram?

### DIFF
--- a/src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
+++ b/src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
+//       SHOULD ONLY CONSIST OF PREPROCESSOR DIRECTIVES
+// -----------------------------------------------------
+
+// This header may be included by other board headers as "boards/pico.h"
+
+#ifndef _BOARDS_RP2040_LCD_0_96_H
+#define _BOARDS_RP2040_LCD_0_96_H
+
+// For board detection
+#define WAVESHARE_RP2040_LCD_0_96
+
+// --- UART ---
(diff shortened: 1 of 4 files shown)


src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
+//       SHOULD ONLY CONSIST OF PREPROCESSOR DIRECTIVES
+// -----------------------------------------------------
+
+// This header may be included by other board headers as "boards/pico.h"
```
lurch, line 12: I think you want to get rid of this line! :wink:

src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
```diff
@@ -0,0 +1,109 @@
...
+#define PICO_DEFAULT_SPI_RX_PIN 16
+#endif
+#ifndef PICO_DEFAULT_SPI_CSN_PIN
+#define PICO_DEFAULT_SPI_CSN_PIN 17
+#endif
+
+// --- LCD ---
+
+#ifndef PICO_DEFAULT_LCD_SPI
+#define PICO_DEFAULT_LCD_SPI 1
```
lurch, line 64: I wonder if this should be just `PICO_LCD_SPI` (i.e. remove the `DEFAULT`) as I assume there's no way to change which SPI bus the LCD screen is connected to?
@kilograham Should this be using a prefix other than `PICO_` for "board-specific" values which aren't otherwise used in the SDK? (compare to e.g. https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/pimoroni_keybow2040.h )

EngineerWill: ok.
I am a little busy these days and did not reply in time. 
I will pull Request it again soon.

PR #619, opened 2021-10-18
Added some waveshare boards definition

EngineerWill:
Added some waveshare boards definition
RP2040-LCD-0.96
The other boards are not on the shop yet, but I've got them.

This is my first time to pull request. I may have made a mistake in several operations. Please forgive me

### DIFF
--- a/src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
+++ b/src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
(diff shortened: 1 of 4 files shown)


src/boards/include/boards/waveshare_rp2040_lcd_0.96.h
```diff
@@ -0,0 +1,107 @@
...
+#endif
+#ifndef PICO_DEFAULT_SPI_RX_PIN
+#define PICO_DEFAULT_SPI_RX_PIN 16
+#endif
+#ifndef PICO_DEFAULT_SPI_CSN_PIN
+#define PICO_DEFAULT_SPI_CSN_PIN 17
+#endif
+
+// --- LCD ---
+#ifndef PICO_LCD_SPI
```
lurch, line 61: @kilograham As I mentioned in one of the previous iterations of this PR, I dunno if we ought to be allowing this to introduce a `PICO_LCD_SPI` define (which IMHO feels a bit generic), or requesting that it instead uses a more board-specific name e.g. `WAVESHARE_RP2040_LCD_SPI` ?

lurch:
One comment, but otherwise LGTM.

PR #644, opened 2021-10-31
allow override of XOSC_MHZ

kilograham:


liamfraser (approved):

src/rp2_common/hardware_xosc/xosc.c
```diff
@@ -13,8 +13,8 @@
 #include "hardware/regs/xosc.h"
 #include "hardware/xosc.h"
 
-#if XOSC_MHZ < 1 || XOSC_MHZ > 15
-#error XOSC_MHZ must be in the range 1-15
+#if XOSC_MHZ < 1 || XOSC_MHZ > 50
```
lurch, line 16: Section 2.16.1. in https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf says "The RP2040 supports
1MHz to 15MHz crystals" ?

lurch: Ahhh, just seen the comment in #367 about XIN, and indeed the datasheet says "it is possible to drive an external clock directly into XIN (aka XI), and disable the oscillator circuit. In this mode XIN can be driven at up to 50MHz."

PR #655, opened 2021-11-12
pio_set_irqn_source_mask_enabled() sets wrong irq

alien999999999:
if irq_index == 0, irq_index evaluates to false; which uses pio_set_irq1_source_mask_enabled() ; and vice versa


lurch (approved):
Well spotted! Your new version now matches the logic used in `pio_set_irqn_source_enabled`

lurch:
This fixes #656

kilograham (approved):
oops

Issue #654, opened 2021-11-11
IVT Syscall Support Request/Feature


spit4520:
Hi All,

I am looking to play around with triggering software interrupts on the RP2040, and I have gotten some conflicting documentation on weather or not this would be possible. The IVT should be able to be configured from the V6 spec from ARM, but I can't quite determine if that is possible on the 2040.

I want to experiment with "thread safety" and airlift some of the code at my job out of deep gluey MCU code into more isolated processes. I know that RTOS exists, this is more of a learning exercise and I would like to add a few examples to the main repo.

Could someone point me at the correct place in either the ARM docs or the Pico docs on how to manually configure an internal software interrupt? (I am trying to avoid the hack of using the PIO between multiple cores to use that for interrupting)

Thanks!

Scott

lurch:
"2.3.2. Interrupts" in https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf and "4.1.14. hardware_irq" in https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf ?

kilograham:
this is not an SDK issue (if the documentation is insufficient then open an issue in pico-feedback)

PR #658, opened 2021-11-12
improve comment

kilograham:


src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -258,6 +258,10 @@ irq_handler_t irq_get_vtable_handler(uint num);
 /*! \brief Clear a specific interrupt on the executing core
  *  \ingroup hardware_irq
  *
+ * This method is only useful for "software" IRQs that are not connected to hardware (i.e. IRQs 26-31)
```
lurch, line 261: The `\param` part refers to `\ref interrupt_nums` and that only lists IRQs 0-25 ? (should that table also lists the software IRQs?)

lurch:
Should a similar comment also be added to the `irq_set_pending` function? (which I assume is the "inverse function" to `irq_clear`?)

lurch:
I assume this is inspired by #654 ?

liamfraser (approved):

PR #661, opened 2021-11-15
Make Error: duplicate 'inline'

### LINKED ISSUE #659: Compiling Error: duplicate 'inline'
When running `make`, the SDK fails to build. The error thrown is:

```shell
error: duplicate 'inline'
```
on file [pico-sdk/src/rp2_common/pico_platform/include/pico/platform.h](https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/pico_platform/include/pico/platform.h)

splch:
Fixes #659
Replaces #660

When building the Pico SDK, a `duplicate 'inline'` error is thrown. This error is thrown because `__always_inline`, `inline`, and `__force_inline` are used together. Removing __always_inline fixes the build error and compilation.

### DIFF
--- a/src/rp2_common/pico_platform/include/pico/platform.h
+++ b/src/rp2_common/pico_platform/include/pico/platform.h
@@ -234,7 +234,7 @@ extern "C" {
  *
  */
 #if defined(__GNUC__) && __GNUC__ <= 7
-#define __force_inline inline __always_inline
+#define __force_inline inline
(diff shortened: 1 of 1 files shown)


kilograham:
this is not the right fix; it must be a problem with the particular compiler version you are using

PR #665, opened 2021-11-22
Add wiznet_w5100s_evb_pico support

irinakim12:

PR #666, opened 2021-11-24
Add wiznet_w5100s_evb_pico support

irinakim12:
#665 Change the Branch for PR


src/boards/include/boards/wiznet_w5100s_evb_pico.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifndef PICO_FLASH_SIZE_BYTES
+#define PICO_FLASH_SIZE_BYTES (2 * 1024 * 1024)
+#endif
+
+// Drive high to force power supply into PWM mode (lower ripple on 3V3 at light loads)
+#define PICO_SMPS_MODE_PIN 23
+
+#ifndef PICO_RP2040_B0_SUPPORTED
+#define PICO_RP2040_B0_SUPPORTED 1
```
lurch, line 76: You can probably set this to 0, as I believe only B1 and B2 are available as bare-chips (e.g. for usage in 3rd-party boards).

lurch:
The schematic at https://www.mouser.co.uk/new/wiznet/wiznet-w5100sevbpico-eval-board/ suggests that there's an LED connected to GPIO25, so you could probably add:
```
#ifndef PICO_DEFAULT_LED_PIN
#define PICO_DEFAULT_LED_PIN 25
#endif
```
?

Also, you might want to add some defines for your `INTn` and `RSTn` pins? e.g.
```
#define WIZNET_W5100S_EVB_PICO_RSTN_PIN 20
#define WIZNET_W5100S_EVB_PICO_INTN_PIN 21
```
:man_shrugging:  (see https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/pimoroni_keybow2040.h for something similar)

irinakim12:
@kilograham I added defined pins that are board dependent.

src/boards/include/boards/wiznet_w5100s_evb_pico.h
```diff
@@ -0,0 +1,105 @@
...
+
+#ifndef W5100S_EVB_PICO_A0_PIN
+#define W5100S_EVB_PICO_A0_PIN 26
+#endif
+#ifndef W5100S_EVB_PICO_A1_PIN
+#define W5100S_EVB_PICO_A1_PIN 27
+#endif
+#ifndef W5100S_EVB_PICO_A2_PIN
+#define W5100S_EVB_PICO_A2_PIN 28
+#endif
```
lurch, line 44: I don't think there's any point adding defines for these, because these GPIO numbers are "hardwired" on RP2040 and can't be changed on _any_ board.

src/boards/include/boards/wiznet_w5100s_evb_pico.h
```diff
@@ -0,0 +1,105 @@
...
+
+
+#ifndef _BOARDS_WIZNET_W5100S_EVB_PICO_H
+#define _BOARDS_WIZNET_W5100S_EVB_PICO_H
+
+// For board detection
+#define WIZNET_W5100S_EVB_PICO
+// --- BOARD SPECIFIC ---
+#ifndef W5100S_EVB_PICO_INTN_PIN
+#define W5100S_EVB_PICO_INTN_PIN 20
```
lurch, line 21: The schematics at https://www.mouser.co.uk/new/wiznet/wiznet-w5100sevbpico-eval-board/ say that INTn is on GPIO21 - so are the schematics wrong, or is the define here wrong? :man_shrugging:

src/boards/include/boards/wiznet_w5100s_evb_pico.h
```diff
@@ -0,0 +1,105 @@
...
+#endif
+
+#ifndef W5100S_EVB_PICO_RSTN_PIN
+#define W5100S_EVB_PICO_RSTN_PIN 21
+#endif
+
+#ifndef W5100S_EVB_PICO_LED_PIN
+#define W5100S_EVB_PICO_LED_PIN 25
```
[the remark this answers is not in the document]
irinakim12: I thought I need to rename all the pin names with the board names except for the same pins I used on the PICO board.
If you don't have to change pin name, I'll leave it as it is in the name of PICO_DEFAULT_LED_PIN  as you told me.

src/boards/include/boards/wiznet_w5100s_evb_pico.h
irinakim12: Sorry, it is the define wrong.  RSTN and INTN pin numbers were defined opposite to each other.

src/boards/include/boards/wiznet_w5100s_evb_pico.h
irinakim12: > 

I saw for something similar.
(https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/pimoroni_tiny2040.h )

Pin **W5100S_EVB_PICO_ADC_VREF_PIN**  will be deleted as it is judged that it can be deleted.

lurch: Oh interesting.
Although digging into it some more, I see that https://shop.pimoroni.com/products/tiny-2040 has "A0", "A1" and "A2" silkscreened onto the PCB, which is probably why they chose to add defines for those to the header file? Whereas for your board it probably just makes sense to stick to the GPIO numbers, i.e. no extra defines for these pins? :shrug:

src/boards/include/boards/wiznet_w5100s_evb_pico.h
```diff
@@ -0,0 +1,102 @@
...
+// -----------------------------------------------------
+
+
+
+#ifndef _BOARDS_WIZNET_W5100S_EVB_PICO_H
+#define _BOARDS_WIZNET_W5100S_EVB_PICO_H
+
+// For board detection
+#define WIZNET_W5100S_EVB_PICO
+// --- BOARD SPECIFIC ---
```
lurch, line 19: Only a minor niggle, but could you add an extra blank line here please?

irinakim12:
I added an blank line for readability.

src/boards/include/boards/wiznet_w5100s_evb_pico.h
kilograham: definitely worth setting this to 0 if you don't use B0 chips (which you probably don't). It saves flash space if your code uses any double precision code (which includes printf)

kilograham: ...

lurch: @irinakim12 I still think that this ought to be set to 0, unless you have a reason you need to support B0 chips?
Everything else looks good to me.

irinakim12: > @irinakim12 I still think that this ought to be set to 0, unless you have a reason you need to support B0 chips? Everything else looks good to me.

HI~ @lurch  
I have now checked the contents.
The definition value(PICO_RP2040_B0_SUPPORTED) was changed to '0'.
Is there anything I need to modify ?

irinakim12:
Don't you have the action I need to do?

kilograham:
is this good @lurch

lurch:
I approved it 22 days ago :wink:

kilograham:
i was looking at the previous comment from 12 days ago, but I will take it as you are happy

kilograham (approved):

PR #672, opened 2021-12-02
Added board definition for upcoming Pimoroni Tiny 2040 2MB variant

ZodiusInfuser:


src/boards/include/boards/pimoroni_tiny2040.h
```diff
@@ -14,6 +14,7 @@
 
 // For board detection
 #define PIMORONI_TINY2040
+#define PIMORONI_TINY2040_8MB
```
lurch, line 17: In a similar vein, _perhaps_ it's worth adding `#define PIMORONI_PICOLIPO` to `pimoroni_picolipo_4mb.h` and `pimoroni_picolipo_16mb.h` ? :shrug: 

Hmmm, and (merely for the sake of "symmetry") I wonder if it might be worth adding a `pimoroni_tiny2040_8mb.h` which just `#include`s `pimoroni_tiny2040.h` ?

lurch:
@kilograham Do you think it might be worth trying to do some sort of refactoring, for (all) the boards which are otherwise-identical bar the size of the Flash chip? Or would that be too much effort for too little gain?

src/boards/include/boards/pimoroni_tiny2040.h
ZodiusInfuser: I'm in favour of adding `#define PIMORONI_PICOLIPO` but thought it best to do that as a separate PR from this?

As for the second suggestion. I would have but as the 8MB product is already out there I did not want to break compatibility for anyone.

ZodiusInfuser:
Is this okay to get merged in, or would you like to do the refactoring you suggested @lurch? We're looking to launch this as soon as Adafruit approve our CircuitPython definitions that have just gone in.

lurch:
It's entirely up to @kilograham .... (and perhaps it's also worth adding this to https://github.com/raspberrypi/pico-sdk/milestone/9 ? )

kilograham:
yeah, i don't think there is a need to refactor these headers

kilograham (approved):

PR #676, opened 2021-12-07
Add board definition for Adafruit KB2040

dshadoff:
New board from Adafruit:
https://www.adafruit.com/product/5302

Similar to ItsyBitsy, but fewer LED options and switched I2C connections.  Same Flash chip.


lurch (approved):
LGTM

dhalbert (approved):
I cross-checked the pin definitions with CircuitPython, and they look OK too. Thanks!

kilograham (approved):

PR #695, opened 2022-01-03
fix pioasm python output (#694)

### LINKED ISSUE #694: pioasm: Bad python blocking irq instruction
pioasm generated python uses wait instead of block for irq instructions, resulting in the error:

`TypeError: unsupported types for __and__: 'bound_method', 'int'`

Demonstration code:

```
$ cat test.asm
.program test
    irq wait 0

$ pioasm -o python test.asm
# -------------------------------------------------- #
# This file is autogenerated by pioasm; do not edit! #
# -------------------------------------------------- #

import rp2
from machine import Pin
# ---- #
# test #
# ---- #

@rp2.asm_pio()
def test():
    wrap_target()
    irq(wait, 0)                          # 0
    wrap()

```

MichaelWOlson:
Fixes #694
Issue:
`    irq wait 0`
was being translated as:
`    irq(wait, 0)                          # 0`
wait is incorrect, it should be block. After change:
`    irq(block, 0)                         # 0`

### DIFF
--- a/tools/pioasm/python_output.cpp
+++ b/tools/pioasm/python_output.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ * Copyright (c) 2021 Raspberry Pi (Trading) Ltd.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -283,9 +283,7 @@ struct python_output : public output_format {
                     if (arg1 & 0x2u) {
                         guts += "clear, ";
                     } else if (arg1 & 0x1u) {
-                        guts += "wait, ";
-                    } else {
-                        guts += "nowait, ";
+                        guts += "block, ";
                     }
                     auto irq = std::to_string(arg2 & 7u);
                     if (arg2 & 0x10u) {


lurch:
According to https://docs.micropython.org/en/latest/library/rp2.html#module-rp2 the `mode` parameter to `irq()` should (only) be `block` or `clear`. Presumably that means that the
```c
                    } else {
                        guts += "nowait, ";
                    }
```
part ought to be removed too? ping also @jagerkin

lurch:
...although https://github.com/micropython/micropython/blob/master/ports/rp2/modules/rp2.py#L212 _suggests_ that `noblock` might be a valid `mode` to `irq()`? (In which case I guess the MP docs would need updating too?) ping @dpgeorge

MichaelWOlson:
> According to https://docs.micropython.org/en/latest/library/rp2.html#module-rp2 the `mode` parameter to `irq()` should (only) be `block` or `clear`. Presumably that means that the
> 
> ```c
>                     } else {
>                         guts += "nowait, ";
>                     }
> ```
> 
> part ought to be removed too? ping also @jagerkin

Yeah, that should be either noblock (https://github.com/micropython/micropython/blob/master/ports/rp2/modules/rp2.py#L213) or just omitted (https://github.com/micropython/micropython/blob/master/ports/rp2/modules/rp2.py#L174). 

I'll add that as well.

MichaelWOlson:
Please take another look:

Current state.
```
    irq wait 0
    irq 0
```
were translated as:
```
    irq(wait, 0)                          # 0
    irq(nowait, 0)                        # 1
```
after change:
```
    irq(block, 0)                         # 0
    irq(0)                                # 1
```

kilograham (approved):
I'm assuming this is correct

PR #696, opened 2022-01-03
Add board definition for RP2040 Stamp

arturo182:
More details about the board available here: https://rp2040-stamp.solder.party/

### DIFF
--- a/src/boards/include/boards/solderparty_rp2040_stamp.h
+++ b/src/boards/include/boards/solderparty_rp2040_stamp.h
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2022 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
+//       SHOULD ONLY CONSIST OF PREPROCESSOR DIRECTIVES
+// -----------------------------------------------------
+//
+//------------------------------------------------------------------------------------------
+// Board definition for the Solder Party RP2040 Stamp
+//
+// This header may be included by other board headers as "boards/solderparty_rp2040_stamp.h"
+
+#ifndef _BOARDS_SOLDERPARTY_RP2040_STAMP_H
+#define _BOARDS_SOLDERPARTY_RP2040_STAMP_H
+
+// For board detection
+#define SOLDERPARTY_RP2040_STAMP
+
+// --- UART ---
+#ifndef PICO_DEFAULT_UART
+#define PICO_DEFAULT_UART 0
+#endif
+#ifndef PICO_DEFAULT_UART_TX_PIN
+#define PICO_DEFAULT_UART_TX_PIN 0
+#endif
+#ifndef PICO_DEFAULT_UART_RX_PIN
+#define PICO_DEFAULT_UART_RX_PIN 1
+#endif
+
+// --- Neopixel ---
+#ifndef PICO_DEFAULT_WS2812_PIN
+#define PICO_DEFAULT_WS2812_PIN 21
+#endif
+
+// --- I2C ---
+#ifndef PICO_DEFAULT_I2C
+#define PICO_DEFAULT_I2C       0
+#endif
+#ifndef PICO_DEFAULT_I2C_SDA_PIN
+#define PICO_DEFAULT_I2C_SDA_PIN   4
+#endif
+#ifndef PICO_DEFAULT_I2C_SCL_PIN
+#define PICO_DEFAULT_I2C_SCL_PIN   5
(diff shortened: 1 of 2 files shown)


src/boards/include/boards/solderparty_rp2040_stamp.h
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
rivimey, line 2: Change the copyright date to 2021?

src/boards/include/boards/solderparty_rp2040_stamp.h
```diff
@@ -0,0 +1,88 @@
...
+#ifndef PICO_DEFAULT_UART_RX_PIN
+#define PICO_DEFAULT_UART_RX_PIN 1
+#endif
+
+// Red LED on the Carrier
+#ifndef PICO_DEFAULT_LED_PIN
+#define PICO_DEFAULT_LED_PIN   20
+#endif
+
+// Neopixel on the Stamp
```
rivimey, line 39: My version of this was:

```
// no PICO_DEFAULT_LED_PIN
// WS2812 can be jumpered out of circuit (JP1/JP2).
#ifndef PICO_NO_WS2812
#define PICO_DEFAULT_WS2812_PIN 21
#endif
#endif // PICO_NO_WS2812
```

I decided not to set 'LED PIN' because it's only on the carrier, not the stamp itself.

rivimey (changes requested):

src/boards/include/boards/solderparty_rp2040_stamp.h
arturo182: TBH most of these are specific to the Carrier, the Stamp itself can use any pin for any function (that the RP2040 allows), but the defaults are those that are used by the Carrier cause for many people that will be their entry point :)

src/boards/include/boards/solderparty_rp2040_stamp.h
```diff
@@ -0,0 +1,88 @@
...
+#endif
+#ifndef PICO_DEFAULT_SPI_RX_PIN
+#define PICO_DEFAULT_SPI_RX_PIN 16
+#endif
+#ifndef PICO_DEFAULT_SPI_CSN_PIN
+#define PICO_DEFAULT_SPI_CSN_PIN 17
+#endif
+
+// --- FLASH ---
+#define PICO_BOOT_STAGE2_CHOOSE_GENERIC_03H 1
```
kilograham, line 73: r you sure you want this?

src/boards/include/boards/solderparty_rp2040_stamp.h
rivimey: I would personally think that this board definition should be specific to the stamp on its own; if the carrier adds useful things to that then IMO it should be another board.

While I am aware that the stamp can use any pin, the point of these definitions is to provide sane defaults for the various examples and a good basis for building a project on top of it, so defining things which are not available unless you have a carrier board is not helpful.

src/boards/include/boards/solderparty_rp2040_stamp.h
rivimey: I noticed this difference from the version I made (copied from another board) and presumed it was intended. Can you elaborate on what the difference actually means?

src/boards/include/boards/solderparty_rp2040_stamp.h
kilograham: Yes, there should probably be a separate header for the carrier board that (pre) overrides anything it needs to, and includes the stamp header at the end to pick up the rest of the defaults

src/boards/include/boards/solderparty_rp2040_stamp.h
kilograham: This select the *slow* generic flash stage2 boot, which is probably only what you want if a faster specific variant does not work. I'm pretty sure the stage2 boot used on the pico worked on the stamp too (when i tried it), but @arturo182 would know better

arturo182: I was not able to get any other variant working for the production GD25Q64 flash. IIRC I sent you one stamp with that flash and one with Winbond, which did you use? I will retest all other defs one more time to verify.

kilograham:
[hello.zip](https://github.com/raspberrypi/pico-sdk/files/7836138/hello.zip)
Yeah i tried with the GDw5q64 version.. here is a little UF2 i just tried with (which for my test purposes both has boot2_w25q080 and ups the clock speed to 270Mhz)... it prints stuff to USB console

arturo182:
@kilograham you were right, it works with w25q080, not sure what I did wrong last time :)

kilograham (approved):

arturo182:
Thank you :)

PR #711, opened 2022-01-17
Use puts instead of putc for semihosting

robbederks:
We have to make sure it's zero-terminated, and we also add a `\n`.

Haven't looked too much in `stdio.c` on how to implement this as a proper out-/input, but this makes it usable for general debugging at least.


kilograham:
closing in favor of #1013

PR #717, opened 2022-01-23
gpio.h: docs: add mask column to gpio events table, fixes #716

### LINKED ISSUE #716: Documentation for gpio_set_irq_enabled doesn't specify the mask values, only bit number
The documenation for gpio_set_irq_enabled is a little bit unhelpful in that it doesn't speicfy the mask numbers (which are the numbers that someoe is interested in when they read the docs for this function), it only specifies the bit values.

It also fails to mention the enum gpio_set_irq_enabled which defines the mask values.

I have a PR ready that fixes this.

lexbailey:
The bit column is not the most helpful number to have in the documentation, really the mask value is what you're interested in when reading that part of the docs, so I added a column for it.

I have also added a mention of gpio_irq_level too, since that appears to define the mask values.

### DIFF
--- a/src/rp2_common/hardware_gpio/include/hardware/gpio.h
+++ b/src/rp2_common/hardware_gpio/include/hardware/gpio.h
@@ -358,12 +358,14 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
  *
  * Events is a bitmask of the following:
  *
- * bit | interrupt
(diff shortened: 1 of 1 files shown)


kilograham (approved):

kilograham:
not going to merge this; the table has been updated, and i don't think it is super important/helpful

Issue #712, opened 2022-01-17
Move (or copy ?) set_sys_clock_khz() from stdlib to clocks.h


MikeDB1:
I would like to not include stdlib but for some reason set_sys_clock_khz() is only present in it.  Moving or copying this function that most people need to clocks.h would allow stdlib not to be included.  Kilograham agreed this would be a good idea.

kilograham:
yes, this code is definitely in the wrong place

kilograham:
fixed

PR #718, opened 2022-01-26
Removed unused pico_stdio target from pico_stdlib.

Molorius:
Removed unused pico_stdio target from pico_stdlib. This allows a program to use the stdlib without including stdio, which allows for a program to write its own versions of _read and _write. This issue is discussed in https://forums.raspberrypi.com/viewtopic.php?t=304905. 

This will cause potential compatibility issues if a program is expecting pico_stdio functions but only includes pico_stdlib. But it is a quick fix, and all of the examples compile without further changes needed.

### DIFF
--- a/src/common/pico_stdlib/include/pico/stdlib.h
+++ b/src/common/pico_stdlib/include/pico/stdlib.h
@@ -8,7 +8,6 @@
 #define _PICO_STDLIB_H
 
 #include "pico.h"
-#include "pico/stdio.h"
 #include "pico/time.h"
 #include "hardware/gpio.h"
 #include "hardware/uart.h"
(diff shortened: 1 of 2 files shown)


lurch:
I guess this is along similar thought-patterns to #712 ?

Molorius:
Yes similar idea but for stdio.

kilograham:
`pico_stdio` *is* going to be expected by almost everyone. It may be "unused" but that is the point of pico_stdlib - it is an aggregation of useful functionality.

If you need to replace it, either create your own library akin to pico_stdlib, or more simply
do

```
set(SKIP_PICO_STDIO 1)
```
at the top of your `CMakeLists.txt` and define your own `pico_stdio` interface library; either empty or with the new functions

Issue #568, opened 2021-09-21
Rx timeout should be controlled individually


xiongyu0523:
I saw this PR https://github.com/raspberrypi/pico-sdk/pull/504/commits/6e6cb52192b3a9455d0714e606633d82283062b4 and some discussion in issue https://github.com/raspberrypi/pico-sdk/issues/500. 

It is not flexible to have both RX and RX timeout interrupts enabled together. In some cases when UART is used together with DMA, we just want to have the RX timeout (IDLE) to tell CPU that a frame is received without triggering a DMA interrupt. 

I would suggest using a single function to control each interrupt bit. Even the original one is breaking the single-purpose design principle.

Something like

`uart_set_tx_irq_enable`
`uart_set_rx_irq_enable`
`uart_set_idle_irq_enable`
`uart_set_error_irq_enable`

xiongyu0523:
@Wren6991 @MustBeArt

Wren6991:
The RX IRQ function is trying to give a simplified view of RX interrupts, satisfying these rules:

- When at least one character has been received and not read, there will eventually be an interrupt
- When all characters have been read, there will be no more interrupts until more are received

Which can be serviced with simple interrupt handler logic, "when I get an interrupt, read all characters and return". It's not perfect but it hides some of the surprising behaviour details of the PL011, and we could write similar functions for another UART (say, PIO) which satisfy the same conditions above.

The functions you suggest give better control, but they seem quite specific to the way PL011 interrupts work.

> It is not flexible to have both RX and RX timeout interrupts enabled together. In some cases when UART is used together with DMA, we just want to have the RX timeout (IDLE) to tell CPU that a complete frame is received without triggering a DMA interrupt, and definitely, we do not want each character or FIFO threshold trigger a interrupt

Sorry, could you give a little more detail here? I'm not sure I fully understand the use case. The DREQ glue does handle SINGLE requests, so there should be no need for the processor to come and tidy up odd bytes from the FIFO when the receiver goes idle.

Are you saying the DMA is handling all of the data movement, and you want to use the idle IRQ to prompt the processor to scan through the DMA RX buffer at meaningful times?

At some point, if you are using PL011-specific functionality, you may just want to write your own wrappers

```c
void pl011_set_idle_irq(uart_hw_t *hw, bool irq) {
    if (irq) {
        hw_set_bits(hw->imsc, UART_UARTIMSC_RTIM_BITS);
    } else {
[...]

Issue #500, opened 2021-06-29
UART interrupts incompletely enabled for FIFO mode


MustBeArt:
Using the SDK interface routines to the UART, enabling FIFO mode doesn't work correctly on receive.

Symptom: the last few characters of a burst are not received promptly.

Explanation: uart_set_irq_enables() only enables the receive data ready interrupt for receive. In FIFO mode, that interrupt only fires when the receive FIFO fills up to the threshold. If a few more characters are received, but not enough to reach the FIFO threshold, that does not trigger another receive data ready interrupt (see 4.2.6.2 of the RP2040 data sheet). In order to get an interrupt for those last few characters, we also need to enable the receive timeout interrupt (see 4.2.6.4).

Proposed solution: in uart_set_irq_enables():

```
    uart_get_hw(uart)->imsc = (!!tx_needs_data << UART_UARTIMSC_TXIM_LSB) |
                              (!!rx_has_data << UART_UARTIMSC_RXIM_LSB) |
                              (!!rx_has_data << UART_UARTIMSC_RTIM_LSB);  // <--- there it is
```

lurch:
Interesting...
4.2.7. in the RP2040 datasheet shows that the FIFOs _are_ enabled, https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_uart/include/hardware/uart.h#L216 shows that the `UARTIFLS_RXIFLSEL` is set to `0`, the register-description for `UARTIFLS` says that a `RXIFSEL` of 0 triggers the receive interrupt when "Receive FIFO
becomes >= 1 / 8 full", and the UART introduction says "Separate 32x8 Tx and 32x12 Rx FIFOs".

All of which combined (using my naive understanding), suggests that the receive interrupt will only trigger when the receive fifo contains at least `32 / 8` (i.e. `4`) characters?

@kilograham Maybe we want a test similar to https://github.com/raspberrypi/pico-examples/blob/master/uart/uart_advanced/uart_advanced.c but which **doesn't** turn off the FIFOs?

kilograham:
cc @liamfraser @Wren6991

Wren6991:
```
[...]

PR #719, opened 2022-01-27
Add function to control UART FIFO trigger levels

nmk456:
The RP2040 has a register to change the UART FIFO level at which an interrupt will be generated, but this is not exposed in the SDK. The only time it is used is within the `uart_set_irq_enables` function, where it is set to the minimum value. I wrote a function `uart_set_irq_fifo_levels` to configure it to any of the possible values (1/8, 1/4, 1/2, 3/4, and 7/8 full).

There is a slight inconvenience that `uart_set_irq_enables` resets them to 1/8, so every time that function is called `uart_set_irq_fifo_levels` must be called again to set the chosen values. This could potentially be fixed by removing the lines in `uart_set_irq_enables` that does this, but I didn't want to change the functionality of an existing function to keep backwards compatibility.

I wrote some code to test this function, which can be found [here](https://gist.github.com/nmk456/68e56a52dfa75990f447cc9a9be18c30), along with it's output.

### DIFF
--- a/src/rp2_common/hardware_uart/include/hardware/uart.h
+++ b/src/rp2_common/hardware_uart/include/hardware/uart.h
@@ -122,6 +122,17 @@ typedef enum {
     UART_PARITY_ODD
 } uart_parity_t;
 
+/** \brief UART interrupt FIFO level select
+ *  \ingroup hardware_uart
+ */
+typedef enum {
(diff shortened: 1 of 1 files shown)


lurch:
I'm not sure if they're relevant to your change here, but see also #548 and #500 and #568

SPMThePlayer (approved):

PR #721, opened 2022-01-30
Adding 10-bit i2c addressing support to the SDK

### LINKED ISSUE #720: 10-bit i2c addressing is missing from pico-sdk, but does in fact work.
Over at my [repos readme](https://github.com/zbaltzer228/pico-sdk-i2c-10bit-explained/blob/develop/README.md) I explore and explain how 10-bit i2c addressing is fully supported on the RP2040 but the RP2040 datasheet suggests otherwise in the case of slave mode operation. I have another [forked repo](https://github.com/zbaltzer228/pico-sdk) with just the updates required to update the SDK to fully support 10-bit i2c addressing.

Basically I left all existing functions alone (obviously), and added about 13 new functions with identical names to the existing i2c functions but with a _10bit qualifier on the end. 

But I should ask before doing so, why was 10-bit i2c addressing left out of the SDK ? Was it simply because it's not used much?

If you all decide it's something worth adding to the SDK, I'll open a PR against this issue.

zbaltzer228:
Fixes #720 

# Raspberry Pi Pico SDK with 10-bit i2c addressing support
   As of v1.3.0 of the SDK, neither 10-bit i2c master mode nor 10-bit i2c slave mode is implemented.

   The [RP2040 Datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf)
   suggests, in two places that I can find, that 10-bit i2c addressing is not supported in slave mode.

   |  |
   |:--:|
   | Figure 1 - Suggesting 10-bit addressing is only available in Master mode|

   |  |
   |:--:|
   | Figure 2 - Also suggesting 10-bit addressing is only available in Master mode |

## Introducing 10-bit i2c addressing support

In the revised [i2c.h](https://github.com/zbaltzer228/pico-sdk-i2c-10bit-explained/blob/develop/src/rp2_common/hardware_i2c/include/hardware/i2c.h)
there are 10 new functions

- i2c_init_10bit
- i2c_set_slave_mode_10bit
- i2c_write_blocking_until_10bit
- i2c_read_blocking_until_10bit
- i2c_write_timeout_us_10bit
- i2c_write_timeout_per_char_us_10bit
- i2c_read_timeout_us_10bit
- i2c_read_timeout_per_char_us_10bit
- i2c_write_blocking_10bit
- i2c_read_blocking_10bit

In the revised [i2c.c](https://github.com/zbaltzer228/pico-sdk-i2c-10bit-explained/blob/develop/src/rp2_common/hardware_i2c/i2c.c)
there are 3 new functions

- i2c_reserved_addr_10bit
- i2c_write_blocking_internal_10bit
- i2c_read_blocking_internal_10bit

All prototypes are identical to their non-10-bit-qualified counterparts,
except wherever there is a **uint8_t addr** parameter, it has been replaced with
**uint16_t addr**.

Their functionality is also identical, but adapted to 10-bit addressing.

Example usage is found at [RP2040_10bit_i2c_host.c](https://github.com/zbaltzer228/pico-sdk-i2c-10bit-explained/blob/develop/Examples/RP2040_10bit_i2c_host/RP2040_10bit_i2c_host.c)
 and [RP2040_10bit_i2c_periph.c](https://github.com/zbaltzer228/pico-sdk-i2c-10bit-explained/blob/develop/Examples/RP2040_10bit_i2c_periph/RP2040_10bit_i2c_periph.c)

### Overview of a 10-bit i2c transaction using the examples

RP2040_10bit_i2c_host and RP2040_10bit_i2c_periph were built using the modified SDK proposed here.

Due to the limitations of the scope used to measure the i2c bus, the transaction
will be split across 6 different pictures.

In whole, the i2c host writes 2 bytes to the bus for address 0x02CF, then asks
the device with address 0x02CF for 3 bytes in response.

|  |
|:--:|
| Figure 3 - Start of the i2c host write |

The host initiates a transaction by dropping SDA while SCL is held high.

The next 5 bits are a specific sequence, 0b11110, reserved by the i2c spec to signal a 10-bit transaction,
followed by the 2 most significant bits of the 10-bit address, in this case 0x2.

The next byte delivers the rest of the 10-bit address, 0xCF.

|  |
|:--:|
| Figure 4 - End of the i2c host write |

Here, the 2 bytes are written to the bus, 0xA5 proceeded by 0x5A with the required ACK inbetween.

|  |
|:--:|
| Figure 5 - Start of the i2c host read request |

The host then starts a read request by transmitting the peripheral address, 0x02CF in a similar manner to Figure 3.
The R/W bit suggests a write, even though this is a read request. This is solved by
transmitting a 3rd byte with the Read bit set.

|  |
|:--:|
| Figure 6 - Start of the i2c host read request |

The host transmits its 3rd and final byte for the read request, this is a repeat of the first
start byte, with the reserved 0b11110 and 2 most significant bits of the periph address,
but with the Read bit set.

The peripheral then holds SDA low while it prepares to transmit its data.

|  |
|:--:|
| Figure 7 - Start of the i2c periph write |

The periph transmits its data and gets the required ACK, it transmits 3 bytes in total.
The first being a constant 0x5A, and the last two bytes are a counter that is incremented
each time a i2c transaction takes place.

|  |
|:--:|
| Figure 8 - End of the i2c transaction |

At the end, the host has acknowledged the bytes its expects and stops driving SCL.

## In conclusion
The RP2040 does, in fact, support 10-bit addressing in both host and peripheral modes.
The changes required to the SDK to fully support 10-bit addressing are provided in [i2c.h](https://github.com/zbaltzer228/pico-sdk/blob/develop/src/rp2_common/hardware_i2c/include/hardware/i2c.h)
and [i2c.c](https://github.com/zbaltzer228/pico-sdk/blob/develop/src/rp2_common/hardware_i2c/i2c.c)

### DIFF
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -30,6 +30,11 @@ static inline void i2c_unreset(i2c_inst_t *i2c) {
 static inline bool i2c_reserved_addr(uint8_t addr) {
     return (addr & 0x78) == 0 || (addr & 0x78) == 0x78;
 }
+//The "standard" reserved addresses are identical in 10-bit mode.
+//See Section 3.1.12 of v7.0 of the i2c spec.
+static inline bool i2c_reserved_addr_10bit(uint16_t addr) {
+    return (addr & 0x0378) == 0 || (addr & 0x0378) == 0x78;
+}
 
 uint i2c_init(i2c_inst_t *i2c, uint baudrate) {
     i2c_reset(i2c);
@@ -44,6 +49,32 @@ uint i2c_init(i2c_inst_t *i2c, uint baudrate) {
             I2C_IC_CON_MASTER_MODE_BITS |
             I2C_IC_CON_IC_SLAVE_DISABLE_BITS |
             I2C_IC_CON_IC_RESTART_EN_BITS |
+            I2C_IC_CON_TX_EMPTY_CTRL_BITS;
+
+    // Set FIFO watermarks to 1 to make things simpler. This is encoded by a register value of 0.
+    i2c->hw->tx_tl = 0;
+    i2c->hw->rx_tl = 0;
+
+    // Always enable the DREQ signalling -- harmless if DMA isn't listening
+    i2c->hw->dma_cr = I2C_IC_DMA_CR_TDMAE_BITS | I2C_IC_DMA_CR_RDMAE_BITS;
+
+    // Re-sets i2c->hw->enable upon returning:
+    return i2c_set_baudrate(i2c, baudrate);
+}
+uint i2c_init_10bit(i2c_inst_t *i2c, uint baudrate) {
+    i2c_reset(i2c);
+    i2c_unreset(i2c);
+    i2c->restart_on_next = false;
+
+    i2c->hw->enable = 0;
+
+    // Configure as a fast-mode master with RepStart support, 10-bit addresses
+    i2c->hw->con =
+            I2C_IC_CON_SPEED_VALUE_FAST << I2C_IC_CON_SPEED_LSB |
+            I2C_IC_CON_MASTER_MODE_BITS |
+            I2C_IC_CON_IC_SLAVE_DISABLE_BITS |
+            I2C_IC_CON_IC_RESTART_EN_BITS |
+            I2C_IC_CON_IC_10BITADDR_MASTER_BITS |
             I2C_IC_CON_TX_EMPTY_CTRL_BITS;
 
     // Set FIFO watermarks to 1 to make things simpler. This is encoded by a register value of 0.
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -112,7 +143,7 @@ uint i2c_set_baudrate(i2c_inst_t *i2c, uint baudrate) {
 }
 
 void i2c_set_slave_mode(i2c_inst_t *i2c, bool slave, uint8_t addr) {
-    invalid_params_if(I2C, addr >= 0x80); // 7-bit addresses
+    invalid_params_if(I2C, addr >= 0x400); // 7-bit addresses
```
lurch, line 146: `addr` here is a `uint8_t`, so it's impossible for it to be `>= 0x400` ?  (i.e. is this a typo?)

zbaltzer228: You are absolutely correct - it should be uint16_t, not uint8_t.

nonchip: @zbaltzer228 are you gonna fix that so we can get your feature, oooor...?

JolonB: It'd be great if this could be merged, but it's worth pointing out that this line should not be uint16_t, as this is the 7-bit addressing function. This should stay as 0x80. In addition, I believe the `invalid_params_if` in the 10-bit function should be 0x400 instead.

PR #722, opened 2022-02-04
Mark __aeabi_lmul section as allocatable and executable

KarlK90:
When using a linker script which does not provide a `.time_critical.` section that is marked "ax" and wrongly define PICO_INT64_OPS_IN_RAM=1 this would generate a malformed executable with strange runtime bugs in the worst case.

I ran into this issue when porting the RP2040 to ChibiOS which does not provide this section in its default linker scripts. The result was `__aeabi_lmul` being put outside of the flash region. But more importantly this did only show in subtle runtime bug, not a hardfault as one would suspect.

```
38 .time_critical.__aeabi_lmul 00000032 00000000 00000000 
```

Therefore I used the pattern as seen in the other pico supplied functions.

### DIFF
--- a/src/rp2_common/pico_int64_ops/pico_int64_ops_aeabi.S
+++ b/src/rp2_common/pico_int64_ops/pico_int64_ops_aeabi.S
@@ -10,11 +10,16 @@
 
 #include "pico/asm_helper.S"
 
+.macro int64_section name
 #if PICO_INT64_OPS_IN_RAM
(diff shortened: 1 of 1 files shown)


kilograham (commented):
you might want to call the macro

KarlK90:
> you might want to call the macro

You are right, apparently it was a bit late yesterday. Fixed.

kilograham (approved):

PR #709, opened 2022-01-15
add board definition for eetree gamekit rp2040

kaidegit:
This is a board made by eeTree tech

The schematic, gerber files and more infomations are as follows:
https://www.eetree.cn/project/detail/698

The board uses a Puya flash named P25Q16. We tested that it works well with `#define PICO_BOOT_STAGE2_CHOOSE_W25Q080 1`

PR #732, opened 2022-02-16
add board definition of eetree gamekit rp2040

kaidegit:
This is a board made by eeTree tech

The schematic, gerber files and more infomations are as follows:
https://www.eetree.cn/project/detail/698

The board uses a Puya flash named P25Q16. We tested that it works well with `#define PICO_BOOT_STAGE2_CHOOSE_W25Q080 1`


lurch:
Replaces #709

kilograham (approved):

PR #741, opened 2022-02-27
Enable 'Files' in Doxygen menu list, added source, caller graph and call graph to Doxigen documentation.

papaathome:
1. Doxyfile.in: added log file to Doxygen for warnings and errors.
2. Doxyfile.in: added call graph and caller graph to Doxygen output.
3. Doxyfile.in: added all source code to Doxygen output. (Probably not a good idea on a public server.)
4. DoxygenLayout.xml: fixed all problem tags by putting them in comment.
5. DoxygenLayout.xml: added menu option 'Files' by making it visible and giving it a title.

### DIFF
--- a/docs/Doxyfile.in
+++ b/docs/Doxyfile.in
@@ -61,3 +61,18 @@ PREDEFINED             = __not_in_flash_func(x) \
                          __not_in_flash(x)= \
                          __no_inline_not_in_flash(x)= \
                          __attribute__(x)=
+
+
+# Additional configuration.
+WARN_LOGFILE           = Doxygen-warnings.log
+
+HAVE_DOT               = YES
+DOT_GRAPH_MAX_NODES    = 75    # default 50 is too small for pico.h
+CALL_GRAPH             = YES   # requires HAVE_DOT = YES
+CALLER_GRAPH           = YES   # requires HAVE_DOT = YES
+
+SOURCE_BROWSER         = YES
+INLINE_SOURCES         = YES
+STRIP_CODE_COMMENTS    = NO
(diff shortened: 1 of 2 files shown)


papaathome:
See also: [C/C++ discussion therad](https://forums.raspberrypi.com/viewtopic.php?t=330422)

Kind regards,

lurch:
`100644 → 100755`
No need to mark the files as executable :)

lurch:
> Which file(s)?

According to the "Files changed" tab at the top of this PR, it's both of the files that you've modified.

papaathome:
Not done intentionally.

The git checkout was done on a samba share and is accessable for both my Debian machine and my windows 10 machine. That might be the cause of this change.

kilograham:
cc @aallan @lurch

lurch:
Could you clarify what you mean by "(Probably not a good idea on a public server.)" please?
The Doxygen-output that gets built from this is visible at https://raspberrypi.github.io/pico-sdk-doxygen/

aallan:
The doxygen also gets sucked into our PDF toolchain via XSLT and auto generates a bunch of asciidoc which is built into the Pico C/C++ SDK book. Unclear how this PR would affect that, or the micro site. 

I'd be disinclined to make arbitrary changes here. In the long term we're going to work on the micro site skin, and look at improving the XSLT and the Doxygen to Asciidoc import. 

However the effects of this are unclear. I'd be extremely reluctant to accept this PR without a lot of investigation into the (not public ally visible) knock on effects, which I don't think either @lurch or myself have time for right now. 

Sorry.

PR #745, opened 2022-03-04
Make clocks_init weak so a user can provide their own version

MichaelBell:
The `clocks_init` function in the SDK unconditionally sets up the external crystal and PLLs.  This function is called as part of runtime_init before the user's main function.

It would be nice to be able to override this behaviour so that the system could be started from ROSC only, or simply to avoid bringing up the PLLs.  This would make it straightforward to create a project that did not use an external crystal using the standard SDK.

I propose to fix this by making the `clocks_init` function weak.  The user project can then provide their own version that configures ROSC and the clocks they need appropriately, and their version will be preferred by the linker.

### DIFF
--- a/src/rp2_common/hardware_clocks/clocks.c
+++ b/src/rp2_common/hardware_clocks/clocks.c
@@ -118,7 +118,8 @@ bool clock_configure(enum clock_index clk_index, uint32_t src, uint32_t auxsrc, 
 }
 /// \end::clock_configure[]
 
-void clocks_init(void) {
+// Weak so that user can replace with their own custom version, e.g. if they are not using XOSC
+void __attribute__((weak)) clocks_init(void) {
     // Start tick in watchdog
     watchdog_start_tick(XOSC_MHZ);
 


kilograham:
I agree that the current rather monolithic and fixed runtime startup is not super flexible, and plan to reorganize it in a more customizable fashion.

Thus far we have shied away from overriding of `__weak` symbols (as opposed to having them be present or NULL) because I believe the various compiler/linkers do not support this correctly. I will take another look when we deal with `clang` support.

In any case, I'm not going to accept this for now. Feel free to open an issue to track the requirement.

You should be able to workaround it for now by setting

```
set(SKIP_HARDWARE_CLOCKS 1)
```
at the top of your `CMakeLists.txt` and providing your own replacement implementation of `hardware_clocks`. not ideal, but is should work

lurch:
> You should be able to workaround it for now by ...

@kilograham Should that be mentioned in https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf somewhere, or is it too much of a "hack" to document?

MichaelBell:
Thanks for the info on the `SKIP` settings.  I've been able to acheive what I wanted using `SKIP_PICO_RUNTIME` and copying that directory out into my own project and making a couple of tweaks.  This was nicer than just overriding `clocks_init`, as I no longer have to bring the USB clock up just to take the USB device out of reset, as I can tweak `rutime_init` not to take it out of reset.

Overall I agree that the init code just needs to be more configurable - I've raised an issue with some discussion and ideas.  Closing this request.

kilograham:
> > You should be able to workaround it for now by ...
> 
> @kilograham Should that be mentioned in https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf somewhere, or is it too much of a "hack" to document?

note i *thought* it was documented, but perhaps not very well !?

PR #747, opened 2022-03-04
irq_add_shared_handler assertion fix #724

### LINKED ISSUE #724: irq_add_shared_handler() causes assertion
Calling  irq_add_shared_handler() on same IRQ causes assertion error. It happens for both head and v1.3. Here is some sample code:

```
#include "pico/stdlib.h"
#include "hardware/dma.h"
#include "hardware/irq.h"

void dma_isr1()
{
    dma_hw->ints0 = 0; // actual code clear the bit only
}

void dma_isr2()
{
    dma_hw->ints0 = 0; // actual code clear the bit only
}

int main() 
{
    irq_add_shared_handler(DMA_IRQ_0, dma_isr1, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
    irq_add_shared_handler(DMA_IRQ_0, dma_isr2, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
    while(1) {}
}
```

Second call for irq_add_shared_handler causes assertion error at 
irq_add_shared_handler in irq.c line 240
make_branch() in irq.c line 139 (SDK 1.3)  
assert(!(delta >> 11u));

kilograham:
fixes #724

### DIFF
--- a/src/rp2_common/hardware_irq/irq.c
+++ b/src/rp2_common/hardware_irq/irq.c
@@ -135,9 +135,9 @@ irq_handler_t irq_get_exclusive_handler(uint num) {
 static uint16_t make_branch(uint16_t *from, void *to) {
     uint32_t ui_from = (uint32_t)from;
     uint32_t ui_to = (uint32_t)to;
-    uint32_t delta = (ui_to - ui_from - 4) / 2;
-    assert(!(delta >> 11u));
-    return (uint16_t)(0xe000 | (delta & 0x7ff));
+    int32_t delta = (int32_t)(ui_to - ui_from - 4);
+    assert(delta >= -2048 && delta <= 2046 && !(delta & 1));
+    return (uint16_t)(0xe000 | ((delta >> 1) & 0x7ff));
 }
 
 static void insert_branch_and_link(uint16_t *from, void *to) {


bruelltuete:
I'm still hitting this assert (after applying this patch)...

ui_from = 0x20000324
ui_to = 0x20000314
delta = 0x7ffffff6

...while attempting to add a 2nd shared handler to the same irq.

bruelltuete:
> I'm still hitting this assert (after applying this patch)...
> 
> ui_from = 0x20000324 ui_to = 0x20000314 delta = 0x7ffffff6
> 
> ...while attempting to add a 2nd shared handler to the same irq.

No, a case of stale code somewhere along cmake, picoprobe and the target.
Works now.

src/rp2_common/hardware_irq/irq.c
```diff
@@ -135,9 +135,9 @@ irq_handler_t irq_get_exclusive_handler(uint num) {
 static uint16_t make_branch(uint16_t *from, void *to) {
     uint32_t ui_from = (uint32_t)from;
```
bruelltuete, line 136: or cast straight to signed int here?
does the calculation of delta change if ui_from and ui_to are signed? eyeballing i would say no.

Wren6991 (approved):

Issue #71, opened 2021-02-01
Callback use of user_data is inconsistent or unavailable across SDK


ndinsmore:
The use of user_data in callbacks is of importance because it can allow callback to a class method when using C++/OOP via a non capturing Lambda function, which is great.

There is an inconsistency with how the user_data is implemented or if it is implemented in some cases.  
The two different schemes are the alarm scheme where user_data is passed as a second argument to the callback:
```c
typedef int64_t (*alarm_callback_t)(alarm_id_t id, void *user_data);
```

And the repeating timer scheme where the user_data is passed to the callback inside the repeating_timer_t:
```c
typedef bool (*repeating_timer_callback_t)(repeating_timer_t *rt);
struct repeating_timer {
    int64_t delay_us;
    alarm_pool_t *pool;
    alarm_id_t alarm_id;
    repeating_timer_callback_t callback;
    void *user_data;
};
```
Having a consistent scheme between the two would be great, with the alarm scheme being more idiomatic from what I can tell.

Also of concern is the lack of user_data on the following callbacks and types.
```c
typedef void (*gpio_irq_callback_t)(uint gpio, uint32_t events); 

typedef void(* resus_callback_t )(void);

typedef void (*rtc_callback_t)(void);

typedef void(* hardware_alarm_callback_t )(uint alarm_num)

```
With the lack on IRQs my primary concern as it makes for example a Sensor class for a hall effect sensor difficult.

kilograham:
> The use of user_data in callbacks is of importance because it can allow callback to a class method when using C++/OOP via a non capturing Lambda function, which is great.
> 
> There is an inconsistency with how the user_data is implemented or if it is implemented in some cases.
> The two different schemes are the alarm scheme where user_data is passed as a second argument to the callback:
> 
> ```c
> typedef int64_t (*alarm_callback_t)(alarm_id_t id, void *user_data);
> ```
> 
> And the repeating timer scheme where the user_data is passed to the callback inside the repeating_timer_t:
> 
> ```c
> typedef bool (*repeating_timer_callback_t)(repeating_timer_t *rt);
> struct repeating_timer {
>     int64_t delay_us;
>     alarm_pool_t *pool;
>     alarm_id_t alarm_id;
>     repeating_timer_callback_t callback;
>     void *user_data;
> };
> ```
> 
> Having a consistent scheme between the two would be great, with the alarm scheme being more idiomatic from what I can tell.
> 
[...]

PR #754, opened 2022-03-10
Partial fix for #71: Add user_data to gpio IRQ.

ShellAddicted:
Partial fix for #71 
Add user_data to `gpio_set_irq_enabled_with_callback()` making it much easier to use in an OOP context.

### DIFF
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -14,7 +14,7 @@
 #include "pico/binary_info.h"
 #endif
 
-static gpio_irq_callback_t _callbacks[NUM_CORES];
+static gpio_irq_callback_entry_t _callbacks[NUM_CORES];
 
 // Get the raw value from the pin, bypassing any muxing or overrides.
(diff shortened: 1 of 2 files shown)


lurch:
Looks like this "fix" would actually break any code using the existing functions? IIRC @kilograham puts great emphasis on maintaining backwards-compatibility.

ShellAddicted:
@lurch 
Yes, this is a breaking change.
Would you prefer a new `gpio_set_irq_enabled_with_callback_wudata()` instead?

The point of this PR is to make the SDK (a bit) more consistent, creating a new method is the opposite of that (in my opinion)
Would this be the only breaking change for the next release?

I can move the changes to the new method if taking the breaking change is not an option.

JamesH65:
My preference would be a new function, I cannot see us deliberately breaking an older function with the consequent pain for users and documentation, but that is down to @kilograham.

kilograham:
We will be improving the GPIO IRQ handling API in the near future, so won't be making additional separate changes before then.

I have added #756 to track.

PR #763, opened 2022-03-22
Align uf2 blocks to flash erase pages

### LINKED ISSUE #762: UF2 files with address gaps are not flashed correctly
The rp2040 bootrom uses the uf2 block number to determine when to erase a flash page:
https://github.com/raspberrypi/pico-bootrom/blob/c09c7f08550e8a36fc38dc74f8873b9576de99eb/bootrom/virtual_disk.c#L205

When there is a gap in target addresses between two uf2 blocks this logic does not work out.
As a workaround align all uf2 blocks to page erase boundary, so that the bootrom always removes the correct flash page.

timokroeger:
Fixes #762

The rp2040 bootrom uses the uf2 block number to determine when to erase a flash sector:
https://github.com/raspberrypi/pico-bootrom/blob/c09c7f08550e8a36fc38dc74f8873b9576de99eb/bootrom/virtual_disk.c#L205

When there is a gap in target addresses between two uf2 blocks this logic does not work out.
As a workaround align all uf2 blocks to page erase boundary, so that the bootrom always removes the correct sector.

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -6,6 +6,7 @@
 
 #include <cstdio>
 #include <map>
+#include <set>
 #include <vector>
 #include <cstring>
 #include <cstdarg>
@@ -20,6 +21,8 @@ typedef unsigned int uint;
 #define ERROR_INCOMPATIBLE -3
 #define ERROR_READ_FAILED -4
 #define ERROR_WRITE_FAILED -5
+
+#define FLASH_SECTOR_ERASE_SIZE 4096u
 
 static char error_msg[512];
 static bool verbose;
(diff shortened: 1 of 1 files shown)


kilograham:
Argh; good catch on the bootrom bug.

Two things:

1) The changes should be limited to flash UF2 generation
2) I'm not sure the fix is quite right, because the block_num needs to change whenever the flash sector number does, and this is relative to the start of flash not the previous UF2 block address. I think a simpler, and the easier one conceptually, would be just to include zeroed data for any unused pages in a sector (they are going to be erased anyway).

timokroeger:
Implemented your feedback, the code is much more readable now.

kilograham:
Actually see #800

PR #766, opened 2022-03-23
fix __ctzdi2 for values with non-zero low 32 bits (fixes #765)

### LINKED ISSUE #765: pico_bit_ops implementation of __ctzdi2 is broken
While trying to use std::countr_zero with uint64_t values, I noticed that the result is wrong for any value with any of the low 32 bits set. For instance, std::countr_zero(~0ull) should return zero but the pico implementation returns 32.

Reading the implementation, it seems likely that the conditional branch in __dtzdi2 should be beq, not bne.
```
// test.cpp:
// Build with gcc -std=c++20 test.cpp
#include <bit>
#include <cassert>
int main() {
    // With pico-sdk, the assert fails - the value returned is 32
    assert(std::countr_zero(~0ull) == 0);
}
```

dfr:


dfr:
I rebased against develop - hope that helps

kilograham (approved):

Issue #734, opened 2022-02-21
Missing 'pwm_config_set_clkdiv_int_fract()' function?


ZodiusInfuser:
I've been dealing with the PWM hardware a lot in recent weeks for an upcoming Pimoroni product, and noticed that there seems to be no way set a clock divisor with a fractional integer argument during configuration.

There's a `pwm_config_set_clkdiv_int(pwm_config * c, uint div)` but no `int_fract` equivalent. Instead I am having to either cast my ints to floats to use `pwm_config_set_clkdiv(pwm_config * c, float div)`, or activate the PWM and change the frequency using `pwm_set_clkdiv_int_frac(uint slice_num, uint8_t integer, uint8_t fract)`.

Best as I can tell, there is no hardware limitation for this so it may just be an oversight. If this is the case would you like me to raise a PR to add it? I would leave the existing function alone for compatibility of course.

lurch:
In a similar vein, there's no `pwm_set_clkdiv_int` either :smiling_imp: 
So yeah, looks like a small oversight that the `pwm_config_set_` functions have `clkdiv` and `clkdiv_int`; but the `pwm_set_` functions have `clkdiv` and `clkdiv_int_frac` ? :shrug:

ZodiusInfuser:
You make a valid point re: missing `pwm_set_clkdiv_int()` 😆 .

I mention the lack of the `fract` version more so because hardware features like PIO have both a `sm_config_set_clkdiv_int_frac()` and `pio_sm_set_clkdiv_int_frac()`, leading me to believe that's the intended convention.

I don't think there's an `int` only version anywhere else in the API, but I'd be happy to be proven wrong 🙂

lurch:
Yup, looks like you're right:
```bash
$ git grep "_int(" | grep -v resolvable_int
src/common/pico_binary_info/include/pico/binary_info/code.h:#define bi_int(_tag, _id, _value) \
src/common/pico_binary_info/include/pico/binary_info/code.h:#define bi_binary_end(end) bi_int(BINARY_INFO_TAG_RASPBERRY_PI, BINARY_INFO_ID_RP_BINARY_END, end)
src/rp2_common/boot_stage2/pad_checksum:def any_int(x):
src/rp2_common/hardware_pwm/include/hardware/pwm.h:static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
[...]

PR #768, opened 2022-03-24
Added int_frac function for configuring PWM

ZodiusInfuser:
It's a bit late, but finally got around to raising a PR for issue https://github.com/raspberrypi/pico-sdk/issues/734. 

I've done some testing, but perhaps worth your team testing too before merging.

Also, I don't know whether its worth mirroring the `pwm_config_set_clkdiv_int` with a `pwm_set_clkdiv_int` too? As mentioned in the issue, it seems somewhat redundant, since PIO only has the `int_frac` version.

### DIFF
--- a/src/rp2_common/hardware_pwm/include/hardware/pwm.h
+++ b/src/rp2_common/hardware_pwm/include/hardware/pwm.h
@@ -146,6 +146,7 @@ static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
  * before passing them on to the PWM counter.
  */
 static inline void pwm_config_set_clkdiv_int_frac(pwm_config *c, uint8_t integer, uint8_t fract) {
+    valid_params_if(PWM, fract < 16);
     c->div = (((uint)integer) << PWM_CH0_DIV_INT_LSB) | (((uint)fract) << PWM_CH0_DIV_FRAC_LSB);
 }
 
--- a/src/rp2_common/hardware_pwm/include/hardware/pwm.h
+++ b/src/rp2_common/hardware_pwm/include/hardware/pwm.h
@@ -134,6 +134,21 @@ static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
     c->div = div << PWM_CH0_DIV_INT_LSB;
 }
 
+/** \brief Set PWM clock divider in a PWM configuration using an 8:4 fractional value
+ *  \ingroup hardware_pwm
+ *
+ * \param c PWM configuration struct to modify
+ * \param integer 8 bit integer part of the clock divider. Must be greater than or equal to 1.
+ * \param fract 4 bit the fractional part of the clock divider
+ *
+ * If the divide mode is free-running, the PWM counter runs at clk_sys / div.
+ * Otherwise, the divider reduces the rate of events seen on the B pin input (level or edge)
+ * before passing them on to the PWM counter.
+ */
+static inline void pwm_config_set_clkdiv_int_frac(pwm_config *c, uint8_t integer, uint8_t fract) {
+    c->div = (((uint)integer) << PWM_CH0_DIV_INT_LSB) | (((uint)fract) << PWM_CH0_DIV_FRAC_LSB);
+}
+
 /** \brief Set PWM counting mode in a PWM configuration
  *  \ingroup hardware_pwm
  *
--- a/src/rp2_common/hardware_pwm/include/hardware/pwm.h
+++ b/src/rp2_common/hardware_pwm/include/hardware/pwm.h
@@ -139,7 +139,7 @@ static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
  *
  * \param c PWM configuration struct to modify
  * \param integer 8 bit integer part of the clock divider. Must be greater than or equal to 1.
- * \param fract 4 bit the fractional part of the clock divider
+ * \param fract 4 bit fractional part of the clock divider
  *
  * If the divide mode is free-running, the PWM counter runs at clk_sys / div.
  * Otherwise, the divider reduces the rate of events seen on the B pin input (level or edge)


src/rp2_common/hardware_pwm/include/hardware/pwm.h
```diff
@@ -134,6 +134,21 @@ static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
     c->div = div << PWM_CH0_DIV_INT_LSB;
 }
 
+/** \brief Set PWM clock divider in a PWM configuration using an 8:4 fractional value
+ *  \ingroup hardware_pwm
+ *
+ * \param c PWM configuration struct to modify
+ * \param integer 8 bit integer part of the clock divider. Must be greater than or equal to 1.
+ * \param fract 4 bit the fractional part of the clock divider
```
lurch, line 142: Might be worth chucking in a `valid_params_if` to check that `fract` isn't too big?

src/rp2_common/hardware_pwm/include/hardware/pwm.h
```diff
@@ -134,6 +134,21 @@ static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
     c->div = div << PWM_CH0_DIV_INT_LSB;
 }
 
+/** \brief Set PWM clock divider in a PWM configuration using an 8:4 fractional value
+ *  \ingroup hardware_pwm
+ *
+ * \param c PWM configuration struct to modify
+ * \param integer 8 bit integer part of the clock divider. Must be greater than or equal to 1.
```
lurch, line 141: It's a `uint8_t`, so the `8 bit` at the front might be superfluous?

src/rp2_common/hardware_pwm/include/hardware/pwm.h
lurch: And I suspect "4 bit the fractional part..." ought to be "The 4-bit fractional part..."

lurch:
I agree that `pwm_set_clkdiv_int` would be redundant - if the user wants that they can just pass in a `0` for the fractional part :slightly_smiling_face:

src/rp2_common/hardware_pwm/include/hardware/pwm.h
ZodiusInfuser: That text was copied mostly verbatim from the existing `int_frac` function. https://github.com/raspberrypi/pico-sdk/blob/d54104a1e5cff04979fe442cadd3018ac583d5d2/src/rp2_common/hardware_pwm/include/hardware/pwm.h#L368

Happy to change it if you'd prefer.

src/rp2_common/hardware_pwm/include/hardware/pwm.h
ZodiusInfuser: Ah good idea. I see this is included in the existing function, but I didn't think to copy it over
https://github.com/raspberrypi/pico-sdk/blob/d54104a1e5cff04979fe442cadd3018ac583d5d2/src/rp2_common/hardware_pwm/include/hardware/pwm.h#L379

src/rp2_common/hardware_pwm/include/hardware/pwm.h
lurch: Ahh, I didn't realise that sorry. Yeah, it probably makes sense to stick to what the existing doxygen comments elsewhere say, so you just need to fix "4 bit the fractional part" to say "4 bit fractional part"

lurch:
...although whilst we're here, I wonder if it might make sense to change the implementation of `pwm_config_set_clkdiv` to more closely look like https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_pwm/include/hardware/pwm.h#L391 ? :shrug: 
(and perhaps it makes sense to change `pwm_config_set_clkdiv_int` to just call `pwm_config_set_clkdiv_int_frac(c, div, 0);` ? )

ZodiusInfuser:
@lurch so change them to be like?
```
static inline void pwm_config_set_clkdiv(pwm_config *c, float div) {
    valid_params_if(PWM, div >= 1.f && div < 256.f);
    uint8_t i = (uint8_t)div;
    uint8_t f = (uint8_t)((div - i) * (0x01 << 4));
    pwm_config_set_clkdiv_int_frac(c, i, f);
}

static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
     pwm_config_set_clkdiv_int_frac(c, (uint8_t)div, 0);
}
```

src/rp2_common/hardware_pwm/include/hardware/pwm.h
ZodiusInfuser: Done

src/rp2_common/hardware_pwm/include/hardware/pwm.h
ZodiusInfuser: Ah, good spot.

lurch:
> so change them to be like?

Yup, as long as @kilograham is okay with that suggestion? :slightly_smiling_face:

src/rp2_common/hardware_pwm/include/hardware/pwm.h
```diff
@@ -134,6 +134,22 @@ static inline void pwm_config_set_clkdiv_int(pwm_config *c, uint div) {
...
+ * \param fract 4 bit fractional part of the clock divider
+ *
+ * If the divide mode is free-running, the PWM counter runs at clk_sys / div.
+ * Otherwise, the divider reduces the rate of events seen on the B pin input (level or edge)
+ * before passing them on to the PWM counter.
+ */
+static inline void pwm_config_set_clkdiv_int_frac(pwm_config *c, uint8_t integer, uint8_t fract) {
+    valid_params_if(PWM, fract < 16);
```
lurch, line 149: I wonder if it's also worth adding `valid_params_if(PWM, integer >= 1);`  ? @kilograham

kilograham: by all means add a separate issue for that... i'm gonna merge this one

kilograham (approved):

PR #769, opened 2022-03-24
Added board definition for upcoming Pimoroni servo driver

ZodiusInfuser:
A teaser can be seen here 🙂
https://twitter.com/i/status/1489298380270575616

With any luck we'll be launching next week


lurch:
Looks like you need to rebase something, as this PR shouldn't be changing `pwm.h` ?

src/boards/include/boards/pimoroni_servo2040.h
```diff
@@ -0,0 +1,236 @@
...
+#define SERVO2040_ADC_ADDR_0_PIN 22
+#endif
+
+#ifndef SERVO2040_ADC_ADDR_1_PIN
+#define SERVO2040_ADC_ADDR_1_PIN 24
+#endif
+
+#ifndef SERVO2040_ADC_ADDR_2_PIN
+#define SERVO2040_ADC_ADDR_2_PIN 25
+#endif
```
lurch, line 133: Just out of curiosity, what does the `ADC` in these pin names signify?

src/boards/include/boards/pimoroni_servo2040.h
```diff
@@ -0,0 +1,236 @@
...
+#ifndef SERVO2040_NUM_ADC_PINS
+#define SERVO2040_NUM_ADC_PINS 3
+#endif
+
+#ifndef SERVO2040_SHARED_ADC_PIN
+#define SERVO2040_SHARED_ADC_PIN 29
+#endif
+
+#ifndef SERVO2040_SENSOR_1_ADDR
+#define SERVO2040_SENSOR_1_ADDR 0b000
```
lurch, line 156: Not that it matters, but any particular reason why these are specified in binary?

src/boards/include/boards/pimoroni_servo2040.h
```diff
@@ -0,0 +1,236 @@
...
+#ifndef SERVO2040_A2_PIN
+#define SERVO2040_A2_PIN 28
+#endif
+
+#ifndef SERVO2040_NUM_ADC_PINS
+#define SERVO2040_NUM_ADC_PINS 3
+#endif
+
+#ifndef SERVO2040_SHARED_ADC_PIN
+#define SERVO2040_SHARED_ADC_PIN 29
```
lurch, line 152: "SHARED_ADC"? :confused:

src/boards/include/boards/pimoroni_servo2040.h
ZodiusInfuser: The board has an ADC mux that connects to the SHARED_ADC pin, so these are the three digital address pins.

src/boards/include/boards/pimoroni_servo2040.h
ZodiusInfuser: It was just convenience for my mental processing of the 3 bits (and thus pins) that are being set.

src/boards/include/boards/pimoroni_servo2040.h
ZodiusInfuser: We needed more pins than the RP2040 has 😉

ZodiusInfuser:
> Looks like you need to rebase something, as this PR shouldn't be changing `pwm.h` ?

Oh, my bad. Guess I branched from the PWM stuff I was doing earlier rather than develop. Fixed it now.

src/boards/include/boards/pimoroni_servo2040.h
ZodiusInfuser: Don't be surprised if some future boards also use this technique too

src/boards/include/boards/pimoroni_servo2040.h
lurch: As I said, it doesn't matter, so feel free to leave them as-is.

src/boards/include/boards/pimoroni_servo2040.h
lurch: Perhaps it'd be less confusing if these were e.g. `SERVO2040_SHARED_ADC_ADDR_0_PIN` ?

Also, is it likely that these would _ever_ be set to anything different, i.e. perhaps the `#ifndef` wrappers are unnecessary? :shrug:  (but again, this doesn't really matter so feel free to leave as-is)

ZodiusInfuser: It would be marginally clearer, but the current name matches how the pin is labelled in the schematic. And I will confess, I've already had a PR approved for CircuitPython with the old name, so would be nice to keep them consistent.

ZodiusInfuser: Though on that point. `LED_DAT` should be `LED_DATA`, so let me fix that

src/boards/include/boards/pimoroni_servo2040.h
```diff
@@ -0,0 +1,236 @@
...
+
+// --- UART ---
+// no PICO_DEFAULT_UART
+// no PICO_DEFAULT_UART_TX_PIN
+// no PICO_DEFAULT_UART_RX_PIN
+
+// --- LED ---
+// no PICO_DEFAULT_LED_PIN
+#ifndef PICO_DEFAULT_WS2812_PIN
+#define PICO_DEFAULT_WS2812_PIN SERVO2040_LED_DAT_PIN
```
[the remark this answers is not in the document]
ZodiusInfuser: 🤦 Thanks for spotting!

lurch (approved):

kilograham (approved):

PR #772, opened 2022-03-30
pico_sdk_import: don't recurse git submodules

### LINKED ISSUE #771: pico_sdk_import: don't recursively update submodules
When cloning from git, CMake will recursively update submodules by default. Updating all of `tiny-usb`'s submodules takes ages, and isn't required for pico-sdk. The official instructions for manual SDK installation don't do this, in fact. CMake has an option to fix this as of 3.17, and using it should prove harmless for those on older versions (the project officially only requires 3.13, but it will silently ignore this unknown option).

willeccles:
When cloning the pico-sdk repo manually, one normally would do
`git submodule update --init`, which is non-recursive. However, when
cloning automatically, CMake will recursively update submodules by
default. Updating all of tiny-usb's submodules takes an extremely long
time. Luckily, CMake 3.17 added an option we can specify for
FetchContent to tell it not to recursively update submodules. Older
CMake versions will not use this flag.

Fixes #771.

### DIFF
--- a/external/pico_sdk_import.cmake
+++ b/external/pico_sdk_import.cmake
@@ -29,11 +29,22 @@ if (NOT PICO_SDK_PATH)
         if (PICO_SDK_FETCH_FROM_GIT_PATH)
             get_filename_component(FETCHCONTENT_BASE_DIR "${PICO_SDK_FETCH_FROM_GIT_PATH}" REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
         endif ()
-        FetchContent_Declare(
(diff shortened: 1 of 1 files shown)


willeccles:
While I am quite certain this shouldn't be an issue for those using an older version of CMake, it ought to be tested by someone using an older version than 3.17 (preferably 3.13).

willeccles:
Okay, I finally got around to testing this on an older version of CMake (3.16). It fails to work on 3.16, so I will add a condition for this.

kilograham (approved):

PR #793, opened 2022-04-24
Add gpio_deinit

### LINKED ISSUE #792: Should there be gpio_deinit()?
Should there be `gpio_deinit(pin)`, for symmetry with `gpio_init(pin)` and similarity to `i2c_deinit()`?

I think this would just be a call to `gpio_set_function(pin, GPIO_FUNC_NULL)`.

also see #415

bruelltuete:
Add a `gpio_deinit(pin)`, for symmetry with `gpio_init(pin)` and similarity to `i2c_deinit()`.

Fixes #792.


kilograham (approved):
Thanks... i don't know how many people are screaming fro this, but it can't hurt.

PR #798, opened 2022-05-01
Adapt assembler syntax to eliminate Clang errors. (LLVM 14.0.0)

### LINKED ISSUE #797: Assembler syntax rejected by Clang
Clang (LLVM 14.0.0) barfs on some of the assembler syntax in SDK files.

- 'rsbs' instruction not accepted, equivalent 'negs' is.
- Some argument syntax isn't accepted (e.g '=(SIO_BASE)' isn't, '=SIO_BASE' is; '=#blah' isn't, '=blah' is.
- 'adr' argument can't be an expression, but can be a .equ that uses an expression.
- Declaration of 'd2fix_a' shouldn't use 'regular_func' macro, because 'd2fix_a' should be a weak symbol.

eskimo-software:
Fixes #797

Clang (LLVM 14.0.0) assembler doesn't like some syntax that GCC assembler accepts:
- 'rsbs' instruction not accepted, equivalent 'negs' is.
- Some argument syntax isn't accepted (e.g '=(SIO_BASE)' isn't, '=SIO_BASE' is; '=#blah' isn't, '=blah' is.
- 'adr' argument can't be an expression, but can be a .equ that uses an expression.
- Declaration of 'd2fix_a' shouldn't use 'regular_func' macro, because 'd2fix_a' should be a weak symbol.

### DIFF
--- a/src/rp2_common/pico_divider/divider.S
+++ b/src/rp2_common/pico_divider/divider.S
@@ -103,7 +103,7 @@ regular_func div_s32s32
 regular_func divmod_s32s32
 #if !PICO_DIVIDER_DISABLE_INTERRUPTS
     // to support IRQ usage (or context switch) we must save/restore divider state around call if state is dirty
-    ldr r2, =(SIO_BASE)
+    ldr r2, =SIO_BASE
     ldr r3, [r2, #SIO_DIV_CSR_OFFSET]
     lsrs r3, #SIO_DIV_CSR_DIRTY_SHIFT_FOR_CARRY
     bcs divmod_s32s32_savestate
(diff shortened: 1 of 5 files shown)


kilograham:
Awesome, thank you, will take a look in more detail.

What is the problem with `regular_func` on weak (me having not looked closely)

eskimo-software:
From memory, Clang considers the .global declaration in the regular_func macro expansion to conflict with the .weak declaration preceding it.

eskimo-software:
Regarding 'rsbs' - it turns out that while the syntax 'rsbs Rn, #0' is not accepted, 'rsbs Rn, Rn, #0' is.

See https://github.com/llvm/llvm-project/issues/55215

eskimo-software:
Also see https://github.com/llvm/llvm-project/issues/55218, which I raised regarding the 'adr' instruction.

kilograham (approved):

kilograham:
All looks good to me, so I am including for 1.3.1; the `RSBS` thing is a Clang bug as you note, however I'm fine with using `NEGS`.

PR #800, opened 2022-05-03
Pad all but last sector in flash binary to write whole of sector

kilograham:
Additional changes on top of #763

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -6,6 +6,7 @@
 
 #include <cstdio>
 #include <map>
+#include <set>
 #include <vector>
 #include <cstring>
 #include <cstdarg>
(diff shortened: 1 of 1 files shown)


tools/elf2uf2/main.cpp
```diff
@@ -293,6 +296,28 @@ int elf2uf2(FILE *in, FILE *out) {
                         MAIN_RAM_START, sp);
         }
 #endif
+    } else {
+        // Fill in empty dummy uf2 pages to align the binary to flash sectors (except for the last sector which we don't
+        // need to pad, and choose not to to avoid making all SDK UF2s bigger)
```
lurch, line 301: For this last sector, it makes sense that we don't want/need to add any zero-padding _after_ the data in that sector; but do any holes _before_ the data in the last sector still need to be filled with zeroes?

kilograham: good point :-)

kilograham: actually it doesn't matter from a correctness point of view (the actual data of the last sector will be written correctly), however as it stands a gap in the last sector will not be filled with zeroes.

kilograham:
fixes #762

Wren6991 (approved):

PR #802, opened 2022-05-04
Removed unnecessary ifdefs from Pimoroni boards

ZodiusInfuser:
Following some discussion with @lurch and @kilograham in #795, I have gone back through our board definitions and removed any `#ifdef`'s for pins and constants that are specific to those boards, and thus would not be user changeable unlike the `PICO` defines.


lurch (approved):
Nice :+1:

kilograham (approved):

PR #809, opened 2022-05-04
Fix bug in stdio_set_driver_enabled

### LINKED ISSUE #808: stdio_set_driver_enabled(.., false) misbehaves if there is only 1 driver
When there's one stdio driver registered (e.g. stdio_uart), calling `stdio_set_driver_enabled(..., false)` to disable it breaks the linked list represented by https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_stdio/stdio.c#L32: the head of the list (`drivers`) will continue pointing to the just disabled driver.
see https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_stdio/stdio.c#L182

bruelltuete:
Fixes #808: the head of the linked list would be left dangling if there was only 1 driver registered at the time of calling stdio_set_driver_enabled(..., false)

### DIFF
--- a/src/rp2_common/pico_stdio/stdio.c
+++ b/src/rp2_common/pico_stdio/stdio.c
@@ -180,21 +180,30 @@ int _write(int handle, char *buffer, int length) {
 }
 
 void stdio_set_driver_enabled(stdio_driver_t *driver, bool enable) {
-    stdio_driver_t *prev = drivers;
-    for (stdio_driver_t *d = drivers; d; d = d->next) {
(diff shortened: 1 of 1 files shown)


kilograham (commented):
Thanks; i think i'd prefer not to change the order things are added in. (i.e. it used to be appended)

maybe rework using `stdio_driver_t **prev = &drivers` instead to avoid the special cases for the head

kilograham:
closed in favor of #822

PR #811, opened 2022-05-05
Add stdio_uart_deinit()

### LINKED ISSUE #810: Should there be a stdio_uart_deinit()?
I would like to disable UART console output after running for a while. Idea is to dump boot diagnostics to console but once everything is up and running (say after 1 minute or so) I would like to disable UART output to save power running off batteries.
I'd call `stdio_uart_deinit` and then clock-off UART. My crusty amp meter says clock-off will save about 0.5mA.

Stumbled over #808 while trying this.

I got a related question: what is supposed to happen (wrt to printf) if I disable `CLOCKS_WAKE_EN1_CLK_SYS_UART0_BITS` and `CLOCKS_WAKE_EN1_CLK_PERI_UART0_BITS`? Somewhere down the call stack is a call to `uart_write_blocking()`. If the UART controller is not clocked, is there always space in the tx fifo?

bruelltuete:
Fixes #810 by adding stdio_uart_deinit.

I would have liked to undo the pin init from `stdio_uart_init_full()` as well but we dont keep track of the pins. Could probably go through all pins, check which ones have uart as function and deinit those. But is that something we want?

### DIFF
--- a/src/rp2_common/pico_stdio_uart/stdio_uart.c
+++ b/src/rp2_common/pico_stdio_uart/stdio_uart.c
@@ -80,6 +80,7 @@ void stdio_uart_deinit() {
     if (uart_instance != NULL)
     {
         stdio_set_driver_enabled(&stdio_uart, false);
+        uart_tx_wait_blocking(uart_instance);
         uart_deinit(uart_instance);
         uart_instance = NULL;
     }
--- a/src/rp2_common/pico_stdio_uart/stdio_uart.c
+++ b/src/rp2_common/pico_stdio_uart/stdio_uart.c
@@ -9,7 +9,7 @@
 #include "pico/binary_info.h"
 #include "hardware/gpio.h"
 
-static uart_inst_t *uart_instance;
+static uart_inst_t *uart_instance = NULL;
 
 #if PICO_NO_BI_STDIO_UART
 #define stdio_bi_decl_if_func_used(x)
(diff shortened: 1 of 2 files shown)


a-gavin:
Changing [this prototype](https://github.com/raspberrypi/pico-sdk/blob/93472b84c6fb08c184af984f948b045e4dba9edc/src/rp2_common/pico_stdio_uart/include/pico/stdio_uart.h#L71) to `void stdio_uart_deinit(void);` should fix the build error

lurch:
> I would have liked to undo the pin init from `stdio_uart_init_full()` as well but we dont keep track of the pins.

I guess that would apply if the function you were adding was named `stdio_uart_deinit_full()`; but given that you're adding `stdio_uart_deinit()`, perhaps it's safe to assume that it's the inverse of `stdio_uart_init()`, and therefore you could use the same [logic](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_stdio_uart/stdio_uart.c#L20) to determine which pins are set up as TX and RX, and "deinit" those? :shrug: 

(and I guess if you wanted to be completionist, you could also add `stdout_uart_deinit()`, `stdin_uart_deinit()` and `stdout_uart_deinit_full(int tx_pin, int rx_pin)` ?)

src/rp2_common/pico_stdio_uart/stdio_uart.c
```diff
@@ -76,6 +76,15 @@ void stdio_uart_init_full(struct uart_inst *uart, uint baud_rate, int tx_pin, in
     stdio_set_driver_enabled(&stdio_uart, true);
 }
 
+void stdio_uart_deinit() {
+    if (uart_instance != NULL)
+    {
+        stdio_set_driver_enabled(&stdio_uart, false);
+        uart_deinit(uart_instance);
```
bruelltuete, line 83: this needs a `uart_tx_wait_blocking(uart_instance);` here, otherwise we'll start loosing output (and wsl's minicom goes all wonky)

bruelltuete:
> > I would have liked to undo the pin init from `stdio_uart_init_full()` as well but we dont keep track of the pins.
> 
> I guess that would apply if the function you were adding was named `stdio_uart_deinit_full()`; but given that you're adding `stdio_uart_deinit()`, perhaps it's safe to assume that it's the inverse of `stdio_uart_init()`, and therefore you could use the same [logic](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_stdio_uart/stdio_uart.c#L20) to determine which pins are set up as TX and RX, and "deinit" those? 🤷

Indeed!

kilograham:
there is a `stdio_uart_deinit` in SDK 2.0.0

PR #814, opened 2022-05-06
Documentation: Clarify GPIO interrupt behaviour

Malvineous:
Just some small documentation updates to clarify a few things I had to investigate to find answers to, related to how the GPIO interrupts work.


src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -401,7 +401,8 @@ void gpio_set_dormant_irq_enabled(uint gpio, uint32_t events, bool enabled);
  *
  * \param gpio GPIO number
  * \param events Bitmask of events to clear. See \ref gpio_set_irq_enabled for details.
-  *
+ *
+ * \note \ref gpio_set_irq_enabled_with_callback calls this function before the callback.
```
lurch, line 405: `gpio_set_irq_enabled_with_callback` doesn't directly call `gpio_acknowledge_irq`. Did you mean something like "The default GPIO IRQ-handler automatically calls this function before calling any callback-function" ?

Malvineous: Sorry yes, I meant to imply that callbacks set with `gpio_set_irq_enabled_with_callback` don't need to acknowledge the IRQ as it's done before the user callback runs.  I've reworded it, hopefully it reads better now.

kilograham:
Not picking this up yet, as this API may get reworked shortly.

Malvineous:
@kilograham You could merge it now and then just let it get replaced if/when the API gets redeveloped 😀

PR #816, opened 2022-05-09
Add documentation to dma_channel_abort regarding errata RP2040-E13

kilograham:
See https://github.com/raspberrypi/pico-feedback/issues/224

Note this is a documentation only fix, as `dma_channel_abort` does not have enough information to efficiently workaround the issue on the callers behalf.

### DIFF
--- a/src/rp2_common/hardware_dma/include/hardware/dma.h
+++ b/src/rp2_common/hardware_dma/include/hardware/dma.h
@@ -506,7 +506,7 @@ static inline void dma_channel_abort(uint channel) {
     dma_hw->abort = 1u << channel;
     // Bit will go 0 once channel has reached safe state
     // (i.e. any in-flight transfers have retired)
-    while (dma_hw->abort & (1ul << channel)) tight_loop_contents();
+    while (dma_hw->ch[channel].ctrl_trig & DMA_CH0_CTRL_TRIG_BUSY_BITS) tight_loop_contents();
 }
 
 /*! \brief  Enable single DMA channel's interrupt via DMA_IRQ_0
--- a/src/rp2_common/hardware_dma/include/hardware/dma.h
+++ b/src/rp2_common/hardware_dma/include/hardware/dma.h
@@ -473,6 +473,32 @@ static inline void dma_channel_start(uint channel) {
  *
  * Function will only return once the DMA has stopped.
  *
+ * Note that due to to errata RP2040-E13, aborting a channel which has transfers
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -473,6 +473,32 @@ static inline void dma_channel_start(uint channel) {
...
+ * transfers will trigger a completion interrupt once they complete.
+ *
+ * The effect of this is that you \em may see a spurious completion interrupt
+ * on the channel as a result of calling this method.
+ *
+ * The calling code should be sure to ignore a completion IRQ as a result of this method. This may
+ * not require any additional work, as aborting a channel which may be about to complete, when you have a completion
+ * IRQ handler registered, is inherently race-prone, and so code is likely needed to disambiguate the two occurrences.
+ *
+ * If that is not the case, but you do have an channel completion IRQ handler registered, you can simply
```
lurch, line 488: "an channel completion" -> "a channel completion"

Wren6991:
I fixed the typo pointed out by @lurch. I also made another change, which might be more contentious, so feel free to back out: poll BUSY rather than ABORT to check for ABORT completion, to guarantee the channel is returned to a quiescent state before the function exits. ABORT clears prematurely due to the erratum (in fact this *is* the erratum).

The difference in timing is the time taken for the in-flight transfers to drain from the bus FIFOs, which should be a few cycles.

Wren6991 (dismissed):
Approval for the docs updates, @kilograham please merge if you are happy with the ABORT -> BUSY polling change

kilograham:
Does the ABORT->BUSY change have an effect on what happens if the channel is disabled?

Wren6991:
No. BUSY clears once the in-flight and not-yet-issued transfer counts both reach 0. The not-yet-issued count is cleared by the abort, and the EN bit being cleared does not prevent in-flight transfers from completing.

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -473,14 +473,40 @@ static inline void dma_channel_start(uint channel) {
  *
  * Function will only return once the DMA has stopped.
  *
+ * Note that due to to errata RP2040-E13, aborting a channel which has transfers
```
bruelltuete, line 476: Another typo: "to" is twice.

kilograham:
Note I force merged this as the only change after approval was a type in a comment

PR #817, opened 2022-05-10
fix bug in pico_float_test

kilograham:


lurch:
Huh? On lines 392-394 you do:
```c
        float s, c;
        sincosf(x, &s, &c);
        printf("FSINCOS %10.18f %10.18f\n", s, c);
```
and then on lines 404-405 (still at the same scope level) you do:
```c
        sincosf(x, &s, &c);
        printf("SINCOS %10.18f %10.18f\n", s, c);
```
?


And then on lines 440-442 (in a different scope level) you do:
```c
        float s, c;
//        sincosf(x, &s, &c);
        printf("FSINCOS %10.18f %10.18f\n", s, c);
```
:confused:

Issue #780, opened 2022-04-07
pico-sdk and tinyusb fail to build with GCC 11.2


lurch:
I'm not sure exactly where the problem is, so pinging @kilograham @liamfraser and @hathach 

If I build _just_ `pico-sdk` with GCC 11.2 from https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/downloads then it builds fine, but if I also clone the tinyusb sumbodule, then when I try to build it fails with:
```
[ 75%] Building C object test/kitchen_sink/CMakeFiles/kitchen_sink_extra_stdio.dir/__/__/lib/tinyusb/src/portable/raspberrypi/rp2040/rp2040_usb.c.obj
/pico-sdk/lib/tinyusb/src/portable/raspberrypi/rp2040/rp2040_usb.c: In function 'rp2040_usb_init':
/pico-sdk/lib/tinyusb/src/portable/raspberrypi/rp2040/rp2040_usb.c:61:3: error: 'memset' offset [0, 155] is out of the bounds [0, 0] [-Werror=array-bounds]
   61 |   memset(usb_hw, 0, sizeof(*usb_hw));
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/pico-sdk/lib/tinyusb/src/portable/raspberrypi/rp2040/rp2040_usb.c:62:3: error: 'memset' offset [0, 4095] is out of the bounds [0, 0] [-Werror=array-bounds]
   62 |   memset(usb_dpram, 0, sizeof(*usb_dpram));
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [test/kitchen_sink/CMakeFiles/kitchen_sink_extra_stdio.dir/build.make:781: test/kitchen_sink/CMakeFiles/kitchen_sink_extra_stdio.dir/__/__/lib/tinyusb/src/portable/raspberrypi/rp2040/rp2040_usb.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:2069: test/kitchen_sink/CMakeFiles/kitchen_sink_extra_stdio.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```
:cry: 

To reproduce:
(I'm doing everything inside docker just to rule out any other part of the system interfering with things)
```bash
sudo docker run -it ubuntu:20.04
```
and then inside docker:
[...]

PR #819, opened 2022-05-10
Suppress false GCC 11 warning on TinyUSB

kilograham:
Note this should be undone when fixed in upstream TinyUSB


kilograham:
fixes #780

kilograham:
Note this will leave (non-fatal) warnings in pico-examples, as that uses the TinyUSB CMake build directly (TinyUSB will need to be fixed for those).

Users of our `suppress_tinyusb_warnings` CMake function however will build cleanly.

Wren6991 (approved):

PR #821, opened 2022-05-10
cleanup pico_stdlib_test. add actual check for __builtin bitops

kilograham:
Seems like this was a bit of a dumping ground for test code. 

Added an actual test of the __builtin ctz etc, as there was previously a bug there!

### DIFF
--- a/test/pico_stdlib_test/pico_stdlib_test.c
+++ b/test/pico_stdlib_test/pico_stdlib_test.c
@@ -8,11 +8,67 @@
 #include <inttypes.h>
 #include "pico/stdlib.h"
 #include "pico/bit_ops.h"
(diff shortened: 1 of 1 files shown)


test/pico_stdlib_test/pico_stdlib_test.c
```diff
@@ -8,11 +8,67 @@
...
+        printf("FAILED (expected count %d\n", (int) expected);
+        exit(1);
+    }
+}
 
 int main() {
     setup_default_uart();
 
-    puts("Hellox, worlxxcd!");
+    puts("Hellox, world!");
```
lurch, line 71: Hellox?

PR #825, opened 2022-05-12
Fix bug in irq_remove_shared_handler and add test #823

### LINKED ISSUE #823: Bug in irq_remove_handler() and/or irq_add_shared_handler()?
I'm still trying to figure out whats going on but the shared handler code looks really complicated and its hard to tell when something goes wrong.
Here is my symptom: I have a few shared DMA handlers, all on the same DMA IRQ. After a round of add/remove-shared-handler one of those handlers is forever called in an endless loop, without any DMA interrupt pending. Looks like one of the shared handlers is just returning to itself over and over again.
After poking around it bit, I've come to the conclusion that the contents of `irq_handler_chain_slots` look fishy.

The code looks equivalent to:
```
irq_add_shared_handler(DMA_IRQ_0, handler1, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
irq_add_shared_handler(DMA_IRQ_0, handler2, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
// do stuff with dma
irq_remove_handler(DMA_IRQ_0, handler1);
irq_remove_handler(DMA_IRQ_0, handler2);
// sleep or do something unrelated
irq_add_shared_handler(DMA_IRQ_0, handler3 /*!!!*/, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
irq_add_shared_handler(DMA_IRQ_0, handler2, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
// do stuff to trigger dma
```

At this point `handler3` is nowhere to be found in `irq_handler_chain_slots`. And `handler2` is looping on itself.

kilograham:
fixes #823


Wren6991 (approved):
So the bug is that we search for the correct slot pointer to free, but then use an index calculated with the initial value of the pointer? Fix seems fine.

I added a few comments to parts of irq_handler_chain.S that were a bit mysterious to me when I read through the code.

PR #828, opened 2022-05-14
Pass PIOASM_EXTRA_SOURCE_FILES to Pioasm sub-cmake and add OUTPUT_FORMAT option (fixes #827)

### LINKED ISSUE #827: PIOASM_EXTRA_SOURCE_FILES is not passed through to the sub-cmake
The PIOASM_EXTRA_SOURCE_FILES variable is not passed through to the inner CMake call, leading to it being impossible to extend `pioasm` from the outside.

This could be desirable for people who want to extend `pioasm` with additional generators (since those can be added without editing existing source files).

Ferdi265:
This PR changes `FindPioasm.cmake` to pass through the PIOASM_EXTRA_SOURCE_FILES variable, so that it is actually able to be changed by users of the SDK without modifying it.

This PR also adds an OUTPUT_FORMAT option to `pico_generate_pio_header` that sets the output format used so users can also easily use their custom output formats.

Example usage:
```
file(GLOB pioasm-sources pioasm-extensions/*.cpp)
set(PIOASM_EXTRA_SOURCE_FILES ${pioasm-sources})

pico_generate_pio_header(my-target thing.pio OUTPUT_FORMAT sdk-custom)
```

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -4,14 +4,15 @@ function(_pico_init_pioasm)
     find_package(Pioasm REQUIRED)
 endfunction()
 
+# PICO_DEFAULT_PIOASM_OUTPUT_FORMAT, default output format used by pioasm when using pico_generate_pio_header, group=build
 function(pico_generate_pio_header TARGET PIO)
     _pico_init_pioasm()
     cmake_parse_arguments(pico_generate_pio_header "" "OUTPUT_FORMAT;OUTPUT_DIR" "" ${ARGN} )
 
     if (pico_generate_pio_header_OUTPUT_FORMAT)
         set(OUTPUT_FORMAT "${pico_generate_pio_header_OUTPUT_FORMAT}")
-    elseif(DEFINED PIOASM_DEFAULT_OUTPUT_FORMAT)
-        set(OUTPUT_FORMAT "${PIOASM_DEFAULT_OUTPUT_FORMAT}")
+    elseif(DEFINED PICO_DEFAULT_PIOASM_OUTPUT_FORMAT)
+        set(OUTPUT_FORMAT "${PICO_DEFAULT_PIOASM_OUTPUT_FORMAT}")
     else()
         set(OUTPUT_FORMAT "c-sdk")
     endif()
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -6,7 +6,13 @@ endfunction()
 
 function(pico_generate_pio_header TARGET PIO)
     _pico_init_pioasm()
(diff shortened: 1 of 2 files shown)


tools/CMakeLists.txt
```diff
@@ -6,7 +6,13 @@ endfunction()
 
 function(pico_generate_pio_header TARGET PIO)
     _pico_init_pioasm()
-    cmake_parse_arguments(pico_generate_pio_header "" "OUTPUT_DIR" "" ${ARGN} )
+    cmake_parse_arguments(pico_generate_pio_header "" "OUTPUT_FORMAT;OUTPUT_DIR" "" ${ARGN} )
+
+    if (usdk_generate_pio_header_OUTPUT_FORMAT)
```
kilograham, line 11: The variable name should be something like `PIOASM_DEFAULT_OUTPUT_FORMAT`

Ferdi265: I just noticed a typo in there: it says "usdk_generate_pio_header_OUTPUT_FORMAT", which is incorrect, should be "pico_...". That variable is set by the cmake_parse function above and is just a temporary.

IMO it could make sense to set the output format to PIOASM_DEFAULT_OUTPUT_FORMAT if the OUTPUT_FORMAT option isn't set (as opposed to hardcoding c-sdk) as an additional way to customize it.

I'll fixup the patch series to fix the typo, and I can also add the PIOASM_DEFAULT_OUTPUT_FORMAT variable if you want.

Ferdi265: typo fixed in a8ce4cd  
default variable added in 546b89d

kilograham (commented):
Can you please rename `PIOASM_DEFAULT_OUTPUT_FORMAT` to `PICO_DEFAULT_PIOASM_OUTPUT_FORMAT` and also add a line 

```
# PICO_DEFAULT_PIOASM_OUTPUT_FORMAT, default output format used by pioasm when using pico_generate_pio_header, group=build
```

above the pico_generate_pio_header (that way it will be automatically added to tthe docs!)

Ferdi265:
sure!

tools/CMakeLists.txt
```diff
@@ -4,9 +4,18 @@ function(_pico_init_pioasm)
     find_package(Pioasm REQUIRED)
 endfunction()
 
+# PICO_DEFAULT_PIOASM_OUTPUT_FORMAT, default output format used by pioasm when using pico_generate_pio_header, group=build
```
lurch, line 7: Could you please add this to the end of the line too?
```
, default=c-sdk
```

kilograham: i'm going to merge it now; @lurch you can add a PR to update that!

kilograham: oh, found handy new edit feature in github!

kilograham (approved):

PR #830, opened 2022-05-16
Add busy_wait_at_least_cycles method

kilograham:
This method will busy wait for at least the given number of system clock cycles.


Wren6991:
I think this fails when the sleep value is less than 3. The carry won't set, and the register will wrap to 4 billion. Maybe limit the range to 2^31, and check the neg flag?

Wren6991:
This might work as an alternative:

```
    negs r0
1:
    adds r0, #3
    bcc 1b
```

Wren6991 (approved):

PR #832, opened 2022-05-17
Update rtc_set_datetime docs to point out that the written value may not be immediately visible

kilograham:


lurch:
I guess this relates to the "All RTC register reads and writes are done from the processor clock domain clk_sys. All data are synchronised back and forth between the domains. Writing to the RTC will take 2 clk_rtc clock periods to arrive, additional to the clk_sys domain. This should be taken into account especially when the reference is slow (e.g. 1 Hz)." note in section 4.8.4 of the RP2040 datasheet?

lurch (approved):

PR #836, opened 2022-05-18
gate inclusion of pico.h in binary info, so as not to break picotool or other non SDK builds

kilograham:


kilograham:
fixes regression due to #693

PR #837, opened 2022-05-22
Add board definition for the RP2040 Stamp Round Carrier

arturo182:
The file modifies a few pins and then includes the stamp board file.

Details about the board available here: https://www.solder.party/docs/rp2040-stamp/round-
carrier/

### DIFF
--- a/src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
+++ b/src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
@@ -26,6 +26,9 @@
 #endif
 
 // --- UART ---
+#ifndef PICO_DEFAULT_UART
+#define PICO_DEFAULT_UART 0
+#endif
 #ifndef PICO_DEFAULT_UART_TX_PIN
 #define PICO_DEFAULT_UART_TX_PIN 16
 #endif
@@ -39,6 +42,9 @@
 #endif
 
 // --- I2C ---
+#ifndef PICO_DEFAULT_I2C
+#define PICO_DEFAULT_I2C       0
+#endif
 #ifndef PICO_DEFAULT_I2C_SDA_PIN
 #define PICO_DEFAULT_I2C_SDA_PIN   0
 #endif
--- a/src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
+++ b/src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
@@ -36,9 +36,14 @@
 #define PICO_DEFAULT_UART_RX_PIN 17
 #endif
 
-// --- Neopixel ---
+// --- Neopixel Ring ---
 #ifndef PICO_DEFAULT_WS2812_PIN
 #define PICO_DEFAULT_WS2812_PIN 24
+#endif
+
+// --- On-Stamp Neopixel ---
+#ifndef PICO_STAMP_WS2812_PIN
+#define PICO_STAMP_WS2812_PIN 21
 #endif
 
 // --- I2C ---
--- a/src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
+++ b/src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
@@ -42,8 +42,8 @@
 #endif
 
 // --- On-Stamp Neopixel ---
-#ifndef PICO_STAMP_WS2812_PIN
-#define PICO_STAMP_WS2812_PIN 21
+#ifndef SOLDERPARTY_RP2040_STAMP_WS2812_PIN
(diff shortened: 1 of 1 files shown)


src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef PICO_DEFAULT_UART_TX_PIN
+#define PICO_DEFAULT_UART_TX_PIN 16
+#endif
+#ifndef PICO_DEFAULT_UART_RX_PIN
+#define PICO_DEFAULT_UART_RX_PIN 17
+#endif
+
+// --- Neopixel ---
+#ifndef PICO_DEFAULT_WS2812_PIN
+#define PICO_DEFAULT_WS2812_PIN 24
```
lurch, line 38: Looking at the product-page you mentioned, I guess the on-board NeoPixel doesn't stop working when the Stamp is on the Carrier Board, so perhaps the  Carrier Board NeoPixel pin should get its own `#define`, rather than overriding the `PICO_DEFAULT_WS2812_PIN` from `solderparty_rp2040_stamp.h`? :shrug:

src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
```diff
@@ -0,0 +1,68 @@
...
+#define PICO_DEFAULT_LED_PIN   25
+#endif
+
+// --- UART ---
+#ifndef PICO_DEFAULT_UART_TX_PIN
+#define PICO_DEFAULT_UART_TX_PIN 16
+#endif
+#ifndef PICO_DEFAULT_UART_RX_PIN
+#define PICO_DEFAULT_UART_RX_PIN 17
+#endif
```
lurch, line 34: Given that this is redefining all the UART pins, perhaps it's worth specifying the `PICO_DEFAULT_UART` definition, just to be on the safe side?

src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
```diff
@@ -0,0 +1,68 @@
...
+#define PICO_DEFAULT_WS2812_PIN 24
+#endif
+
+// --- I2C ---
+#ifndef PICO_DEFAULT_I2C_SDA_PIN
+#define PICO_DEFAULT_I2C_SDA_PIN   0
+#endif
+#ifndef PICO_DEFAULT_I2C_SCL_PIN
+#define PICO_DEFAULT_I2C_SCL_PIN   1
+#endif
```
lurch, line 47: Given that this is redefining all the I2C pins, perhaps it's worth specifying the `PICO_DEFAULT_I2C` definition, just to be on the safe side?

src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
arturo182: I thought about it, but I think I'd rather have the Carrier NeoPixel ring take priority over the single NeoPixel on the Stamp 😅

src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
arturo182: Sure, will do :)

src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
lurch: Okay, so perhaps you want to add an additional define to the carrier's header-file with the definition of the stamp's neopixel pin, so that code written to target the carrier can write to both (defined) neopixel pins? :slightly_smiling_face:

src/boards/include/boards/solderparty_rp2040_stamp_round_carrier.h
```diff
@@ -36,11 +36,16 @@
...
 
-// --- Neopixel ---
+// --- Neopixel Ring ---
 #ifndef PICO_DEFAULT_WS2812_PIN
 #define PICO_DEFAULT_WS2812_PIN 24
 #endif
 
+// --- On-Stamp Neopixel ---
+#ifndef PICO_STAMP_WS2812_PIN
+#define PICO_STAMP_WS2812_PIN 21
```
lurch, line 46: This should probably be something like `SOLDERPARTY_RP2040_STAMP_...` rather than `PICO_STAMP_...`, as it's a feature specific to this board and not to any other Pico boards.

arturo182: Gotcha 👍

lurch (approved):
LGTM! And hopefully @kilograham agrees with all my suggestions :wink:

kilograham (approved):

PR #848, opened 2022-06-04
Add sem_try_acquire(). Fixes #846

### LINKED ISSUE #846: Add sem_try_acquire()
sem_try_acquire() attempts to acquire a permit without blocking, returns true on success, or false if no permits are available. I have a PR coming up.

slimhazard:
`sem_try_acquire()` attempts to acquire a permit without blocking, returns true on success, or false if no permits are available.

I added a pico_test app. Maybe that's overkill, but I thought the PR should have some kind of verification.


src/common/pico_sync/sem.c
```diff
@@ -52,6 +52,17 @@ bool __time_critical_func(sem_acquire_block_until)(semaphore_t *sem, absolute_ti
     } while (true);
 }
 
+bool __time_critical_func(sem_try_acquire)(semaphore_t *sem) {
+    uint32_t save = spin_lock_blocking(sem->core.spin_lock);
+    if (sem->permits == 0) {
+        spin_unlock(sem->core.spin_lock, save);
+        return false;
+    }
+    sem->permits--;
```
lurch, line 61: Looking at the implementations of `sem_acquire_blocking` and `sem_acquire_block_until`, should this check that `sem->permits > 0` before decrementing it?

src/common/pico_sync/include/pico/sem.h
```diff
@@ -122,6 +122,16 @@ bool sem_acquire_timeout_us(semaphore_t *sem, uint32_t timeout_us);
  */
 bool sem_acquire_block_until(semaphore_t *sem, absolute_time_t until);
 
+/*! \brief Attempt to acquire a permit from a semaphore without blocking
```
lurch, line 125: Apologies if this is a stupid question (it's been many many years since I wrote any semaphore-related code), but is this basically the same as calling `sem_acquire_timeout_us` but with a 0 timeout value?

src/common/pico_sync/sem.c
slimhazard: Good point, I think it would be better to reverse the logic -- check for `permits > 0`, if so then acquire the permit and return true. Otherwise release the lock and return false.

src/common/pico_sync/include/pico/sem.h
slimhazard: If you call `sem_acquire_block_until()` with the timeout set to expire immediately, and there are no permits available, then it will call `lock_internal_spin_unlock_with_best_effort_wait_or_timeout()` with the immediate expiration time and return false. That amounts to nearly the same thing, except that the timeout has to be evaluated. `lock_internal_etc_etc` is a macro that expands to call `spin_unlock()`, as we have here, and then `best_effort_wfe_or_timeout(until)`.

So you're right that it's very similar, I hadn't noticed that. But `sem_try_acquire()` saves a few cycles, since there's no need to set an "immediate timeout" value that is then evaluated as expiring right away.

lurch: > But `sem_try_acquire()` saves a few cycles, since there's no need to set an "immediate timeout" value that is then evaluated as expiring right away.

Indeed, and that's always a good thing :+1:

src/common/pico_sync/sem.c
slimhazard: @lurch or @kilograham if you agree I'll make this change, say the word.

kilograham: sounds good

src/common/pico_sync/sem.c
```diff
@@ -52,6 +52,17 @@ bool __time_critical_func(sem_acquire_block_until)(semaphore_t *sem, absolute_ti
...
 
+bool __time_critical_func(sem_try_acquire)(semaphore_t *sem) {
+    uint32_t save = spin_lock_blocking(sem->core.spin_lock);
+    if (sem->permits > 0) {
+        sem->permits--;
+        lock_internal_spin_unlock_with_notify(&sem->core, save);
+        return true;
+    }
+    spin_unlock(sem->core.spin_lock, save);
+    return false;
```
lurch, line 63: I wonder if it's worth sticking this inside an `else` block? That'd then make this function look _very_ similar to `sem_release()` :slightly_smiling_face: 

(Just out of curiosity @kilograham , why does `sem_release` use a separate `count` variable, rather than just doing
```c
    if (sem->permits < sem->max_permits) {
        sem->permits++;
```
?
Also, I guess it might make sense for `sem_init` to `assert(initial_permits <= max_permits)` ? )

slimhazard: > I wonder if it's worth sticking this inside an `else` block? That'd then make this function look very similar to `sem_release()`

IMO that would be superfluous, since the `if`-block ends with `return`. This way it's more similar to `sem_acquire_blocking()`.

> Also, I guess it might make sense for `sem_init` to `assert(initial_permits <= max_permits)` ?

+1

slimhazard:
I noticed that some minor tweaks were needed -- I had forgotten the doxygen doc for the return value, and the copyright year in the pico_test app was wrong.

The commits for this PR should probably be squashed into just one. @lurch or @kilograham let me know if you agree.

kilograham (approved):

PR #850, opened 2022-06-05
Add new GPIO APIs for adding shared GPIO handlers, and improve docs

kilograham:
Note this PR includes the get_core_num() change as they are dependent

### DIFF
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -182,7 +182,6 @@ void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled) {
 }
 
 void gpio_set_irq_enabled_with_callback(uint gpio, uint32_t events, bool enabled, gpio_irq_callback_t callback) {
-    // this is the original behavior
     gpio_set_irq_enabled(gpio, events, enabled);
     gpio_set_irq_callback(callback);
     if (enabled) irq_set_enabled(IO_IRQ_BANK0, true);
@@ -197,9 +196,7 @@ void gpio_set_irq_callback(gpio_irq_callback_t callback) {
         callbacks[core] = callback;
     } else if (callback) {
         callbacks[core] = callback;
-        // todo - remove comment; note we could have added a #define to resort to the old exclusive handler, but since the
-        //  new handler is likely faster anyway, i don't think there is much point (it's only a few cycles for a shared one)
-        irq_add_shared_handler(IO_IRQ_BANK0, gpio_default_irq_handler, GPIO_CALLBACK_IRQ_ORDER_PRIORITY);
+        irq_add_shared_handler(IO_IRQ_BANK0, gpio_default_irq_handler, GPIO_IRQ_CALLBACK_ORDER_PRIORITY);
     }
 }
 
@@ -209,7 +206,7 @@ void gpio_add_raw_irq_handler_with_order_priority_masked(uint gpio_mask, irq_han
 }
 
 void gpio_add_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler) {
-    gpio_add_raw_irq_handler_with_order_priority_masked(gpio_mask, handler, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
+    gpio_add_raw_irq_handler_with_order_priority_masked(gpio_mask, handler, GPIO_RAW_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
 }
 
 void gpio_remove_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler) {
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -148,7 +148,6 @@ static void gpio_default_irq_handler(void) {
         uint32_t events8 = irq_ctrl_base->ints[gpio >> 3u];
         // note we assume events8 is 0 for non-existent GPIO
         for(uint i=gpio;events8 && i<gpio+8;i++) {
-            events8 >>= 4;
             uint32_t events = events8 & 0xfu;
             if (events && !(raw_irq_mask[core] & (1u << i))) {
                 gpio_acknowledge_irq(i, events);
@@ -156,6 +155,7 @@ static void gpio_default_irq_handler(void) {
                     callback(gpio, events);
                 }
             }
+            events8 >>= 4;
         }
     }
 }
@@ -201,6 +201,7 @@ void gpio_set_irq_callback(gpio_irq_callback_t callback) {
 }
 
 void gpio_add_raw_irq_handler_with_order_priority_masked(uint gpio_mask, irq_handler_t handler, uint8_t order_priority) {
+    hard_assert(!(raw_irq_mask[get_core_num()] & gpio_mask)); // should not add multiple handlers for the same event
     raw_irq_mask[get_core_num()] |= gpio_mask;
     irq_add_shared_handler(IO_IRQ_BANK0, handler, order_priority);
 }
@@ -210,6 +211,7 @@ void gpio_add_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler) {
 }
 
 void gpio_remove_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler) {
+    assert(raw_irq_mask[get_core_num()] & gpio_mask); // should not remove handlers that are not added
     irq_remove_handler(IO_IRQ_BANK0, handler);
     raw_irq_mask[get_core_num()] &= ~gpio_mask;
 }
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -14,11 +14,13 @@
 #include "pico/binary_info.h"
 #endif
 
-static gpio_irq_callback_t _callbacks[NUM_CORES];
+static gpio_irq_callback_t callbacks[NUM_CORES];
+// a 1 bit means the IRQ is handled by a raw IRQ handler
+static uint32_t raw_irq_mask[NUM_CORES];
 
 // Get the raw value from the pin, bypassing any muxing or overrides.
 int gpio_get_pad(uint gpio) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     hw_set_bits(&padsbank0_hw->io[gpio], PADS_BANK0_GPIO0_IE_BITS);
     return (iobank0_hw->io[gpio].status & IO_BANK0_GPIO0_STATUS_INFROMPAD_BITS)
             >> IO_BANK0_GPIO0_STATUS_INFROMPAD_LSB;
@@ -28,7 +30,7 @@ int gpio_get_pad(uint gpio) {
 // Select function for this GPIO, and ensure input/output are enabled at the pad.
 // This also clears the input/output/irq override bits.
 void gpio_set_function(uint gpio, enum gpio_function fn) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     invalid_params_if(GPIO, ((uint32_t)fn << IO_BANK0_GPIO0_CTRL_FUNCSEL_LSB) & ~IO_BANK0_GPIO0_CTRL_FUNCSEL_BITS);
     // Set input enable on, output disable off
     hw_write_masked(&padsbank0_hw->io[gpio],
@@ -42,14 +44,14 @@ void gpio_set_function(uint gpio, enum gpio_function fn) {
 /// \end::gpio_set_function[]
 
 enum gpio_function gpio_get_function(uint gpio) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     return (enum gpio_function) ((iobank0_hw->io[gpio].ctrl & IO_BANK0_GPIO0_CTRL_FUNCSEL_BITS) >> IO_BANK0_GPIO0_CTRL_FUNCSEL_LSB);
 }
 
 // Note that, on RP2040, setting both pulls enables a "bus keep" function,
 // i.e. weak pull to whatever is current high/low state of GPIO.
 void gpio_set_pulls(uint gpio, bool up, bool down) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     hw_write_masked(
             &padsbank0_hw->io[gpio],
             (bool_to_bit(up) << PADS_BANK0_GPIO0_PUE_LSB) | (bool_to_bit(down) << PADS_BANK0_GPIO0_PDE_LSB),
@@ -59,7 +61,7 @@ void gpio_set_pulls(uint gpio, bool up, bool down) {
 
 // Direct override for per-GPIO IRQ signal
 void gpio_set_irqover(uint gpio, uint value) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     hw_write_masked(&iobank0_hw->io[gpio].ctrl,
                    value << IO_BANK0_GPIO0_CTRL_IRQOVER_LSB,
                    IO_BANK0_GPIO0_CTRL_IRQOVER_BITS
@@ -68,7 +70,7 @@ void gpio_set_irqover(uint gpio, uint value) {
 
 // Direct overrides for pad controls
 void gpio_set_inover(uint gpio, uint value) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     hw_write_masked(&iobank0_hw->io[gpio].ctrl,
                    value << IO_BANK0_GPIO0_CTRL_INOVER_LSB,
                    IO_BANK0_GPIO0_CTRL_INOVER_BITS
@@ -76,7 +78,7 @@ void gpio_set_inover(uint gpio, uint value) {
 }
 
 void gpio_set_outover(uint gpio, uint value) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     hw_write_masked(&iobank0_hw->io[gpio].ctrl,
                    value << IO_BANK0_GPIO0_CTRL_OUTOVER_LSB,
                    IO_BANK0_GPIO0_CTRL_OUTOVER_BITS
@@ -84,7 +86,7 @@ void gpio_set_outover(uint gpio, uint value) {
 }
 
 void gpio_set_oeover(uint gpio, uint value) {
-    invalid_params_if(GPIO, gpio >= NUM_BANK0_GPIOS);
+    check_gpio_param(gpio);
     hw_write_masked(&iobank0_hw->io[gpio].ctrl,
                    value << IO_BANK0_GPIO0_CTRL_OEOVER_LSB,
                    IO_BANK0_GPIO0_CTRL_OEOVER_BITS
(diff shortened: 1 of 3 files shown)


src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+
+/*! \brief Convenience function which performs multiple GPIO IRQ related initializations
+ *  \ingroup hardware_gpio
+ *
+ * This method is a slightly eclectic mix of initialization, that:
+ *
+ * \li Updates whether the specified events for the specified GPIOs cause an interrupt on the calling core bease
+ * on the enable flag
+ *
+ * \li Sets the callback handler for the calling core to callback (or clears the handler it callback is NULL)
```
lurch, line 411: "it callback is NULL" -> " if callback is NULL" ?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ * This method is commonly used to perform a one time setup, and following that any additional IRQs/events are enabled
+ * via \ref gpio_set_irq_enabled. All GPIOs/events added in this way on the same core share the same callback; for multiple
+ * independent handlers for different GPIOs you should use \ref gpio_add_raw_irq_handler and related functions.
+ *
+ * This method is equivalent to.
+ * \code{/c}
+ * gpio_set_irq_enabled(gpio, events, enabled);
+ * gpio_set_irq_callback(callback);
+ * if (enabled) irq_set_enabled(IO_IRQ_BANK0, true);
+ * \endocde
```
lurch, line 424: Typo? https://www.doxygen.nl/manual/commands.html#cmdendcode

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -177,22 +182,50 @@ void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled) {
 }
 
 void gpio_set_irq_enabled_with_callback(uint gpio, uint32_t events, bool enabled, gpio_irq_callback_t callback) {
+    // this is the original behavior
```
lurch, line 185: Not clear what this comment is signifying?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ * \param callback default user function to call on GPIO irq. Note only one of these can be set per processor.
+ */
+void gpio_set_irq_callback(gpio_irq_callback_t callback);
+
+/*! \brief Convenience function which performs multiple GPIO IRQ related initializations
+ *  \ingroup hardware_gpio
+ *
+ * This method is a slightly eclectic mix of initialization, that:
+ *
+ * \li Updates whether the specified events for the specified GPIOs cause an interrupt on the calling core bease
```
lurch, line 408: This function takes a single `uint gpio`, so should "specified GPIOs" be "specified GPIO" ?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ *
+ * This method is equivalent to.
+ * \code{/c}
+ * gpio_set_irq_enabled(gpio, events, enabled);
+ * gpio_set_irq_callback(callback);
+ * if (enabled) irq_set_enabled(IO_IRQ_BANK0, true);
+ * \endocde
  *
- * \note Currently the GPIO parameter is ignored, and this callback will be called for any enabled GPIO IRQ on any pin.
+ * \note The IO IRQs are independent per-processor. This configures affects the processor that calls the function.
```
lurch, line 426: "This configures affects the..." doesn't sound right?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
  *
+ * \param gpio GPIO number
+ * \param events Which events will cause an interrupt. See \ref gpio_irq_level for details.
```
lurch, line 429: Is it worth clarifying that this is a bitmask of events, or is that fairly self-evident?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ *     if (gpio_get_irq_events(my_gpio_num) & my_gpio_events) {
+ *        gpio_acknowledge_irq(my_gpio_num, my_gpio_events);
+ *       // handle the IRQ
+ *     }
+ *     if (gpio_get_irq_events(my_gpio_num2) & my_gpio_events2) {
+ *        gpio_acknowledge_irq(my_gpio_num2, my_gpio_events2);
+ *       // handle the IRQ
+ *     }
+ * }
+ * \code
```
lurch, line 492: `\endocde` ?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ * \param events Bitmask of events to clear. See \ref gpio_irq_level for details.
  */
 void gpio_acknowledge_irq(uint gpio, uint32_t events);
 
+/*! \brief Adds a raw GPIO IRQ handler for the specified GPIOs on the current core
+ *  \ingroup hardware_gpio
+ *
+ * In addition to the default mechanism of a single GPIO IRQ event callback per core (see \ref gpio_set_irq_callback),
+ * it is possible to add explicit GPIO IRQ handlers which are called independent of the default callback. The order
+ * relative to the default callback can be controller via the order_priority parameter. (The default callback has the priority
```
lurch, line 474: "can be controller" -> "can be controlled"

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -177,22 +182,50 @@ void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled) {
...
+    }
+}
+
+void gpio_add_raw_irq_handler_with_order_priority_masked(uint gpio_mask, irq_handler_t handler, uint8_t order_priority) {
+    raw_irq_mask[get_core_num()] |= gpio_mask;
+    irq_add_shared_handler(IO_IRQ_BANK0, handler, order_priority);
+}
+
+void gpio_add_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler) {
+    gpio_add_raw_irq_handler_with_order_priority_masked(gpio_mask, handler, PICO_SHARED_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
```
lurch, line 212: Should this have been `GPIO_CALLBACK_IRQ_ORDER_PRIORITY`, or is there a reason for it being different?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ * \ref GPIO_CALLBACK_IRQ_ORDER_PRIORITY)
+ *
+ * This method adds such a explicit GPIO IRQ handler, and disables the "default" callback for the specified GPIOs.
```
lurch, line 477: nit: "a explicit" -> "an explicit"

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ *
+ * A raw handler should check for whichever GPIOs and events it handles, and acknowledge them itself; it might look something like:
```
lurch, line 479: This is probably a dumb question, but if the user has to manually check for GPIO numbers inside their callback-function themselves, what's the advantage of using these "raw handlers" over the "shared handler"? From looking at the implementation of `gpio_add_raw_irq_handler_with_order_priority_masked` it seems like if you added multiple raw-handlers (each with a separate callback), then _each_ of the callbacks would get called for *every* GPIO IRQ that's registered to be handled by a raw-handler?
So you could just do the same thing with the shared-handler, by calling a different function depending on the GPIO number that it was passed?
No doubt I'm missing something obvious / fundamental :confused:

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ *     }
+ * }
+ * \code
+ *
+ * @param gpio_mask a bit mask of the GPIO numbers that will no longer be passed to the default callback for this core
+ * @param handler the handler to add to the list of GPIO IRQ handlers for this core
+ */
+void gpio_add_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler);
+
+/*! \brief Adds a raw GPIO IRQ handler for a specific GPIO on the current code
```
lurch, line 558: current code -> current core

src/rp2_common/hardware_gpio/gpio.c
kilograham: nope it is different, but i'm adding a another constant

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+/*! \brief Removes a raw GPIO IRQ handler for the specified GPIOs on the current core
+ *  \ingroup hardware_gpio
+ *
+ * In addition to the default mechanism of a single GPIO IRQ event callback per core (see \ref gpio_set_irq_callback),
+ * it is possible to add explicit GPIO IRQ handlers which are called independent of the default event callback.
+ *
+ * This method removes such a callback, and enables the "default" callback for the specified GPIOs.
+ *
+ * @param gpio_mask a bit mask of the GPIO numbers that will now be passed to the default callback for this core
+ * @param handler the handler to remove from the list of GPIO IRQ handlers for this core
```
lurch, line 595: If you use `gpio_add_raw_irq_handler_masked` to e.g. set up GPIOs 2, 3 and 4 to be handled by `my_callback1`, is it possible that you might want to switch GPIOs 2 and 4 back to the default handler, but **not** remove the `my_callback1` handler that's still being used by GPIO 3?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ *
+ * \li Updates whether the specified events for the specified GPIOs cause an interrupt on the calling core bease
+ * on the enable flag
+ *
+ * \li Sets the callback handler for the calling core to callback (or clears the handler it callback is NULL)
+ *
+ * \li Enables GPIO IRQs on the current core if enabled is true.
+ *
+ * This method is commonly used to perform a one time setup, and following that any additional IRQs/events are enabled
+ * via \ref gpio_set_irq_enabled. All GPIOs/events added in this way on the same core share the same callback; for multiple
```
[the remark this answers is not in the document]
kilograham: the events may be different, and I was disfavoring the original method only, so don't think ill change it

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ */
+void gpio_remove_raw_irq_handler_masked(uint gpio_mask, irq_handler_t handler);
+
+/*! \brief Removes a raw GPIO IRQ handler for the specified GPIO on the current core
+ *  \ingroup hardware_gpio
+ *
+ * In addition to the default mechanism of a single GPIO IRQ event callback per core (see \ref gpio_set_irq_callback),
+ * it is possible to add explicit GPIO IRQ handlers which are called independent of the default event callback.
+ *
+ * This method removes such a callback, and enables the "default" callback for the specified GPIOs.
```
lurch, line 605: specified GPIOs -> specified GPIO

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
  *   1 | High level
  *   2 | Edge low
  *   3 | Edge high
+ *
+ *  Which are specified in \ref gpio_irq_level
  */
 void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled);
 
-/*! \brief Enable interrupts for specified GPIO
+#ifndef GPIO_CALLBACK_IRQ_ORDER_PRIORITY
```
lurch, line 383: Should there be a PICO_CONFIG stanza for this, or is it something that you don't expect users to change?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -124,32 +126,35 @@ enum gpio_slew_rate gpio_get_slew_rate(uint gpio) {
...
-            gpio_irq_callback_t callback = _callbacks[get_core_num()];
-            if (callback) {
-                callback(gpio, events);
+static void gpio_default_irq_handler(void) {
+    uint core = get_core_num();
+    gpio_irq_callback_t callback = callbacks[core];
+    io_irq_ctrl_hw_t *irq_ctrl_base = core ? &iobank0_hw->proc1_irq_ctrl : &iobank0_hw->proc0_irq_ctrl;
+    for (uint gpio = 0; gpio < NUM_BANK0_GPIOS; gpio+=8) {
+        uint32_t events8 = irq_ctrl_base->ints[gpio >> 3u];
+        // note we assume events8 is 0 for non-existent GPIO
```
lurch, line 149: What does "non-existent GPIO" mean here, given that you're iterating from `0` to `NUM_BANK0_GPIOS` ?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
lurch: Ahhh, "libraries that need to hook a GPIO should add their own handler" was the bit I was missing, thanks :smiley: 
(is that worth clarifying in the doxygen?)

src/rp2_common/hardware_gpio/gpio.c
lurch: Ahhh, so I guess the fix for that would be to put a `if (i >= NUM_BANK0_GPIOS) break;` inside the inner for-loop? (or even change the inner for-loop itself to `for(uint i=gpio;events8 && i<gpio+8 && i<NUM_BANK0_GPIOS;i++) {` )

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,45 +375,249 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+ * \li Sets the callback handler for the calling core to callback (or clears the handler if the callback is NULL)
+ *
+ * \li Enables GPIO IRQs on the current core if enabled is true.
  *
- * \note Currently the GPIO parameter is ignored, and this callback will be called for any enabled GPIO IRQ on any pin.
+ * This method is commonly used to perform a one time setup, and following that any additional IRQs/events are enabled
+ * via \ref gpio_set_irq_enabled. All GPIOs/events added in this way on the same core share the same callback; for multiple
+ * independent handlers for different GPIOs you should use \ref gpio_add_raw_irq_handler and related functions.
  *
+ * This method is equivalent to.
```
lurch, line 425: nit: I guess the full-stop here ought to be a colon?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -178,21 +183,46 @@ void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled) {
...
+            irq_remove_handler(IO_IRQ_BANK0, gpio_default_irq_handler);
+        }
+        callbacks[core] = callback;
+    } else if (callback) {
+        callbacks[core] = callback;
+        irq_add_shared_handler(IO_IRQ_BANK0, gpio_default_irq_handler, GPIO_IRQ_CALLBACK_ORDER_PRIORITY);
+    }
+}
+
+void gpio_add_raw_irq_handler_with_order_priority_masked(uint gpio_mask, irq_handler_t handler, uint8_t order_priority) {
```
Wren6991, line 203: Should we assert that there is not already a raw IRQ for this GPIO?

Reasoning is that we unconditionally clear the raw bit in the `remove` function, so going `add` -> `add` -> `remove` on the same GPIO leaves it marked as non-raw even though it does have a raw handler installed.

src/rp2_common/hardware_gpio/gpio.c
Wren6991: Not convinced there is anything that needs fixing here. It's not totally clear your proposed change would make anything faster -- seems like six of one, half a dozen the other.

```
// note we assume events8 is 0 for non-existent GPIO
```

This assumption is correct btw, and I believe always will be.

Wren6991 (commented):
Happy with this, other than discussion about whether to assert on double-hooking the same pin, and the comment nits.

At risk of bikeshedding, I'm not crazy about the name `raw`, as opposed to e.g. `per_pin`, but I don't feel strongly enough about this to not approve

src/rp2_common/hardware_gpio/include/hardware/gpio.h
Wren6991: Good point! I think this code works fine as long as you add/remove with the same masks, and some cases where you fail to do this will be caught by an assert on removing the same IRQ twice, but in this case you would actually just lock up the core because nobody would clear the IRQ status.

I don't see any way to fix this without excessive bookkeeping, so maybe we just need to document that add/remove must use the same mask?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -124,32 +126,35 @@ enum gpio_slew_rate gpio_get_slew_rate(uint gpio) {
...
-                callback(gpio, events);
+static void gpio_default_irq_handler(void) {
+    uint core = get_core_num();
+    gpio_irq_callback_t callback = callbacks[core];
+    io_irq_ctrl_hw_t *irq_ctrl_base = core ? &iobank0_hw->proc1_irq_ctrl : &iobank0_hw->proc0_irq_ctrl;
+    for (uint gpio = 0; gpio < NUM_BANK0_GPIOS; gpio+=8) {
+        uint32_t events8 = irq_ctrl_base->ints[gpio >> 3u];
+        // note we assume events8 is 0 for non-existent GPIO
+        for(uint i=gpio;events8 && i<gpio+8;i++) {
+            events8 >>= 4;
```
kilograham, line 151: oops; these are the wrong way around

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -365,46 +375,243 @@ enum gpio_drive_strength gpio_get_drive_strength(uint gpio);
...
+    io_ro_32 *status_reg = &irq_ctrl_base->ints[gpio >> 3u];
+    return (*status_reg >> (4 * (gpio & 7u))) & 0xfu;
+}
+
+/*! \brief Acknowledge a GPIO interrupt for the specified events
+ *  \ingroup hardware_gpio
+ *
+ * \param gpio GPIO number
+ * \param events Bitmask of events to clear. See \ref gpio_irq_level for details.
  */
```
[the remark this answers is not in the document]
kilograham: i have added some blurb

src/rp2_common/hardware_gpio/include/hardware/gpio.h
kilograham: i think these comments are outdated as I don't see the text "libraries that need to"

src/rp2_common/hardware_gpio/include/hardware/gpio.h
kilograham: fixed

kilograham:
> Happy with this, other than discussion about whether to assert on double-hooking the same pin, and the comment nits.
> 
> At risk of bikeshedding, I'm not crazy about the name `raw`, as opposed to e.g. `per_pin`, but I don't feel strongly enough about this to not approve

note it isn't per pin anyway, as it can be more than one pin :-) sticking with raw for now

PR #851, opened 2022-06-05
Add some optional header includes, to make per-target changes to config easier

kilograham:
The main motivation here is to allow inclusion of FreeRTOS on a per target basis, rather than as a project wide thing.
It also allows FreeRTOS to be added at any point in the `CMakeLists.txt`s (rather than just before `pico_sdk_init`) which is much more natural if only a sub-part of your project uses it

### DIFF
--- a/src/common/pico_base/include/pico.h
+++ b/src/common/pico_base/include/pico.h
@@ -22,7 +22,7 @@
 #include "pico/types.h"
 #include "pico/version.h"
 
-// PICO_CONFIG: PICO_CONFIG_HEADER, path to header include in place of the default pico/config.h which may be desirable for build systems which can't easily generate the config_autogen header, type=string, default=none, group=pico_base
+// PICO_CONFIG: PICO_CONFIG_HEADER, unquoted path to header include in place of the default pico/config.h which may be desirable for build systems which can't easily generate the config_autogen header, default=none, group=pico_base
 #ifdef PICO_CONFIG_HEADER
 #include __PICO_XSTRING(PICO_CONFIG_HEADER)
 #else
--- a/src/common/pico_base/include/pico.h
+++ b/src/common/pico_base/include/pico.h
@@ -16,9 +16,18 @@
  * This header may be included by assembly code
 */
 
+#define	__PICO_STRING(x)	#x
+#define	__PICO_XSTRING(x)	__PICO_STRING(x)
+
 #include "pico/types.h"
 #include "pico/version.h"
+
+// PICO_CONFIG: PICO_CONFIG_HEADER, path to header include in place of the default pico/config.h which may be desirable for build systems which can't easily generate the config_autogen header, type=string, default=none, group=pico_base
+#ifdef PICO_CONFIG_HEADER
+#include __PICO_XSTRING(PICO_CONFIG_HEADER)
+#else
 #include "pico/config.h"
+#endif
 #include "pico/platform.h"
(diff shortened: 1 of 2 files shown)


src/common/pico_base/include/pico/config.h
```diff
@@ -18,4 +18,9 @@
 
 #include "pico/config_autogen.h"
 
+// PICO_CONFIG: PICO_CONFIG_RTOS_ADAPTER_HEADER, path to header include in the default pico/config.h for RTOS integration defines that must be included in all sources, type=string, default=none, group=pico_base
```
lurch, line 21: Is this definitely specific to RTOS integration, or could it have other uses too?

src/common/pico_base/include/pico/config.h
```diff
@@ -18,4 +18,9 @@
...
+#ifdef PICO_CONFIG_RTOS_ADAPTER_HEADER
+#include __PICO_XSTRING(PICO_CONFIG_RTOS_ADAPTER_HEADER)
```
lurch, line 23: Does the use of `__PICO_XSTRING` here mean that `pico/config.h` will _only_ ever be included directly by `pico.h` ?

src/common/pico_base/include/pico/config.h
kilograham: It was intended to be specific (you could use it for something else, but then you have no idea if you're colliding with something else). There is no easy way to make a generic include interface (except for the CMake auto-gen which we already have but are deliberately not using), so I just picked this one out since it is the one I know that is needed now

src/common/pico_base/include/pico/config.h
kilograham: Good point, but I think it is fair. The user (and all our headers) are supposed to include `pico.h` as this includes all the other stuff in the right order.

src/common/pico_base/include/pico.h
```diff
@@ -16,9 +16,18 @@
  * This header may be included by assembly code
 */
 
+#define	__PICO_STRING(x)	#x
+#define	__PICO_XSTRING(x)	__PICO_STRING(x)
+
 #include "pico/types.h"
 #include "pico/version.h"
+
+// PICO_CONFIG: PICO_CONFIG_HEADER, path to header include in place of the default pico/config.h which may be desirable for build systems which can't easily generate the config_autogen header, type=string, default=none, group=pico_base
```
lurch, line 25: When I run the `tools/extract_configs.py` script against this PR, it complains that `string` isn't a valid config type - it currently only expects `int`, `bool` or `enum`.
I _could_ add `string` as an additional config type, but https://github.com/raspberrypi/pico-project-generator/blob/master/pico_project.py#L547 (which is what makes use of the TSV files that  `extract_configs` creates) doesn't support `string` yet either.

Let me know if you want me to add `string` config items to `extract_configs` (would we also want minimum or maximum string-length validation?), and I'll add a corresponding issue to `pico-project-generator`.

Hmmm, although looking at the CONFIG descriptions, I wonder if `filename` would be a more suitable type that just `string`? :thinking:

Wren6991 (dismissed):

src/common/pico_base/include/pico.h
kilograham: i will just remove the type for now - feel free to add it back later

Wren6991 (approved):

PR #852, opened 2022-06-05
move get_core_num() into platform.h

kilograham:
It was previously in hardware_sync because it was believed to be only used for inter core stuff, but it is also useful elsewhere

### DIFF
--- a/src/rp2_common/pico_platform/include/pico/platform.h
+++ b/src/rp2_common/pico_platform/include/pico/platform.h
@@ -17,6 +17,8 @@
  */
 
 #include "hardware/platform_defs.h"
+#include "hardware/regs/addressmap.h"
+#include "hardware/regs/sio.h"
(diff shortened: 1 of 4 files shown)


Wren6991 (dismissed):

src/rp2_common/pico_platform/include/pico/platform.h
```diff
@@ -423,6 +425,15 @@ static inline void busy_wait_at_least_cycles(uint32_t minimum_cycles) {
     );
 }
 
+/*! \brief Get the current core number
+ *  \ingroup hardware_sync
```
lurch, line 429: Does this ingroup need updating?

kilograham: yes; and i fixed that yesterday (online) - bad github

PR #864, opened 2022-06-12
Wrap realloc() call with malloc_mutex in multicore

### LINKED ISSUE #863: Realloc() is not wrapped for multi-core safety
[Sorry about busted issue before, I hit return by accident instead of tab]

In `pico_malloc.c` we wrap `malloc`, `free`, and `calloc` to protect the heap in the case of multicore operation.

The standard C function `realloc` is, however, unwrapped.  This can lead to very interesting crashes due to heap corruption if both cores run memory code at the same time (https://github.com/maxgerhardt/platform-raspberrypi/issues/7).

I'll submit a PR fixing this later today, but wanted to track it with an issue in case someone else runs into the same thing I did.  We had some code which only failed under very specific timing circumstances and the two cores only exactly aligned and hit it with a particular GCC `-O` level.

earlephilhower:
Protect against heap corruption by mutex-protecting the realloc() call
(like malloc/free are already).

Fixes #863
Fixes https://github.com/maxgerhardt/platform-raspberrypi/issues/7
Fixes https://github.com/earlephilhower/arduino-pico/issues/614

### DIFF
--- a/src/rp2_common/pico_malloc/CMakeLists.txt
+++ b/src/rp2_common/pico_malloc/CMakeLists.txt
@@ -10,6 +10,7 @@ if (NOT TARGET pico_malloc)
 
     pico_wrap_function(pico_malloc malloc)
     pico_wrap_function(pico_malloc calloc)
+    pico_wrap_function(pico_malloc realloc)
     pico_wrap_function(pico_malloc free)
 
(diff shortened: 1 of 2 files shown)


kilograham:
nice; thanks

kilograham (approved):

PR #869, opened 2022-06-16
Increase PLL min VCO from 400MHz to 750MHz for improved stability across operating conditions

liamfraser:


src/rp2_common/hardware_clocks/scripts/vcocalc.py
```diff
@@ -5,11 +5,13 @@
 parser = argparse.ArgumentParser(description="PLL parameter calculator")
 parser.add_argument("--input", "-i", default=12, help="Input (reference) frequency. Default 12 MHz", type=float)
 parser.add_argument("--vco-max", default=1600, help="Override maximum VCO frequency. Default 1600 MHz", type=float)
-parser.add_argument("--vco-min", default=400, help="Override minimum VCO frequency. Default 400 MHz", type=float)
+parser.add_argument("--vco-min", default=750, help="Override minimum VCO frequency. Default 750 MHz", type=float)
 parser.add_argument("--low-vco", "-l", action="store_true", help="Use a lower VCO frequency when possible. This reduces power consumption, at the cost of increased jitter")
 parser.add_argument("output", help="Output frequency in MHz.", type=float)
 args = parser.parse_args()
 
+assert((args.vco_min >= 750) and (args.vco_max <= 1600))
```
lurch, line 13: It's just occurred to me that we should probably also assert that `(args.vco_min <= args.vco_max)` :wink:

Wren6991: Can we remove this assert please. The limits are there by default, but there's no reason to stop people from calculating values beyond the limits. We're not cops

Wren6991, line 13: Can we remove this assert please. The limits are set and checked by default, but I don't see any reason to stop people from raising limits they want to. We're not cops.

Wren6991 (changes requested):

src/rp2_common/hardware_clocks/scripts/vcocalc.py
Wren6991: Yes, which checks against the limits set by PICO_CONFIG_ defines. This assert was checking the limits themselves, which is different.

kilograham: > Hmmm, so should the `PICO_PLL_VCO_` defines get turned into "full" PICO_CONFIG defines by adding the appropriate comment-lines like e.g. https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_platform/include/pico/platform.h#L24 ?

the reason i didn't is that i didn't expect many people would need to (i was going to include it in the release notes)

PR #870, opened 2022-06-16
make use of TUD_QUEUE_SEND_CB_SUPPORTED if available. Add an assert o…

kilograham:
Avoid use of (1ms) timer for handling TinyUSB polling, by polling in response to something being posted to the event queue.


kilograham:
Converting to draft, until patch to TinyUSB is finalized

Issue #838, opened 2022-05-23
Allow use of pico_stdio_usb with explicit TinyUSB use


kilograham:
Currently pico_stdio_usb requires that the user not be using tinyusb, as it exposes its own USB descriptor(s) and runs its own polling loop. 

The majority of pico_stdio_usb would be fine if the user has defined their own CDC device and it would probably be useful to allow that.

lurch:
Just out of curiosity, would that interfere with the picotool-reset-over-CDC mechanism?

kilograham:
It wouldn't interfere with the reset over CDC mechanism, but that is not what picotool uses. It uses a separate interface, which the user would have to add separately (so the short answer is probably yes)

kilograham:
(The CDC mechanism is to set the baud rate, however using that from the host requires you to know how to translate a USB id into a higher level comms device you can talk to which is non-trivial and different on all platforms)

lurch:
> It uses a separate interface...

Right, that's what I was thinking of, I guess my mention of CDC must have been a mis-remembering.

kilograham:
merged into develop

popy2k14:
Hey guys,

Running tiny USB in device mode as hid keyboard and pico tool doesn't work. I am using no CDC in tiny USB. 

When I get it correctly, this should be possible now. Read your thread here and the commits but just don't what to configure to get CDC for pico tool with standard USB lib alongside tiny USB.

Any hints?

kilograham:
> Hey guys,
> 
> Running tiny USB in device mode as hid keyboard and pico tool doesn't work. I am using no CDC in tiny USB.
> 
> When I get it correctly, this should be possible now. Read your thread here and the commits but just don't what to configure to get CDC for pico tool with standard USB lib alongside tiny USB.
> 
> Any hints?

um CDC != picotool ... you cannot run the picotool interface alongside your app (it is a separate thing int the bootrom)

popy2k14:
Added CDC to TinyUSB descriptors and copied and adapted the function:

```
// Support for default BOOTSEL reset by changing baud rate
void tud_cdc_line_coding_cb(__unused uint8_t itf, cdc_line_coding_t const* p_line_coding) {
    if (p_line_coding->bit_rate == 110) {    
        //board_led_write(1);
        reset_usb_boot(gpio_mask, 0);
    }
}
```

Would be nice if this function from sdk reset interface does'nt get's defined out when using TinyUSB.

[...]

PR #871, opened 2022-06-17
stdio_usb improvements

kilograham:
* Use shared IRQ if available to avoid 1ms timer. 
* Allow use of stdio_usb with user's tinyusb setup if it has CDC

### DIFF
--- a/src/rp2_common/hardware_irq/include/hardware/irq.h
+++ b/src/rp2_common/hardware_irq/include/hardware/irq.h
@@ -252,6 +252,14 @@ void irq_add_shared_handler(uint num, irq_handler_t handler, uint8_t order_prior
  */
 void irq_remove_handler(uint num, irq_handler_t handler);
 
+/*! \brief Determine if the current handler for the given number is shared
+ *  \ingroup hardware_irq
+ *
+ * \param num Interrupt number \ref interrupt_nums
+ * \param return true if the specified IRQ has a shared handler
+ */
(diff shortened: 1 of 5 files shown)


kilograham:
fixes #838

Wren6991 (approved):

eustas:
Hi. This PR has added a check that callback is executed on the right core... But what is the way to setup default alarm pool to fire on core1? Thanks in advance.

thewh1teagle:
Hi
How can I use it?
I tried to set 
```
pico_enable_stdio_usb(server 1)
```
but got the warning
```
stdio USB was configured along with user use of TinyUSB device mode, but CDC is not enabled
```
Trying to make the PI as HID + Serial at the same time

lurch:
All of the Raspberry Pi engineers are currently on holiday, so if none of the examples at https://github.com/raspberrypi/pico-examples/blob/master/README.md#usb-device help, you'll probably get a quicker response from https://forums.raspberrypi.com/

peterharperuk:
Don't configure stdio to use usb. Use a UART instead

thewh1teagle:
@peterharperuk 
But I want to use the same cable for both HID + CDC (for serial) without special tools.

thewh1teagle:
I found a way to do that and it works. the PICO act as HID + CDC device and I can read serial output in my PC in the same time.
[example here](https://github.com/thewh1teagle/pico-ble-server/tree/usb-hid-cdc-ble)

PR #880, opened 2022-06-24
Fix GPIO # callback parameter

### LINKED ISSUE #879: gpio_default_irq_handler() calls callback with incorrect gpio parameter
In the develop branch, after #850 was merged, it seems like the GPIO sent to the user callback is not correct.

I have called the init function as follows for any change on GPIO 9:
````
gpio_set_irq_enabled_with_callback(9 /*gpio #9*/, 12 /* rise or fall */, true /*enabled*/, _gpioInterruptDispatcher /* userland irq handler */);
````

When I set a breakpoint in GDB I can see that a toggle on GPIO #9 does result in a callback to the SDK `gpio_default_irq_handler()`, so the interrupt enables are set properly, but the callback routine gets called with `8` as the interrupting GPIO, not `9`.

Right now the code is only returning GPIO 0, 8, 16, or 24:
https://github.com/raspberrypi/pico-sdk/blob/4c49427bf3048a12569f9890188c4be33e082838/src/rp2_common/hardware_gpio/gpio.c#L152-L157

Line 155 should be
````
         callback(i, events); 
````

earlephilhower:
Fixes #879

Correctly returns `i`, the actual GPIO number, and not the GPIO set-of-8 we're scanning.

Tested to fix https://github.com/earlephilhower/arduino-pico/issues/651

### DIFF
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -152,7 +152,7 @@ static void gpio_default_irq_handler(void) {
             if (events && !(raw_irq_mask[core] & (1u << i))) {
                 gpio_acknowledge_irq(i, events);
                 if (callback) {
-                    callback(gpio, events);
+                    callback(i, events);
                 }
             }
             events8 >>= 4;


kilograham (approved):
oops; thanks

lurch:
@kilograham I wonder if it's worth changing `for (uint gpio = 0; ...` to `for (uint gpio8 = 0; ...`, and `for(uint i=gpio;...` to `for(uint gpio=gpio8;...` to make the code slightly more readable? :shrug:  (and mirroring the `events8` and `events` variable naming)

PR #883, opened 2022-06-27
Update TinyUSB version to latest master

liamfraser:


kilograham (dismissed):

joshuag1000:
Just something to consider but would it not be better to update tiny usb to its latest release (0.13.0) rather than just the latest commit at the time.
Also worth noting that even though the build failed on the bot when using the develop branch with these same changes the builds pass as expected.

kilograham:
> unless there is a reason as to why you are using the latest commit i missed

Yeah, there are some updates for rp2040 we would like to pick up

liamfraser:
Fine on gcc 9.2.1 and gcc 11.2.1

joshuag1000:
> Fine on gcc 9.2.1 and gcc 11.2.1

It seems fine on my pi too now it has been fixed. You might want to update to commit [a62c2640b5217e416ba48aa2011d997cfa7d81a5](https://github.com/hathach/tinyusb/commit/a62c2640b5217e416ba48aa2011d997cfa7d81a5) instead though because it just removes some redundant rp2040 boards from tinyusb as they are already supported here.

kilograham:
duplicate of merged #886

kilograham:
re-opening this as I backed out the change I committed which duplicated this

kilograham:
closed in favor of #937

PR #894, opened 2022-07-01
Add mbedtls to pico-sdk

### LINKED ISSUE #893: Need mbedtls support
It would be helpful to add mbedtls support to the sdk to make it easier to use tls with picow

peterharperuk:
Clone mbedtls somewhere and define PICO_MBEDTLS_PATH to point to the
absolute path, e.g.
export PICO_MBEDTLS_PATH=~/mbedtls

The mbedtls library must be mbedtls-2.28, e.g.
git clone git@github.com:Mbed-TLS/mbedtls.git -b mbedtls-2.28

Rebuild the sdk, then link your code to pico_lwip_mbedtls and
pico_mbedtls. See tls_client example in pico-examples

Fixes #893


lurch:
Is it worth adding this as an optional git submodule, like is done with TinyUSB and lwIP ? :shrug:

peterharperuk:
> Is it worth adding this as an optional git submodule, like is done with TinyUSB and lwIP ? 🤷

I think there's a desire not to add mbedtls to the pico-sdk. But I'll let @kilograham comment.

maxnet:
>I think there's a desire not to add mbedtls to the pico-sdk.

But it does is pico_lwip_mbedtls -which lives in the SDK- that needs mbedtls...

Downside of having mbedtls in every application instead of in the SDK, is that if you update your lwip stuff to a newer version, that may require a different version of mbedtls, and everyone has to update there stuff again.

kilograham:
more of a discussion than "a desire not to"; we will likely add it

tomicooler:
I think a git submodule would make things easier.

I tried to do the same based on how it was done in micropython: [fork](https://github.com/tomicooler/pico-sdk/commit/cefa36599c5734fc84ffccfe75869680f1b727d1).

src/rp2_common/pico_mbedtls/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        ssl_msg.c
+        ssl_srv.c
+        ssl_ticket.c
+        ssl_tls.c
+        ssl_tls13_keys.c
+    )
+
+    set(MBEDTLS_SRC "")
+    foreach(FILE IN LISTS src_crypto src_x509 src_tls)
+        list(APPEND MBEDTLS_SRC ${PICO_MBEDTLS_PATH}/library/${FILE})
```
kilograham, line 117: seems like it would make sense to make multiple different libraries, and then have `pico_mbedtls` depend on all of them

kilograham (commented):
note you didn't change kitchen_sink to include the mbed tls headers, which is mostly what it is there for

src/rp2_common/pico_mbedtls/CMakeLists.txt
peterharperuk: Ok, changed as suggested I think.

peterharperuk:
Do you want me to add mbedtls as a submodule then?

test/kitchen_sink/CMakeLists.txt
```diff
@@ -142,4 +142,19 @@ if (TARGET pico_cyw43_arch)
...
             ${CMAKE_CURRENT_LIST_DIR})
 
+    if (PICO_MBEDTLS_PATH)
+        target_link_libraries(kitchen_sink_lwip_poll
+            pico_lwip_mbedtls
+            pico_mbedtls)
+        target_compile_definitions(kitchen_sink_lwip_poll INTERFACE
+            LIB_PICO_MBEDTLS=1)
+        target_link_libraries(kitchen_sink_lwip_background
+            pico_lwip_mbedtls
```
kilograham, line 152: what is `pico_lwip_mbedtls`? i don't see it created in this PR

kilograham:
Agree on moving the random function into the SDK, but probably as a separate PR (`src/common/pico_random` for the header` and `src/rp2_common/pico_random` for the impl

kilograham:
> Do you want me to add mbedtls as a submodule then?

Yes; i think that is fine, it only looks to be about 15M ... if people can complain we can start suggesting which submodules to init

test/kitchen_sink/CMakeLists.txt
peterharperuk: It's the mbedtls lwip "app" that we use - the alternate socket that gives us tls support. You actually added the source for this when adding lwip. See ./src/rp2_common/pico_lwip/CMakeLists.txt

kilograham:
I'm also curious about threading - mbedtls supports a threading model (basically mutexes)... currently I assume people using it with lwIP would be protected by the lwIP concurrency we provide? i.e. they follow the same rules as for lwip?

Is this good enough, or do we need to provide an implementation of `MBEDTLS_THREADING_ALT` in some situations?

test/kitchen_sink/CMakeLists.txt
kilograham: oh duh... sorry

maxnet:
>currently I assume people using it with lwIP would be protected by the lwIP concurrency we provide?

Eh, what concurrency?
Note that all your examples use the lwip  "raw" API methods, and those are only supposed to be ever called from the main thread/core.

https://www.nongnu.org/lwip/2_0_x/raw_api.html

(lwip does also supports higher level APIs that do offer some thread-safety, but I very much doubt you implemented support for those.
Recall that if you want to use those lwip expects a multi-threading OS and that it can have a thread of its own.)

kilograham:
See https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_cyw43_arch/include/pico/cyw43_arch.h

We support the raw API in general, yes either as raw polling, or in "thread-safe background" mode who's main points are to avoid having to do a user polling loop, but also supporting both cores. This requires that we protect the raw API with a mutex. Then we have the full on FreeRTOS support.

kilograham:
Note, the other reason I'm wondering about threading and `mbedtls` is that ppl might use `pico_mbedtls` independently of wanting `lwip`/networking. I think it is fair to see this is pretty fair for now, because they are perhaps not going to be doing both at the same time, and if they are, may not be sharing any keys etc. but still something to think about

maxnet:
> I think it is fair to see this is pretty fair for now, because they are perhaps not going to be doing both at the same time, and if they are, may not be sharing any keys etc.

Could be that the random number generator stuff is a problem, if both threads want to do crypthographic stuff that requires random.
With that I do not mean mbedtls_hardware_poll(), as that is only used to get some initial entropy.
But functions like this that likely have the mutex for a reason: https://github.com/Mbed-TLS/mbedtls/blob/development/library/ctr_drbg.c#L578..L597

src/rp2_common/pico_mbedtls/CMakeLists.txt
```diff
@@ -0,0 +1,155 @@
...
+        ssl_tls.c
+        ssl_tls13_keys.c
+    )
+    list(TRANSFORM src_tls PREPEND ${PICO_MBEDTLS_PATH}/library/)
+    add_library(pico_mbedtls_tls INTERFACE)
+    target_sources(pico_mbedtls_tls INTERFACE ${src_tls})
+
+    add_library(pico_mbedtls INTERFACE)
+    target_link_libraries(pico_mbedtls INTERFACE pico_mbedtls_crypto pico_mbedtls_x509 pico_mbedtls_tls)
+    target_compile_definitions(pico_mbedtls INTERFACE MBEDTLS_CONFIG_FILE="mbedtls_config.h")
```
kilograham, line 132: we should probably add a `PICO_MBEDTLS_CONFIG_FILE` Cmake var which defaults to "mbedtls_config.h"

kilograham: this is the requested change (go github)

kilograham (changes requested):

src/rp2_common/pico_mbedtls/CMakeLists.txt
peterharperuk: I think I've done this now

kilograham (approved):

PR #911, opened 2022-07-09
Make NO_LWIP panic callbacks weak

### LINKED ISSUE #910: No way to support custom LWIP implementations when NO_LWIP enabled
The Pico SDK gives a good, simple set up for LWIP use, but it combines the WiFi state control and the TCP stack setup in `cyw43_lwip` https://github.com/georgerobotics/cyw43-driver/blob/195dfcc10bb6f379e3dea45147590db2203d3c7b/src/cyw43_lwip.c and in the SDK proper in `pico_cyw43_arch`.

Unfortunately, I think this makes it impossible, without duplicating large portions of the CYW43 support in the SDK, to do more advanced setup (i.e. configure the AP to other addresses, start with a static IP address, use IPV6) and use.

The SDK can be configured with `CYW43_LWIP=0` while still enabling the CYW43 driver (I assume mainly for LED support?) and all the WiFi link-level support, but any use of LWIP at the app-level is denied because there are predefined callbacks the CYW43 driver calls that are hard-coded to `panic` the machine:
https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/rp2_common/pico_cyw43_arch/cyw43_arch_threadsafe_background.c#L300-L317

If these `panic` CBs were made `weak`, then the default behavior would still happen (as an aside, does it make sense to crash w/a`panic`in runtime on the device vs. just failing to link at build time?) but a user application could but an application could override them and do their own LWIP from-the-ground-up implementation.

The alternative now would be to build with `CYW43_LWIP=0` but not use anything from `pico_cyw43_arch` and  delete the fake CBs in the SDK .c library and implement your own `cyw43_cb_process_ethernet`, etc.

earlephilhower:
I expect this to be more a starting point than a final answer for the issue of using a fully configurable LWIP

---
Fixes #910

When CYW43_LWIP=0, make the default ethernet callbacks (which panic the
machine) weak functions.  User applications can then write their own
strong versions of the nevessary callbacks to allow them to set any LWIP
config needed (IPv6, static IPs, etc.)

### DIFF
--- a/src/rp2_common/pico_cyw43_arch/cyw43_arch_threadsafe_background.c
+++ b/src/rp2_common/pico_cyw43_arch/cyw43_arch_threadsafe_background.c
@@ -20,6 +20,7 @@
 #include "lwip/timeouts.h"
 #endif
 
+
 // note same code
 #if PICO_CYW43_ARCH_THREADSAFE_BACKGROUND
 
@@ -68,6 +69,10 @@ static low_prio_irq_dispatch_t low_priority_irq_dispatch_slots[CYW43_DISPATCH_SL
 static recursive_mutex_t cyw43_mutex;
 semaphore_t cyw43_irq_sem;
 
+#if !CYW43_LWIP
+void nolwip_periodic_cb();
+#endif
+
 // Called in low priority pendsv interrupt only to do lwip processing and check cyw43 sleep
 static void periodic_worker(void)
 {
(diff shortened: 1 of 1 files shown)


earlephilhower:
Closing this as it's actually more complicated and this is necessary but not sufficient.

kilograham:
cool; keep us informed... certainly we may want to also split our use of LWIP in the SDK to use a new variable vs CYW43_LWIP which should be whether the driver provides LWIP

PR #914, opened 2022-07-09
Implement LWIP ntohl() and ntohs() with rev and rev16 instructions. Fixes #913

### LINKED ISSUE #913: Use rev and rev16 instructions for LWIP ntohl() and nthos()
One of the ways that the LWIP project recommends optimization for a platform is to provide platform-specific implementations of the byte swap functions. If none are provided, LWIP does it by default with bit twiddling (shifts and ors).

On the RP2040, these can be done in one instruction each with `rev` and `rev16`. PR is on the way.

slimhazard:
With `rev` and `rev16`, byte swapping for network order is done in one cycle, rather than the 3 to 7 cycles or thereabouts needed for bit twiddling.

I tested the patch by building examples for pico_w in pico-examples. In particular, the test in the tcp_server example is successful, and iperf shows bandwidth results with few or no TCP retries.

Unfortunately I can't evaluate the impact on performance, since my home WiFi is not lab conditions. (The PicoW's bandwidth appears to be overwhelmingly affected by how many other devices are also connected to the 2.4 GHz network, and what they're doing.) The iperf bandwidth results with the patch are at any rate not worse, but if there is an improvement, I'm not seeing it as so impactful that it overcomes all of the other factors.

Nevertheless I think the patch makes sense, because `rev` and `rev16` are there for just this sort of thing.

Fixes #913

### DIFF
--- a/src/rp2_common/pico_lwip/include/arch/cc.h
+++ b/src/rp2_common/pico_lwip/include/arch/cc.h
@@ -32,6 +32,8 @@
 #ifndef __CC_H__
 #define __CC_H__
 
+#include <stdint.h>
+
 #if NO_SYS
 // todo really we should just not allow SYS_LIGHTWEIGHT_PROT for nosys mode (it doesn't do anything anyway)
 typedef int sys_prot_t;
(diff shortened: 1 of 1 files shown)


slimhazard:
Feeling a bit foolish now, since I realized after the PR that the compiler probably recognizes the byte swap idiom and emits `rev` and `rev16` anyway. After a build against 1.4.0 and a look in the .dis files, I see that it does. (No wonder I didn't see any difference in iperf.)

So it's your call as to whether you want to make that explicit with the patch. One difference might be that without the patch, the `rev` instructions might not be emitted at a level like `-O0`, but with the patch they will be at any optimization level. That may or may not be a problem, offhand I don't have a clear opinion.

Sorry if this was just noise!

kilograham:
since the compiler handles this except when -O0 (where performance is horrible anyway) I shall close for now

PR #918, opened 2022-07-13
Add call to tud_task in getchar()

kilograham:
* Depending on USB timing, the low priority may never call `tud_task()` during `getchar()
* Also fixed a 1.4.0 regression which may cause `tud_task()` to not be called after an IRQ if the mutex wasn't available (previously we tried again automatically ever 1ms)

### DIFF
--- a/src/rp2_common/pico_stdio_usb/stdio_usb.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb.c
@@ -38,7 +38,6 @@ static uint8_t low_priority_irq_num;
 #endif
 
 static int64_t timer_task(__unused alarm_id_t id, __unused void *user_data) {
-    assert(stdio_usb_core_num == get_core_num()); // if this fails, you have initialized stdio_usb on the wrong core
     int64_t repeat_time;
     if (critical_section_is_initialized(&one_shot_timer_crit_sec)) {
         critical_section_enter_blocking(&one_shot_timer_crit_sec);
@@ -147,6 +146,12 @@ stdio_driver_t stdio_usb = {
 };
 
 bool stdio_usb_init(void) {
+    if (get_core_num() != alarm_pool_core_num(alarm_pool_get_default())) {
+        // included an assertion here rather than just returning false, as this is likely
+        // a coding bug, rather than anything else.
+        assert(false);
+        return false;
+    }
 #ifndef NDEBUG
     stdio_usb_core_num = (uint8_t)get_core_num();
 #endif
--- a/src/rp2_common/pico_stdio_usb/stdio_usb.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb.c
@@ -53,14 +53,20 @@ static int64_t timer_task(__unused alarm_id_t id, __unused void *user_data) {
 }
 
 static void low_priority_worker_irq(void) {
-    // if the mutex is already owned, then we are in user code
-    // in this file which will do a tud_task itself, so we'll just do nothing
-    // until the next tick; we won't starve
     if (mutex_try_enter(&stdio_usb_mutex, NULL)) {
         tud_task();
         mutex_exit(&stdio_usb_mutex);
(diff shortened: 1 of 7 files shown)


liamfraser (dismissed):

kilograham:
ha, as I looked at this again, i realize it should be canceling the existing alarm.. i will add a fix

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -139,8 +175,11 @@ bool stdio_usb_init(void) {
     if (irq_has_shared_handler(USBCTRL_IRQ)) {
         // we can use a shared handler to notice when there may be work to do
         irq_add_shared_handler(USBCTRL_IRQ, usb_irq, PICO_SHARED_IRQ_HANDLER_LOWEST_ORDER_PRIORITY);
+        critical_section_init_with_lock_num(&one_shot_timer_crit_sec, next_striped_spin_lock_num());
     } else {
-        rc = add_alarm_in_us(PICO_STDIO_USB_TASK_INTERVAL_US, timer_task, NULL, true);
+        rc = add_alarm_in_us(PICO_STDIO_USB_TASK_INTERVAL_US, timer_task, NULL, true) >= 0;
+        // we use initialization state of the one_shot_timer_critsec as a flag
+        memset(&one_shot_timer_crit_sec, 0, sizeof(one_shot_timer_crit_sec));
```
lurch, line 182: So AFAICT you're setting `one_shot_timer_crit_sec` to NULL, here, and then elsewhere you're doing `critical_section_is_initialized(&one_shot_timer_crit_sec)`. But given that `critical_section_is_initialized` checks `crit_sec->spin_lock != 0`, won't that lead to a null-pointer dereference? :man_shrugging:

kilograham: setting the contents of one_shot_timer_crit_sec to zero.

liamfraser (dismissed):
lgtm but maybe get @Wren6991 to review also before merging

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -25,32 +25,65 @@ static uint8_t stdio_usb_core_num;
...
+static volatile bool one_shot_timer_pending;
 #ifdef PICO_STDIO_USB_LOW_PRIORITY_IRQ
 static_assert(PICO_STDIO_USB_LOW_PRIORITY_IRQ >= NUM_IRQS - NUM_USER_IRQS, "");
 #define low_priority_irq_num PICO_STDIO_USB_LOW_PRIORITY_IRQ
 #else
 static uint8_t low_priority_irq_num;
 #endif
 
+static int64_t timer_task(__unused alarm_id_t id, __unused void *user_data) {
+    assert(stdio_usb_core_num == get_core_num()); // if this fails, you have initialized stdio_usb on the wrong core
```
Wren6991, line 41: > if this fails, you have initialized stdio_usb on the wrong core

Which core is this supposed to be? The one taking the IRQs for the default timer pool? Maybe it's worth making that constraint explicit in the doc comment for `stdio_usb_init`, especially if the outcome of this assertion is that you won't be getting any stdout.

Might it also be worth failing earlier, at init time, if the USB IRQ is not shared and the init is called on a different core than the default timer IRQ is serviced on?

kilograham: I have moved the assertion (it remains an assertion)... i added a return code to `stdio_init_all()` as it wasn't even previously passed thru.

Note, i also notice that `stdio_usb.c` won't even compile if `PICO_TIME_DEFAULT_ALARM_POOL_DISABLED=1` though no one has apparently needed this. It might be worth adding the ability to initialize `stdio_usb` with your own alarm pool instead, but that is for another day

Wren6991 (approved):

PR #931, opened 2022-07-18
Update .gitignore

ChristianIvarsson:
Exclude the examples folder from the base project.
(This way you don't have to deal with +++ "modified" files in github desktop and the likes)

I guess another solution would be to just include the examples repo as a submodule?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 cmake-*
 .DS_Store
 build
+examples


fsievers:
You don't need to clone the pico-examples folder into the pico-sdk folder. If the SDK path environment variable is correctly set you can clone the examples into folder on your system.

lurch:
The way we typically recommend doing things (see e.g. section 2.1 of https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf ) is having them as "sibling directories", rather than having one as a subdirectory of the other, e.g.
```
pico/
pico/pico-sdk/
pico/pico-examples/
```

ChristianIvarsson:
I have no idea how I could've missed that. I'm so sorry.
I've kinda been rocking this setup since the beginning so I didn't think of actually rtfm now that it's mature.

> The way we typically recommend doing things (see e.g. section 2.1 of https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf ) is having them as "sibling directories", rather than having one as a subdirectory of the other, e.g.
> 
> ```
> pico/
> pico/pico-sdk/
> pico/pico-examples/
> ```

lurch:
The Fine Manuals have been mature since the Raspberry Pi Pico's launch day :books: :blush:

Issue #906, opened 2022-07-08
CDC doesn't respect no flow control setting (DTR control needed to receive data)


MrGreensWorkshop:
### Operating System

Windows, Mac OS, I think in all OS es.

### Board

Raspberry pi pico

### Firmware

just simple app to test cdc behavior.

https://github.com/MrGreensWorkshop/RasPiPicoSDK_cdc_test

- Download
- Compile and run. (or just use binary named PicoTest.uf2)
- Use Terminal.exe or any other terminal you like to check DTR behavior. 

### What happened ?

When handshake is set to none, to receive data we need to enable DTR. We shouldn't be force to enable DTR to receive data. (other flow control status shouldn't be changed in cdc side if it is the case.)

I believe this caused by tinyusb lib. That's why I open this issue also in [tinyusb repo](https://github.com/hathach/tinyusb/issues/1548).

please watch the gif file.

![raspi pico serial dtr](https://user-images.githubusercontent.com/76992231/177780477-e542724c-630b-4e83-a234-a48680f823a0.gif)

### How to reproduce ?

https://github.com/MrGreensWorkshop/RasPiPicoSDK_cdc_test

- Download
- Compile and run. (or just use binary named PicoTest.uf2)
- Use Terminal.exe or any other terminal you like to check DTR behavior as shown in gif file.

### Screenshots

This the demonstration that 2 FDTI USB serial converter connected each other with 3 pins.
```
Rx <- Tx
Tx -> Rx
GND <-> GND
```
It shouldn't be forced to enable DTR to receive data either sides. Please watch the gif file.

![usual usb serial](https://user-images.githubusercontent.com/76992231/177780995-f4d3242d-b360-488a-8ac6-d4a83e4ffb36.gif)

lurch:
Are you sure that this isn't a quirk of your particular serial terminal program?  I never had to fiddle with any DTR settings when using PuTTY as detailed in https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf

MrGreensWorkshop:
PuTTY or Teraterm terminals are made for VT100, not plain text transmission. You can use if you want. But those are uses other pins too. DTR -> DSR  and DSR <-  DTR. That's the problem. Asking host application for DSR on is reduce the compatibility with other apps. I'm saying this because if you check 2 FDTI USB serial converter example gif, it doesn't uses DTR. Most 3rd party apps only uses RX, TX, GND if no flow control involved.

[...]

Issue #921, opened 2022-07-14
stdio over USB needs DTR signal to be activated from terminal program


RobertK66:
I think, this is no bug but a feature ;-).
But at least, IMHO it would be worth a note in the documentation. (Chapter 4.4 in 'Getting started ...')

I wasted some time in debugging my Windows 11 - USB driver Stack and tried to understand your documented issues with the 'USB Enumeration workaround' .....

In the end, I found that using the HTerm terminal program is my issue! This terminal software does not set DTR signal on connect per default! Also my own code using NET Serial Port needed the line:

`  Port.DtrEnable = true;`

to be added in order to work. Using putty iso HTerm worked out of the box.
( But once you know the problem, there is a button fot DTR in HTerm 😄 ... )

lurch:
This was also discussed recently in #906 - I've not been following it very closely, but it seems to be more of a tinyusb thing than a pico-sdk thing? https://github.com/hathach/tinyusb/issues/1548 :shrug:

RobertK66:
Ok, Thx for pointing me there.
I am not sure, if "DTR" is considered part of 'Flow Control' but I also didn't expect the behaviour it currently shows....

RobertK66:
Note (to others having problems to get this up and running): 
I still need the 
`TUD_OPT_RP2040_USB_DEVICE_ENUMERATION_FIX=1`
to make the com port working on my Windows 11 Notebook. If I leave this to default or 0, then the COM port shows up in device manager but is 'not connectable' (Hterm hangs on connect).

lurch:
Whether you need the USB_DEVICE_ENUMERATION_FIX or not depends which other USB devices you have connected at the same time. And it's also not needed if you're using a board using RP2040 B2, e.g. Raspberry Pi Pico W.

kilograham:
merged into develop

PR #932, opened 2022-07-19
Enable/disable connection check made with DTR

MrGreensWorkshop:
This gives users the option to disable DTR check.
this fixes issues #906, #921

### Problem:

tud_cdc_connected is using DTR bit. If not set by host (PC terminal or other device), we can't send anything. 

[src/class/cdc/cdc_device.h](https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.h)
```
static inline bool tud_cdc_connected (void)
{
  return tud_cdc_n_connected(0);
}
```

[src/class/cdc/cdc_device.h](https://github.com/hathach/tinyusb/blob/master/src/class/cdc/cdc_device.c)

```
bool tud_cdc_n_connected(uint8_t itf)
{
  // DTR (bit 0) active  is considered as connected
  return tud_ready() && tu_bit_test(_cdcd_itf[itf].line_state, 0);
}
```

### how to fix
PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR option added to config file for who needs to enable.
It doesn't change current behavior because it's enabled by default. Just gives user to option to disable it.

### Usage
add this to your cmake file and DTR check will be disabled.
add_definitions(-DPICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR=0)

### DIFF
--- a/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
+++ b/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
@@ -97,6 +97,11 @@
 #define PICO_STDIO_USB_RESET_RESET_TO_FLASH_DELAY_MS 100
 #endif
 
+// PICO_CONFIG: PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR, Enable/disable connection check made with DTR checking, type=bool, default=1, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR
+#define PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR 1
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/src/rp2_common/pico_stdio_usb/stdio_usb.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb.c
@@ -60,7 +60,7 @@ static void stdio_usb_out_chars(const char *buf, int length) {
         if (owner == get_core_num()) return; // would deadlock otherwise
         mutex_enter_blocking(&stdio_usb_mutex);
     }
-    if (tud_cdc_connected()) {
+    if (stdio_usb_connected()) {
         for (int i = 0; i < length;) {
             int n = length - i;
             int avail = (int) tud_cdc_write_available();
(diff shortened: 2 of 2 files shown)


lurch:
I guess this might affect the behaviour of the `PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS` and `PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS` options? :shrug:

MrGreensWorkshop:
> I guess this might affect the behaviour of the `PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS` and `PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS` options? 🤷

sorry about that. I will come with better one. just min.

MrGreensWorkshop:
@lurch 
It depends on future plans for `stdio_usb_connected` so I respect that and change the code but I think it doesn't make any difference. 
I mean I don't think this affect the behavior of the those 2 defs.

lurch:
I dunno, it's entirely up to @kilograham , but I was just wondering if it might be e.g. worth doing a compile-time `#warning` if `PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS` is non-zero, and `PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR` is zero? :shrug: 

And IMHO you probably _don't_ want a sleep inside of `stdio_usb_connected`, given how often it gets called?

MrGreensWorkshop:
to be clear `PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR`is default enabled. so sleep will be called if `PICO_STDIO_USB_CONNECTION_CHECK_WITH_DTR` disabled and if `PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS` is set other than 0.

in here, when stdio_usb_connected called it will be always true so, if PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS != 0 I though its best wait it in stdio_usb_connected.

```
        do {
            if (stdio_usb_connected()) {
#if PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS != 0
                sleep_ms(PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS);
#endif
                break;
            }
            sleep_ms(10);
        } while (!time_reached(until));
```

MrGreensWorkshop:
@kilograham Thanks. LGTM.

MrGreensWorkshop:
@kilograham Thanks, 
I add a fix,
1. There was a typo on comments. 
1. PICO_STDIO_USB_CONNECTION_WITHOUT_DTR was enabled by default -> disabled
if you want to revert let me know.

kilograham:
nice, thanks

kilograham (approved):

PR #936, opened 2022-07-20
fix delayed_by_us and delayed_by_ms to not return times > at_the_end_…

### LINKED ISSUE #687: multicore_lockout_start_blocking assertion failure
```
void draw_screen()
{
    multicore_lockout_victim_init();
    while(1)
    {
// in the actual application, I'm doing some i2c work here, but this code also fails.
        tight_loop_contents();
    }
}

int main()
{
    stdio_init_all();
    multicore_launch_core1(draw_screen);
    sleep_ms(100);
    multicore_lockout_start_blocking();
    multicore_lockout_end_blocking();
}
```

with the above code, I'm getting the following assertion failure:
```
assertion "us_since_boot <= INT64_MAX" failed: file "C:\Users\...\pico-sdk\src\common\pico_base\include/pico/types.h", line 63, function: update_us_since_boot
```

I'm on sdk 1.3, and using pico probe to debug... at some point, this code was running for me.

kilograham:
fixes #687

### DIFF
--- a/src/common/pico_time/include/pico/time.h
+++ b/src/common/pico_time/include/pico/time.h
@@ -95,8 +95,9 @@ static inline absolute_time_t delayed_by_us(const absolute_time_t t, uint64_t us
     absolute_time_t t2;
     uint64_t base = to_us_since_boot(t);
     uint64_t delayed = base + us;
-    if (delayed < base) {
-        delayed = (uint64_t)-1;
+    if ((int64_t)delayed < 0) {
+        // absolute_time_t (to allow for signed time deltas) is never greater than INT64_MAX which == at_the_end_of_time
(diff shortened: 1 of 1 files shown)


Wren6991 (approved):

Wren6991:
This change looks ok, though I'm not seeing the path where a negative (or in this case I guess unsigned-wrapped) delta made it into that function. Initially I thought `multicore_lockout_handshake` was the culprit, but seems like we check the sign there.

kilograham:
Yes, this was just a bespoke and vestigial use of -1 for at_the_end_of_time, whereas very soon after SDK1.0 at_the_end_of_time was changed from UINT32_MAX to INT32_MAX

PR #954, opened 2022-08-03
Use system include directories

### LINKED ISSUE #82: Mark headers as SYSTEM includes in CMake
The include paths are set using:

```
    target_include_directories(pico_stdlib_headers INTERFACE include)
```

(e.g. in `src/common/pico_stdlib/CMakeLists.txt`)

As they're not marked as `SYSTEM`,  enabling warnings  (e.g. `-Wall`) will erroneously be tagged in the system headers.

```
pico-sdk/src/rp2040/hardware_regs/include/hardware/regs/addressmap.h:56:20: error: type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
   56 | #define TIMER_BASE 0x40054000
```

for example. If the headers are tagged `SYSTEM` then CMake uses `-Isystem` or equivalent which tells the compiler to suppress warnings coming from those includes.

A potential fix is:

```
$ git grep -l target_include_directories | xargs perl -pi -e 's/target_include_directories(.*)INTERFACE/target_include_directories$1SYSTEM INTERFACE/'
```

### LINKED ISSUE #924: Include directories should be added as system include directories
Currently all include directories are added as none system include directories. This causes warnings and errors when compiling my own source code

```
In file included from /pico-sdk/src/common/pico_stdlib/include/pico/stdlib.h:13,
                 from /blink.c++:2:
/pico-sdk/src/rp2_common/hardware_gpio/include/hardware/gpio.h:882:33: error: ISO C++ does not permit named variadic macros [-Werror=variadic-macros]
  882 | #define CU_REGISTER_DEBUG_PINS(p...) \
      |                                 ^~~

```

ooxi:
Currently all include directories are added as none system include directories. This might causes warnings and errors when compiling user code.

Fixes #82 and #924

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -102,7 +102,7 @@ macro(pico_simple_hardware_headers_target NAME)
     if (NOT TARGET hardware_${NAME}_headers)
         add_library(hardware_${NAME}_headers INTERFACE)
 
-        target_include_directories(hardware_${NAME}_headers INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
+        target_include_directories(hardware_${NAME}_headers SYSTEM INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
         target_link_libraries(hardware_${NAME}_headers INTERFACE pico_base_headers)
         if (NOT PICO_NO_HARDWARE)
             target_link_libraries(hardware_${NAME}_headers INTERFACE hardware_structs hardware_claim_headers)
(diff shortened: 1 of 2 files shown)


ooxi:
@kilograham I have updated the PR to work with the develop branch based upon 1.5.0

Would be great if we could merge this since it's the only remaining issue keeping me from running upstream pico-sdk :-)

peterharperuk:
I don't quite follow why this is needed. I'm surprised if our headers are generating warnings?

PR #958, opened 2022-08-03
Header hardware targets depend on hardware_claim_headers

### LINKED ISSUE #957: A target created by pico_simple_hardware_headers_target should depend on hardware_claim_headers not hardware_claim
A target created by [`pico_simple_hardware_headers_target`](https://github.com/raspberrypi/pico-sdk/blob/master/src/CMakeLists.txt#L59-L72) should depend on `hardware_claim_headers` not `hardware_claim`.

The reason for this change request is `hardware_claim` adding `claim.c` to my sources, which I do not want (otherwise I would have depended on e.g. `hardware_gpio` not `hardware_gpio_headers`).

This is not an issue for `hardware_structs` since this target does not define sources.

ooxi:
A target created by [`pico_simple_hardware_headers_target`](https://github.com/raspberrypi/pico-sdk/blob/master/src/CMakeLists.txt#L59-L72) must depend on `hardware_claim_headers` not `hardware_claim`.

This avoids adding `claim.c` to the user sources. If the user wanted `claim.c` to be added to the sources, the user would have used the targed defined by `pico_simple_hardware_impl_target`.

This is not an issue for `hardware_structs` since this target does not define sources.

Fixes #957

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -58,7 +58,7 @@ endfunction()
 
 # create an INTERFACE library named hardware_NAME_headers INTERFACE library if it doesn't already exist,
 #        and add include/ relative to the calling directory to the includes.
-#        and hardware_structs and hardware_claim as dependencies of the library
+#        and hardware_structs and hardware_claim_headers as dependencies of the library
 macro(pico_simple_hardware_headers_target NAME)
     if (NOT TARGET hardware_${NAME}_headers)
         add_library(hardware_${NAME}_headers INTERFACE)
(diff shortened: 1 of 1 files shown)


kilograham:
presumably you should now also add `hardware_claim` as a dependency during `pico_simple_hardware_impl_target`

ooxi:
@kilograham thanks for the feedback, I will update the patch accordingly

ooxi:
@kilograham I have updated the patch accordingly and tested the change in my application

kilograham:
closed in favor of #1211

PR #962, opened 2022-08-07
Properly terminate cyw43 task when `cyw43_arch_deinit()` is called

### LINKED ISSUE #961: FreeRTOS crashes when `cyw43_arch_deinit()` is called
Calling `cyw43_arch_deinit()` results in a '*** PANIC *** not supported' message in projects using FreeRTOS.

attilawagner:
Fixes #961


kilograham (commented):
Thanks

kilograham (approved):

PR #969, opened 2022-08-10
tcpip_init fix

### LINKED ISSUE #968: A new instance of `tcpip_thread` get spawned for each call to `cyw43_arch_init()`
1. Connect to WiFi
2. Disconnect by calling cyw43_arch_deinit()
3. Repeat
For each repetition, a new FreeRTOS task gets created, eventually leading to a crash or hang when they try to modify the same data structures.

attilawagner:
Fixes #968

It's more a workaround than a proper fix, but I've added a flag to prevent calling `tcpip_init()` multiple times during a single run.

I've added the same workaround to the `threadsafe_background` version for `lwip_init()`, as I've found that the `poll` version also contained it.

### DIFF
--- a/src/rp2_common/pico_cyw43_arch/cyw43_arch_freertos.c
+++ b/src/rp2_common/pico_cyw43_arch/cyw43_arch_freertos.c
@@ -123,9 +123,13 @@ int cyw43_arch_init(void) {
     cyw43_worker_ran_sem = xSemaphoreCreateBinary();
 
 #if CYW43_LWIP
-    SemaphoreHandle_t init_sem = xSemaphoreCreateBinary();
-    tcpip_init(tcpip_init_done, init_sem);
-    xSemaphoreTake(init_sem, portMAX_DELAY);
+    static bool tcpip_init_called = false;
+    if (!tcpip_init_called) {
+      tcpip_init_called = true;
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_cyw43_arch/cyw43_arch_freertos.c
```diff
@@ -123,9 +123,13 @@ int cyw43_arch_init(void) {
     cyw43_worker_ran_sem = xSemaphoreCreateBinary();
 
 #if CYW43_LWIP
-    SemaphoreHandle_t init_sem = xSemaphoreCreateBinary();
-    tcpip_init(tcpip_init_done, init_sem);
-    xSemaphoreTake(init_sem, portMAX_DELAY);
+    static bool tcpip_init_called = false;
+    if (!tcpip_init_called) {
+      tcpip_init_called = true;
```
lurch, line 128: Does access to this variable need to be protected by some kind of mutex, to prevent race conditions? :shrug:

attilawagner: Ideally `cyw43_arch_init()` should only be called from one location, so I don't think a mutex is required here.

kilograham:
fixed by #1177 instead

Issue #974, opened 2022-08-17
dma_irqn_acknowledge_channel incorrectly clears ALL channel interrupts


NomiChirps:
After spending a couple of days chasing this down, I've discovered that attempting to use hw_set_bits (or hw_clear_bits, or hw_xor_bits) with the DMA INTS0/INTS1 interrupt flag register causes ALL of the active channel interrupts to be cleared, not just the ones in the bitmask. This happens even when an empty bitmask is used.

The correct behavior is obtained by simply assigning a bitmask to the register, e.g. `dma_hw->ints0 = 1 << channel_num`. Unfortunately the [current implementation of dma_irqn_acknowledge_channel](https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/rp2_common/hardware_dma/include/hardware/dma.h#L690) uses hw_set_bits.

Discussion: https://forums.raspberrypi.com/viewtopic.php?p=2029939 
Minimal reproduction: https://github.com/NomiChirps/missed_interrupt 

The DMA INTS0/INTS1 register is marked as "WC" type in the RP2040 datasheet:

> WC
> This is a single bit that is typically set by a piece of hardware and then written to by the processor to clear the bit. The
> bit is cleared by writing a 1, using either a normal write or the clear alias. See Section 2.1.2 for more information about
> the clear alias.

So it's advertised that hw_clear_bits() should work with this register, and not clear that hw_set_bits() does anything well-defined. My testing has found that none of the hw_*_bits() aliases function as advertised here, though; they all clear the entire register no matter what bitmask is used.

In conclusion...
- I'd be happy to author a pull request to correct dma_irqn_acknowledge_channel, if you'd like.
- I think it might be worth checking to make sure no other "WC"-typed registers are accidentally misused(?) this way by the SDK.
- It's a mystery why hw_clear_bits() doesn't work with this register, although the data sheet seems to advertise that it should.

Thanks for reading :)

Wren6991:
Thanks for raising this, this is an inconsistency in the behaviour of fields that are implemented with our register block generator (most of them) and those which either were handwritten, or are vendor register blocks with an external atomic shim added.

For the following blocks:

* DMA
* XIP CTRL
* XIP SSI
* I2C
* UART
* SPI

[...]

PR #978, opened 2022-08-17
Apply change for #974 to PIO for consistency

daveythacher:
It is not known if this is required. This is done for consistency purposes. Related to #974 

Another instance of all WC was found in PIO.h. What is the recommendation for registers with multiple WC with RW? I saw one register with a single WC and multiple RW, which should be fine. However, if there are multiple WC there could be an issue.


kilograham (approved):
thanks (also for correctness probably)

daveythacher:
hw_set_bits on ADC register CS could be problematic for START_ONE. Undefined behavior for ERR_STICKY. Is this true?

daveythacher:
ADC register FCS has multiple WC with RW.

daveythacher:
@kilograham The review was lost when I changed baseline.

kilograham (approved):

PR #989, opened 2022-08-25
Make sure OUT_STICKY is not set when setting pins / pindirs

### LINKED ISSUE #988: pio_sm_set_pindirs_with_mask misbehaves if OUT_STICKY is set
We fixed an issue in the rust rp2040 support which is also present in the C SDK: https://github.com/rp-rs/rp-hal/pull/432

If the OUT_STICKY bit is set, pio_sm_set_pindirs_with_mask and similar functions will change the out or outdir values of unrelated pins.
The values written using the injected SET instruction will be sticky, ie. they will be applied again each PIO cycle.
This still happens after resetting `pinctrl` with `pio->sm[sm].pinctrl = pinctrl_saved`. At this point, depending on the previous value of `pinctrl`, the values will be written to some other unintended pin.

jannic:
To avoid the issue described in #988, temporarily clear the OUT_STICKY bit in the affected functions, and restore the value later.
(Note that I didn't test this fix with the C SDK. It's a port of https://github.com/rp-rs/rp-hal/pull/432 which was tested on real hardware, but I might have introduced bugs when porting it to C - my C has become quite rusty, pun intended... Please review carefully)

Fixes #988

### DIFF
--- a/src/rp2_common/hardware_pio/pio.c
+++ b/src/rp2_common/hardware_pio/pio.c
@@ -158,7 +158,7 @@ void pio_sm_set_pins(PIO pio, uint sm, uint32_t pins) {
     check_sm_param(sm);
     uint32_t pinctrl_saved = pio->sm[sm].pinctrl;
     uint32_t execctrl_saved = pio->sm[sm].execctrl;
-    pio->sm[sm].execctrl = pio->sm[sm].execctrl & ~(1 << PIO_SM0_EXECCTRL_OUT_STICKY_LSB);
+    hw_clear_bits(&pio->sm[sm].execctrl, 1u << PIO_SM0_EXECCTRL_OUT_STICKY_LSB);
     uint remaining = 32;
     uint base = 0;
     while (remaining) {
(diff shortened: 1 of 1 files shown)


ithinuel (dismissed):

src/rp2_common/hardware_pio/pio.c
```diff
@@ -157,6 +157,8 @@ void pio_sm_set_pins(PIO pio, uint sm, uint32_t pins) {
     check_pio_param(pio);
     check_sm_param(sm);
     uint32_t pinctrl_saved = pio->sm[sm].pinctrl;
+    uint32_t execctrl_saved = pio->sm[sm].execctrl;
+    pio->sm[sm].execctrl = pio->sm[sm].execctrl & ~(1 << PIO_SM0_EXECCTRL_OUT_STICKY_LSB);
```
kilograham, line 161: please use `hw_clear_bits`

kilograham (changes requested):

jannic:
(Sorry, updated only one instance at first, fixed now.)

kilograham (approved):

PR #992, opened 2022-08-26
Avoid initializing full sdio_stack_buffer object with zeros

kilograham:
Noticed this when debugging something else, that the compiler calls memset here to clear the whole structure (including the character buffer)


lurch:
This PR also bumps the `tinyusb` submodule - was that accidental?

kilograham:
> This PR also bumps the `tinyusb` submodule - was that accidental?

twas. now fixed

liamfraser (approved):

Issue #1000, opened 2022-09-01
sdk fails to build with latest cyw43-driver


peterharperuk:
In file included from /home/peterh/source/pico/pico-sdk/lib/cyw43-driver/src/cyw43_ll.c:40:
/home/peterh/source/pico/pico-sdk/lib/cyw43-driver/src/cyw43_config.h:51:39: fatal error: firmware/wifi_nvram_43439.h: No such file or directory
   51 | #define CYW43_WIFI_NVRAM_INCLUDE_FILE "firmware/wifi_nvram_43439.h"

kilograham:
merged into develop

PR #1001, opened 2022-09-01
Fix cyw43_driver build

peterharperuk:
The latest version of the cyw43-driver fails to build with the pico-sdk
Caused by CYW43_WIFI_NVRAM_INCLUDE_FILE which requires the parent of the
firmware folder to be on the include path.

Fixes https://github.com/raspberrypi/pico-sdk/issues/1000

### DIFF
--- a/src/rp2_common/pico_cyw43_arch/include/cyw43_configport.h
+++ b/src/rp2_common/pico_cyw43_arch/include/cyw43_configport.h
@@ -70,4 +70,8 @@
 #define CYW43_SPI_PIO 1
 #endif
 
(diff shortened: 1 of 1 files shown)


kilograham:
Note I think we wanted to make this user settable anyway, so should perhaps conditionally set it inside our cyw43_configport.h

peterharperuk:
> Note I think we wanted to make this user settable anyway, so should perhaps conditionally set it inside our cyw43_configport.h

Yes - done now.

kilograham (approved):

PR #1011, opened 2022-09-08
Mdns changes

peterharperuk:
I need a new function for an mDns example I'm writing. I also spotted issues when testing different versions of lwip (the mDns API has changed in new versions of lwip)

### DIFF
--- a/src/rp2_common/pico_lwip/CMakeLists.txt
+++ b/src/rp2_common/pico_lwip/CMakeLists.txt
@@ -53,7 +53,6 @@ if (EXISTS ${PICO_LWIP_PATH}/${LWIP_TEST_PATH})
 
     add_library(pico_lwip_core4 INTERFACE)
     target_sources(pico_lwip_core4 INTERFACE
-            ${PICO_LWIP_PATH}/src/core/ipv4/acd.c
             ${PICO_LWIP_PATH}/src/core/ipv4/autoip.c
             ${PICO_LWIP_PATH}/src/core/ipv4/dhcp.c
             ${PICO_LWIP_PATH}/src/core/ipv4/etharp.c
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_cyw43_arch/cyw43_arch.c
```diff
@@ -123,6 +123,18 @@ void cyw43_hal_get_mac(__unused int idx, uint8_t buf[6]) {
     memcpy(buf, cyw43_state.mac, 6);
 }
 
+// e.g. 112233445566
+// chr_off: index of character in mac to start
+// chr_len: length of result
```
lurch, line 128: Is it possible for a bad combination of `chr_off` and `chr_len` values to cause you to "fall off the end" of the mac, before reading the full `chr_len`? Perhaps it's worth adding an assert to guard against that?

lurch:
> Instructions: (please delete)

:wink:

src/rp2_common/pico_cyw43_arch/cyw43_arch.c
peterharperuk: ok - I've had a go at fixing that.

src/rp2_common/pico_cyw43_arch/cyw43_arch.c
```diff
@@ -123,6 +123,24 @@ void cyw43_hal_get_mac(__unused int idx, uint8_t buf[6]) {
...
+// e.g. 112233445566
+// chr_off is index of character in mac to start
+// chr_len is length of result
+// chr_off=8 and chr_len=4 would return "5566"
+// Return number of characters put into destination
+size_t cyw43_hal_get_mac_ascii(int idx, size_t chr_off, size_t chr_len, char *dest_in) {
+    static const char hexchr[16] = "0123456789ABCDEF";
+    uint8_t mac[6];
+    char *dest = dest_in;
+    assert(chr_len <= (2 * sizeof(mac)) && chr_off + chr_len <= (2 * sizeof(mac)));
```
lurch, line 136: Not that it really matters, but I guess if `chr_off + chr_len <= (2 * sizeof(mac))` then `chr_len <= (2 * sizeof(mac))` is implicitly true? :wink:

peterharperuk: yes - let me get rid of it

kilograham:
I'm not that convinced that the new method belongs here... the caller could get the mac, and then do character manipulation themselves. do we expect this to be useful elsewhere?

peterharperuk:
Removed the change that added the function cyw43_hal_get_mac_ascii

kilograham (approved):

PR #1018, opened 2022-09-10
Avoid compound assignment with volatile left operand as described in #1017

### LINKED ISSUE #1017: C++ compiler warning: compound assignment with 'volatile'-qualified left operand is deprecated
Including `hardware/pio.h` from a C++ file generates warnings like the following. My compiler version is `arm-none-eabi-g++ (GNU Toolchain for the Arm Architecture 11.2-2022.02 (arm-11.16)) 11.2.1 20220111`.

```
In file included from <snip>/ws2812.cpp:1:
<snip>/pio.h: In function 'void pio_sm_restart(PIO, uint)':
<snip>/pio.h:583:15: error: compound assignment with 'volatile'-qualified left operand is deprecated [-Werror=volatile]
  583 |     pio->ctrl |= 1u << (PIO_CTRL_SM_RESTART_LSB + sm);
      |     ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<snip>/pio.h: In function 'void pio_restart_sm_mask(PIO, uint32_t)':
<snip>/pio.h:598:15: error: compound assignment with 'volatile'-qualified left operand is deprecated [-Werror=volatile]
  598 |     pio->ctrl |= (mask << PIO_CTRL_SM_RESTART_LSB) & PIO_CTRL_SM_RESTART_BITS;
<snip>/pio.h: In function 'void pio_sm_clkdiv_restart(PIO, uint)':

... and so on.
```

This is because operations such as `|=` and `&=` on volatile variables is deprecated in C++20. Here is a summary of the change: https://blog.feabhas.com/2021/05/modern-embedded-c-deprecation-of-volatile/

What the C++ committee wants you to do is to rewrite each such statement as follows:

```diff
- pio->ctrl |= 1u << (PIO_CTRL_SM_RESTART_LSB + sm);
+ pio->ctrl = pio->ctrl | (1u << (PIO_CTRL_SM_RESTART_LSB + sm));
```

Sample of responses from other projects:

- [Zephyr](https://github.com/zephyrproject-rtos/zephyr/pull/30868) silenced the warning by adding `-Wno-volatile`, pointing to the fact that they do not control the CMSIS header files.
- [CMSIS](https://github.com/ARM-software/CMSIS_5/issues/1544) will not fix the issue in their files as they do not  officially support C++. The maintainer's suggestion that it should work with `extern "C"` is not correct since that's just a linkage directive. 
- [esp-idf](https://github.com/espressif/esp-idf/issues/9170) updated some macros which generated this warning.
- [STMCubeWB](https://github.com/STMicroelectronics/STM32CubeWB/issues/22) has not done anything

jonathangjertsen:
Fixes #1017

Checked by compiling a C++ file that includes all the same files included by `kitchen_sink.c`.

### DIFF
--- a/src/rp2_common/hardware_interp/include/hardware/interp.h
+++ b/src/rp2_common/hardware_interp/include/hardware/interp.h
@@ -296,7 +296,7 @@ static inline void interp_set_config(interp_hw_t *interp, uint lane, interp_conf
  */
 static inline void interp_set_force_bits(interp_hw_t *interp, uint lane, uint bits) {
     // note cannot use hw_set_bits on SIO
-    interp->ctrl[lane] |= (bits << SIO_INTERP0_CTRL_LANE0_FORCE_MSB_LSB);
+    interp->ctrl[lane] = interp->ctrl[lane] | (bits << SIO_INTERP0_CTRL_LANE0_FORCE_MSB_LSB);
 }
 
 typedef struct {
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -580,7 +580,7 @@ static inline void pio_set_sm_mask_enabled(PIO pio, uint32_t mask, bool enabled)
 static inline void pio_sm_restart(PIO pio, uint sm) {
     check_pio_param(pio);
     check_sm_param(sm);
-    pio->ctrl = pio->ctrl | (1u << (PIO_CTRL_SM_RESTART_LSB + sm));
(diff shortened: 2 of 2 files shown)


kilograham:
As per #1017 this C++ change will be undone, however...

All except the `interp` one should be using `hw_set_bits` anyway, so feel free to update this PR to do that, and then leave the `interp = interp | ` as you have it here, as one such occurrence is probably not worth adjusting warnings for

kilograham (changes requested):
as per my comment above

kilograham (approved):

Issue #1021, opened 2022-09-13
SDK portability to IAR toolchain


bavison:
The SDK is a useful basis for Pico projects built using toolchains other than GCC, but it currently relies upon a number of GCC-only features. The IAR toolchain is one of those that is most divergent from GCC (certainly compared to the likes of Clang) but nevertheless it looks practical to extend the SDK so that it can be built with either GCC or IAR.

I'm creating this issue as a reference point for associated PRs (the changes are quite substantial, so I think it makes sense to split them into multiple PRs).

PR #1022, opened 2022-09-13
C aspects of portability to IAR toolchain

bavison:
This PR represents the bulk of the changes to C source files required to permit the SDK to build with the IAR toolchain (as verified with the pico-examples repository). Where possible, code has been refactored to use generic ISO C, but in other places I have used conditional compilation, based on either the `__GNUC__` or `__ICCARM__` preprocessor predefines.

I have tried to roughly group similar issues either into the same commit or into nearby commits in the history, and issues are generally ranked according to the severity of the build failure caused with IAR, from more to less serious. I'm aiming to have all the example projects building cleanly, without any warnings being emitted.

Partially addresses #1021.

### DIFF
--- a/src/common/pico_binary_info/include/pico/binary_info/code.h
+++ b/src/common/pico_binary_info/include/pico/binary_info/code.h
@@ -16,7 +16,7 @@
 #include "pico/binary_info/structure.h"
 
 #if !PICO_NO_BINARY_INFO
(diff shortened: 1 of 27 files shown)


kilograham:
I am going to use this as a basis for some changes for 1.5.1 (IAR specifics will be kept in a separate branch for now)

kilograham:
closed in favor of #1285

PR #1028, opened 2022-09-18
Add packet "build-essential" to required packages.

dm3jan:
`hello_world.c` tutorial fails on a fresh installed Debian machine becurse no compiler is installed, so CMake complains about `No CMAKE_CXX_COMPILER could be found. `

The packet build-essential contains gcc/g++ and with it installed the Kids in the programming course can start without any error on their Laptops. ;-)


lurch:
(and BTW this PR needs to be targetted against `develop` rather than `master`)

PR #1033, opened 2022-09-25
Some optimizations for the hardware divider

boretrk:
save_div_state_and_lr_64 had an unnecessary delay loop, the code before it guaranteed that the divider was done with more than 8 cycles.
Interesting tidbit here is that the needed 8 cycle delay is from when divisor/dividend is written to when result is ready.
Reading back divisor/dividend shouldn't require any delay.

hw_divider_save_state/hw_divider_restore_state becomes faster and uses less stack if done in two steps instead of pushing registers first.

hw_divider_divmod_s32/hw_divider_divmod_u32 can share the same exit code without any performance loss.
The preceding align didn't seem to be used for anything. The regular_func_with_section starts a new section anyway.

### DIFF
--- a/src/rp2_common/hardware_divider/divider.S
+++ b/src/rp2_common/hardware_divider/divider.S
@@ -44,33 +44,24 @@ regular_func_with_section hw_divider_divmod_u32
     bx lr
 // end::hw_div_u32[]
 
-#if SIO_DIV_CSR_READY_LSB == 0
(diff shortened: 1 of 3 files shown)


kilograham:
passes `pico_divider_nesting_test` which is a good sign!

kilograham (approved):

PR #1040, opened 2022-09-30
Add timeout to stdout_serialize_begin

### LINKED ISSUE #1039: stdout_serialize_begin can cause a deadlock
See https://github.com/georgerobotics/cyw43-driver/issues/33

peterharperuk:
stdout_serialize_begin can cause a deadlock in some cases. Testing suggests 7ms is the max time it can wait, so give up after 15ms.

Fixes #1039

### DIFF
--- a/src/rp2_common/pico_stdio/stdio.c
+++ b/src/rp2_common/pico_stdio/stdio.c
@@ -41,6 +41,10 @@ static stdio_driver_t *filter;
 #if PICO_STDOUT_MUTEX
 auto_init_mutex(print_mutex);
 
+#ifndef PICO_STDOUT_SERIALIZE_DELAY_MS
+#define PICO_STDOUT_SERIALIZE_DELAY_MS 15
+#endif
+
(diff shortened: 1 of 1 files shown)


kilograham:
not sure this is the right fix.

kilograham:
> not sure this is the right fix.

Actually, it is the right sort of fix... the issue is that we allow printf in IRQs without disabling IRQs while printf-ing. The existing code makes an attempt to avoid deadlocks by comparing the lock_owner with the core_num, however

* This is not valid under FreeRTOS anyway. 
* It does not help in the case of multiple locks (as evidenced here)

We have no way in general to detect if taking the lock in the IRQ will cause a deadlock, so we should time out.

Note similar code is needed when taking the usb_mutex too (there is already the same broken code there), so probably adding an API makes sense. Assigning to myself to think some more, because I want to decide if the behavior should be different in/out of IRQ.

kilograham:
closed in favor of #1224

Issue #1046, opened 2022-10-04
Unexpected (CMake) behavior with PICO_BOARD variable.


tjko:
When changing PICO_BOARD variable between subsequence "cmake" invocations, there seems to be unexpected (wrong?) behavior when it comes to setting/unsetting PICO_CYW43_SUPPORTED variable.

At first this seems to work fine, passing "-DPICO_BOARD=pico_w" to cmake causes PICO_CYW43_SUPPORTED to get set.
However, if re-run cmake and specify "-DPICO_BOARD=pico", this does not unset the PICO_CYW43_SUPPORTED. And this in turn leads to CYW43 stuff being still included in the build....(which seems "wrong").

Following patch to .cmake files would seem to correct this behavior:
```
diff --git a/src/boards/generic_board.cmake b/src/boards/generic_board.cmake
index 54b43d1..188cea3 100644
--- a/src/boards/generic_board.cmake
+++ b/src/boards/generic_board.cmake
@@ -21,3 +21,5 @@ else()
     string(CONCAT msg ${msg} "   Looked for ${PICO_BOARD}.cmake in ${DIRS} (additional paths specified by PICO_BOARD_CMAKE_DIRS)")
     message(FATAL_ERROR ${msg})
 endif()
+
+set(PICO_CYW43_SUPPORTED "0" CACHE INTERNAL "Exclude support for PICO_CYW43")
diff --git a/src/boards/pico_w.cmake b/src/boards/pico_w.cmake
index 15b901d..1191731 100644
--- a/src/boards/pico_w.cmake
+++ b/src/boards/pico_w.cmake
@@ -1,2 +1,2 @@
-set(PICO_CYW43_SUPPORTED "1" CACHE INTERNAL "Try to add support for PICO_CYW43")
 include(${CMAKE_CURRENT_LIST_DIR}/generic_board.cmake)
+set(PICO_CYW43_SUPPORTED "1" CACHE INTERNAL "Try to add support for PICO_CYW43")
```

(more details here: https://forums.raspberrypi.com/viewtopic.php?t=340817)

peterharperuk:
Looks good to me. Do you want to make a pull request?

kilograham:
merged into develop

PR #1050, opened 2022-10-06
Fix unexpected (CMake) behavior with PICO_BOARD variable.

tjko:
Changing PICO_BOARD value from "pico_w" to "pico" did not unset PICO_CYW43_SUPPORTED variable. Which could lead CYW43 driver included in builds for "pico".  (See issue https://github.com/raspberrypi/pico-sdk/issues/1046)

_Instructions: (please delete)_
 - _please do not submit against `master`, use `develop` instead_
 - _please make sure there is an associated issue for your PR, and reference it via "Fixes #num" in the description_
 - _please enter a detailed description_

### DIFF
--- a/src/boards/generic_board.cmake
+++ b/src/boards/generic_board.cmake
@@ -21,5 +21,3 @@ else()
     string(CONCAT msg ${msg} "   Looked for ${PICO_BOARD}.cmake in ${DIRS} (additional paths specified by PICO_BOARD_CMAKE_DIRS)")
     message(FATAL_ERROR ${msg})
 endif()
-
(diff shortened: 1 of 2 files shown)


lurch:
* please do not submit against `master`, use `develop` instead

kilograham:
Does this actually work as expected; I would think you need `FORCE` on the set commands

tjko:
I'm not really familiar with CMake so not sure if this is "correct" way to fix the issue.  But it definitely fixed the issue I was having...

kilograham:
I updated the fix to have the `PICO_CYW43_SUPPORTED` not be a cache variable

kilograham (approved):

PR #948, opened 2022-07-28
Add pico_lwip_mqtt library

TPRoberts:
When trying to use the [LwIP MQTT client application library ](https://www.nongnu.org/lwip/2_1_x/group__mqtt.html)I observed the required source files were not built, which causes linking errors. This PR adds the `pico_lwip_mqtt` library which builds the LwIP mqtt source. This means a user of the `pico-sdk` can include `"lwip/apps/mqtt.h"` and link `pico_lwip_mqtt` against their desired project.

I believe this PR address issue #944

PR #1056, opened 2022-10-11
add mqtt

mikejonesey:
it seems MQTT was mixed up with MBEDTLS, this meant the app/mqtt was never made availible.

no code change, just a tweek to the cmake to add mqtt code under pico_lwip_mqtt


kilograham:
Please submit PRs against `develop` branch - which as you now see already has this fix (#948)

mikejonesey:
thanks

PR #1067, opened 2022-10-17
fix build without mbedtls submodule initialized, and fix deoptimized debug build

kilograham:


peterharperuk:
Thanks - that explains why I had to do a clean build with this change I guess

PR #1074, opened 2022-10-18
Add "-b <addr>" option to elf2uf2.

### LINKED ISSUE #1071: Add "-b <addr>" option to elf2uf2
The input file is treated as a binary blob to be packaged at the specified address.

Handy when the application expects to find data at a known location in flash.
e.g. Config data, display resources, etc.

esky-software:
The input file is treated as a binary blob to be packaged at the specified address.

Fixes #1071

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -26,6 +26,7 @@ typedef unsigned int uint;
 
 static char error_msg[512];
 static bool verbose;
+static bool binary;
 
 static int fail(int code, const char *format, ...) {
     va_list args;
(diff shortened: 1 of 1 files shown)


lurch:
I may have misunderstood what this is doing, but what if you want an ELF _and_ a binary blob packaged into the same UF2 file?

kilograham:
note this is in back burner as we plan to release an improved tool at some point

kilograham:
converting to draft as we have some upcoming changes which will affect this

will-v-pi:
This has been added to `picotool uf2 convert` (which replaces elf2uf2) as the `-o` option

PR #1088, opened 2022-11-02
Fix RX vs TX typo for clearing PIO FIFOs.

### LINKED ISSUE #1087: Copy-pasta mistake in pio_sm_clear_fifos()
See https://github.com/raspberrypi/pico-sdk/blob/5258ee640bf08909cd8d3e9c3d1c75afb51d9352/src/rp2_common/hardware_pio/include/hardware/pio.h#L1192

Looks like one of those two PIO_SM0_SHIFTCTRL_FJOIN_RX_BITS should have been PIO_SM0_SHIFTCTRL_FJOIN_TX_BITS.

bruelltuete:
Fixes #1087.

Note: I can't test that function right now. But this looks like it was the intended bit flag.

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -1189,7 +1189,7 @@ static inline void pio_sm_clear_fifos(PIO pio, uint sm) {
     check_pio_param(pio);
     check_sm_param(sm);
     hw_xor_bits(&pio->sm[sm].shiftctrl, PIO_SM0_SHIFTCTRL_FJOIN_RX_BITS);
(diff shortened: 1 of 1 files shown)


alastairpatrick:
This change looks wrong. I believe the goal of the modified code is to toggle one of the FJOIN bits and then toggle that same bit back to its original value. Doing so clears the FIFOs because, from the RP2040 datasheet, "Changing FJOIN discards any data present in the state machine’s FIFOs. If this data is irreplaceable, it must be drained beforehand."

The proposed change flips both the RX and TX FJOIN states, leading to the FJOIN state permanently changing, which is not something pio_sm_clear_fifos() should do.

bruelltuete:
Thanks for checking!
You are right, important detail i've overlooked: changing *either* of the fjoin bits clears *both* fifos.

PR #1105, opened 2022-11-20
Feature to Set USB device as self powered device

### LINKED ISSUE #1104: Feature to Set USB device as self powered device
### Overview
I was working on a project that required me to use the Raspberry Pi Pico as a self-powered USB device. Power is supplied from VSYS, and D1 is removed. With this, it is possible to make self-powered USB devices.

We also need to set USB Configuration Descriptor attribute (bmAttributes) to make it happen.

https://www.keil.com/pack/doc/mw/USB/html/_u_s_b__configuration__descriptor.html
https://www.beyondlogic.org/usbnutshell/usb5.shtml

Please check the PR #1105 

#### USBD_MAX_POWER_MA

Screenshot 2022-11-21 at 12 43 41

Btw, even if we set the USB as self powered device there are still 2 resistor on the VBUS side.
If we calculate the power consumption of the VBUS side,

```
 R1 + R10 is 15.6KΩ 
 5V /  15.6KΩ = 0.3mA
 ```

That's why USBD_MAX_POWER_MA is set to 1. 

```c
#define USBD_MAX_POWER_MA (1)
```

#### Usage
add this to your CMakeLists.txt file and USB will be self power device in USB Configuration Descriptor.
add_definitions(-DPICO_STDIO_USB_DEVICE_SELF_POWERED=1)

#### Test

```shell
Bus powered
~$ lsusb -s 3:3 -v | grep bmAttributes
    bmAttributes         0x80
        bmAttributes            3
        bmAttributes            2
        bmAttributes            2
        
Self powered (-DPICO_STDIO_USB_DEVICE_SELF_POWERED=1)
~$ lsusb -s 3:6 -v | grep bmAttributes
    bmAttributes         0xc0
        bmAttributes            3
        bmAttributes            2
        bmAttributes            2
 ```

### PR DESCRIPTION
**Pull request by MrGreensWorkshop**
Fixes #1104

### CODE REVIEW

File: src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h [hunk 1]
```diff
@@ -102,6 +102,11 @@
 #define PICO_STDIO_USB_CONNECTION_WITHOUT_DTR 0
 #endif
 
+// PICO_CONFIG: PICO_STDIO_USB_DEVICE_SELF_POWERED, Set USB device as self powered device, type=bool, default=0, group=pico_stdio_usb
```
lurch, line 105: I know very little about the USB specs, but is this a "generic USB" thing, or is it definitely specific to just pico_stdio_usb ?

MrGreensWorkshop: I'm not sure that I understand your question. I thought I was clear at the issue.

lurch: Ahh, sorry if I wasn't very clear. By "generic USB" I think I was trying to ask if something like `PICO_USB_DEVICE_SELF_POWERED` would be a more suitable name than `PICO_STDIO_USB_DEVICE_SELF_POWERED` ?

(But I'm probably just confusing things, so I'll drop out now)

kilograham (approved):

PR #1111, opened 2022-11-21
Pico rand

andygpz11:
Add improved random number generator for RP2040


src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,283 @@
...
+    // else - deliberately leave what is already in 'rand_storage'
+
+    // Note! The safety of the length assumption here is protected by a 'static_assert' above
+    pico_get_unique_board_id(&board_id);
+    rand_storage.s[0] ^= splitmix64(*(uint64_t *)&board_id);
+
+    // If the RAM has not been hashed or a (re)hash is required, (re)hash the RAM
+    if (!ram_is_hashed || force_ram_rehash) {
+
+        // Note: By design, do *not* initialise the static 'ram_hash' variable
```
peterharperuk, line 236: won't ram_hash be initialised to zero?

peterharperuk (commented):
Looks good to me

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,283 @@
...
+
+	s1 ^= s0;
+	rand_storage.s[0] = rotl(s0, 24) ^ s1 ^ (s1 << 16); // a, b
+	rand_storage.s[1] = rotl(s1, 37); // c
+
+	return result;
+}
+
+
+// Note: This takes approximately 80 milliseconds to execute!
```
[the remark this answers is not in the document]
andygpz11: I'm happy to take it out although this code will need "adjusting" or even removing for any future devices _either_ with a (T)RNG _and/or_ a different amount of RAM.  So this is pretty much RP2040 only anyway.
What do you think?

src/rp2_common/pico_rand/rand.c
andygpz11: Ah, it probably will come to think of it :)  So the comment is wrong, I will fix it.

andygpz11:
Comments fixed and pushed (I thought).

kilograham:
also please fix compiler warnings/errors failing the build...

just rebuild everything from the pico-sdk SDK repo.

andygpz11:
Yes, of course.  Sorry -  I was _not_ expecting build failures based on everything building here...
I will go round the loop again tomorrow - I have a number of different branches of pico-examples and pico-sdk I need to rationalise (one reason that I want this merged, once - it is correct of course).
Should I be using:
https://github.com/raspberrypi/pico-examples/develop
and
https://github.com/raspberrypi/pico-sdk/develop
Thanks

kilograham:
yes develop against the `develop` branches! I rebased your PR onto develop here already

alastairpatrick:
I think the intent of not explicitly initializing ram_hash is to get 64 new random bits after each power cycle? In which case, could do:
```static uint64_t ram_hash __attribute__((section(".uninitialized_data")));```
As written, it will go in the BSS section by default and be initialized to zero.

Although the system clock will commonly be derived from the crystal oscillator when measure_rosc_freq() and update_rosc_samples() are called, the system clock could potentially be derived from the relaxation oscillator, in which case the results will not be random. It might be worth panicking or asserting if the system clock is not suitable, unless the other sources of randomness are adequate on their own.

File: src/rp2_common/pico_rand/include/pico/rand.h [hunk 3]
```diff
@@ -0,0 +1,88 @@
...
+ * with the locally generated seed to produce the initial value used
+ * to start the PRNG.
+ *
+ * This function can be called as often as required.  Note: If the RAM has
+ * already been hashed, either becuase this function has previously been called
+ * or a random number has been already requested, it is optional to hash the
+ * RAM contents again, not doing so reduces the re-seeding time to ~1ms.
+ *
+ * NOTE:  Code within the rand module will automatically call this function
+ * the first time a RN is requested *if* it has not previously been called.
```
lurch, line 50: Presumably 'RN' here stands for Random Number? Might be better to avoid the abbreviation, for the sake of clarity.

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,88 @@
...
+ * It is not recommended to rely on this hook because the time taken may
+ * be unexpected at that point in program execution.
+ *
+ * \param user_seed  NULL _OR_ pointer to user seed of length PICO_RAND_SEED_SIZE_BYTES
+ * \param force_ram_rehash  If true, *re*hash RAM contents and incorporate in new reed
```
lurch, line 55: reed -> seed ? :wink:

andygpz11:
Thanks for all the feedback, very useful and now, all I _think_, fixed*
@alastairpatrick I've moved both ram_hash and also the actual 128-bit state used by the XORshift algo uninitialised data, that will offer a useful improvement on reboots - thanks.
*I get your comment about clock source; I'm not sure of the best answer.  The unpredictability of when this gets called, even with the same clock, should still provide a chance of randomness from that source and there are still the temporal aspects provided by system time and hopefully the fact we don't know where the PRNG machine started in the first place.
My feeling is not to panic/assert since that doesn't help the thing work.  
Arguably we need a large disclaimer somewhere say the entropy of the system _might_ be reduced if we aren't using a separate clock.  We offer no guarantee of anything, just scrape round for everything that _might_ add to it.
The one thing in our favour is that entropy is cumulative.  If you incorporate* sources in the correct way you may or may not add entropy but you won't reduce it.  cf: Almost everything else I've seen in engineering degrades with more handling (precision, purity, spectral growth in RF systems) and you can never get it back.
*Obviously collating poor entropy sources doesn't a good RNG make but the maths here is designed to make it harder to work that out remotely. 
Summary: I'm open to suggestions about the best fix.
@kilograham Sorry - I had not spotted the fact that 'kitchensink' ran tests to a higher stander than just forcing the SDK to get remade in support of example code etc - the things that reported were all useful as well, now fixed I hope.

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,293 @@
...
+    rosc_hw->ctrl = (ROSC_CTRL_ENABLE_VALUE_ENABLE   << ROSC_CTRL_ENABLE_LSB) |
+                    (ROSC_CTRL_FREQ_RANGE_VALUE_HIGH << ROSC_CTRL_FREQ_RANGE_LSB);
+    while(!(rosc_hw->status & ROSC_STATUS_STABLE_BITS)) {
+        tight_loop_contents();
+    }
+
+    rosc_freq = frequency_count_rosc();
+
+    // restore rosc (reduces power)
+    rosc_hw->ctrl = snapshot;
```
alastairpatrick, line 177: This can could cause to clock glitch, which could be an issue if ROSC is the source for the system clock.

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,293 @@
...
+    // We further seed the state so the user seed is optional.
+    if (user_seed != NULL) {
+        struct prng_128_t *user_seed_128 = (struct prng_128_t *)user_seed;
+        rand_storage.s[0] ^= user_seed_128->s[0];
+        rand_storage.s[1] ^= user_seed_128->s[1];
+    }
+    // else - deliberately leave what is already in 'rand_storage'
+
+    // Note! The safety of the length assumption here is protected by a 'static_assert' above
+    pico_get_unique_board_id(&unique_id.board_id_native);
```
alastairpatrick, line 242: This is less code (the unique_id_u_t union declaration isn't needed) and doesn't rely on a GCC's type punning extension for unions.
```pico_unique_board_id_t unique_id;
pico_get_unique_board_id(&unique_id);
uint64_t unique_id_int;
memcpy(&unique_id_int, &unique_id, sizeof(unique_id_int));
rand_storage.s[0] ^= splitmix64(unique_id_int);```

File: src/rp2_common/pico_rand/rand.c [hunk 7]
```diff
@@ -0,0 +1,293 @@
...
+    rand_storage.s[0] ^= splitmix64(rosc_samples);
+    rand_storage.s[1] ^= splitmix64(time_us_64());
+
+    // Generate a 64-bit RN from the modified PRNG state.
+    // Note: This also "churns" the 128-bit state for next time.
+    return xoroshiro128ss();
+}
+
+
+void pico_initialise_rand(void *user_seed, bool force_ram_rehash) {
```
alastairpatrick, line 228: I think this should be:
```void pico_initialise_rand(const struct prng_128_t* user_seed, bool force_ram_rehash)```

`const` because `*user_seed` is not modified. `struct prng_128_t*` rather than `void*` to prevent accidental type punning.

**Comment by andygpz11** [comment]
I have just pushed updated code the this branch, I _think_ this should address the outstanding review comments

**Comment by peterharperuk** [comment]
That "Merge branch" commit is a bit odd?

**Comment by andygpz11** [comment]
@peterharperuk Yes, probably.  I took it as "I've updated my branch from develop", which is what I intended.

**Comment by lurch** [comment]
I think the way people normally tend to do things is to rebase their feature-branch onto `develop` and then force-push, rather than merging the latest version of `develop` into their feature-branch? :shrug:

**Comment by JamesH65** [comment]
Rather than merge dev back to your branch, I do a `get pull --rebase origin develop` on the branch then force push. I think that might be what @lurch is saying?

**Comment by lurch** [comment]
@JamesH65 Yes. Except my muscle-memory always does:
```
git fetch origin
git rebase origin/develop
```

Always more than one way to skin the (octo)cat! :octocat:

**Comment by peterharperuk** [comment]
I tested this with mbedtls and it seems to be working.

File: src/rp2_common/hardware_sync/include/hardware/sync.h [hunk 8]
```diff
@@ -70,6 +70,11 @@ typedef volatile uint32_t spin_lock_t;
 #define PICO_SPINLOCK_ID_HARDWARE_CLAIM 11
 #endif
 
+// PICO_CONFIG: PICO_SPINLOCK_ID_RAND, Spinlock ID for Random Number Generator, min=0, max=31, default=11, group=hardware_sync
```
lurch, line 73: Huh? Says "default=11" here, but then "PICO_SPINLOCK_ID_RAND 12" a couple of lines further down?

andygpz11: Thanks for the spot.  Now fixed.

kilograham (changes requested):

src/rp2_common/pico_rand/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+pico_add_impl_library(pico_rand)
+
+target_sources(pico_rand INTERFACE
+        ${CMAKE_CURRENT_LIST_DIR}/rand.c
+)
+
+target_include_directories(pico_rand INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
+
+target_link_libraries(pico_rand INTERFACE pico_time pico_unique_id)
```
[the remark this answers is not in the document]
andygpz11: Agreed about time but I do think I need:
target_link_libraries(pico_rand INTERFACE **pico_unique_id**)
As, during testing with a "real" example, the build fails for the lack of it.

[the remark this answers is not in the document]
kilograham: what i meant was, use the libraries you DO require (which is pretty clear from what you `#include`)

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,302 @@
...
+    for (unsigned int i = RAM_HASH_START; i < RAM_HASH_END; i++) {
+
+        uint8_t c = *(uint8_t *) i;
+        hash = (uint64_t) c + (hash << 6) + (hash << 16) - hash;
+    }
+    return hash;
+}
+
+
+static uint32_t frequency_count_rosc(void) {
```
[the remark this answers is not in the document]
andygpz11: Most of the module ends up being called from _get_new_rand_64()_ which applies the lock.
I've tagged any function that are not deterministic in their execution times.

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,74 @@
...
+ * samples but system time is also used to add a little extra entropy.
+ *
+ * All entropy sources are hashed before application to the PRNG state machine.
+ *
+ * Note: The *first* time a random number is requested, the 128-bit PRNG state
+ * must be seeded.  The seed generation takes approximately 1 millisecond with
+ * subsequent random numbers taking between 10 and 20 microseconds to generate.
+ */
+
+// PICO_CONFIG: PICO_RAND_MIN_ROSC_SAMPLE_TIME_US, Define a default minimum time between sampling the ROSC random bit, min=5, max=20, group=pico_rand
```
lurch, line 38: Just out of curiosity, is there a reason for these particular min/max values?
Probably makes sense to also include the `default=10` that you use in the `.c` file?

andygpz11: I did a lot of testing using https://github.com/Fourmilab/ent_random_sequence_tester on results collected over a serial port (over many hours : )

And, for whatever reason, observed better results with samples spaced 10us apart of more, than faster samples.
The limits I have specified are something of an estimation, based on the fact that 10us was a decent data point.

Do I have the option not to specify min and max I wonder?

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,303 @@
...
+    // Modify PRNG state with the two run-time entropy sources,
+    // hashed to reduce correlation with previous modifications.
+    rand_storage.r[0] ^= splitmix64(rosc_samples);
+    rand_storage.r[1] ^= splitmix64(time_us_64());
+
+    // Generate a 64-bit RN from the modified PRNG state.
+    // Note: This also "churns" the 128-bit state for next time.
+    rand64 = xoroshiro128ss();
+
+    spin_unlock(lock, save);
```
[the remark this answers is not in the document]
andygpz11: It does need to be under lock.  Did you see my comment above?

_Most of the module ends up being called from get_new_rand_64() which applies the lock.
I've tagged any function that are not deterministic in their execution times._

I have no issue with tagging everything under lock which would also include the splitmix64 calls as well of course.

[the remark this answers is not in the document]
lurch: > I have no issue with tagging everything under lock

That sounds like a Graham-decision to me...

src/rp2_common/pico_rand/include/pico/rand.h
lurch: Yeah, I think the min and max are only so that https://github.com/raspberrypi/pico-project-generator can display suitable options. Should be fine to leave them off if you want.

andygpz11: Thanks.  I might stiffen the text a tiny bit to indicate 10us was as a result of testing.

src/rp2_common/pico_rand/rand.c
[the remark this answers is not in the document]
kilograham: yeah i mean for consistency, then yes.

If every static function requires lock, then possibly just documenting the entire group is fine, but I have no problem with being explicit in all function names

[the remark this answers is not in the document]
andygpz11: Sorry, how do you mean, "documenting the entire group"?
Just with a comment or is there a "more substantial" mechanism I've missed please?
FYI:  _All_ the functions from the start of the module until  get_new_rand_64()  (where the lock is applied) are locked.
Perhaps that helps explain this.

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,74 @@
...
+ * to help provide a large number space.
+ *
+ * The main source of entropy at run time is a provided by ROSC 'random bit'
+ * samples but system time is also used to add a little extra entropy.
+ *
+ * All entropy sources are hashed before application to the PRNG state machine.
+ *
+ * Note: The *first* time a random number is requested, the 128-bit PRNG state
+ * must be seeded.  The seed generation takes approximately 1 millisecond with
+ * subsequent random numbers taking between 10 and 20 microseconds to generate.
```
lurch, line 35: Presumably this time is also dependent upon what `PICO_RAND_MIN_ROSC_SAMPLE_TIME_US` has been set to?

andygpz11: Absolutely

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,303 @@
...
+
+    uint32_t rosc_freq;
+
+    uint32_t snapshot = rosc_hw->ctrl;
+
+    check_rosc_asserts();
+
+    // speed up rosc, wait until stable
+    rosc_hw->ctrl = (ROSC_CTRL_ENABLE_VALUE_ENABLE   << ROSC_CTRL_ENABLE_LSB) |
+                    (ROSC_CTRL_FREQ_RANGE_VALUE_HIGH << ROSC_CTRL_FREQ_RANGE_LSB);
```
bruelltuete, line 187: what does speeding up here do for the random number generator?
does the sped-up frequency only go into initialisation as a tiny bit of startup entropy?
could we make this optional? there might be stuff running from the rosc (other than clk_sys).

kilograham: yeah, i removed this

src/rp2_common/pico_platform/include/pico/platform.h
```diff
@@ -153,14 +153,14 @@ extern "C" {
  *
  * For example a `uint32_t` foo that will retain its value if the program is restarted by reset.
  *
- *     uint32_t __uninitialized_ram("my_group_name") foo;
+ *     uint32_t __uninitialized_ram("foo");
  *
  * The section attribute is `.uninitialized_ram.<group>`
```
lurch, line 158: Does this comment need updating to `.uninitialized_data.<group>` to match the change a few lines below, or am I misunderstanding? :shrug:   (I don't know anything about linker stuff)

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,237 @@
...
+// requested from interrupt context only a small amount of RAM should be hashed
+// hence the 1k default here which will take ~300us on RP2040 at 125Mhz
+
+// PICO_CONFIG: PICO_RAND_DISABLE_ROSC_CHECKS, Dsiable assertions that the ROSC is running when generating random numbers which will drastically reduce randomness, type=bool, default=0, group=pico_rand
+
+// PICO_CONFIG: PICO_RAND_RAM_HASH_END, end of address in RAM rand (non-inclusive) to hash during pico_rand initialization, default=SRAM_END, group=pico_rand
+#ifndef PICO_RAND_RAM_HASH_END
+#define PICO_RAND_RAM_HASH_END     SRAM_END
+#endif
+// PICO_CONFIG: PICO_RAND_RAM_HASH_START, end of address in RAM rand (non-inclusive) to hash during pico_rand initialization, default=SRAM_END, group=pico_rand
```
lurch, line 45: "end of address in RAM rand" -> "start of address in RAM rand" ? Also, the `default=` seems incorrect?

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,237 @@
...
+#include "hardware/clocks.h"
+#include "hardware/structs/rosc.h"
+#include "hardware/sync.h"
+
+// Hashing the selected area of RAM is done the first time
+// a random number is requested. If there is any chance the first RN will be
+// requested from interrupt context only a small amount of RAM should be hashed
+// hence the 1k default here which will take ~300us on RP2040 at 125Mhz
+
+// PICO_CONFIG: PICO_RAND_DISABLE_ROSC_CHECKS, Dsiable assertions that the ROSC is running when generating random numbers which will drastically reduce randomness, type=bool, default=0, group=pico_rand
```
lurch, line 39: "Dsiable" -> "Disable"

src/rp2_common/pico_platform/include/pico/platform.h
kilograham: no that comment was correct, the one i changed wasn;t

src/rp2_common/pico_platform/include/pico/platform.h
```diff
@@ -153,14 +153,14 @@ extern "C" {
...
  *
  * \param group a string suffix to use in the section name to distinguish groups that can be linker
  *              garbage-collected independently
  */
-#define __uninitialized_ram(group) __attribute__((section(".uninitialized_ram." #group))) group
+#define __uninitialized_ram(group) __attribute__((section(".uninitialized_data." #group))) group
```
lurch, line 163: @kilograham This is the "change a few lines below" that I was referring to in my previous comment, but like I said I don't really understand all this stuff.

Looking at the other defines in this file, is line 177 missing a `.` off the end of the name, i.e. should `".flashdata"` be `".flashdata."` ? :shrug:

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,241 @@
...
+// PICO_CONFIG: PICO_RAND_RAM_HASH_END, end of address in RAM (non-inclusive) to hash during pico_rand initialization, default=SRAM_END, group=pico_rand
+#ifndef PICO_RAND_RAM_HASH_END
+#define PICO_RAND_RAM_HASH_END     SRAM_END
+#endif
+// PICO_CONFIG: PICO_RAND_RAM_HASH_START, start of address in RAM (inclusive) to hash during pico_rand initialization, default=PICO_RAND_RAM_HASH_END-1024, group=pico_rand
+#ifndef PICO_RAND_RAM_HASH_START
+#define PICO_RAND_RAM_HASH_START   (PICO_RAND_RAM_HASH_END - 1024u)
+#endif
+
+// PICO_CONFIG: PICO_RAND_MIN_ROSC_SAMPLE_TIME_US, Define a default minimum time between sampling the ROSC random bit, min=5, max=20, group=pico_rand
```
lurch, line 53: Should probably add `default=10` here

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,241 @@
...
+#include "hardware/clocks.h"
+#include "hardware/structs/rosc.h"
+#include "hardware/sync.h"
+
+// Hashing the selected area of RAM is done the first time
+// a random number is requested. If there is any chance the first RN will be
+// requested from interrupt context only a small amount of RAM should be hashed
+// hence the 1k default here which will take ~100us on RP2040 at 125Mhz
+
+// PICO_CONFIG: PICO_RAND_DISABLE_ROSC_CHECKS, Disable assertions that the ROSC is running when generating random numbers which will drastically reduce randomness, type=bool, default=0, group=pico_rand
```
lurch, line 39: This comment _might_ be misread that simply disabling assertions reduces randomness? :wink: 
Could perhaps be something like "Disable assertions that the ROSC is running when generating random numbers. Not having the ROSC running will drastically reduce randomness" ?

lurch: And does this need an explicit
```
#ifndef PICO_RAND_DISABLE_ROSC_CHECKS
#define PICO_RAND_DISABLE_ROSC_CHECKS 0
#endif
```
Or is that superfluous for (boolean) PICO_CONFIG items that default to 0?

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,241 @@
...
+
+static void update_rosc_samples_under_lock(void) {
+    static uint32_t previous_sample_time_us = 0;
+
+    check_rosc_asserts();
+
+    // Ensure that the ROSC random bit is not sampled too quickly,
+    // ROSC may be ticking only a few times a microsecond.
+    // Note: In general (i.e. sporadic) use, very often there will be no delay here.
+    while ((time_us_32() - previous_sample_time_us) < PICO_RAND_MIN_ROSC_SAMPLE_TIME_US) {
```
lurch, line 161: If the ROSC **isn't** enabled, is there any point `in having this delay? (Or even in calling `update_rosc_samples_under_lock` at all?)  But perhaps tweaking that is an optimisation that's not worth it.

kilograham (approved):

src/rp2_common/pico_rand/rand.c
kilograham, line 161: Actually we need to fix this; this is sleeping with interrupts disabled

kilograham (changes requested):

kilograham (dismissed):

src/rp2_common/pico_rand/rand.c
kilograham: it is now undocumented, as not strictly needed

src/rp2_common/pico_rand/rand.c
```diff
@@ -140,100 +116,186 @@ static uint64_t sdbm_hash64_sram(uint64_t hash) {
...
 
-static void update_rosc_samples_under_lock(void) {
-    static uint32_t previous_sample_time_us = 0;
+#if PICO_RAND_SEED_ENTROPY_SRC_ROSC | PICO_RAND_ENTROPY_SRC_ROSC
+/* gather an additional n bits of entropy, and shift them into the 64 bit entropy counter */
+static uint64_t capture_additional_rosc_samples(uint n) {
+    static absolute_time_t next_sample_time;
 
-    check_rosc_asserts();
+    // provide an override if someone really wants it, but disabling ROSC as an entroy source makes more sense
```
P33M, line 126: Typo: `entropy`

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -18,23 +18,128 @@ extern "C" {
...
+ *
+ * \note The \em first time a random number is requested, the 128-bit PRNG state
+ * must be seeded. Multiple entropy sources are also available for the seeding operation:
+ *
+ *  - The Ring Oscillator (ROSC) (\ref PICO_RAND_SEED_ENTROPY_SRC_ROSC == 1):
+ *    64 bits are gathered from the ring oscillator "random bit" and mixed into the seed.
+ *  - Time (\ref PICO_RAND_SEED_ENTROPY_SRC_TIME == 1): The 64-bit microsecond timer is mixed into the seed.
+ *  - Board Identifier (PICO_RAND_SEED_ENTROPY_SRC_BOARD_ID == 1): The board idea via \ref pico_get_unique_board_id
+ *    is mixed into the seed.
+ *  - RAM has (\ref PICO_RAND_SEED_ENTROPY_SRC_RAM_HASH (\ref PICO_RAND_SEED_ENTROPY_SRC_RAM_HAS): The hashed contents of a
```
P33M, line 50: Typo: `RAM hash`

lurch: Also `PICO_RAND_SEED_ENTROPY_SRC_RAM_HAS`

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+ *    \ref PICO_RAND_ROSC_BIT_SAMPLE_COUNT bits are gathered from the ring oscillator "random bit" and mixed in each
+ *    time. This should not be used if the ROSC is off, or the processor is running from
+ *    the ROSC.
+ *    \note the maximum throughput of ROSC bit sampling is controlled by PICO_RAND_MIN_ROSC_BIT_SAMPLE_TIME_US which defaults
+ *    to 10us, i.e. 100,000 bits per second.
+ *  - Time (\ref PICO_RAND_ENTROPY_SRC_TIME == 1): The 64-bit microsecond timer is mixed in each time.
+ *  - Bus Performance Counter (\ref PICO_RAND_ENTROPY_SRC_BUS_PERF_COUNTER == 1): One of the bus fabric's performance
+ *    counters is mixed in each time.
+ *
+ * \note all entropy sources are hashed before application to the PRNG state machine.
```
lurch, line 40: Minor nitpick, but should probably be a capital 'A' on 'all'.

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * \note The \em first time a random number is requested, the 128-bit PRNG state
+ * must be seeded. Multiple entropy sources are also available for the seeding operation:
+ *
+ *  - The Ring Oscillator (ROSC) (\ref PICO_RAND_SEED_ENTROPY_SRC_ROSC == 1):
+ *    64 bits are gathered from the ring oscillator "random bit" and mixed into the seed.
+ *  - Time (\ref PICO_RAND_SEED_ENTROPY_SRC_TIME == 1): The 64-bit microsecond timer is mixed into the seed.
+ *  - Board Identifier (PICO_RAND_SEED_ENTROPY_SRC_BOARD_ID == 1): The board idea via \ref pico_get_unique_board_id
```
lurch, line 48: "board idea" -> "board id"

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+#ifndef PICO_RAND_SEED_ENTROPY_SRC_ROSC
+#define PICO_RAND_SEED_ENTROPY_SRC_ROSC PICO_RAND_ENTROPY_SRC_ROSC
+#endif
+
+// PICO_CONFIG: PICO_RAND_SEED_ENTROPY_SRC_TIME, Enable/disable use of hardware timestamp as an entropy source for the random seed, type=bool, default=1, group=pico_rand
+#ifndef PICO_RAND_SEED_ENTROPY_SRC_TIME
+#define PICO_RAND_SEED_ENTROPY_SRC_TIME PICO_RAND_ENTROPY_SRC_TIME
+#endif
+
+// PICO_CONFIG: PICO_RAND_SEED_ENTROPY_SRC_BOARD_ID, Enable/disable use of as part of the random seed, type=bool, default=1, group=pico_rand
```
lurch, line 92: Missing word(s) between "of" and "as" ?

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+// ---------------------------------
+// PICO_RAND_ENTROPY_SRC_ROSC CONFIG
+// ---------------------------------
+
+// PICO_CONFIG: PICO_RAND_ROSC_BIT_SAMPLE_COUNT, Number of samples to take of the ROSC random bit per random number generation , min=1, max=64, default=1, group=pico_rand
+#ifndef PICO_RAND_ROSC_BIT_SAMPLE_COUNT
+#define PICO_RAND_ROSC_BIT_SAMPLE_COUNT 1
+#endif
+
+// PICO_CONFIG: PICO_RAND_MIN_ROSC_BIT_SAMPLE_TIME_US, Define a default minimum time between sampling the ROSC random bit, min=5, max=20, group=pico_rand
```
lurch, line 111: Should add `default=10,` here

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+// PICO_CONFIG: PICO_RAND_BUS_PERF_COUNTER_EVENT, Bus performance counter event to use for sourcing entropy, default=arbiter_sram5_perf_event_access, group=pico_rand
+#ifndef PICO_RAND_BUS_PERF_COUNTER_EVENT
+#define PICO_RAND_BUS_PERF_COUNTER_EVENT arbiter_sram5_perf_event_access
+#endif
+
+// ------------------------------------------
+// PICO_RAND_SEED_ENTROPY_SRC_RAM_HASH CONFIG
+// ------------------------------------------
+
+// PICO_CONFIG: PICO_RAND_RAM_HASH_END, end of address in RAM (non-inclusive) to hash during pico_rand SEEDization, default=SRAM_END, group=pico_rand
```
lurch, line 134: Fairly sure that "SEEDization" isn't a proper word - I dunno if this was deliberate or is a search'n'replace error?

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+#ifndef PICO_RAND_RAM_HASH_END
+#define PICO_RAND_RAM_HASH_END     SRAM_END
+#endif
+// PICO_CONFIG: PICO_RAND_RAM_HASH_START, start of address in RAM (inclusive) to hash during pico_rand SEEDization, default=PICO_RAND_RAM_HASH_END-1024, group=pico_rand
```
lurch, line 138: I _guess_ you could add MIN and MAX values for the `PICO_RAND_RAM_HASH_START` and `PICO_RAND_RAM_HASH_END` PICO_CONFIG entries?

src/rp2_common/pico_rand/include/pico/rand.h
```diff
@@ -0,0 +1,173 @@
...
+#ifndef PICO_RAND_MIN_ROSC_BIT_SAMPLE_TIME_US
+// (Arbitrary / tested) minimum time between sampling the ROSC random bit
+#define PICO_RAND_MIN_ROSC_BIT_SAMPLE_TIME_US 10u
+#endif
+
+// ---------------------------------------------
+// PICO_RAND_ENTROPY_SRC_BUS_PERF_COUNTER CONFIG
+// ---------------------------------------------
+
+// PICO_CONFIG: PICO_RAND_BUS_PERF_COUNTER_INDEX, Bus performance counter index to use for sourcing entropy, group=pico_rand
```
lurch, line 121: `min=0, max=3` ?

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,303 @@
...
+        // Do not provide 'RNs' until the system has been initialised.  Note:
+        // The first initialisation can be quite time-consuming depending on
+        // the amount of RAM hashed, see RAM_HASH_START and RAM_HASH_END
+        initialise_rand();
+    }
+
+    static volatile uint16_t check_byte;
+    rng_128_t local_rng_state = rng_state;
+    uint8_t local_check_byte = check_byte;
+    // Modify PRNG state with the two run-time entropy sources,
```
lurch, line 258: "the two run-time entropy sources" -> "the run-time entropy sources"

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,303 @@
...
+#endif
+#if PICO_RAND_ENTROPY_SRC_BUS_PERF_COUNTER
+    uint32_t bus_counter_value = bus_ctrl_hw->counter[bus_counter_idx].value;
+    // counter is saturating, so clear it if it has reached saturation
+    if (bus_counter_value == BUSCTRL_PERFCTR0_BITS) {
+        bus_ctrl_hw->counter[bus_counter_idx].value = 0;
+    }
+    local_rng_state.r[which] &= splitmix64(bus_counter_value);
+    which ^= 1;
+#endif
```
lurch, line 277: Should there be a compile-time warning/error if PICO_RAND_ENTROPY_SRC_TIME, PICO_RAND_ENTROPY_SRC_ROSC and PICO_RAND_ENTROPY_SRC_BUS_PERF_COUNTER are **all** disbaled? Or is that too much hand-holding?

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,303 @@
...
+
+uint64_t get_rand_64(void) {
+    if (!rng_initialised) {
+        // Do not provide 'RNs' until the system has been initialised.  Note:
+        // The first initialisation can be quite time-consuming depending on
+        // the amount of RAM hashed, see RAM_HASH_START and RAM_HASH_END
+        initialise_rand();
+    }
+
+    static volatile uint16_t check_byte;
```
lurch, line 255: `uint8_t`, to match local_check_byte ?

src/rp2_common/pico_rand/rand.c
```diff
@@ -0,0 +1,303 @@
...
+    Written in 2015 by Sebastiano Vigna (vigna@acm.org)
+    To the extent possible under law, the author has dedicated all copyright
+    and related and neighboring rights to this software to the public domain
+    worldwide. This software is distributed without any warranty.
+
+    See <http://creativecommons.org/publicdomain/zero/1.0/>
+*/
+
+#include "pico/rand.h"
+#include "pico/unique_id.h"
```
lurch, line 28: Is it worth making these includes conditional, based upon what entropy sources have been selected?  e.g. could have
```
#if PICO_RAND_SEED_ENTROPY_SRC_BOARD_ID
#include "pico/unique_id.h"
#endif
```

kilograham: yeah, i thought about it, but is more clutter, and only really helps if we remove the requirement in the CMakeLists.txt dependencies (where we don't have the `#define`

src/rp2_common/pico_rand/include/pico/rand.h
kilograham: nah

src/rp2_common/pico_rand/rand.c
kilograham: agreed

P33M (approved):

PR #1114, opened 2022-11-23
Disable linker warning

### LINKED ISSUE #1029: Linker warning when compiling with GCC 12
When compiling a [minimal example](https://github.com/ooxi/fedora-36-gcc-12-pico-sdk-linker-warning) with GCC 12.2 I'm getting the following linker warning:

```
[1/5] Performing build step for 'ELF2UF2Build'
ninja: no work to do.
[5/5] Linking CXX executable hello_world.elf
/usr/lib/gcc/arm-none-eabi/12.2.0/../../../../arm-none-eabi/bin/ld: warning: hello_world.elf has a LOAD segment with RWX permissions
/usr/lib/gcc/arm-none-eabi/12.2.0/../../../../arm-none-eabi/bin/ld: warning: hello_world.elf has a LOAD segment with RWX permissions
```

This is confusing for novices. Unfortunately my linker script foo is not sufficient enough to propose a fix.

GCC 12.2 is shipped by Fedora 36, the current stable release. You can see the linker warning with full compile output in [my GitHub Actions build](https://github.com/ooxi/fedora-36-gcc-12-pico-sdk-linker-warning/actions/runs/3084298252/jobs/4986283377#step:4:133)

peterharperuk:
In arm-gnu-toolchain-12.2 we see this warning, that's not relevant to pico. Disable it.

warning: blink.elf has a LOAD segment with RWX permissions

Fixes #1029

### DIFF
--- a/src/rp2_common/pico_standard_link/CMakeLists.txt
+++ b/src/rp2_common/pico_standard_link/CMakeLists.txt
@@ -94,4 +94,20 @@ if (NOT TARGET pico_standard_link)
         target_compile_options(pico_standard_link INTERFACE -ffunction-sections -fdata-sections)
         target_link_options(pico_standard_link INTERFACE "LINKER:--gc-sections")
     endif()
+
+    # Ignore warnings about rwx segments introduced in binutils 2.39
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -print-prog-name=ld RESULT_VARIABLE RUN_C_RESULT OUTPUT_VARIABLE FULL_LD_PATH
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if (${RUN_C_RESULT} EQUAL 0)
+        execute_process(COMMAND ${FULL_LD_PATH} --help RESULT_VARIABLE RUN_LD_RESULT OUTPUT_VARIABLE LD_HELP_OUTPUT
+            OUTPUT_STRIP_TRAILING_WHITESPACE)
+        if (${RUN_LD_RESULT} EQUAL 0)
+                set(RWX_WARNING "no-warn-rwx-segments")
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -94,4 +94,8 @@ if (NOT TARGET pico_standard_link)
         target_compile_options(pico_standard_link INTERFACE -ffunction-sections -fdata-sections)
         target_link_options(pico_standard_link INTERFACE "LINKER:--gc-sections")
     endif()
+
+    if (CMAKE_C_COMPILER_VERSION VERSION_GREATER 12)
```
aswild, line 98: This isn't a great check. A) the warning actually comes from binutils 2.39 rather than any particular C compiler version, and B) clang also exists with version numbers greater than 12.

Unfortunately cmake doesn't give a good way to get the linker version so a proper check here probably needs some sort of way to parse the `ld --version` output.

peterharperuk: Thanks. I'll take another look.

peterharperuk:
> Would be great if we could fix the root issue
Evidentially it's just not a relevant issue for Pico

peterharperuk:
I don't think your change works correctly on my system as it'll match on the number 12.2

GNU ld (Arm GNU Toolchain 12.2.MPACBTI-Bet1 (Build arm-12-mpacbti.16)) 2.39.50.20220821

I'll try something else.

peterharperuk:
Now just runs ld --help and searches for the option.

lurch:
I noticed last night that https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads was updated to 12.2.Rel1 on December 22 2022, so at least this patch will be easier to test now :wink:  (although I've not done that myself)

kilograham (approved):

PR #1130, opened 2022-12-02
Allow changing the USB VID, PID, manufacturer, and product via CMake, fixes #1129

### LINKED ISSUE #1129: Allow changing the USB vendor ID, product ID, manufacturer, and product name using build settings
Presently, it's not possible to change the USB identifiers without forking the SDK and modifying `src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c`. This can be a challenge when shipping a product or even just developing multiple RP2040-based devices at the same time, since it's very difficult to figure out which device is which when multiple are connected.

This ideally should be possible to change through `CMakeLists.txt` like this:

```Cmake
target_compile_definitions(project PUBLIC USBD_VID=0xCAFE USBD_PID=0xCAFE USBD_MANUFACTURER="Widgets, Inc" USBD_PRODUCT="Widget")
```

theacodes:
Small modification to `stdio_usb_descriptors.c` to allow changing the USB vendor ID, product ID, manufacturer name, and product name through build definitions. See #1129.

### DIFF
--- a/src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb_descriptors.c
@@ -32,8 +32,21 @@
 #include "pico/stdio_usb/reset_interface.h"
 #include "pico/unique_id.h"
 
+#ifndef USBD_VID
 #define USBD_VID (0x2E8A) // Raspberry Pi
+#endif
+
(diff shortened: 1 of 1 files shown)


kilograham (approved):

theacodes:
Thank you so much, @kilograham!

PR #1134, opened 2022-12-05
Allow cyw43 worker stack size to be changed

### LINKED ISSUE #1128: Can we have an option for CYW43 Worker task stack size?
Currently it uses configMINIMAL_STACK_SIZE. The only other use for this is the IDLE task(s) that need very little stack. I have found that CYW43 Worker task needs more than 300 entries. Seems a waste of RAM for 2 IDLE tasks to take so much stack as well.

peterharperuk:
Add CYW43_TASK_STACK_SIZE

Fixes #1128


src/rp2_common/pico_cyw43_arch/cyw43_arch_freertos.c
```diff
@@ -39,6 +39,10 @@
 
 #define CYW43_GPIO_IRQ_HANDLER_PRIORITY 0x40
 
+#ifndef CYW43_TASK_STACK_SIZE
+#define CYW43_TASK_STACK_SIZE configMINIMAL_STACK_SIZE
```
kilograham, line 43: yeah, i wonder if `configMINIMAL_STACK_SIZE` is ever a good value ;-) we should maybe set it to something reasonable that works for the variety of normal uses that suggested this issue

kilograham: (i.e. `configMINIMAL_STACK_SIZE` i think is intended for basically a task which hardly uses any stack at all)

kilograham:
please add a PICO_CONFIG comment so it shows up in the docs.

peterharperuk:
Apologies. Fixed.

kilograham (changes requested):
may need rework (or be obsolete) due to async_context changes

peterharperuk:
The only difference between what this commit does and what's in the async branch is that it set the default stack size to 4Kb rather than using min stack size (1Kb).  I think that was a bit excessive. The current iperf and ping examples seem to have about 200bytes of stack free. We could increase the stack size in future but in the meantime CYW43_TASK_STACK_SIZE can be set.

So I don't really think I should do anything with this PR., I might close this and the issue can be resolved once the async changes are merged.

PR #1138, opened 2022-12-06
rework some header library dependencies

kilograham:
fix #958 #882

Issue #882, opened 2022-06-27
Mirror SDK `xxx` library dependencies in `xxx_headers`


kilograham:
All libraries with dependencies should have the correct dependencies in their "header only" versions

This can probably be done semi-automatically.

kilograham:
relates to #881

kilograham:
implemented by #1211

kilograham:
merged into develop

PR #1155, opened 2022-12-23
pico_time_headers must not depend on non header target

### LINKED ISSUE #1154: pico_time_headers must not depend on hardware_timer
The [`pico_time_headers`](https://github.com/raspberrypi/pico-sdk/blob/master/src/common/pico_time/CMakeLists.txt#L6) target currently depends on `hardware_timer` instead of `hardware_timer_headers`, pulling in implementation when only headers are required.

ooxi:
The [pico_time_headers](https://github.com/raspberrypi/pico-sdk/blob/master/src/common/pico_time/CMakeLists.txt#L6) target currently depends on hardware_timer instead of hardware_timer_headers, pulling in implementation when only headers are required.

Fixes #1154


ooxi:
Most likely duplicate of #1138

kilograham:
duplicate of #882

Issue #713, opened 2022-01-19
spi_set_slave should disable SPI


daveythacher:
According to PrimeCell documentation copied into the RP2040 datasheet SSPCR1::MS should not be modified unless disabled.
https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf
```
Master or slave mode select. This bit can be modified only
when the PrimeCell SSP is disabled, SSE=0: 0 Device
configured as master, default. 1 Device configured as
slave.
```
https://developer.arm.com/documentation/ddi0194/h/programmer-s-model/register-descriptions/control-register-1--sspcr1
```
Master or slave mode select.This bit can be modified only when the PrimeCell SSP is disabled, SSE=0:0Device configured as master, default.1Device configured as slave.
```

Looking at this, it would appear that we are to call spi_init followed by spi_set_slave. https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/hardware_spi/include/hardware/spi.h#L193-L207

Documentation from https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__spi.html#ga1dba14fb5488b5e4a69b30678f774df0
```
Initialise SPI instances.

Puts the SPI into a known state, and enable it. Must be called before other functions.
```
SPI is enabled by spi_init. https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/hardware_spi/spi.c#L21-L33

Does the API support SPI? Deinit says I have to call init. I could modify it manually; however, I think this should be corrected. Unless I am missing something.

daveythacher:
This would be valid work around?
```C
    spi_set_format(spi0, 8, SPI_CPOL_0, SPI_CPHA_0, SPI_MSB_FIRST);
    hw_set_bits(&spi_get_hw(spi0)->dmacr, SPI_SSPDMACR_TXDMAE_BITS | SPI_SSPDMACR_RXDMAE_BITS);
    spi_set_slave(spi0, true);
    hw_set_bits(&spi_get_hw(spi0)->cr1, SPI_SSPCR1_SSE_BITS);
```

daveythacher:
> This would be valid work around?
> 
> ```c
>     spi_set_format(spi0, 8, SPI_CPOL_0, SPI_CPHA_0, SPI_MSB_FIRST);
[...]

PR #1158, opened 2022-12-26
#713 Hack to add SPI slave support

daveythacher:
Three SDK functions potentially attempt to change the SPI configuration while it is enabled. I hacked it to disable the module then change the configuration. It will attempt to re-enable the SPI if enabled. I do not save and restore other settings. I am not sure if this is required or not.

Tested with the following code: (Quick and dirty)
```C++
#include <stdio.h>
#include "pico/stdio.h"
#include "hardware/gpio.h"
#include "hardware/spi.h"

int main() {
    // Setup USB Serial
    stdio_init_all();

    // Setup SPI0 Pins
    for (int i = 0; i < 4; i++) {
        gpio_init(i + 2);
        gpio_set_function(i + 2, GPIO_FUNC_SPI);
    }
    gpio_set_dir(3, true);

    // Setup and enable SPI0-slave
    spi_init(spi0, 1000000);
    spi_set_format(spi0, 8, SPI_CPOL_0, SPI_CPHA_1, SPI_MSB_FIRST);
    spi_set_slave(spi0, true);

    while(1) {
        // Test SPI0-slave function
        if (spi_is_readable(spi0)) {
            uint8_t c;
            spi_read_blocking(spi0, 0, &c, 1);
            printf("%c ", c);
        }
    }
}
```

### DIFF
--- a/src/rp2_common/hardware_spi/include/hardware/spi.h
+++ b/src/rp2_common/hardware_spi/include/hardware/spi.h
@@ -183,7 +183,7 @@ static inline void spi_set_format(spi_inst_t *spi, uint data_bits, spi_cpol_t cp
     invalid_params_if(SPI, cpha != SPI_CPHA_0 && cpha != SPI_CPHA_1);
 
     // Disable the SPI
-    uint32_t enable_mask = spi_get_hw(spi)->cr1 & SPI_SSPCR1_SSE_BITS);
+    uint32_t enable_mask = spi_get_hw(spi)->cr1 & SPI_SSPCR1_SSE_BITS;
     hw_clear_bits(&spi_get_hw(spi)->cr1, SPI_SSPCR1_SSE_BITS);
 
     hw_write_masked(&spi_get_hw(spi)->cr0,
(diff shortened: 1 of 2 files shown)


daveythacher:
FYI, do not use more the 1Mbps with that test. It can be modified to support 10Mbps.

daveythacher:
@kilograham Any chance of this on 1.4.1?

daveythacher:
This is one approach to solving this. This is the best path without increasing the major version?

andygpz11 (approved):
Thanks - looks good to me.

kilograham:
re-opened as #1227

andygpz11:
Hi @daveythacher I think your change looks plausible.  But, I do have to ask:
1. If you saw problems provably caused by **_not_** taking the extra steps you have added here?
2. That those issues were then fixed by adding your extra steps?
Thanks.

daveythacher:
I based this on documentation from ARM SPI, which stated explicitly not to do this. This change allows the API to be used without register programming.  In #713 I found a way to use the SDK to implement this with register programming. (Not documented in SDK to my knowledge.)

There are few possible issues which could happen depending on the SPI implementation. However I am assuming that this is not the case. This is not mentioned in the ARM documentation and I do not recall it being an issue in other SPI implementations. Switching from master to slave while enabled is almost always forbidden.

In a version 2.0 of the SDK, the SPI API may need more work. However for the time being this is likely fine, in my opinion. Addressing this with documentation should also work. Not all features are exposed by the SPI API but that is not a major issue. My personal preference is for a disable, configure and enable workflow. However this likely is not required and would actually be annoying if required. With this approach this can be left to the programmer to decide/handle.

One thing that my change does not do and this probably should be documented, is waiting for idle. If the SPI port needs to change roles, a pending transfer could be aborted. However this should be handled in application logic as the SDK should not be expected to handle DMA. A note for the programmer to see in the SDK, should handle this.

PR #1174, opened 2023-01-14
add 0xcb helios

Conor-Burns:
Adding 0xCB Helios

[source](https://github.com/0xCB-dev/0xCB-Helios)

### DIFF
--- a/src/boards/include/boards/0xcb_helios.h
+++ b/src/boards/include/boards/0xcb_helios.h
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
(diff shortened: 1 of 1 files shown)


src/boards/include/boards/0xcb_helios.h
```diff
@@ -0,0 +1,84 @@
...
+// -----------------------------------------------------
+//
+//------------------------------------------------------------------------------------------
+// Board definition for the 0xCB Helios
+
+#ifndef _BOARDS_0XCB_HELIOS_H
+#define _BOARDS_0XCB_HELIOS_H
+
+// For board detection
+#define 0XCB_HELIOS
```
kilograham, line 19: Um; that surely doesn't compile!

kilograham (changes requested):
as per comment, this header won't compile

kilograham:
please fix this if you want it

kilograham:
As OP doesn't seem to be around, I suggest we just add an `_` to the define and commit this (i would have done so in passing, but i can't edit the commit in the original repo)

lurch:
> As OP doesn't seem to be around, I suggest we just add an `_` to the define and commit this (i would have done so in passing, but i can't edit the commit in the original repo)

Done that in #1724

kilograham:
#1724 has been merged into develop

PR #1178, opened 2023-01-18
change clock_gpio_init to take float and add clock_gpio_init_int_frac

howard0su:
The current API only accept a div while the hardware supports frac as well. Make div as float so that we can support FRAC without breaking the existing applications. (C will silent cast int to float to make old code still compile)

### DIFF
--- a/src/rp2_common/hardware_clocks/clocks.c
+++ b/src/rp2_common/hardware_clocks/clocks.c
@@ -314,7 +314,7 @@ void clocks_enable_resus(resus_callback_t resus_callback) {
     clocks_hw->resus.ctrl = CLOCKS_CLK_SYS_RESUS_CTRL_ENABLE_BITS | timeout;
 }
 
-void clock_gpio_init(uint gpio, uint src, uint div) {
+void clock_gpio_init(uint gpio, uint src, float div) {
     // Bit messy but it's as much code to loop through a lookup
     // table. The sources for each gpout generators are the same
     // so just call with the sources from GP0
@@ -327,10 +327,14 @@ void clock_gpio_init(uint gpio, uint src, uint div) {
         invalid_params_if(CLOCKS, true);
     }
 
+    // set up div
+    invalid_params_if(CLOCKS, div <= 0);
+    uint udiv = div * (1u << 8u);
+
     // Set up the gpclk generator
     clocks_hw->clk[gpclk].ctrl = (src << CLOCKS_CLK_GPOUT0_CTRL_AUXSRC_LSB) |
                                  CLOCKS_CLK_GPOUT0_CTRL_ENABLE_BITS;
-    clocks_hw->clk[gpclk].div = div << CLOCKS_CLK_GPOUT0_DIV_INT_LSB;
+    clocks_hw->clk[gpclk].div = udiv;
 
     // Set gpio pin to gpclock function
     gpio_set_function(gpio, GPIO_FUNC_GPCK);
--- a/src/rp2_common/hardware_clocks/clocks.c
+++ b/src/rp2_common/hardware_clocks/clocks.c
@@ -314,7 +314,7 @@ void clocks_enable_resus(resus_callback_t resus_callback) {
     clocks_hw->resus.ctrl = CLOCKS_CLK_SYS_RESUS_CTRL_ENABLE_BITS | timeout;
 }
 
-void clock_gpio_init_with_frac(uint gpio, uint src, uint32_t div, uint8_t frac) {
+void clock_gpio_init_with_frac(uint gpio, uint src, uint32_t div_int, uint8_t div_frac) {
     // Bit messy but it's as much code to loop through a lookup
     // table. The sources for each gpout generators are the same
     // so just call with the sources from GP0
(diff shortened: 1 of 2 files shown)


howard0su:
This is useful when I am working with a I2S chip which requires MCLK. I can use gpout to output the clock.

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -327,10 +327,14 @@ void clock_gpio_init(uint gpio, uint src, uint div) {
         invalid_params_if(CLOCKS, true);
     }
 
+    // set up div
+    invalid_params_if(CLOCKS, div <= 0);
+    uint udiv = div * (1u << 8u);
```
lurch, line 332: Might make sense to replace the `8u` here with `CLOCKS_CLK_GPOUT0_DIV_INT_LSB` ?

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -327,10 +327,14 @@ void clock_gpio_init(uint gpio, uint src, uint div) {
...
+
     // Set up the gpclk generator
     clocks_hw->clk[gpclk].ctrl = (src << CLOCKS_CLK_GPOUT0_CTRL_AUXSRC_LSB) |
                                  CLOCKS_CLK_GPOUT0_CTRL_ENABLE_BITS;
-    clocks_hw->clk[gpclk].div = div << CLOCKS_CLK_GPOUT0_DIV_INT_LSB;
+    clocks_hw->clk[gpclk].div = udiv;
```
lurch, line 337: _Might_ be worth modifying this code to match the style of e.g. https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_adc/include/hardware/adc.h#L164 ?

kilograham:
I would prefer to change the implementation in the C file to be `clock_gpio_init_int_frac`, and add a static inline `clock_gpio_init` to the header that calculates the int and frac, and calls the `clock_gpio_init_int_frac`

This will avoid pulling in floating point code for constant integers in the source, and matches many of the current APIs

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -173,8 +173,23 @@ void clocks_enable_resus(resus_callback_t resus_callback);
...
+ *  \ingroup hardware_clocks
+ *
+ * \param gpio The GPIO pin to output the clock to. Valid GPIOs are: 21, 23, 24, 25. These GPIOs are connected to the GPOUT0-3 clock generators.
+ * \param src  The source clock. See the register field CLOCKS_CLK_GPOUT0_CTRL_AUXSRC for a full list. The list is the same for each GPOUT clock generator.
+ * \param div  The amount to divide the source clock by. This is useful to not overwhelm the GPIO pin with a fast clock.
+ */
+static inline void clock_gpio_init(uint gpio, uint src, float div)
+{
+    uint div_int = (uint)div;
+    uint frac = (uint8_t)((div - (float)div_int) * (1u << CLOCKS_CLK_GPOUT0_DIV_INT_LSB));
```
kilograham, line 190: should be a `uint8_t` ; will give compiler warnings

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -173,8 +173,23 @@ void clocks_enable_resus(resus_callback_t resus_callback);
  * \param gpio The GPIO pin to output the clock to. Valid GPIOs are: 21, 23, 24, 25. These GPIOs are connected to the GPOUT0-3 clock generators.
  * \param src  The source clock. See the register field CLOCKS_CLK_GPOUT0_CTRL_AUXSRC for a full list. The list is the same for each GPOUT clock generator.
  * \param div  The amount to divide the source clock by. This is useful to not overwhelm the GPIO pin with a fast clock.
```
kilograham, line 175: prefer changing the names to `div_int` and `div_frac` and updating the doxygen to be a bit more specific about integer, and fraction between 0 and 255

kilograham (changes requested):

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -172,22 +172,22 @@ void clocks_enable_resus(resus_callback_t resus_callback);
  *
  * \param gpio The GPIO pin to output the clock to. Valid GPIOs are: 21, 23, 24, 25. These GPIOs are connected to the GPOUT0-3 clock generators.
  * \param src  The source clock. See the register field CLOCKS_CLK_GPOUT0_CTRL_AUXSRC for a full list. The list is the same for each GPOUT clock generator.
- * \param div  The amount to divide the source clock by. This is useful to not overwhelm the GPIO pin with a fast clock.
- * \param frac The frac to divide the source clock by.
+ * \param div_int  The integer part of dividing the source clock by. This is useful to not overwhelm the GPIO pin with a fast clock. this is in range of 1..2^24-1.
```
lurch, line 175: Could add an `invalid_params_if` to the .c file to check that `div_int < (1<<24)` ? With the 24 calculated as `CLOCKS_CLK_GPOUT0_DIV_INT_MSB - CLOCKS_CLK_GPOUT0_DIV_INT_LSB + 1` ?
(but if @kilograham disagrees, please listen to him rather than me :laughing: )

kilograham (approved):
lgtm thanks

PR #1179, opened 2023-01-18
cleanup header guard names to be consistent

kilograham:


ooxi:
I agree that having consistent header guards is good, but underscore followed by an upper case character are reserved names. I suggest simply using `PICO_CONFIG_H` etc.

PR #1186, opened 2023-01-23
SYS funcs and GCC12 fixes

kilograham:
* Add implementation of _gettimeofday and settimeofday
* Remove seme GCC warnings about unimplemented SYS functions (e.g. _open) by making them weak
* Removed _exit from crt0.S since we have a weak version in runtime.c and we don't want two weak impls since the linker can't pick If the user omits runtime.c then they'll need to provide _exit or get the error
* Add sys/time.h to arch/cc.h for lwIP as it seems under GCC12 this is not getting included


liamfraser (approved):
LGTM

PR #1187, opened 2023-01-23
elf2uf2: Use LMA (instead of VMA) of entry_point to determine whether binary is flash/RAM

### LINKED ISSUE #689: elf2uf2 incorrectly detects binaries using LMAs
When trying to convert a binary, I got this strange error:
```
ERROR: Memory segment 10000000->10000100 is outside of valid address range for device
```

Running with `-v` outputs `Detected RAM binary`, despite my binary storing all sections in flash.

I'm using LMAs in my linker script to indicate that sections are stored in flash, but copied to sram before execution. It seems like elf2uf2 incorrectly detects this as being a RAM binary, since the entry address points to SRAM.

kilograham:
fixes #689

### DIFF
--- a/tools/elf2uf2/main.cpp
+++ b/tools/elf2uf2/main.cpp
@@ -129,7 +129,7 @@ int check_address_range(const address_ranges& valid_ranges, uint32_t addr, uint3
     for(const auto& range : valid_ranges) {
         if (range.from <= addr && range.to >= addr + size) {
             if (range.type == address_range::type::NO_CONTENTS && !uninitialized) {
-                return fail(ERROR_INCOMPATIBLE, "ELF contains memory contents for uninitialized memory");
+                return fail(ERROR_INCOMPATIBLE, "ELF contains memory contents for uninitialized memory at 0x%p", addr);
             }
             ar = range;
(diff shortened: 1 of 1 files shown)


tools/elf2uf2/main.cpp
```diff
@@ -142,61 +142,64 @@ int check_address_range(const address_ranges& valid_ranges, uint32_t addr, uint3
...
+                uint file_offset = entry.offset;
+                while (remaining) {
+                    uint off = addr & (PAGE_SIZE - 1);
+                    uint len = std::min(remaining, PAGE_SIZE - off);
+                    auto &fragments = pages[addr - off]; // list of fragments
+                    // note if filesz is zero, we want zero init which is handled because the
+                    // statement above creates an empty page fragment list
+                    // check overlap with any existing fragments
+                    for (const auto &fragment : fragments) {
+                        if ((off < fragment.page_offset + fragment.bytes) !=
```
peterharperuk, line 186: np. I think I follow this. Could perhaps have been written more clearly.

peterharperuk (approved):
Looks ok. I can't spot any issues

PR #1177, opened 2023-01-18
Add new async_context abstraction and refactor cyw43_arch to use it

kilograham:
Extract all poll/threadsafe_background/freertos from cyw43_arch into new abstraction async_context:

* provides support for asynchronous events (timers/IRQ notifications) to be handled in a safe context.
* now guarantees all callbacks happen on a single core.
* is reusable by multiple different libraries (stdio_usb can now be ported to this but hasn't been yet).
* supports multiple independent instances (independent instances will not block each other).

cyw43_arch libraries cleaned up to use the new abstraction. Note each distinct cyw43_arch type is now a very thin layer
that creates the right type of context and adds cyw43_driver and lwip support as appropriate.

PR #1188, opened 2023-01-24
Keep retrying if cyw43_wifi_join fails

### LINKED ISSUE #1054: Pico W cyw43_arch_wifi_connect_timeout_ms gives up before timeout
### Problem
Pico W cyw43_arch_wifi_connect_timeout_ms gives up before timeout.

### Expected behavior
Calling cyw43_arch_wifi_connect_timeout_ms is expected to try to connect to wifi within the given timeout.

### Steps to reproduce:
1. Make sure your AP is turned off. (you can use smartphone hotspot AP too)
2. Plug Raspberry Pi Pico W to USB.
3. After 5 second, Turn your AP on.

### Test code
I created test code to check cyw43_arch_wifi_connect_timeout_ms function behavior. [Full project](https://github.com/raspberrypi/pico-sdk/files/10067783/RasPiPicoSDK_wifi_con_test-main.zip)(on macOS open in VS Code run ./build.sh)

```C
#include <stdio.h>

#include "pico/stdlib.h"
#include "pico/cyw43_arch.h"

char ssid[] = "A Network";
char pass[] = "A Password";
const uint32_t timeout = 1000 * 60 * 10;// 10min

int main() {
    stdio_init_all();

    if (cyw43_arch_init_with_country(CYW43_COUNTRY_UK)) {
        printf("failed to initialise\n");
        return 1;
    }
    printf("initialised\n");

    cyw43_arch_enable_sta_mode();

    int ret = cyw43_arch_wifi_connect_timeout_ms(ssid, pass, CYW43_AUTH_WPA2_AES_PSK, timeout);
    if (ret) {
        printf("failed to connect, resp: %d\n", ret);
        return 1;
    }

    printf("connected\n");
}
```
if AP is there at the beginning, It connects within 10 secs.

### Log
```
connect status: joining
connect status: no ip
connect status: link up
connected
```

if AP isn't there at the beginning It fails within 10 secs .
```
connect status: joining
connect status: network fail
failed to connect, resp: -2
```
`PICO_ERROR_GENERIC = -2 `

### Conclusion
This is how it's called. (I'm sure you already know it)

https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_cyw43_arch/cyw43_arch.c

```
cyw43_arch_wifi_connect_timeout_ms -> cyw43_arch_wifi_connect_until -> 
{
   cyw43_arch_wifi_connect_async -> called once
   checks link status with cyw43_tcpip_link_status and cyw43_arch_poll
}
```

I think the way its checking connection is wrong. 

Since link fails, I think, cyw43_arch_wifi_connect_async must be called again periodically.

this could be related to #917 but I'm not sure.

peterharperuk:
Currently, if the network can't be found we fail almost immediately. Also fix error codes.

Fixes #1054


peterharperuk:
Note: This is based on async_context as it hits the same code https://github.com/raspberrypi/pico-sdk/pull/1177
I'll rebase on develop when that's submitted.

PR #1192, opened 2023-01-25
CI Improvements. Add multi-gcc workflow

liamfraser:
Tests every ARM gcc available

### DIFF
--- a/.github/workflows/scripts/generate_multi_gcc_workflow.py
+++ b/.github/workflows/scripts/generate_multi_gcc_workflow.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+import os
+from collections import OrderedDict
+import subprocess
+import re
+
+toolchain_dir = "/opt/arm"
+toolchains = os.listdir(toolchain_dir)
+
+gcc_versions = OrderedDict()
+
+for toolchain in toolchains:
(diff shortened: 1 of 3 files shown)


kilograham (approved):

lurch:
Nice :+1: 
However they **all** do `-DPICO_BOARD=pico_w`. Perhaps the very last `gcc_version` should _additionally_ do builds with `-DPICO_BOARD=pico` as there may be some code-paths that are only compiled for `pico` and not `pico_w`? (e.g. anything to do with DEFAULT_LED)

.github/workflows/scripts/generate_multi_gcc_workflow.py
```diff
@@ -0,0 +1,71 @@
...
+
+import os
+from collections import OrderedDict
+import subprocess
+import re
+
+toolchain_dir = "/opt/arm"
+toolchains = os.listdir(toolchain_dir)
+
+gcc_versions = OrderedDict()
```
lurch, line 11: Minor nitpick, but this doesn't need to be an OrderedDict, because `gcc_versions_sorted` gets constructed as an OrderedDict :wink:   (but of course leaving it as an OrderedDict doesn't hurt anything.)

PR #1194, opened 2023-01-25
update multi-gcc workflow

kilograham:


lurch:
I see that these **all** do `-DPICO_BOARD=pico_w` - should some of them do `-DPICO_BOARD=pico` as there may be some code-paths that are only compiled for `pico` and not `pico_w`? (e.g. anything to do with DEFAULT_LED)

lurch:
Looking at #1192 I see that this `multi-gcc.yml` is itself autogenerated, in which case it's probably worth making these same fixes to the `generate_multi_gcc_workflow.py` script too.

kilograham:
ah; crap

PR #1201, opened 2023-01-26
Move more code from pico_cyw43_arch to pico_cyw43_driver

kilograham:
Basically the integration code (cyw43_config.h and related implementations) are now in the driver. cyw43_arch now just has
  * async_context creation per CYW43_ARCH_TYPE
  * pre-existing cyw43_arch methods for connect etc.


src/rp2_common/pico_runtime/runtime.c
```diff
@@ -238,7 +238,7 @@ __attribute((weak)) int _times(struct tms *tms) {
 #if CLOCKS_PER_SEC >= 1000000
     tms->tms_utime = to_us_since_boot(get_absolute_time()) * (CLOCKS_PER_SEC / 1000000);
 #else
-    tms->tms_utime = to_us_since_boot(get_absolute_time()) / (1000000 / CLOCKS_PER_SEC);
+    tms->tms_utime = (clock_t)to_us_since_boot(get_absolute_time()) / (1000000 / CLOCKS_PER_SEC);
```
peterharperuk, line 241: Do you need to cast the call to to_us_since_boot above as well?

kilograham: yes; i've picked up your fix now

peterharperuk (approved):

peterharperuk:
Nice tidyup!

PR #1202, opened 2023-01-27
Tweaks to section-attribute stuff in platform.h

lurch:
I _think_ this is right, based on a careful reading of the comments and the code, but I can't be 100% sure.
I also think that the `\param group ...` stuff in the `__uninitialized_ram` section is now wrong, but I'm not sure what it should be changed to.


kilograham (approved):
thanks

PR #1204, opened 2023-01-27
add pico_rand to kitchen_sink

kilograham:


test/kitchen_sink/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ target_link_libraries(kitchen_sink_libs INTERFACE
     pico_mem_ops
     pico_multicore
     pico_platform
+    pico_rand
```
lurch, line 38: `pico_rand` now appears both here and also on line 44

test/kitchen_sink/kitchen_sink.c
```diff
@@ -41,6 +41,7 @@
 #include "pico/malloc.h"
 #include "pico/multicore.h"
 #include "pico/printf.h"
+#include "pico/rand.h"
```
lurch, line 44: `#include "pico/rand.h"` now appears both here and also on line 51

kilograham: crap; ok; tsk people not keeping thing is alphabetical order :-(

Issue #1208, opened 2023-01-29
Automated build on Windows & Mac


flaviut:
At the moment, automated builds of this are only done for Linux. It'd be nice if these were also done on Windows & Mac.

kilograham:
we have them now

PR #1209, opened 2023-01-29
Add windows CI

flaviut:
Adds some basic automated builds on Windows, #1208


flaviut:
re. "make manual for now", you may have other reasons, but I believe that CI is free on github for public repos, and there's no security concerns here since there's no secrets and this entirely runs on github's infrastructure.

PR #1211, opened 2023-01-29
Rework build targets to mirror libraries as _headers versions, and fix dependency issues

kilograham:
Rework build targets such that:

1. Pico SDK libraries now have the correct dependencies on other SDK libraries
2. Pico SDK libraries all have _headers variants to include the just headers. This may facilitate building user STATIC libraries without pulling in SDK code, though care will still need to be taken w.r.t. values of #defines
3. Make sure the _headers versions also have the correct dependencies

Note: There are a few exceptions to 1. for some non code libraries like pico_standard_link and pico_cxx_options

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,6 +27,29 @@ function(pico_add_subdirectory subdir)
     pico_promote_common_scope_vars()
 endfunction()
 
+# takes dependencies after the target
+function(pico_mirrored_target_link_libraries TARGET SCOPE)
+    if (${ARGC} LESS 3)
+        message(FATAL_ERROR "expected a target, scope and at least one dependency")
+    endif()
+    if (NOT TARGET ${TARGET}_headers)
+        message(FATAL_ERROR "${TARGET} does not have headers")
+    endif()
+    # library should depend on its own header
+    target_link_libraries(${TARGET} ${SCOPE} ${TARGET}_headers)
+    foreach(DEPENDENCY IN LISTS ARGN)
+        if (DEPENDENCY MATCHES ".*_headers")
+            message(FATAL_ERROR "should not use headers with pico_mirrored_target_link_libraries")
+        endif()
+        # note, it would be nice to only add the dependency if it exists, but we do
+        # not necessarily add libraries in reverse dependency order, so we do this
+        # unconditionally, so this function should only be passed dependencies that
+        # have headers, or link will fail with a missing library -lfoo_headers
+        target_link_libraries(${TARGET}_headers ${SCOPE} ${DEPENDENCY}_headers)
+        target_link_libraries(${TARGET} ${SCOPE} ${DEPENDENCY})
+    endforeach()
+endfunction()
+
 # add a link option to wrap the given function name; i.e. -Wl:wrap=FUNCNAME for gcc
 function(pico_wrap_function TARGET FUNCNAME)
     target_link_options(${TARGET} INTERFACE "LINKER:--wrap=${FUNCNAME}")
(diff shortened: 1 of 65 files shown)


src/CMakeLists.txt
```diff
@@ -42,18 +65,34 @@ function(pico_add_map_output TARGET)
...
 function(pico_add_impl_library target)
     add_library(${target} INTERFACE)
     string(TOUPPER ${target} TARGET_UPPER)
-    target_compile_definitions(${target} INTERFACE LIB_${TARGET_UPPER}=1)
+    if (${ARGC} GREATER 1)
+        if (NOT "${ARGV1}" STREQUAL "NOFLAG")
+            message(FATAL_ERROR "Unknown parameter ${ARGV1}")
+        endif()
+    else()
+        target_compile_definitions(${target} INTERFACE LIB_${TARGET_UPPER}=1)
```
peterharperuk, line 86: You might have to explain to me what these LIB_* defs are for

src/CMakeLists.txt
```diff
@@ -66,7 +105,7 @@ macro(pico_simple_hardware_headers_target NAME)
         target_include_directories(hardware_${NAME}_headers INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
         target_link_libraries(hardware_${NAME}_headers INTERFACE pico_base_headers)
         if (NOT PICO_NO_HARDWARE)
-            target_link_libraries(hardware_${NAME}_headers INTERFACE hardware_structs hardware_claim)
+            target_link_libraries(hardware_${NAME}_headers INTERFACE hardware_structs hardware_claim_headers)
```
peterharperuk, line 108: Should this be hardware_structs_headers? See other examples

src/CMakeLists.txt
```diff
@@ -80,13 +119,17 @@ macro(pico_simple_hardware_headers_only_target NAME)
...
+        add_library(hardware_${NAME}_headers INTERFACE)
 
-        target_include_directories(hardware_${NAME} INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
-        target_link_libraries(hardware_${NAME} INTERFACE pico_base_headers)
+        # a headers only target should still have an explicit _headers library for consistency
+        target_include_directories(hardware_${NAME}_headers INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
+        target_link_libraries(hardware_${NAME}_headers INTERFACE pico_base_headers)
         if (NOT PICO_NO_HARDWARE)
-            target_link_libraries(hardware_${NAME} INTERFACE hardware_structs)
+            target_link_libraries(hardware_${NAME}_headers INTERFACE hardware_structs)
```
peterharperuk, line 128: Should this be hardware_structs_headers?

peterharperuk (approved):
Some questions, but basically looks ok

src/CMakeLists.txt
kilograham: The idea was to allow inter-dependent libraries to be aware of each other, and to change/add behavior or guard includes

so you can do 

```
#if LIB_PICO_FOO
#include "pico/foo.h"
#endif
```

as well - you can search on them; they are using a lot in the `stdio` stuff and with some of the swapable libraries. I'm kind of leaning in favor of making these more explicit than implicit, and thus have also started adding the new `NOFLAG` option to avoid setting them... of course it is hard to know who has taken advantage of them, but probably few if any beyond SDK and examples/extras/playground

Note: you might think (as did I) the the LIB_ flag therefore belongs on the header, however in practice we only check them in the C files

src/CMakeLists.txt
kilograham: yeah, i almost did - the libraries are identical

kilograham:
> Apologies for the stupid question, but is this all "internal to pico-sdk only", i.e. it has no impact on anybody _using_ the SDK in their projects?

The intention is for it to have no impact

peterharperuk:
I approve of this.

PR #1213, opened 2023-01-30
Add stdio_set_chars_available_callback

peterharperuk:
There's currently no way to be notified in a "stdio" agnostic way whether there's an incoming character available. You can poll with getchar_timeout_us, but that's far from ideal.

Add a method that takes a callback to notify that a character might be available.

### DIFF
--- a/src/rp2_common/pico_stdio/include/pico/stdio.h
+++ b/src/rp2_common/pico_stdio/include/pico/stdio.h
@@ -109,6 +109,14 @@ int putchar_raw(int c);
  */
 int puts_raw(const char *s);
 
+/*! \brief get notified when there are input characters available
+ * \ingroup pico_stdio
+ *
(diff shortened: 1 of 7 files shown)


kilograham (commented):
Actually now i think about it, let's have `PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK` #ifdef out all of the function implementations and static vars too, and then add a
`PICO_STDIO_UART_SUPPORT_CHARS_AVAILABLE_CALLBACK` #ifdef too
because it is a bunch (small amount) of new code that is taking up space even if not used

peterharperuk:
Do you want it disabled by default? Currently I made it enabled by default.

kilograham:
i think enabled is fine, people may just want to turn it off

peterharperuk:
Done as requested

kilograham (approved):

PR #1221, opened 2023-02-05
Allow external pico-sdk-tools package

ndabas:
This allows us to create a package named `pico-sdk-tools` somewhere in the CMake search path, and have that package provide pre-compiled binaries for pioasm and elf2uf2. This is useful, especially on Windows, where a native compiler might not be available.


kilograham (approved):
LGTM thanks

Issue #1078, opened 2022-10-26
"cw43 warn" causing deadlock on freertos smp


p8a:
I'm running into a deadlock on raspberry pico w and freertos smp kernel due to CYW43 WARN printing functions using vprintf - this is the stack:
```
Thread 1 received signal SIGINT, Interrupt.
0x100125c8 in vPortRecursiveLock (uxAcquire=1, pxSpinLock=0xd000013c, ulLockNum=1) at /Users/p8a/data/test/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h:192
192	                while ( __builtin_expect( !*pxSpinLock, 0 ) );
(gdb) bt
#0  0x100125c8 in vPortRecursiveLock (uxAcquire=1, pxSpinLock=0xd000013c, ulLockNum=1) at /Users/p8a/data/test/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h:192
#1  vTaskEnterCritical () at /Users/p8a/data/test/FreeRTOS-Kernel/tasks.c:5367
#2  0x1001313c in xTaskGetSchedulerState () at /Users/p8a/data/test/FreeRTOS-Kernel/tasks.c:5041
#3  0x10010804 in xEventGroupWaitBits (xEventGroup=0x2000c5b8 <ucHeap+12>, uxBitsToWaitFor=131584, xClearOnExit=xClearOnExit@entry=1, xWaitForAllBits=xWaitForAllBits@entry=0, xTicksToWait=xTicksToWait@entry=4294967295) at /Users/p8a/data/test/FreeRTOS-Kernel/event_groups.c:333
#4  0x100106a4 in vPortLockInternalSpinUnlockWithWait (pxLock=pxLock@entry=0x200003d0 <print_mutex>, ulSave=<optimized out>) at /Users/p8a/data/test/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/port.c:788
#5  0x200000e8 in mutex_enter_blocking (mtx=mtx@entry=0x200003d0 <print_mutex>) at /Users/p8a/pico/pico-sdk/src/common/pico_sync/mutex.c:44
#6  0x1000417a in stdout_serialize_begin () at /Users/p8a/pico/pico-sdk/src/rp2_common/pico_stdio/stdio.c:52
[...]

PR #1224, opened 2023-02-06
stdio hardening + new mutex API

kilograham:
* Harden stdio_usb and stdio in general against deadlocks which could otherwise result from doing printfs from within IRQs
* Add a test for the above
* Add mutex_try_enter_block_until API.
* Make best_effort_wfe_or_timeout not use alarms if called from within IRQ

### DIFF
--- a/src/rp2_common/pico_stdio_usb/stdio_usb.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb.c
@@ -124,7 +124,6 @@ static void stdio_usb_out_chars(const char *buf, int length) {
 }
 
 int stdio_usb_in_chars(char *buf, int length) {
-    uint32_t owner;
     // note we perform this check outside the lock, to try and prevent possible deadlock conditions
     // with printf in IRQs (which we will escape through timeouts elsewhere, but that would be less graceful).
     //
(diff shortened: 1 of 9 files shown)


kilograham:
fixes #1039

kilograham:
fixes #1078 (probably)

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -127,19 +125,28 @@ static void stdio_usb_out_chars(const char *buf, int length) {
...
     int rc = PICO_ERROR_NO_DATA;
-    if (stdio_usb_connected() && tud_cdc_available()) {
-        int count = (int) tud_cdc_read(buf, (uint32_t) length);
-        rc =  count ? count : PICO_ERROR_NO_DATA;
-    } else {
-        // because our mutex use may starve out the background task, run tud_task here (we own the mutex)
-        tud_task();
-    }
-    mutex_exit(&stdio_usb_mutex);
+//    if (stdio_usb_connected() && tud_cdc_available()) {
```
peterharperuk: did you mean to leave this code commented out

peterharperuk: ah - I see the second commit now

kilograham: no, but i had already fixed that!

peterharperuk (approved):

PR #1227, opened 2023-02-06
Improve SPI set-up: Don't change the config whilst it is enabled

kilograham:


kilograham:
cc @daveythacher

peterharperuk (approved):
Looks good to me!

PR #1228, opened 2023-02-06
Fix windows build issue

peterharperuk:
On Windows I see an error that the option "print-prog-name" doesn't exist


peterharperuk:
grrr, why are there 2 commits?!

kilograham:
i still can't understand how you could possibly see this error

kilograham:
I think this is a reasonable change... i'm not at all sure why people are picking up non GCC things here (or binutils frankly) this should always be the GCC cross-compiler...

kilograham (approved):

PR #1236, opened 2023-02-09
Setup clock pin when starting SPI comms.

### LINKED ISSUE #1141: cyw43_driver: explicitly init clock pin?
Looking at https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/cyw43_driver/cyw43_bus_pio_spi.c#L194, should this function also have a `gpio_set_function(CLOCK_PIN, bus_data->pio_func_sel);`?
Because the clock pin is shared with adc, so could have been left in that adc config. (i just tripped over that right now)

peterharperuk:
Because the clock pin (29) also has to be used to read vsys, its properties might be changed.
So reset when starting a spi transaction.

Fixes #1141


kilograham (commented):
should we not perhaps do that before pulling CS low?

peterharperuk:
Can we NOT merge this yet - I want to test some more.

peterharperuk:
Seems ok

kilograham (approved):

PR #1252, opened 2023-02-13
Add some configuration for VBUS / VSYS

### LINKED ISSUE #1222: Unable to measure VSYS/ADC3/GPIO29 with WIFI
I have a setup with Pico W + pimoroni lipo shim.
I am trying to run it on batteries, measure some sensors and submit the data via wifi.
I am also trying to check battery status from time to time to get an idea when charging is needed. There is a micropython example [here](https://github.com/pimoroni/pimoroni-pico/blob/main/micropython/examples/pico_lipo_shim/battery_pico.py), but I found that this also does not work for Pico W.

I also looked on this forum thread where others also struggle with doing this with python in combination with using WIFI.
https://forums.raspberrypi.com/viewtopic.php?p=2078340#p2078340  (my username = Darkness)

I am trying to accomplish this using the SDK.

From the datasheet:
`GPIO29 OP/IP wireless SPI CLK/ADC mode (ADC3) to measure VSYS/3
GPIO25 OP wireless SPI CS - when high also enables GPIO29 ADC pin to read VSYS
WL_GPIO2 IP VBUS sense - high if VBUS is present, else low

Due to pin limitations, some of the wireless interface pins are shared. The CLK is shared with VSYS monitor, so only
when there isn’t an SPI transaction in progress can VSYS be read via the ADC. The Infineon CYW43439 DIN/DOUT and
IRQ all share one pin on the RP2040. Only when an SPI transaction isn’t in progress is it suitable to check for IRQs. The
interface typically runs at 33MHz`

So it is clear that due to pins being shared with the wifi chip, it is not possible to measure VSYS using GPIO29+GPIO25.

* is there a way to know if an SPI transaction is in progress or not to make sure an ADC3 measurement can be done?
* Is there a reliable way to repeatedly enable and disable wifi to make GPIO29 and GPIO25 available? I tryied calling **cyw43_arch_deinit()** but afterwards **cyw43_arch_init()** seems to fail and hang the pico.
* is **cyw43_arch_gpio_get(2);** the correct way to read WL_GPIO2/VBUS sense? pico_w.h does not contain a definition for these pins, unlike the **#define CYW43_WL_GPIO_LED_PIN 0**

This seems like a very common use case, so an example would be very useful for the community.

peterharperuk:
The process for reading VBUS and VSYS on Pico and Pico W is different. It's hard to write code that compiles for both devices. Add some configuration to the board files.

Fixes #1222

### DIFF
--- a/src/boards/include/boards/pico.h
+++ b/src/boards/include/boards/pico.h
@@ -81,4 +81,14 @@
 #define PICO_RP2040_B0_SUPPORTED 1
 #endif
 
+// Pin get VBUS
+#ifndef PICO_VBUS_PIN
+#define PICO_VBUS_PIN 24
+#endif
+
+// Pin used to monitor VSYS using ADC
+#ifndef PICO_VSYS_PIN
+#define PICO_VSYS_PIN 29
+#endif
+
 #endif
--- a/src/boards/include/boards/pico_w.h
+++ b/src/boards/include/boards/pico_w.h
@@ -99,4 +99,19 @@
 #define CYW43_WL_GPIO_LED_PIN 0
 #endif
 
+// CYW43 GPIO to get VBUS
+#ifndef CYW43_WL_GPIO_VBUS_PIN
+#define CYW43_WL_GPIO_VBUS_PIN 2
+#endif
+
+// VSYS pin is shared with CYW43
+#ifndef CYW43_USES_VSYS_PIN
+#define CYW43_USES_VSYS_PIN 1
+#endif
+
+// Pin used to monitor VSYS using ADC
+#ifndef PICO_VSYS_PIN
+#define PICO_VSYS_PIN 29
+#endif
+
 #endif


src/boards/include/boards/pico_w.h
```diff
@@ -99,4 +99,22 @@
 #define CYW43_WL_GPIO_LED_PIN 0
 #endif
 
+// CYW43 GPIO to get VBUS
+#ifndef CYW43_WL_GPIO_VBUS_PIN
+#define CYW43_WL_GPIO_VBUS_PIN 2
```
lurch, line 104: Slight inconsistency in the order-of-words in `PICO_VBUS_GPIO_PIN` and `CYW43_WL_GPIO_VBUS_PIN` but I dunno if that's deliberate and/or if it matters? :wink:

src/boards/include/boards/pico.h
```diff
@@ -81,4 +81,19 @@
...
+#endif
+
+// GPIO pin used to monitor VSYS
+#ifndef PICO_VSYS_GPIO_PIN
+#define PICO_VSYS_GPIO_PIN 29
+#endif
+
+// ADC to use to monitor VSYS
+#ifndef PICO_VSYS_ADC_NUM
+#define PICO_VSYS_ADC_NUM 3
```
lurch, line 96: I wonder if `PICO_VSYS_ADC_CHANNEL` might be less ambiguous, as a name ending in `_NUM` _might_ be misinterpreted to mean "count of"? (IYSWIM!)  :shrug:

peterharperuk: Looking at the api "input" might be better.

src/boards/include/boards/pico_w.h
peterharperuk: I've dropped "GPIO" from the pin defs (except for the WL one). I think that's consistent?

src/boards/include/boards/pico.h
```diff
@@ -81,4 +81,19 @@
...
+
+// Pin used to monitor VSYS
+#ifndef PICO_VSYS_PIN
+#define PICO_VSYS_PIN 29
+#endif
+
+// ADC to use to monitor VSYS
+#ifndef PICO_VSYS_ADC_INPUT
+#define PICO_VSYS_ADC_INPUT 3
+#endif
```
lurch, line 97: Isn't there a 1:1 mapping between GPIO pins and ADC inputs? In which case perhaps having separate defines for both of them is a bit overkill, and it'd be worth having one value calculated from the other one instead?

src/boards/include/boards/pico_w.h
```diff
@@ -99,4 +99,22 @@
...
+#endif
+
+// VSYS pin is shared with CYW43
+#define CYW43_USES_VSYS_PIN 1
```
lurch, line 108: Apologies if this is a stupid question, but should this also be inside an `#ifndef` ?

Alternatively, given that this is the header-file specifically for Pico W, and there's no way to change which pins the CYW43 is connected to, perhaps _none_ of these settings should be inside an `#ifndef` ? :shrug:

src/boards/include/boards/pico.h
peterharperuk: Yes, there's a one to one mapping on Pico boards. I don't think that has to be the case on other boards though. I'm not sure we infer the adc from the pin elsewhere, but I'm prepared to be corrected.

src/boards/include/boards/pico_w.h
peterharperuk: I don't know why we use these ifndef in this file - it does seem pointless. Maybe @kilograham knows.

src/boards/include/boards/pico.h
lurch: IIRC this is actually a chip-level (RP2040) thing rather than a board-level (Pico) thing?

src/boards/include/boards/pico_w.h
lurch: For some settings (e.g. the `DEFAULT_UART` stuff) it _does_ make sense to "ifndef it", because  the UART is available on multiple pins, and so the user might choose to have **their** UART wired up to different pins, and they're able to override it by specifying their own values before including the board-header-file. But for other settings (e.g. anything to do with CYW43) I'm not sure if it makes sense to "ifndef it", because there's no way for the user to override the values. All IMHO obviously!

kilograham: yeah we do by convention always `#ifndef` partly just because, but also because a user can take one board definition and make it a basis for something else

peterharperuk:
@kilograham Does this look ok? I might use this to add some handling to Micropython if it's likely to be merged

kilograham (approved):

PR #1253, opened 2023-02-13
Set PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS

### LINKED ISSUE #1210: Allow changing PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS from the command line
It's useful to be able to switch to SDIO USB with cmake variables. But you can't set PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS  so often miss the first few log lines. Add an option to set this so you don't have to change the code to set it.

peterharperuk:
If PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS is passed to cmake, use it to set the build macro of the same name.

Fixes #1210

### DIFF
--- a/src/rp2_common/pico_stdio_usb/CMakeLists.txt
+++ b/src/rp2_common/pico_stdio_usb/CMakeLists.txt
@@ -18,4 +18,10 @@ if (TARGET tinyusb_device_unmarked)
     target_link_libraries(pico_stdio_usb INTERFACE
         tinyusb_device_unmarked
     )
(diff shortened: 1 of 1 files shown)


kilograham (changes requested):
nice, should probably add a `PICO_CMAKE_CONFIG` line tho

peterharperuk:
Added a PICO_CMAKE_CONFIG for this.

kilograham (approved):

PR #1254, opened 2023-02-13
Add missing busy_wait_ms functino to host platform

kilograham:


src/host/hardware_timer/timer.c
```diff
@@ -13,21 +13,31 @@
...
 void PICO_WEAK_FUNCTION_IMPL_NAME(busy_wait_us)(uint64_t delay_us) {
     absolute_time_t t;
     update_us_since_boot(&t, time_us_64() + delay_us);
     busy_wait_until(t);
 }
 
+PICO_WEAK_FUNCTION_DEF(busy_wait_ms)
+void PICO_WEAK_FUNCTION_IMPL_NAME(busy_wait_ms)(uint32_t delay_ms) {
+#if defined(__unix__) || defined(__APPLE__)
+    usleep(delay_ms * 1000);
```
lurch, line 35: https://man7.org/linux/man-pages/man3/usleep.3.html says that `usleep` fails if "usec is greater than or equal to 1000000", which I guess means this `busy_wait_ms` function will fail if passed a `delay_ms` value >= 1000 ?
Is that something we ought to check for, or are we not really that concerned about the correct operation of "host platform mode"? :man_shrugging:

PR #1257, opened 2023-02-14
Run pre-installed pioasm and elf2uf2 if found in system search path

lurk101:


lurch:
You seem to have `progran` in many places - was that supposed to be `program` ?

lurk101:
> You seem to have `progran` in many places - was that supposed to be `program` ?

No longer used in current patch

lurch:
I'm afraid I don't know much about CMake, but how does this interact with #1221 ?

lurk101:
> I'm afraid I don't know much about CMake, but how does this interact with #1221 ?

I could be wrong but I believe #1221 applies only to Windows. It doesn't seem to have any effect in Linux.

PR #1259, opened 2023-02-14
Add board definition for upcoming Pololu 3pi+ 2040 Robot

pdg137:
This is to support a new Pololu robot that will be released soon.  We have tested this file with C and Micropython, and it is working for us!


pdg137:
The robot is now up for sale:

https://www.pololu.com/product/5001

Please let me know if there's anything we need to do to get this merged!

kilograham:
hadn't seen it - please add our copyright, and whilst #pragma once is fine in general, perhaps match the style of everything other board header and use #ifndef

pdg137:
Okay, I updated it with something like what's at the top of the other ones.  Please let me know if you need anything else.

kilograham (approved):

Issue #1024, opened 2022-09-13
No support for crystals other than 12MHz


arg08:
The current runtime_init() invokes clocks_init() which is entirely hard-wired for crystal oscillator at 12MHz (and sets up PLLs assuming 12MHz base frequency), such that any board with a higher frequency crystal/external oscillator to XIN will fail before the user application gets a chance to reconfigure it.

Issue #748 requests support to heavily customise this startup behaviour (for example, if there is no crystal at all or it is undersirable to use it for power reasons).   However, there is another class of applications that do want all the standard initialisation but simply have a crystal of different frequency (or an external clock wired to XIN).  This could usefully be supported by allowing the board file to specify the crystal frequency.

There is already a symbol XOSC_MHZ defined in platform_defs.h, but this is almost entirely ignored in the setup code in clocks_init().

If the code in clocks_init() were updated to respect XOSC_MHZ, it could be overridden in the board file for boards with non-standard crystals.

One question is what setup is expected in respect of the PLLs.  The minimum requirement is that the PLLs are not set up to out-of-range frequencies, causing a crash before the user program has a chance to do further setup.  The ideal would be to set the desired PLL output frequencies for any crystal, but the PLL configuration tool is currently in the form of a python program not readily invoked during the build process.  A compromise would be to allow the board file to define the PLL parameters, and for the code in clocks_init() to set the PLLs to x1 as a fallback if XOSC_MHZ != 12 and custom PLL parameters have not been defined.

arg08:
> Is the ideal solution to this to run vcocalc.py during the build to get the PLL configuration given the crystal frequency?
> 

Ideal in one sense, but making Python a dependency of the build system may not be considered justified for just this feature.

> I might need this explaining to me...
> 
> > the code in clocks_init() to set the PLLs to x1 as a fallback

I simply meant that if the init code doesn't have access to sensible PLL parameters (by calculating them, or having them passed in through defines in the board file) then it can simply use the crystal frequency as clk_sys as that's always safe, and the application can sort things out later.   Probably that's more sensibly done by bypassing the PLLs rather than literally setting them to a x1 configuration.

peterharperuk:
@andygpz11 I might need your advice on this

andygpz11:
Sure @peterharperuk
Actually I'll assign it to me.

kilograham:
[...]

Issue #951, opened 2022-08-01
vcocalc.py script should consider PLL configurations with REFDIV > 1


alastairpatrick:
This is a feature request. You can assign it to me. I have already written the code and will follow this up with a pull request if there are no objections.

At present, the vcocalc.py script only considers PLL configurations with REFDIV = 1. Sometimes it is beneficial to use PLL configurations with REFDIV > 1. For example, in an audio application using the common 44.1kHz sample rate, one might desire to use a system clock of 88.2MHz or 176.4MHz. At present, vcocalc.py cannot satisfy that exactly:
```
> python vcocalc.py 88.2
Requested: 88.2 MHz
Achieved: 88.28571428571429 MHz
FBDIV: 103 (VCO = 1236 MHz)
PD1: 7
PD2: 2
```
In this case, it would help if it considered REFDIV = 2, which yields an exact match:
```
> python vcocalc.py 88.2
Requested: 88.2 MHz
Achieved: 88.2 MHz
REFDIV: 2
FBDIV: 147 (VCO = 882.0 MHz)
PD1: 5
PD2: 2
```
As an aside, if this change is made, it might be worth adding a variant of set_sys_clock_pll() taking a ref_div parameter.

lurch:
ping @liamfraser

lurch:
This can be closed, as it's included in SDK 1.5.0 ?

kilograham:
fixed in 1.5.0

PR #1272, opened 2023-02-21
Allow support of crystals other than 12MHz (#1024)

andygpz11:
Improve the current hardcoded PLL and clock set-up


andygpz11:
I have just pushed some changes for comment which I think address the XOSC and system frequency related hard-codings in the SDK.  
Making set_sys_clock_pll() and check_sys_clock_khz() properly general purpose wasn’t possible without changing their APIs so, for now, I left them: This is because neither of them currently support ’refdiv’ setting other than 1.  I have coned a new #define to keep things in step along with a tripwire to enforce this.
Having used vcocalc.py it is _not_ clear to me that refdiv settings other than 1 can sensibly be used anyway.
A few ‘pico-examples’ will need updating to use the definitions I have added to platform_defs.h

src/rp2040/hardware_regs/include/hardware/platform_defs.h
```diff
@@ -37,9 +37,39 @@
...
+#define SYS_CLK_MHZ _u(125)
+#endif
+
+// PICO_CONFIG: XOSC_MHZ, The crystal oscillator frequency in MHz, type=int, default=12, advanced=true, group=hardware_base
 #ifndef XOSC_MHZ
 #define XOSC_MHZ _u(12)
+#endif
+
+ // Check for standard set-up
+#if (XOSC_MHZ == 12 && SYS_CLK_MHZ == 125)
```
kilograham, line 54: this does not belong here; better in hardware/clocks.h

src/rp2040/hardware_regs/include/hardware/platform_defs.h
```diff
@@ -37,9 +37,39 @@
...
+#define PLL_SYS_POSTDIV2                    2
+
+#define PLL_USB_VCO_FREQ_MHZ                1200
+#define PLL_USB_POSTDIV1                    5
+#define PLL_USB_POSTDIV2                    5
+
+// Note: Do not trivially change this
+#define PLL_COMMON_REFDIV                   1
+#else
+#error Use vcocalc.py to calculate correct values for the revised XOSC and/or system clock frequencies and define here.
```
kilograham, line 72: this should only be a #error if the six/seven things aren't defined (we do not expect people to edit this header, we expect them to set them in their compile_defintions or via a board header)

also,

separate out the definition checks  USB since it seems to be independent of SYS_CLK_MHZ, i.e. two separate checks for 3 defines, vs 1 for 6.

also presumably PLL_COMMON_REFDIV maybe should be just #ifndef-ed above all of this and defaulted to 1.

src/rp2_common/pico_stdlib/stdlib.c
```diff
@@ -64,23 +64,24 @@ void set_sys_clock_pll(uint32_t vco_freq, uint post_div1, uint post_div2) {
...
-                        48 * MHZ);
+                        USB_CLK_MHZ * MHZ,
+                        USB_CLK_MHZ * MHZ);
     }
 }
 
-bool check_sys_clock_khz(uint32_t freq_khz, uint *vco_out, uint *postdiv1_out, uint *postdiv_out) {
-    uint crystal_freq_khz = clock_get_hz(clk_ref) / 1000;
+bool check_sys_clock_khz(uint32_t freq_khz, uint *vco_out, uint *postdiv1_out, uint *postdiv2_out) {
+    static_assert(PLL_COMMON_REFDIV == 1, " This function implicitly assumes a 'refdiv' value of 1");
```
kilograham, line 73: um, if you've made a define for it, why not use it?

kilograham (changes requested):

src/rp2040/hardware_regs/include/hardware/platform_defs.h
```diff
@@ -37,9 +37,39 @@
...
+#define PLL_SYS_VCO_FREQ_MHZ                1500
+#define PLL_SYS_POSTDIV1                    6
+#define PLL_SYS_POSTDIV2                    2
+
+#define PLL_USB_VCO_FREQ_MHZ                1200
+#define PLL_USB_POSTDIV1                    5
+#define PLL_USB_POSTDIV2                    5
+
+// Note: Do not trivially change this
+#define PLL_COMMON_REFDIV                   1
```
lurch, line 70: https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_pll/pll.c#L55 seems to suggest that each PLL has its own REFDIV, rather than there being a single "global" REFDIV? I.e. should there be separate defines for `PLL_SYS_REFDIV` and `PLL_USB_REFDIV` ? :man_shrugging:

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -145,63 +148,56 @@ void clocks_init(void) {
...
+                    USB_CLK_MHZ * MHZ);
 
-    // CLK RTC = PLL USB (48MHz) / 1024 = 46875Hz
+    // CLK RTC = PLL USB 48MHz / 1024 = 46875Hz
     clock_configure(clk_rtc,
                     0, // No GLMUX
                     CLOCKS_CLK_RTC_CTRL_AUXSRC_VALUE_CLKSRC_PLL_USB,
-                    48 * MHZ,
+                    USB_CLK_MHZ * MHZ,
                     46875);
```
lurch, line 192: Apologies for the naive question, but should this value also get its own `#define`? Or is this a value that can't be (sensibly) changed?

src/rp2040/hardware_regs/include/hardware/platform_defs.h
liamfraser: Each pll does have it's own reference divider. In practice the reference just has to be in a certain range and can be divided if it is not.

src/rp2040/hardware_structs/include/hardware/structs/clocks.h
```diff
@@ -29,7 +29,7 @@ enum clock_index {
     clk_gpout1,         ///< GPIO Muxing 1
     clk_gpout2,         ///< GPIO Muxing 2
     clk_gpout3,         ///< GPIO Muxing 3
-    clk_ref,            ///< Watchdog and timers reference clock
+    clk_ref,            ///< PLLs, Watchdog and timers reference clock
```
liamfraser, line 32: This is actually not true. The PLLs are directly connected to the xosc as their reference clock.

src/rp2_common/hardware_clocks/clocks.c
andygpz11: @lurch I get your point but have no opinion on this.
@kilograham?

kilograham: i have no idea whu the RTC is set to 46875 so probably a define, with integer math if it has a sensible reason

andygpz11 (commented):
Latest version.  I think I have addressed all comments.
I also found that one of the asserts in `pll_init()` was non-functional due to an error of scaling and also that ` check_sys_clock_khz()` should _not_ use `clock_get_hz()` but the `XOSC_MHZ` define instead.
Apparently the PLL ref divs inputs are always connected to the output of the XOSC cell and therefore the (possible) multiplexing and division present in the clock switching is not applicable to them.

src/rp2040/hardware_regs/include/hardware/platform_defs.h
```diff
@@ -37,7 +37,15 @@
 
 #define PIO_INSTRUCTION_COUNT _u(32)
 
-// PICO_CONFIG: XOSC_MHZ, The crystal oscillator frequency in Mhz, type=int, default=12, advanced=true, group=hardware_base
+// For USB operation this *has* to be 48 MHz
+#define USB_CLK_MHZ _u(48)
```
kilograham, line 41: is this 100% true; i mean could you set it a little off

Also i wonder if we should specify this and SYS_CLK in khz

src/rp2_common/pico_stdlib/stdlib.c
```diff
@@ -64,23 +64,23 @@ void set_sys_clock_pll(uint32_t vco_freq, uint post_div1, uint post_div2) {
...
-                        48 * MHZ);
+                        USB_CLK_MHZ * MHZ,
+                        USB_CLK_MHZ * MHZ);
     }
 }
 
-bool check_sys_clock_khz(uint32_t freq_khz, uint *vco_out, uint *postdiv1_out, uint *postdiv_out) {
-    uint crystal_freq_khz = clock_get_hz(clk_ref) / 1000;
+bool check_sys_clock_khz(uint32_t freq_khz, uint *vco_out, uint *postdiv1_out, uint *postdiv2_out) {
+    uint reference_freq_khz = XOSC_MHZ * KHZ / PLL_COMMON_REFDIV;
```
kilograham, line 73: presumably this should be MHZ not KHZ

kilograham (changes requested):

src/rp2040/hardware_regs/include/hardware/platform_defs.h
kilograham: the first part of this is not resolved

src/rp2040/hardware_regs/include/hardware/platform_defs.h
kilograham: > Also i wonder if we should specify this and SYS_CLK in khz

I guess i meant, "let's specify these in khz" since i don't think we want to force people to use multiples of Mhz

kilograham (changes requested):

src/rp2040/hardware_regs/include/hardware/platform_defs.h
lurch: Also, given that these are ultimately derived from XOSC_MHZ, perhaps it makes sense to list them in this file _atfer_ `XOSC_MHZ` rather than before?

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,37 @@ extern "C" {
 #define KHZ 1000
 #define MHZ 1000000
 
+#if (XOSC_MHZ == 12)
+#ifndef PLL_COMMON_REFDIV
+// Software requires that the same 'reference divider' setting is used for both PLLs (although each has its own register).
+// In general, this is not an issue and 1 is the best setting - change with caution!
```
lurch, line 95: #951 provides an example of why you _might_ want to do this? :man_shrugging:

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,37 @@ extern "C" {
...
+ // Pll settings for 48 MHz USB clock are based on being able to get an exact
+ // 1200 MHz VCO, they _may_ need revising with a different reference frequency.
+#define PLL_USB_VCO_FREQ_MHZ                1200
+#define PLL_USB_POSTDIV1                    5
+#define PLL_USB_POSTDIV2                    5
+#else
+#error Use vcocalc.py to calculate correct values for the revised reference clock frequency and define here.
+#endif // XOSC_MHZ == 12
+
+ // Pll settings for standard system clock
```
lurch, line 108: Minor nitpick, but perhaps PLL should always be capitalised as thus?

liamfraser:
This looks good to me, apart from the btstack related commit

andygpz11:
> This looks good to me, apart from the btstack related commit
Thanks and agreed.  I found I could not push to my branch because there was another push that I hadn't got locally.
So I did a 'git pull --rebase' and then a push.  It was not enough for me to rebase my local branch from develop which is what I also tried.

src/rp2040/hardware_regs/include/hardware/platform_defs.h
```diff
@@ -37,19 +37,23 @@
...
-
-// PICO_CONFIG: SYS_CLK_MHZ, The system operating frequency in MHz, type=int, default=125, advanced=true, group=hardware_base
-#ifndef SYS_CLK_MHZ
-#define SYS_CLK_MHZ _u(125)
+// PICO_CONFIG: XOSC_KHZ, The crystal oscillator frequency in kHz, type=int, default=12000, advanced=true, group=hardware_base
+// NOTE:  The system and USB clocks are generated from the frequency using two PLLs.
+// If you override this define, or SYS_CLK_KHZ below, you will *also* need to add your own adjusted PLL set-up defines to
+// overide the defaults which live in src/rp2_common/hardware_clocks/include/hardware/clocks.h
+// Please see the comments there about calculating the new PLL setting values.
+#ifndef XOSC_KHZ
```
lurch, line 45: Given that this is a PICO_CONFIG value, renaming it from `XOSC_MHZ` to `XOSC_KHZ` feels like a backwards-incompatible change? :thinking: 
Perhaps adding
```
#ifdef XOSC_MHZ
#define XOSC_KHZ (XOSC_MHZ * KHZ)
#endif
```
is "enough" to fix that? :man_shrugging:

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,46 @@ extern "C" {
 #define KHZ 1000
 #define MHZ 1000000
 
+/// \tag::pll_settings[]
```
kilograham, line 92: note the "tag" here hoists the comment into the documentation, so it may be worth checking whether this new content make sense in that context (cc @lurch)

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,46 @@ extern "C" {
...
+/// \end::pll_settings[]
+
+#ifndef PLL_COMMON_REFDIV
+// Software requires that the same 'reference divider' setting is used for both PLLs (although each has its own register).
+#define PLL_COMMON_REFDIV                   1
+#endif
+
+// NOTE:  PLL settings for a USB clock of 48MHz  are based on being able to get an
+// exact 1200MHz VCO, they may need revising with a different reference frequency.
+#ifndef PLL_USB_VCO_FREQ_KHZ
```
kilograham, line 111: we should #ifdef setting these defaults with USB lock being 48Mhz

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,46 @@ extern "C" {
...
+#define PLL_USB_VCO_FREQ_KHZ                (1200 * KHZ)
+#endif
+#ifndef PLL_USB_POSTDIV1
+#define PLL_USB_POSTDIV1                    5
+#endif
+#ifndef PLL_USB_POSTDIV2
+#define PLL_USB_POSTDIV2                    5
+#endif
+
+// PLL settings for standard system clock with SYS_CLK_KHZ == 125,000
```
kilograham, line 121: similarly with SYS_CLK_KHZ == 125000

kilograham (changes requested):

src/rp2040/hardware_regs/include/hardware/platform_defs.h
kilograham: and now there is no `#error`

src/rp2040/hardware_regs/include/hardware/platform_defs.h
kilograham: i was referring to USB_CLK_MHZ and SYS_CLK_MHZ (github may have been confusing here as i clicked on USB_CLK_MHZ but it shouws a comment about XOSC_MHZ)

Can XOSC actually be specified in non MHZ increments? (I had assumed not)

kilograham:
> > This looks good to me, apart from the btstack related commit
> 
> Thanks and agreed. I found I could not push to my branch because there was another push that I hadn't got locally. So I did a 'git pull --rebase' and then a push. It was not enough for me to rebase my local branch from develop which is what I also tried.

well you have picked up some other stuff it seems (bt license)

src/rp2_common/hardware_clocks/include/hardware/clocks.h
lurch: The documentation just grabs whatever appears between the start and end tags - see page 232 of https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf for where this tag-content is displayed currently (and its surrounding context).
So yeah the "settings below" part should probably be moved outside of the tags.

src/rp2040/hardware_regs/include/hardware/platform_defs.h
lurch: **bump**

src/rp2_common/hardware_xosc/xosc.c
```diff
@@ -6,18 +6,18 @@
...
+// For frequency related definitions etc
 #include "hardware/clocks.h"
 
 #include "hardware/platform_defs.h"
 #include "hardware/regs/xosc.h"
 #include "hardware/xosc.h"
 
-#if XOSC_MHZ < 1 || XOSC_MHZ > 50
-#error XOSC_MHZ must be in the range 1-50
+#if XOSC_KHZ < (1 * KHZ) || XOSC_KHZ > (15 * KHZ)
```
lurch, line 16: See #644 for context about why this was set to 50MHz ?

src/rp2040/hardware_regs/include/hardware/platform_defs.h
kilograham: indeed, bump

src/rp2040/hardware_regs/include/hardware/platform_defs.h
```diff
@@ -37,11 +37,23 @@
...
+#define XOSC_KHZ _u(12000)
 #endif
 
+// PICO_CONFIG: SYS_CLK_KHZ, The system operating frequency in kHz, type=int, default=125000, advanced=true, group=hardware_base
+#ifndef SYS_CLK_KHZ
+#define SYS_CLK_KHZ _u(125000)
+#endif
+
+// For USB operation this *has* to be 48MHz
+#define USB_CLK_KHZ _u(48000)
```
kilograham, line 55: pretty sure i asked for this to be #ifndef-ed

src/rp2_common/hardware_clocks/include/hardware/clocks.h
kilograham: bump

src/rp2_common/hardware_clocks/include/hardware/clocks.h
kilograham: bump

kilograham (changes requested):
please read thru all pre-existing review comments, and issue comments - they keep getting ignored

src/rp2_common/hardware_clocks/include/hardware/clocks.h
kilograham: "these" defaults for VCO, DIV1, DIV2 only make sense when USB clock is 48Mhz, so #ifdef setting them based on USB clock being 48Mhz

src/rp2_common/hardware_clocks/include/hardware/clocks.h
kilograham: "these" defaults for VCO, DIV1, DIV2 only make sense when USB clock is 125Mhz, so #ifdef setting them based on USB clock being 48Mhz

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,49 @@ extern "C" {
...
+#define PLL_SYS_VCO_FREQ_KHZ                (1500 * KHZ)
+#endif
+#ifndef PLL_SYS_POSTDIV1
+#define PLL_SYS_POSTDIV1                    6
+#endif
+#ifndef PLL_SYS_POSTDIV2
+#define PLL_SYS_POSTDIV2                    2
+#endif
+
+static_assert(SYS_CLK_KHZ == PLL_SYS_VCO_FREQ_KHZ / PLL_SYS_POSTDIV1 / PLL_SYS_POSTDIV2, "see use of `vcocal.py` above");
```
kilograham, line 133: is this assertion strictly true due to rounding?

kilograham (commented):
Note all the new user overridable defines should also have PICO_CONFIG comments

lurch:
The `tag::pll_settings` block has now been totally removed - could it (or something very similar) be restored please? As a reminder, it gets auto-included into Section 2.18.3. of https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf

liamfraser (dismissed):
I think this looks good

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -89,6 +89,70 @@ extern "C" {
...
 
+/// \tag::pll_settings[]
+//
+// There are two PLLs in RP2040:
+// 1. The 'SYS PLL' generates the 125MHz system clock, the frequency is defined by `SYS_CLK_KHZ`.
+// 2. The 'USB PLL' generates the 48MHz USB clock, the frequency is defined by `USB_CLK_KHZ`.
+//
+// The two PLLs use the crystal oscillator output directly as their reference frequency input; the PLLs reference
+// frequency cannot be reduced by the dividers present in the clocks block. The crystal frequency is defined by `XOSC_KHZ` or
+// `XOSC_MKHZ`.
```
lurch, line 100: MKHZ? :thinking:

kilograham (approved):

kilograham:
fixes #1024

PR #1276, opened 2023-02-23
Disable Doxygen XML output

lurch:
We no longer need the Doxygen XML output, so disable it for (slightly) quicker documentation builds.


kilograham:
i think some people are using this

aallan:
It's possible we might also in the future. Right now we're going Doxygen -> HTML -> Asciidoc -> HTML, but in future, we might want to go Doxygen -> XML -> Asciidoc -> HTML.

PR #1284, opened 2023-02-27
Make sure BT device address is set.

### LINKED ISSUE #1269: BT mac address uses default if OTP is not set
The mac address used for WIFI and BT comes from OTP. For WIFI if the OTP is not set we use unique id to set it. We're not doing this for BT.  But why do some devices not have OTP set? https://forums.raspberrypi.com/viewtopic.php?t=347650

`Version: 7.95.61 (abcd531 CY) CRC: 4528a809 Date: Wed 2023-01-11 10:29:38 PST Ucode Ver: 1043.2169 FWID 01-7afb0879
Warning. No mac in otp. Generating mac from board id                       
cyw43 loaded ok, mac 5a:38:83:52:1a:28                                     
BT FW download, version = CYW4343A2_001.003.016.0031.0000_Generic_SDIO_37MHz_wlbga_BU_dl_signed
API: 12.2
Data: RaspberryPi.PicoW
Compiler: 1.29.4
ClmImport: 1.47.1
Customization: v5 22/06/24
Creation: 2022-06-24 06:55:08
Connecting to WiFi "Pi Towers"...
connect status: joining
BTstack up and running on 43:43:A2:12:1F:AC.
connect status: no ip
connect status: link up
Connected in 9s.

Ready, running iperf server at 10.3.15.26
`

peterharperuk:
Some devices seem to not have their OTP set. This is handled for the wifi mac but the BT device address gets set to an always the same default. Always set the bt address to wifi mac + 1

Fixes #1269

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/btstack_chipset_cyw43.c
+++ b/src/rp2_common/pico_cyw43_driver/btstack_chipset_cyw43.c
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2023 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "pico/btstack_chipset_cyw43.h"
+
+static void chipset_set_bd_addr_command(bd_addr_t addr, uint8_t *hci_cmd_buffer) {
+    hci_cmd_buffer[0] = 0x01;
+    hci_cmd_buffer[1] = 0xfc;
+    hci_cmd_buffer[2] = 0x06;
+    reverse_bd_addr(addr, &hci_cmd_buffer[3]);
+}
+
+static const btstack_chipset_t btstack_chipset_cyw43 = {
+    .name = "CYW43",
+    .init = NULL,
+    .next_command = NULL,
+    .set_baudrate_command = NULL,
+    .set_bd_addr_command = chipset_set_bd_addr_command,
+};
+
+const btstack_chipset_t * btstack_chipset_cyw43_instance(void) {
+    return &btstack_chipset_cyw43;
+}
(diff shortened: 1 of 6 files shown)


src/rp2_common/pico_cyw43_driver/btstack_chipset_cyw43.c
```diff
@@ -0,0 +1,26 @@
...
+#include "btstack_chipset.h"
+
+static void chipset_set_bd_addr_command(bd_addr_t addr, uint8_t *hci_cmd_buffer) {
+    hci_cmd_buffer[0] = 0x01;
+    hci_cmd_buffer[1] = 0xfc;
+    hci_cmd_buffer[2] = 0x06;
+    reverse_bd_addr(addr, &hci_cmd_buffer[3]);
+}
+
+static btstack_chipset_t btstack_chipset_cyw43 = {
```
kilograham, line 16: seems like this should be const

kilograham (changes requested):

src/rp2_common/pico_cyw43_driver/btstack_chipset_cyw43.c
peterharperuk: yes - well spotted.

src/rp2_common/pico_cyw43_driver/btstack_chipset_cyw43.c
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2023 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "btstack_chipset.h"
```
kilograham, line 7: oh, and i would suggest that this include your header, and that include "btstacK_chipset.h"

kilograham (changes requested):

peterharperuk:
I've been testing with Micropython and I notice we set the mac differently if OTP is not set. I've included a change that makes the sdk use the same code as MP to be consitent

PR #1305, opened 2023-03-11
Make linker fill alignment gaps with bkpt instruction

bruelltuete:
Request for comment: what do you think of this?
This came out from https://forums.raspberrypi.com/viewtopic.php?p=2089017#p2088888.

The linker fills in some gaps that result from alignment. But it filled in just zeros. Would it be better if those filler bytes would trigger a breakpoint or fault in case they are ever executed?


kilograham:
revisit when we do some cleanup of linker scripts

PR #1309, opened 2023-03-13
Use cyw43_malloc/free macro instead of malloc/free to allow alternative implementations

peterharperuk:
Bluetooth needs a tempoarary buffer to load firmware. Micropython won't work safely with malloc and free so allow these calls to be changed by using cyw43_alloc / free macros.

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cybt_shared_bus/cybt_shared_bus.c
+++ b/src/rp2_common/pico_cyw43_driver/cybt_shared_bus/cybt_shared_bus.c
@@ -62,14 +62,14 @@ static cybt_result_t cybt_fw_download_prepare(uint8_t **p_write_buf, uint8_t **p
     *p_write_buf = NULL;
     *p_hex_buf = NULL;
 
-    *p_write_buf = malloc(BTFW_DOWNLOAD_BLK_SIZE + BTFW_SD_ALIGN);
+    *p_write_buf = cyw43_malloc(BTFW_DOWNLOAD_BLK_SIZE + BTFW_SD_ALIGN);
     if (NULL == *p_write_buf) {
         return CYBT_ERR_OUT_OF_MEMORY;
     }
 
-    *p_hex_buf = malloc(BTFW_MAX_STR_LEN);
+    *p_hex_buf = cyw43_malloc(BTFW_MAX_STR_LEN);
     if (NULL == *p_hex_buf) {
-        free(*p_write_buf);
+        cyw43_free(*p_write_buf);
         return CYBT_ERR_OUT_OF_MEMORY;
     }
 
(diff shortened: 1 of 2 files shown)


peterharperuk:
This MP change prevents us using malloc and free https://github.com/micropython/micropython/commit/c80e7c14e6305e50e3b39f97172d4d8fe1214d3b

dpgeorge:
> This MP change prevents us using malloc and free

Oh yes!  Before making that change I did check to see if pico-sdk used malloc/free at all and came to the conclusion that it was not needed (I looked through the firmware.elf file of MicroPython).  But I did not check with BT enabled...

An alternative is that in MicroPython we could leave a little space for the C heap.  The downside with that approach is that C heap will be mostly unused and hence wasted.  So adding a `cyw43_malloc` hook seems like a better approach.

Another alternative: add an argument to `cyw43_btbus_init()` which is a temporary buffer that the caller must supply.  And also a constant (eg `#define CYW43_BT_BUF_SIZE 1024`) so the caller knows how much buffer space needs to be passed in.  Then the caller (eg MicroPython bindings) can allocate this, eg statically, on the C stack, on the GC heap.

src/rp2_common/pico_cyw43_driver/cybt_shared_bus/cybt_shared_bus.c
```diff
@@ -62,14 +62,14 @@ static cybt_result_t cybt_fw_download_prepare(uint8_t **p_write_buf, uint8_t **p
     *p_write_buf = NULL;
     *p_hex_buf = NULL;
 
-    *p_write_buf = malloc(BTFW_DOWNLOAD_BLK_SIZE + BTFW_SD_ALIGN);
+    *p_write_buf = cyw43_alloc(BTFW_DOWNLOAD_BLK_SIZE + BTFW_SD_ALIGN);
```
dpgeorge, line 65: If `cyw43_alloc` has the semantics of `malloc` then I'd suggest calling it `cyw43_malloc`.

peterharperuk: Will do.

peterharperuk:
Updated to rename cyw43_alloc to cyw43_malloc

dpgeorge (dismissed):

src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
```diff
@@ -157,6 +157,10 @@ void cyw43_post_poll_hook(void);
 
 #define CYW43_POST_POLL_HOOK cyw43_post_poll_hook();
 
+// Allow malloc and free to be changed
+#define cyw43_malloc malloc
```
kilograham, line 161: maybe we should #ifndef these too, so users can change them - esp. in FreeRTOS malloc may not be the favorite thing

kilograham (approved):

PR #1310, opened 2023-03-17
Fw compress merge

peterharperuk:
Changes required for wifi/bt firmware compression.

Use uzlib library added to the sdk.

Change bt firmware loading to use the functions provided for this purpose in cyw43-driver. It calls cyw43_get_firmware_funcs to get the list of functions.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "lib/btstack"]
 	path = lib/btstack
 	url = https://github.com/bluekitchen/btstack.git
+[submodule "lib/uzlib"]
+	path = lib/uzlib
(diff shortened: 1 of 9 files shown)


kilograham:
what is the status of this? it says checks are failing but that may have been intermittant

peterharperuk:
It's dependent on this https://github.com/georgerobotics/cyw43-driver/pull/68 which needs reviewing first.
I think Damien is of the opinion that the time it takes to decompress the firmware is not worth the flash saving.

kilograham:
moving to 1.6.0 as it sounds like this won't make 1.5.1

kilograham:
doh; have i not reviewed/merged this yet? sorry!

PR #1316, opened 2023-03-20
CI: Improve multi-gcc workflow

liamfraser:


kilograham:
looks good; btw you can push to `test_workflow` branch if you want it to run to check

kilograham (approved):

lurch:
I wonder if it's also worth testing with e.g. GCC 10.4 and 9.5 ? :shrug:  https://gcc.gnu.org/releases.html

liamfraser:
@lurch we are using the list of releases from the ARM website. Aka these https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads and these https://developer.arm.com/downloads/-/gnu-rm

PR #1320, opened 2023-03-23
Update mainpage.md

JamesH65:
Remove Trading reference

There is also a date on the same line, not sure that should be changed.

### DIFF
--- a/docs/mainpage.md
+++ b/docs/mainpage.md
@@ -24,5 +24,5 @@ The SDK builds an executable which is bare-metal, i.e. it includes the entirety 
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


lurch:
I think when we switched from "Raspberry Pi (Trading) Ltd." to "Raspberry Pi Ltd" we also lost the dot after "Ltd" ?

JamesH65:
> I think when we switched from "Raspberry Pi (Trading) Ltd." to "Raspberry Pi Ltd" we also lost the dot after "Ltd" ?

Fixed.

kilograham (approved):

PR #1324, opened 2023-03-25
Fix header link to index.html

chromaticOscillation:
Fix the link in the header not redirecting back to index.html in narrow windows


kilograham:
some sort of detailed description/repro would be helpful

lurch:
This change matches up with the `href` on line 48? ping @nelliemckesson

nathan-contino:
More of a @nelliemckesson issue than content.

Issue #1329, opened 2023-03-27
pico/cyw43_arch.h missing cyw43_arch_disable_ap_mode()


michaelbusby:
after initializing cyw43_arch_init(), I enable the access point and then try and connect the sta. If the sta is successful I want to disable the access point. 

I am trying to build a "captive portal" for initial setup and configurations utilizing a pico_w board. 

This issue will link to a pull request adding a new function within cyw43_arch.h cyw43_arch_disable_ap_mode() to simple turn off the AP mode.

michaelbusby:
see pull request #1330

kilograham:
merged into develop

PR #1330, opened 2023-03-27
add cyw43_arch_disable_ap_mode() to cyw43_arch 

michaelbusby:
the pico/cyw43_arch.h doesn't have functionality to turn off the access point mode of wireless controller. To be able to build a captive portal functionality on the pico w controller, the access point will need to be turned off some time after it is turned on.


michaelbusby:
see issue #1329

michaelbusby:
pull into wrong branch, sorry.

PR #1331, opened 2023-03-27
 add cyw43_arch_disable_ap_mode() to cyw43_arch

michaelbusby:
the pico/cyw43_arch.h doesn't have functionality to turn off the access point mode of the wireless controller. To be able to build captive portal functionality on the pico w, you will need to disable the AP mode after some time.


michaelbusby:
see issue #1329

kilograham (approved):

PR #1341, opened 2023-04-05
Customizable USB descriptor string length (Fixes #1339)

### LINKED ISSUE #1339: Arbitrary string length hard coded in tud_descriptor_string_cb() in stdio_usb_descriptors.c
The length of USB descriptor strings is limited to 20 characters through a define statement in the descriptor call back:

```
const uint16_t *tud_descriptor_string_cb(uint8_t index, __unused uint16_t langid) {
    #define DESC_STR_MAX (20)
```

Since version 1.5, it is possible to supply custom manufacturer and product strings by defining USBD_MANUFACTURER and USBD_PRODUCT, but they get truncated at 20 characters.  It would be nice to permit a user-defined maximum string length, e.g.,  by defining USBD_DESC_STR_MAX and changing the call back code to

```
const uint16_t *tud_descriptor_string_cb(uint8_t index, __unused uint16_t langid) {
#ifndef USBD_DESC_STR_MAX
#define USBD_DESC_STR_MAX (20)
#elif USBD_DESC_STR_MAX > 127
#error USBD_DESC_STR_MAX too high (max is 127).
#elif USBD_DESC_STR_MAX < 16
#error USBD_DESC_STR_MAX too low (min is 16).
#endif
    static uint16_t desc_str[USBD_DESC_STR_MAX];
```

The preprocessor statements above include range checking for the maximum descriptor length so it is always long enough to return the serial number and not longer than the maximum possible length.

hubiscode:
Implement a mechanism to set the maximum string length used in tud_descriptor_string_cb() by defining USBD_DESC_STR_MAX. If USBD_DESC_STR_MAX is not defined, the behavior remains unchanged and the previous default value of 20 is used. A compile time error is produced if USBD_DESC_STR_MAX is higher than 127 since the length of the string plus header is returned in a single byte as (2 * len + 2). Similarly, a compile time error is generated if the length is defined as less than 17 in order to ensure that there is enough room for the 16-character serial number plus header.

Fixes #1339


kilograham (changes requested):
Prefer this to be called `PICO_STDIO_USB_DEVICE_DESC_STR_MAX` since it is a user configurable item, and have a `PICO_CONFIG` (see elsewhere) comment

hubiscode:
> Prefer this to be called `PICO_STDIO_USB_DEVICE_DESC_STR_MAX` since it is a user configurable item, and have a `PICO_CONFIG` (see elsewhere) comment
I picked USBD_DESC_STR_MAX for consistency with USBD_MANUFACTURER and USBD_PRODUCT, which are the user-definable strings that get truncated without my changes.

I am not sure what you mean by "PICO_CONFIG (see elsewhere) comment". Please, let me know what I should do.

lurch:
> I am not sure what you mean by "PICO_CONFIG (see elsewhere) comment". Please, let me know what I should do.

See e.g. https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_sync/include/hardware/sync.h#L58
If a setting has a "PICO_CONFIG comment" then it can get extracted by https://github.com/raspberrypi/pico-sdk/blob/develop/tools/extract_configs.py and then included in https://github.com/raspberrypi/pico-project-generator/

kilograham:
> > Prefer this to be called `PICO_STDIO_USB_DEVICE_DESC_STR_MAX` since it is a user configurable item, and have a `PICO_CONFIG` (see elsewhere) comment
> 
> I picked USBD_DESC_STR_MAX for consistency with USBD_MANUFACTURER and USBD_PRODUCT, which are the user-definable strings that get truncated without my changes.

Ok, i agree with you in retrospect - this works as kind of a TinyUSB config

> I am not sure what you mean by "PICO_CONFIG (see elsewhere) comment". Please, let me know what I should do.

ok; yeah i don't think any of these are documented, but that is more of a doxygen thing now; I don't think USBD_xxx belongs in our public list of configs as they may conflict with user stuff if they are using TinyUSB on their own

I will merge this as is for now.

kilograham (approved):

Issue #841, opened 2022-05-30
Changing system clock sets peripheral clock to 48MHz


msearle5:
At boot, the system and peripheral clocks both appear to be 125MHz.
If the system clock is later set (to any speed, even 125MHz), the peripheral clock is set to 48MHz.
I haven't measured the peripheral clock directly, but I have used a modified version of the FatFS "example" app: 
https://github.com/msearle5/no-OS-FatFS-SD-SPI-RPi-Pico - which has a "cpu" command to set the system clock and a "spi" command to set the SPI (divided from peripheral) clock. At boot, SPI clocks of 62.5, 31.25, 20.833 etc. MHz can be set implying 125MHz peripheral clock, while after setting the system clock only clocks of 24MHz, 12MHz etc. can be set implying 48MHz peripheral clock. 
Setting the system clock with set_sys_clock_khz() calls set_sys_clock_pll() in pico-sdk/src/rp2_common/pico_stdlib/stdlib.c. line 39..64 and line 64 looks as if it's setting clk_peri to 48MHz from the USB clock in all cases, while I would expect it to be set to the same as the system clock.
(Maybe the fault is my expectation, and this is intentional. But if nothing else it was surprising to me, so I would recommend documenting it - the relevant section of the manual here https://raspberrypi.github.io/pico-sdk-doxygen/group__pico__stdlib.html doesn't mention the peripheral clock.)

bruelltuete:
Not sure where pico-sdk/src/rp2_common/pico_stdlib/stdlib.c's set_sys_clock_pll() is used. The default sdk clock setup happens in pico-sdk/src/rp2_common/pico_runtime/runtime.c: runtime_init() -> clocks_init(). Which gives you a clock setup like https://www.bruelltuete.com/picoclocks.txt (last time I checked).

msearle5:
Yes, the default clock setup behaves as I would expect with system and peripheral clocks both 125MHz, as in your diagarm (picoclocks.txt).
The problem shows up only when I set the system clock with set_sys_clock_khz(125000). (stdlib.c:126)
This calls set_sys_clock_pll (stdlib.c:129: "set_sys_clock_pll(vco, postdiv1, postdiv2);").
That function is at stdlib.c:39, and afterwards it appears that the clocks are set up like this:

    rosc @ 1-12 MHz

    xosc @ 12 MHz
        |
        \-- clk_ref @ 12 MHz
                |
[...]

PR #1346, opened 2023-04-19
Changing the sys clock also sets peripheral clock to 48MHz

andygpz11:
Currently, changing the system clock frequency also _always_ sets the peripheral clock frequency to 48MHz.
Fix:  Add a PICO_ config control that can be defined to override this behaviour. The current behaviour is unchanged by default.
Fixes issue #841

### DIFF
--- a/src/rp2_common/pico_stdlib/stdlib.c
+++ b/src/rp2_common/pico_stdlib/stdlib.c
@@ -36,6 +36,16 @@ void set_sys_clock_48mhz() {
     }
 }
 
+// When the System clock PLL set-up is changed from the initial setting made
+// during system initialisation, the code automatically switches the peripheral
+// clock to the 48MHz USB clock for safety reasons.
+// The Pico config option below allows a user to change this default behaviour if
+// required by defining the pico config option below as 0 in their board file.
+// PICO_CONFIG: SET_SYS_CLK_PLL_SETS_PERI_CLK_TO_USB_CLK, When the SYS clock PLL is changed, *also* switch the peripheral clock to the USB clock, type=bool, default=1, advanced=true, group=hardware_clocks
+#ifndef SET_SYS_CLK_PLL_SETS_PERI_CLK_TO_USB_CLK
+#define SET_SYS_CLK_PLL_SETS_PERI_CLK_TO_USB_CLK 1
+#endif
+
 void set_sys_clock_pll(uint32_t vco_freq, uint post_div1, uint post_div2) {
     if (!running_on_fpga()) {
         clock_configure(clk_sys,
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_stdlib/stdlib.c
```diff
@@ -36,6 +36,16 @@ void set_sys_clock_48mhz() {
     }
 }
 
+// When the System clock PLL setup is changed from the initial setting made
+// during system initialisation, the code automatically switches the peripheral
+// clock to the 48MHz USB clock to ensure continuity of peripheral operation.
+// The Pico config option below allows a user to change this default behaviour if
+// required by defining the Pico config option below as 0 in their board file.
+// PICO_CONFIG: PICO_CLOCK_AJDUST_PERI_CLOCK_WITH_SYS_CLOCK, When the SYS clock PLL is changed also switch the peripheral clock to the USB clock, type=bool, default=1, advanced=true, group=hardware_clocks
+#ifndef PICO_CLOCK_AJDUST_PERI_CLOCK_WITH_SYS_CLOCK
```
lurch, line 45: Not sure if this is the best name? "AJDUST_PERI_CLOCK_WITH_SYS_CLOCK" _might_ be misinterpreted to mean that the peri-clock will be adjusted to the same value as the sys-clock? From my limited understanding of the PLL and Clocks code, this setting is effectively `PICO_CLOCK_DETACH_PERI_CLOCK_FROM_SYS_CLOCK_WHEN_CHANGING_SYS_CLOCK` ? :wink: 

Perhaps it's worth inverting the logic, and instead adding `PICO_CLOCK_KEEP_PERI_CLOCK_ATTACHED_TO_SYS_CLOCK` which defaults to `0`? :man_shrugging:

kilograham: Yes the default should be zero.
Other note, you should probably move the define to the stdlib.h header (so that other code can see the value)

andygpz11: At this point I can see value in lurch's suggestion:
`PICO_CLOCK_KEEP_PERI_CLOCK_ATTACHED_TO_SYS_CLOCK`
I'd like to use that please.

src/rp2_common/pico_stdlib/stdlib.c
```diff
@@ -39,11 +39,13 @@ void set_sys_clock_48mhz() {
...
 // during system initialisation, the code automatically switches the peripheral
 // clock to the 48MHz USB clock to ensure continuity of peripheral operation.
-// The Pico config option below allows a user to change this default behaviour if
-// required by defining the Pico config option below as 0 in their board file.
-// PICO_CONFIG: PICO_CLOCK_AJDUST_PERI_CLOCK_WITH_SYS_CLOCK, When the SYS clock PLL is changed also switch the peripheral clock to the USB clock, type=bool, default=1, advanced=true, group=hardware_clocks
+// The Pico config option below allows a user to change this default behaviour,
+// if required, by defining the Pico config option below as 1 in their board file.
+// If this option is selected, the peripheral clock will remain connected to the
+// system PLL clock and so will change frequency with that.
+// PICO_CONFIG: PICO_CLOCK_AJDUST_PERI_CLOCK_WITH_SYS_CLOCK, When the SYS clock PLL is changed also switch the peripheral clock to the USB clock, type=bool, default=0, advanced=true, group=hardware_clocks
```
kilograham, line 46: doc is now wrong

andygpz11: It would appear so as rendered by github but it is not in the copy of file I see here even after:
`rm ./src/rp2_common/pico_stdlib/stdlib.c && git checkout -f ./src/rp2_common/pico_stdlib/stdlib.c`
Or in my `gitk` log.

kilograham (approved):

Issue #1274, opened 2023-02-22
Problems with pico-sdk UART transmission: blocking exits early, and (possibly PL011) 2nd stop bit missing


sdbbs:
Hi all,

I have prepared a small pico-sdk project, `rp2040_uart_err_test`, an RP2040 UART TX test, in this gist: [rp2040_uart_err_test](https://gist.github.com/sdbbs/aca6afa809166543588e82a29874b3ae)

I believe that it demonstrates the following two problems:

1. `uart_write_blocking` does block, but returns early: when it returns, two bytes (in this test) still remain to be sent
    * The documentation for [`uart_write_blocking`](https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__uart.html#gada6a512414b1113a59725ddfbe23c4a5) explicitly says:
      > "This function will block until all the data has been sent to the UART"
      
      ... so I would expect either this function to indeed block until all data has been transmitted - or the help text needs to be more precise about what should be expected, otherwise it is misleading as it is.
2. If UART is set up for 2 stop bits, then when transmitting (TX) over UART, 2nd stop bit is never transmitted
    * The documentation for [`uart_set_format`](https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__uart.html#ga2b1663aa78039847d282907b20201310) explicitly says:
      > "stop_bits	Number of stop bits 1..2 "
      
      ... so I would expect that two stop bits are sent, if I have configured the UART for two stop bits.

-------

Few words about the UART TX test project, `rp2040_uart_err_test`:

* I have tried to make it as "low-level" as possible, to avoid possible setup interference with convenience features of "vanilla" projects; so the project avoids `pico_stdlib` and builds only against `pico_runtime`; also `stdio_usb` and `stdio_uart` are disabled
* There is a setup for a UART RX interrupt service routine, `on_uart_rx_isr`, but it is there simply to go through the setup motions - it is otherwise not expected to fire or do anything in this test
* Clock is at 133 MHz

So, the main functionality is in the `main` function:

* There is a 24-byte global array `tx_buffer` with some random binary content
* After setup, `main` goes into the infinite `while` loop, where:
    * it sleeps for 250 ms
    * then it sets debug pin HIGH, executes `uart_write_blocking` with the intent to send/TX all 24 bytes of `tx_buffer` (via reference variables), and sets debug pin LOW

This would allow for the debug pin indicating the duration of `uart_write_blocking`. Here is the `while` loop:

```c
[...]

PR #1347, opened 2023-04-19
Correct writes to the UART LCR register

andygpz11:
The current SDK code allows application code to change UART settings when the UART is busy.  This is not correct and causes some API functionality to silently fail, see issue https://github.com/raspberrypi/pico-sdk/issues/548 and also (a part of) https://github.com/raspberrypi/pico-sdk/issues/1274
A fix was authoured in https://github.com/raspberrypi/pico-sdk/pull/556 and has been slightly tidied and then tested for this change & PR.

### DIFF
--- a/src/rp2_common/hardware_uart/uart.c
+++ b/src/rp2_common/hardware_uart/uart.c
@@ -73,8 +73,8 @@ void uart_deinit(uart_inst_t *uart) {
 static uint32_t uart_disable_before_lcr_write(uart_inst_t *uart) {
     // Notes from PL011 reference manual:
     //
-    // - Before writing LCR, must disable UART and wait for current TX + RX
-    //   char to finish
+    // - Before writing the LCR, if the UART is enabled it needs to be
+    //   disabled and any current TX + RX activity has to be completed
     //
     // - There is a BUSY flag which waits for the current TX char, but this is
     //   OR'd with TX FIFO !FULL, so not usable when FIFOs are enabled and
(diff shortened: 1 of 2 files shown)


lurch:
Does the doxygen-comment for `uart_set_break` need the same "The UART is paused..." stuff?

src/rp2_common/hardware_uart/uart.c
```diff
@@ -163,37 +167,42 @@ void uart_set_format(uart_inst_t *uart, uint data_bits, uint stop_bits, uart_par
...
+    if (was_enabled) {
         uart_get_hw(uart)->cr = cr_save;
+    }
 }
 
 void uart_set_break(uart_inst_t *uart, bool en) {
     bool was_enabled = uart_is_enabled(uart);
-    uint32_t cr_save;
-    if (was_enabled)
+    uint32_t cr_save = 0;  // stifle warning
```
lurch, line 192: Just out of curiosity, how come this "stifle warning" wasn't needed in the other functions that this PR modifies?

andygpz11: I'm not sure!  This one was required, I think the compiler was being more paranoid than it needed to be, no others were.

src/rp2_common/hardware_uart/uart.c
```diff
@@ -69,6 +70,42 @@ void uart_deinit(uart_inst_t *uart) {
...
+    // Anything else would require API changes to permit a non-enabled UART
+    // state after init() where settings can be changed safely.
+
+    uint32_t cr_save = uart_get_hw(uart)->cr;
+    hw_clear_bits(&uart_get_hw(uart)->cr,
+        UART_UARTCR_UARTEN_BITS | UART_UARTCR_TXE_BITS | UART_UARTCR_RXE_BITS);
+
+    uint32_t current_ibrd = uart_get_hw(uart)->ibrd;
+    uint32_t current_fbrd = uart_get_hw(uart)->fbrd;
+    uint64_t baud_period_usec = 1u +
```
kilograham, line 101: why is this `uint64_t`? doesn't seem like it needs to be, and that pulls in a bunch of code

src/rp2_common/hardware_uart/uart.c
```diff
@@ -86,19 +123,88 @@ uint uart_set_baudrate(uart_inst_t *uart, uint baudrate) {
         baud_fbrd = ((baud_rate_div & 0x7f) + 1) / 2;
     }
 
-    // Load PL011's baud divisor registers
+    // Need to cleanly disable UART before touching LCR
```
kilograham, line 126: in all 3 places we do `uart_disable_before_lcr_write` we also do "was_enabled".  since cr is only a 16 bit value, maybe return a negative number for cr_save when uart wasn't enabled, and avoid the code duplication

kilograham (changes requested):

src/rp2_common/hardware_uart/uart.c
andygpz11: I've rolled up the code as suggested.  I would still like to triple check everything is ok.

src/rp2_common/hardware_uart/uart.c
```diff
@@ -69,6 +70,65 @@ void uart_deinit(uart_inst_t *uart) {
...
+    uint32_t current_ibrd = uart_get_hw(uart)->ibrd;
+    uint32_t current_fbrd = uart_get_hw(uart)->fbrd;
+
+    // Note: Maximise precision here. Show working, the compiler will mop this up.
+    // Create a 16.6 fixed-point fractional division ratio; then scale to 32-bits.
+    uint32_t brdiv_ratio = 64u * current_ibrd + current_fbrd;
+    brdiv_ratio <<= 10;
+    // 3662 is ~(15 * 244.14) where 244.14 is 16e6 / 2^16
+    uint32_t scaled_freq = clock_get_hz(clk_peri) / 3662ul;
+    uint32_t wait_time_us = brdiv_ratio / scaled_freq;
```
kilograham, line 108: I'm not convinced this math is right yet (or the comments)

I calc for 133Mhz peri clock and 115200 baud rate, then brdiv_ratio should be 9236 ish.
Thhe above calc then gives me (I think) 260us

However 15 bauds is 130us

src/rp2_common/hardware_uart/uart.c
```diff
@@ -69,6 +70,65 @@ void uart_deinit(uart_inst_t *uart) {
...
+    busy_wait_us(wait_time_us);
+
+    return cr_save;
+}
+
+static void uart_write_lcr_bits_masked(uart_inst_t *uart, uint32_t values, uint32_t write_mask) {
+    invalid_params_if(UART, uart != uart0 && uart != uart1);
+    uint32_t cr_save;
+    bool enabled = uart_is_enabled(uart);
```
kilograham, line 117: I mentioned this in a previous review, but i would like to get rid of the `uart_is_enabled` checks in the caller, and just put it in the uart_disable_before_lcr_write (which can just do the delay if the UART was enabled based on its read of the CR).

it is then fine to unconditionally set the CR at the bottom of the callers too.

kilograham (changes requested):

src/rp2_common/hardware_uart/uart.c
andygpz11: I think this is correct as submitted.
You got to 260 us because there's a clever twist in the Baud rate setting code that divides the integer part of the divide ratio by _128_ using a `>> 7` and *not* by 64 as would be assumed from the 6-bit fractional part...
That's why the the code that calculates the divisor uses 8 and not the 16 required by the UART hardware.
`uint32_t baud_rate_div = (8 * clock_get_hz(clk_peri) / baudrate);`
I suspect that was to avoid wrapping a 32-bit number at low Baud rate with fast `clk_peri` rates.

kilograham: ah ok, thanks; i had assumed the x8 was for 8 bits ;-)

src/rp2_common/hardware_uart/uart.c
lurch: `uart_disable_before_lcr_write` is only called from one place, so I guess an alternative might be to just fold that code back inside `uart_write_lcr_bits_masked` ? :shrug:

andygpz11: @kilograham I've just pushed a fix
@lurch We could roll up those two functions but I don't think it will make a difference to the object; the compiler will inline them because (as you observe) `uart_disable_before_lcr_write` is only called from one place.
I think it is more readable as it is - the intricacies of the UART disabling are removed from the functional code in `uart_write_lcr_bits_masked`

kilograham (approved):
excellent

PR #1369, opened 2023-05-11
Update lwip to tip of master

### LINKED ISSUE #1244: lwip submodule should point to a tag or branch tip in the remote repository
Currently the lwip submodule points to an arbitrary commit c385f31076b27efb8ee37f00cb5568783a58f299. Most git servers, including GNU Savannah, only allow shallow clones of branch tips and tags. This means if you do a shallow recursive clone of pico-extras (for example, in a CI pipeline), you will get the following error:

    Error: error: Server does not allow request for unadvertised object c385f31076b27efb8ee37f00cb5568783a58f299
    Error: fatal: Fetched in submodule path 'lib/lwip', but it did not contain c385f31076b27efb8ee37f00cb5568783a58f299. Direct fetching of that commit failed.

User must instead do a full clone of the submodule, which might mean doing a full clone of pico-extras too. This is slow, wastes bandwidth and electricity, uses up extra Actions minutes, and hammers the GNU servers.

This can be fixed by either changing the submodule to point to an existing tag or branch (if branch, you have to keep on updating the submodule whenever the branch tip changes), or by forking the upstream repository and adding a tag to the commit you want to use, and then adjusting the submodule to point to your fork.

peterharperuk:
Fixes #1244


kilograham (commented):
I'm not 100% sure how this helps the original issue #1244 as this new commit won't be the branch tip for long.

Does this actually fix anything in lwip we care about?

peterharperuk:
> I'm not 100% sure how this helps the original issue https://github.com/raspberrypi/pico-sdk/issues/1244 as this new commit won't be the branch tip for long.

You're right, it doesn't. But I can't reproduce their problem. Master is better than nothing I guess.

> Does this actually fix anything in lwip we care about?

Not that I'm aware of

kilograham:
I'm not sure this usefully solves any problem, and we don't seem to be picking up anything new from lwip for this release, so closing

PR #1370, opened 2023-05-11
Add last_abort_reason

### LINKED ISSUE #1049: Feature request: Expose abort reason in I2C
While calling i2c functions, for example `i2c_write_blocking_until()`, internally `i2c_write_blocking_internal` is called, where `uint32_t abort_reason = 0;` gets defined, but there is no real way to get this value raw as caller. After reading it, `i2c->hw->clr_tx_abrt;` is used, so I can't read the value on my own after calling the function.

Only error that's checkable is `I2C_IC_TX_ABRT_SOURCE_ABRT_TXDATA_NOACK_BITS`, as it will return different value than length of data array. All other errors are hidden behind `PICO_ERROR_GENERIC`, which as name suggests, is generic, not specific.

Knowing specific error would come in handy (for example pooling `AT24Cx` write status - checking for `I2C_IC_TX_ABRT_SOURCE_ABRT_7B_ADDR_NOACK_BITS`).

Personally I would just add `last_error` to `i2c_inst_t`  or add flag to disable calling of `i2c->hw->clr_tx_abrt;` (But I don't know if that would be good idea - I have next to zero expirience with this kind of low level programming).

peterharperuk:
Fixes #1049


andygpz11:
This looks good to me.

andygpz11 (approved):
This looks good to me.

peterharperuk:
Fixed build issue

kilograham:
assigning to myself; i think i'd rather return the abort reason in the top half of a 64 bit rc (and do some API massaging)

peterharperuk:
> assigning to myself; i think i'd rather return the abort reason in the top half of a 64 bit rc (and do some API massaging)

I don't think you liked this change? I could add a variant of the functions to return int64_t if that sounds like what you wanted?

kilograham:
yeah, i guess returning a struct with two elements in a replacement APIs and picking the result out for the existing APIs is good
(Note a two uint32_t struct will be returned correctly in r0/r1 i believe)

src/rp2_common/hardware_i2c/include/hardware/i2c.h
```diff
@@ -246,6 +251,7 @@ static inline i2c_inst_t *i2c_get_instance(uint num) {
  *
  * \return Number of bytes written, or PICO_ERROR_GENERIC if address not acknowledged, no device present, or PICO_ERROR_TIMEOUT if a timeout occurred.
  */
+i2c_result_t i2c_write_result_blocking_until(i2c_inst_t *i2c, uint8_t addr, const uint8_t *src, size_t len, bool nostop, absolute_time_t until);
```
lurch, line 254: Does this (and the other new `_result_` functions) need to be added to the header file _after_ the old functions, in order not to invalidate the documented doxygen `\return` parameter?

src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -130,7 +130,7 @@ void i2c_set_slave_mode(i2c_inst_t *i2c, bool slave, uint8_t addr) {
     i2c->hw->enable = 1;
 }
 
-static int i2c_write_blocking_internal(i2c_inst_t *i2c, uint8_t addr, const uint8_t *src, size_t len, bool nostop,
+static i2c_result_t i2c_write_blocking_internal(i2c_inst_t *i2c, uint8_t addr, const uint8_t *src, size_t len, bool nostop,
```
lurch, line 133: For consistency with function-names and return-types, I wonder if `i2c_write_blocking_internal` ought to be renamed to `i2c_write_result_blocking_internal` ? :thinking:  (It's an internal-only function name, so I think we're free to rename it?)

peterharperuk: out of date

peterharperuk:
I've had another go at this. I decided I couldn't face adding new overloads to an already complicated API, just to get a not-often used reason for an abort. Instead I've added a new PICO_ERROR_ABORT error code. This is only used if PICO_I2C_RETURN_ABORT_REASON=1 in which case the abort reason is taken from this. So you can obtain the reason with a bit of maths.

```
if (ret <= PICO_ERROR_ABORT)
    abort_reason = PICO_ERROR_ABORT - ret; // one of I2C_IC_TX_ABRT_SOURCE_ABRT_*_LSB
```

kilograham:
interesting... i think perhaps i'd prefer to make something more generic (and not overlaying existing/future return codes)...

e.g. assign codes `0x80000000` to `0xbfffffff` as "data" error codes which return 30 bits of function specific result (similar to the way that any positive value can be success)

can then add a `pico_error_has_data(code)` and `pico_error_get_data(code)` macros... the I2C API can add a more specific API that uses this

peterharperuk:
Sounds good, will have a go...

kilograham:
yes; we have hit this elsewhere - hard to change from -1, though we want to move people towards `< 0` .. 
so yeah, you can still leave this optional

src/common/pico_base_headers/include/pico/error.h
```diff
@@ -42,7 +42,8 @@ enum pico_error_codes {
     PICO_ERROR_UNSUPPORTED_MODIFICATION = -18,  ///< Write is impossible based on previous writes; e.g. attempted to clear an OTP bit
     PICO_ERROR_LOCK_REQUIRED = -19,             ///< A required lock is not owned
     PICO_ERROR_VERSION_MISMATCH = -20,          ///< A version mismatch occurred (e.g. trying to run PIO version 1 code on RP2040)
-    PICO_ERROR_RESOURCE_IN_USE = -21            ///< The call could not proceed because requires resourcesw were unavailable
+    PICO_ERROR_RESOURCE_IN_USE = -21,           ///< The call could not proceed because requires resourcesw were unavailable
```
lurch, line 45: nitpick: "requires resourcesw" -> "required resources"

src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -130,6 +130,15 @@ void i2c_set_slave_mode(i2c_inst_t *i2c, bool slave, uint8_t addr) {
     i2c->hw->enable = 1;
 }
 
+// PICO_CONFIG: PICO_I2C_RETURN_ABORT_REASON, change i2c functions to return the abort reason via a return code less than or equal to PICO_ERROR_ABORT, type=bool, default=0, group=harware_i2c
```
lurch, line 133: `harware_i2c` -> `hardware_i2c`

PR #1372, opened 2023-05-12
Add dma_channel_cleanup method and use it to cleanup after pico_cyw43_driver is closed

### LINKED ISSUE #1156: Is cyw43 driver leaving DMA channels in some undefined state?
The cyw43 driver uses DMA to feed the PIO machines, e.g. https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/cyw43_driver/cyw43_bus_pio_spi.c#L275

But does it need to do more to clean up on deinit? See https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/cyw43_driver/cyw43_bus_pio_spi.c#L169

I've noticed that when I deinit the wifi module, my ADC code will re-use one of the DMA channels previously used by the cyw driver. And that re-used DMA channel will pretty much immediately trigger a completion IRQ even when I haven't started the ADC nor the channel itself. So something was left pending.

Should there be an `dma_irqn_acknowledge_channel(DMA_IRQ_0, bus_data->dma_out);` just before dma_channel_unclaim? (same for `DMA_IRQ_1` and `dma_in`).

peterharperuk:
cyw43 seems to be causing a dma channel interrupt to go off if the same channels are reused after cyw43 is deinitialised. Achnowledging them seems to fix it.

Fixes #1156

### DIFF
--- a/src/rp2_common/hardware_dma/dma.c
+++ b/src/rp2_common/hardware_dma/dma.c
@@ -72,10 +72,8 @@ bool dma_timer_is_claimed(uint timer) {
 
 void dma_channel_cleanup(uint channel) {
     check_dma_channel_param(channel);
-    // we definitely want to disable chaining, because that could cause a restart
-    // during abort. Sicne we're doing that, it's just as easy to reset to default config!
-    const dma_channel_config config = dma_channel_get_default_config(channel);
-    dma_channel_set_config(channel, &config, false);
+    // Disable CHAIN_TO, and disable channel, so that it ignores any further triggers 
+    hw_write_masked( &dma_hw->ch[channel].al1_ctrl, (channel << DMA_CH0_CTRL_TRIG_CHAIN_TO_LSB) | (0u << DMA_CH0_CTRL_TRIG_EN_LSB), DMA_CH0_CTRL_TRIG_CHAIN_TO_BITS | DMA_CH0_CTRL_TRIG_EN_BITS );
     // disable IRQs first as abort can cause spurious IRQs
     dma_channel_set_irq0_enabled(channel, false);
     dma_channel_set_irq1_enabled(channel, false);
--- a/src/rp2_common/hardware_dma/dma.c
+++ b/src/rp2_common/hardware_dma/dma.c
@@ -70,6 +70,20 @@ bool dma_timer_is_claimed(uint timer) {
     return hw_is_claimed(&_timer_claimed, timer);
 }
 
+void dma_channel_cleanup(uint channel) {
+    check_dma_channel_param(channel);
+    // we definitely want to disable chaining, because that could cause a restart
+    // during abort. Sicne we're doing that, it's just as easy to reset to default config!
+    const dma_channel_config config = dma_channel_get_default_config(channel);
+    dma_channel_set_config(channel, &config, false);
+    // disable IRQs first as abort can cause spurious IRQs
+    dma_channel_set_irq0_enabled(channel, false);
+    dma_channel_set_irq1_enabled(channel, false);
+    dma_channel_abort(channel);
+    // finally clear the IRQ status, which may have been set during abort
+    dma_hw->intr = 1u << channel;
+}
+
 #ifndef NDEBUG
 
 void print_dma_ctrl(dma_channel_hw_t *channel) {
--- a/src/rp2_common/hardware_dma/include/hardware/dma.h
+++ b/src/rp2_common/hardware_dma/include/hardware/dma.h
@@ -912,8 +912,8 @@ static inline void dma_channel_cleanup(uint channel) {
     check_dma_channel_param(channel);
     // we definitely want to disable chaining, because that could cause a restart
     // during abort. Sicne we're doing that, it's just as easy to reset to default config!
-    uint32_t cfg = dma_channel_default_config(channel);
-    dma_channel_set_config(channel, &cfg, false);
+    uint32_t config = dma_channel_get_default_config(channel);
+    dma_channel_set_config(channel, &config, false);
     // disable IRQs first as abort can cause spurious IRQs
     dma_channel_set_irq0_enabled(channel, false);
     dma_channel_set_irq1_enabled(channel, false);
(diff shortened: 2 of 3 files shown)


kilograham (changes requested):
(I made this comment already but it disappeared!)

I think in general, we should add a `dma_channel_cleanup` method which

1. Disables channel in INTE0 and INTE1
2. Calls `dma_channel_abort`
3. Clears channel IRQ via INTR register

(and call that here)

peterharperuk:
yes - I'll have a look at adding that

peterharperuk:
I added dma_channel_cleanup

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -891,6 +891,18 @@ static inline uint dma_get_timer_dreq(uint timer_num) {
...
+/*! \brief Performs DMA channel cleanup after use
+ *  \ingroup hardware_dma
+ *
+ * \param channel DMA channel
+ */
+static inline void dma_channel_cleanup(uint channel) {
+    dma_channel_set_irq0_enabled(channel, false);
+    dma_channel_set_irq1_enabled(channel, false);
+    dma_channel_abort(channel);
+    hw_clear_bits(&dma_hw->intr, 1u << channel);
```
kilograham, line 903: should just be a regular write of `dma_hw->intr`; it is a WC field

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -891,6 +891,18 @@ static inline uint dma_get_timer_dreq(uint timer_num) {
     return DREQ_DMA_TIMER0 + timer_num;
 }
 
+/*! \brief Performs DMA channel cleanup after use
```
kilograham, line 894: perhaps the doc could be a bit clearer about what a "cleanup"'s purpose/function is

kilograham (changes requested):

src/rp2_common/hardware_dma/include/hardware/dma.h
peterharperuk: Apologies - that was dumb. I think I fixed it now.

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -891,6 +891,37 @@ static inline uint dma_get_timer_dreq(uint timer_num) {
...
+ * if (dma_channel >= 0) {
+ *     dma_channel_cleanup(dma_channel);
+ *     dma_channel_unclaim(dma_channel);
+ *     dma_channel = -1;
+ * }
+ * \endcode
+ *
+ * \param channel DMA channel
+ */
+static inline void dma_channel_cleanup(uint channel) {
```
kilograham, line 911: there's really no point in this being inline; can you move it into dma.c

kilograham (changes requested):

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -891,6 +891,37 @@ static inline uint dma_get_timer_dreq(uint timer_num) {
...
+    check_dma_channel_param(channel);
+    // we definitely want to disable chaining, because that could cause a restart
+    // during abort. Sicne we're doing that, it's just as easy to reset to default config!
+    uint32_t config = dma_channel_get_default_config(channel);
```
lurch, line 915: CI is falling over because this should be `dma_channel_config config` rather than `uint32_t config` ?

src/rp2_common/hardware_dma/include/hardware/dma.h
peterharperuk: Done

src/rp2_common/hardware_dma/include/hardware/dma.h
peterharperuk: fixed

src/rp2_common/hardware_dma/dma.c
```diff
@@ -70,6 +70,20 @@ bool dma_timer_is_claimed(uint timer) {
     return hw_is_claimed(&_timer_claimed, timer);
 }
 
+void dma_channel_cleanup(uint channel) {
+    check_dma_channel_param(channel);
+    // we definitely want to disable chaining, because that could cause a restart
+    // during abort. Sicne we're doing that, it's just as easy to reset to default config!
+    const dma_channel_config config = dma_channel_get_default_config(channel);
```
Wren6991, line 77: What state is the channel expected to be in when this is called? Wiping the config when it is potentially running seems unsafe.

Possibly this should be:

```c
// Disable CHAIN_TO, and disable channel, so that it ignores any further triggers
hw_write_field(
    &dma_hw->ch[channel].al1_ctrl,
    (channel << DMA_CH0_CTRL_TRIG_CHAIN_TO_LSB) | (0u << DMA_CH0_CTRL_TRIG_EN_LSB),
    DMA_CH0_CTRL_TRIG_CHAIN_TO_BITS | DMA_CH0_CTRL_TRIG_EN_BITS
);
```

What do you think?

peterharperuk: We call this when we've finished using it when uninitialising the wifi driver that no longer needs the dma channel or pios

Wren6991: If it's always expected to be called after the channel has been brought to a halt, then the current code is fine, but the abort seems unnecessary, and it's probably worth making this explicit in the doc comment in dma.h.

If it's expected to be possible to call this function at any time, then it's probably not a good idea to write all fields in the CTRL register. For example, this code will turn on read increment, which means there is a period where the channel may read from unexpected addresses (like other PIO FIFOs) before we call `dma_channel_abort()` after clearing the IRQs.

peterharperuk: I see, that sounds bad. Your change makes sense, although I assume you meant hw_write_masked.

Wren6991: > although I assume you meant hw_write_masked
yes, oops

kilograham (approved):

PR #1380, opened 2023-05-18
WIP: Add missing btstack files

### LINKED ISSUE #1379: Some btstack files missing
When we update btstack (#1376) we need to add some missing files.

peterharperuk:
Fixes #1379


peterharperuk:
Needs us to update btstack first

kilograham (changes requested):
small thing, but (and it looks like there are a cople of offenders with md5 and yxml, but the majority of this is in alphabetical order, so best to maintain

peterharperuk:
fixed alpha order and moved the classic file to the right lib!

kilograham (approved):

peterharperuk:
Oh dear - still says WIP. I thought that would block it.

PR #1388, opened 2023-05-22
Add a http client implementation.

### LINKED ISSUE #1386: Add a http client
I'm adding some test code for the lwip http client to pico-examples https://github.com/raspberrypi/pico-examples/issues/318

But I think it's a bit of a cop out to leave it over there. We should really have a pico http client in the sdk that users can use until they find something better?

Some attempt to make it independent from lwip is a bonus.

peterharperuk:
Implemented using the lwip http client.

Fixes #1386

### DIFF
--- a/src/rp2_common/pico_lwip/CMakeLists.txt
+++ b/src/rp2_common/pico_lwip/CMakeLists.txt
@@ -302,5 +302,15 @@ if (EXISTS ${PICO_LWIP_PATH}/${LWIP_TEST_PATH})
             pico_lwip_contrib_freertos
             pico_rand)
 
+    pico_add_library(pico_lwip_http_util NOFLAG)
+    target_sources(pico_lwip_http_util INTERFACE
+            ${CMAKE_CURRENT_LIST_DIR}/http_client_util.c
+            )
+    pico_mirrored_target_link_libraries(pico_lwip_http_util INTERFACE
+            pico_lwip_http
+            pico_lwip_mbedtls
+            pico_mbedtls
+            )
+
     pico_promote_common_scope_vars()
 endif()
--- a/src/rp2_common/pico_lwip/http_client_util.c
+++ b/src/rp2_common/pico_lwip/http_client_util.c
@@ -1,0 +1,141 @@
+/**
+ * Copyright (c) 2023 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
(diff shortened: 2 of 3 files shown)


src/rp2_common/pico_http_client/include/pico/http_client.h
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @param hostname hostname for request
+ * @param client_header_cb function to call with headers (can be null)
+ * @param client_body_cb function to call with http body
+ * @param arg Argument to pass to callbacks
+ * @return state object to pass to other functions, or null on error
+ */
+void *http_client_init(const char *hostname, pico_http_header_cb client_header_cb, pico_http_body_cb client_body_cb, void *arg);
+
+/*! \brief Initialise the http client to perform a https request
```
kilograham, line 34: comment should probably explain something about what _secure means

kilograham:
> Other comments,
> 
> 1/ Passing the hostname in http_client_init means you can't reuse the same "state" to make a request to a different host. Seems a bit restrictive? The only reason I can think for this is that your tls config might be related to the host you're talking to. It's easy enough to allocate "state" for each host?
> 
> 2/ Can't run multiple requests at the same time? You could do this by calling http_client_init multiple times for the number of requests you want to have running at the same time. The lwip api just allocates a new state for each request and frees this in the callback. We could do the same but it seems wasteful when you want to make multiple requests one after another (to the same host).

Yeah, perhaps have a friendly API which combines the underlying "state" init, and the actual request, but expose the underlying methods too.

I guess I need to understand what the thing we are initializing is (it's life time) to know what it should be called (is it actually a client instance, a connection, etc.?)

peterharperuk:
> I guess I need to understand what the thing we are initializing is (it's life time)

It looks like the lwip api is per http request - which is a bit naff. This was just supposed to be a "helper" utility to make the lwip functionality a bit easier to use. Originally it was just part of the example code but I thought that was a bit lazy and should maybe add it to the sdk. Maybe I should forget that.

peterharperuk:
I pushed a new version. To make an http request is something like this...

    PICO_HTTP_REQUEST_T req = {0};
    req.hostname = HOST;
    req.url = URL;
    req.recv_fn = response_function;
    int rc = http_client_request_sync(cyw43_arch_async_context(), &req);

I'm not sure this is adding a lot of value, so I might just move it back to pico-examples. What do you think?

PR #1391, opened 2023-05-24
Work around issue with unreferenced sections marked allocated in the asm being discarded by GNU linker

kilograham:


src/rp2_common/pico_standard_link/crt0.S
```diff
@@ -324,13 +324,29 @@ hold_non_core0_in_bootrom:
 // ----------------------------------------------------------------------------
 // Stack/heap dummies to set size
 
-.section .stack, "a"
+// Prior to SDK 1.5.1 these were `.section .stack` without the `, "a"`... Clang linker gives a warning about this,
+// however setting it eplxicityl to `, "a"` makes GCC *now* discard the section unless it is also KEEP. This
```
lurch, line 328: 'explicitly' typo

Wren6991 (approved):
That GCC behaviour is odd. The workaround makes sense.

PR #1399, opened 2023-05-27
[HUST CSE][rp2_common][pico_malloc.c]Modify to check if rc is a null pointer

### LINKED ISSUE #1397: 【rp2_common】【pico_malloc.c】Modify the code to check if rc is a null pointer before calling rc
The code on lines 41, 58, and 75 of this code may attempt to convert a null pointer to a uint8_t* type and attempt to perform an addition operation on top of that pointer. This behavior is undefined, and the undefined result depends on the specific compiler and runtime environment. In some cases, this may cause the program to crash.

jming912:
The code on lines 41, 58, and 75 of this code may attempt to convert a null pointer to a uint8_t* type and attempt to perform an addition operation on top of that pointer. This behavior is undefined, and the undefined result depends on the specific compiler and runtime environment. In some cases, this may cause the program to crash.
    In this modified version, we first check if rc is NULL. if rc is NULL, we print a message stating that calloc failed to allocate the required memory. Only if rc is not NULL do we attempt pointer arithmetic and proceed to the next conditional step. This avoids undefined pointer operations when rc is NULL.
    The issue of this pr is Fixes #1397

### DIFF
--- a/src/rp2_common/pico_malloc/pico_malloc.c
+++ b/src/rp2_common/pico_malloc/pico_malloc.c
@@ -37,8 +37,13 @@ void *WRAPPER_FUNC(malloc)(size_t size) {
     mutex_exit(&malloc_mutex);
 #endif
 #if PICO_DEBUG_MALLOC
-    if (!rc || ((uint8_t *)rc) + size > (uint8_t*)PICO_DEBUG_MALLOC_LOW_WATER) {
-        printf("malloc %d %p->%p\n", (uint) size, rc, ((uint8_t *) rc) + size);
+    if (!rc) 
+    {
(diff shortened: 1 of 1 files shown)


kilograham:
fixes #1397

jming912:
> fixes #1397

Can you Approve my workflow

jming912:
> fixes #1397

Could you approve my workflow

kilograham:
nope; please comment on the associated issue as to what circumstances this could ever be a problem.

kilograham:
can you please fix code style to match

peterharperuk:
I think this fix has now been implemented. Thanks.

PR #1402, opened 2023-05-29
SUBS requires a lo register, so specify that in assembler constraints

kilograham:


kilograham:
I saw this cause a compile failure in the wild (surprised it hasn't before)

PR #1403, opened 2023-05-29
avoid calling find_package for pico-sdk-tools over and over again

kilograham:
see also https://forums.raspberrypi.com/viewtopic.php?t=351995

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,7 +1,8 @@
+# Assemble the version string from components instead of using PICO_SDK_VERSION_STRING, because the version string
+# potentially has a PRE_RELEASE_ID suffix, which will trip up the find_package call.
+find_package(pico-sdk-tools "${PICO_SDK_VERSION_MAJOR}.${PICO_SDK_VERSION_MINOR}.${PICO_SDK_VERSION_REVISION}" QUIET CONFIG CMAKE_FIND_ROOT_PATH_BOTH)
+
 function(_pico_init_pioasm)
-    # Assemble the version string from components instead of using PICO_SDK_VERSION_STRING, because the version string
-    # potentially has a PRE_RELEASE_ID suffix, which will trip up the find_package call.
-    find_package(pico-sdk-tools "${PICO_SDK_VERSION_MAJOR}.${PICO_SDK_VERSION_MINOR}.${PICO_SDK_VERSION_REVISION}" QUIET CONFIG CMAKE_FIND_ROOT_PATH_BOTH)
     if (NOT Pioasm_FOUND)
         set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PICO_SDK_PATH}/tools)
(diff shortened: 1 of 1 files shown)


kilograham:
cc @ndabas

smuehlst:
I had posted the question in the Raspberry Pi forum. I tried a new cmake configuration run with the `find_package_perf` branch checked out, and this reduces the time from over 15 minutes to 44 seconds.

ndabas:
Thanks, I'll just do a quick test with the Windows installer.

peterharperuk:
Looks ok. I'll test on Windows and see if I have any issues.

ndabas:
It'll probably work fine if you have native GCC or Visual Studio C++ compilers installed. However the point of making these changes in the first place was that the pico-setup-windows installer does not install native C++ compilers, so the SDK should be able to detect and use the pre-compiled versions of the SDK tools. Which is failing at the moment but I'll look deeper into it.

ndabas:
It works if I structure the first function like this: (and remove the `find_package(pico-sdk-tools ...` from global scope):

```
function(_pico_init_pioasm)
    if (NOT Pioasm_FOUND)
        # Assemble the version string from components instead of using PICO_SDK_VERSION_STRING, because the version string
        # potentially has a PRE_RELEASE_ID suffix, which will trip up the find_package call.
        find_package(pico-sdk-tools "${PICO_SDK_VERSION_MAJOR}.${PICO_SDK_VERSION_MINOR}.${PICO_SDK_VERSION_REVISION}" QUIET CONFIG CMAKE_FIND_ROOT_PATH_BOTH)
        if (NOT Pioasm_FOUND)
            set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PICO_SDK_PATH}/tools)
            # todo CMAKE_CURRENT_FUNCTION_LIST_DIR ... what version?
            find_package(Pioasm REQUIRED)
        endif()
    endif()
endfunction()
```

Since the call is now in a conditional, it should not be repeated if we have already found the package somehow (but I have not validated that on Linux.)

Similarly for the `pico_add_uf2_output` function:

```
    if (NOT ELF2UF2_FOUND)
        find_package(pico-sdk-tools "${PICO_SDK_VERSION_MAJOR}.${PICO_SDK_VERSION_MINOR}.${PICO_SDK_VERSION_REVISION}" QUIET CONFIG CMAKE_FIND_ROOT_PATH_BOTH)
        if (NOT ELF2UF2_FOUND)
            set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PICO_SDK_PATH}/tools)
            find_package(ELF2UF2)
        endif()
    endif()
```

If this solution works, while I don't have permission to push to this PR branch, I can open a new one with the fix.

kilograham:
Ha, that fix is exactly what i was going to suggest (and i had thought about that after i did it!)

liamfraser (approved):

Issue #1142, opened 2022-12-13
multicore: multicore_fifo_pop_timeout_us doesn't work as expected


DipSwitch:
While in theory the function `multicore_fifo_pop_timeout_us(timeout, *out)` looks workable, in practice it doesn't work as desired.
Because the usage of `__wfe()` it doesn't respect it's given timeout. Since there is no call to `__sev()` on the other core the `__wfe()` is endless since there is no sev instruction executed.

A solution would be to make it a busy wait.

Don't know if SEVONPEND could help with this problem of some sort? Maybe set the timeout in an alarm without interrupt handler and just wakeup via SEVONPEND?

kilograham:
fixed

PR #1404, opened 2023-05-30
use best_effort_wfe_or_timeout instead of wfe for multicore_pop

kilograham:


kilograham:
fixes #1142

kilograham:
moving to 1.6.0 as this is long pre-existing, and probably needs better testing

PR #775, opened 2022-03-30
Add support for SEGGER RTT STDIO

anaoum:
Fixes #774

This PR adds support for using SEGGER RTT as an STDIO driver.

The code for the stdio_rtt implementation is quite simple and copies the structure of the other STDIO drivers. RTT support can be enabled in the same way as the other STDIO drivers, simply add `pico_enable_stdio_rtt(${TARGET_NAME} 1)` to your CMakeLists.txt and call `stdio_init_all()` in your initialisation code.

The official SEGGER_RTT code is included in the lib directory (the latest version, V762c). I have copied all files exactly as-is, despite many of the files being unnecessary/code samples/etc. I did this to make updates to newer versions more simple (just overwrite the entire directory with the latest version). Unfortunately SEGGER does not distribute this code as a git repo which could have been included as a submodule.

Happy to make improvements/adjustments to this implementation based on feedback.

Issue #774, opened 2022-03-30
No support for SEGGER RTT STDIO


anaoum:
It would be nice to be able to use SEGGER RTT as an STDIO driver in the the pico-sdk.

The main advantages of RTT over semihosting are the speed of data transfer and not halting execution.

RTT is widely used and even has support in OpenOCD.

peterharperuk:
Closing this as it's been done. Add the following to your cmake file...

```
pico_enable_stdio_rtt(hello_serial 1)
```

Or change the default.

```
cmake .. -DPICO_STDIO_RTT=1 -DPICO_STDIO_UART=0
```

PR #1411, opened 2023-06-04
Add support for SEGGER RTT STDIO (updated)

adfernandes:
**Supersedes** #775, **fixes** #774 (ping @kilograham, @anaoum)

* Segger's RTT is now included as a submodule

* RTT is enabled via `pico_enable_stdio_rtt(${TARGET_NAME} 1)`
  * `cr/lf` conversion is supported via `PICO_STDIO_RTT_DEFAULT_CRLF`
  * a `pico_stdio_test_rtt` target was added to the test suite
  
Note the strange `#include "../../../lib/rtt/RTT/SEGGER_RTT.c"` in `segger_rtt.c`.

I did this because, as documented in the source file, cmake will not let us change the `COMPILE_OPTIONS` for a single source file if the target has one source file in the current directory tree and another source file in a directory outside the current directory tree.

Sadly, `SEGGER_RTT.c` produces warnings that are treated as errors, and without this fanciness, RTT will not compile with `arm-gnu-toolchain-12.2.mpacbti-rel1`.


kilograham:
This is probably going to annoy you, but perhaps I'm leaning back towards not using a sub module. I think my main objection before was that we had copied an entire repo, however I _think_ that you are only using `SEGGER_RTT.c` and `SEGGER_RTT.h`(and possibly a config header)... it is probably simpler to just copy these files into the same directory.

adfernandes:
Actually, I agree with you about getting rid of the submodule!  😆 

Normally, with RTT, you need to include the `libc` intercepts and overrides for `printf` and `puts` and all that.

But the pico has it's own _layered_ `stdio` implementation, so it just might be that `SEGGER_RTT.c` is all we need...

Let me check and test.

If it's just the source and few includes, I'll copy them in with a note of the revision.

That will allow me to get rid of the `#include "...c"` ugliness because I will be able to _properly_ set the compile options.

Let me look into it tonight / tomorrow.

kilograham (approved):

kilograham (approved):

PR #1417, opened 2023-06-08
Update svd using latest version of generation tools

### LINKED ISSUE #1076: Please validate rp2040.svd against CMSIS-SVD.xsd
When passing rp2040.svd to a [xml schema validator](https://www.utilities-online.info/xsdvalidation) against [CMSIS-SVD](https://github.com/ARM-software/CMSIS/blob/master/CMSIS/Utilities/CMSIS-SVD.xsd), it appears that there's many issue, like mandatory field missing, incorrect order of elements in a sequence etc.

Please validate your document.

liamfraser:
Should fix #1076. Can @andygpz11 verify this is valid before merging.


andygpz11 (approved):
Following checks, I think this is good to go.  Thanks.

Issue #1304, opened 2023-03-10
Incorrect PICO_FLASH_SPI_CLKDIV for WaveShare RP2040-Zero


ArkBrj:
src\boards\include\boards\waveshare_rp2040_zero.h sets PICO_FLASH_SPI_CLKDIV to 2.
However, according to the board schematics (https://www.waveshare.com/w/upload/4/4c/RP2040_Zero.pdf), this board uses Winbond W25Q16JVUXIQ chip. According to the chip's datasheet (https://www.winbond.com/resource-files/w25q16jv%20spi%20revh%2004082019%20plus.pdf), its max Clock frequency for Read Data instruction (03h) is 50 MHz (Section 10.6 AC Electrical Characteristics). PICO_FLASH_SPI_CLKDIV of 2 sets the clock to around 60 MHz exceeding the spec.
In my experiments with various binaries I see that the code always works right after flashing it, but does not work after restarting the board. This includes the officially released binary for picoprobe.
For home-built projects Debug binaries sometimes work  even after reset but Release do not. Sometimes even Debug binaries do not work after reset.
CircuitPython works fine because it uses PICO_FLASH_SPI_CLKDIV of 4 (confirmed by disassembling the binary).
I think PICO_FLASH_SPI_CLKDIV should be set to 4 in SDK for this board.

AlynxZhou:
I meet this problem too with 2 of 3 RP2040-Zeros I bought, a interesting thing is I found setting `PICO_XOSC_STARTUP_DELAY_MULTIPLIER` to `64` also fixes this bug, I think it makes program waits longer before startup.

But your words are more reasonable, and I tried your suggestion, it does fix the bug, so I submit a PR according to this issue. Thanks!

lurch:
In case it's useful, see the related discussion in #401

AlynxZhou:
> In case it's useful, see the related discussion in #401

Thanks!

My test shows that only change `PICO_FLASH_SPI_CLKDIV` to 4 is enough, there is no need to change both `PICO_FLASH_SPI_CLKDIV` and `PICO_XOSC_STARTUP_DELAY_MULTIPLIER`.

kilograham:
merged into develop

zwenergy:
[...]

PR #1421, opened 2023-06-11
Fix waveshare_rp2040_zero USB unrecognize after unplug and replug

AlynxZhou:
Some RP2040-Zero boards from WaveShare can only be recognized via USB after flashing UF2 into it, and if you unplug and replug it, nothing happens on both USB host and device, RP2040-Zero seems just not booting.

According to @ArkBrj, it seems that setting `PICO_FLASH_SPI_CLKDIV` to `2` makes the clock chip exceed the spec, setting it to `4` fixes this bug.

I tested it with the blink program in pico-examples, this does fix the bug for me.

Fixes <https://github.com/raspberrypi/pico-sdk/issues/1304>.

This also has a minor change to remove a trailing space by my editor.

### DIFF
--- a/src/boards/include/boards/waveshare_rp2040_zero.h
+++ b/src/boards/include/boards/waveshare_rp2040_zero.h
@@ -64,7 +64,7 @@
 #define PICO_BOOT_STAGE2_CHOOSE_W25Q080 1
 
 #ifndef PICO_FLASH_SPI_CLKDIV
-#define PICO_FLASH_SPI_CLKDIV 2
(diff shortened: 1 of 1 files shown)


AlynxZhou:
I am not sure whether the same fix should be applied to other WaveShare boards, I am just a poor customer who bought 3 boards from WaveShare and spent 2 weeks to find why 2 of my boards cannot be recognized after unplug & replug...

@EngineerWill I see you are the original committer of WaveShare boards files, maybe you could test other boards? Thanks.

kilograham (approved):

PR #1424, opened 2023-06-13
Add a cmake function for setting IP addresses

peterharperuk:
The fix for the following issue adds some macros to set default ip addresses. These are expressed in hex which is a bit non-obvious to set. So add a macro to convert from a string to the hex representation.

https://github.com/georgerobotics/cyw43-driver/issues/41

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/CMakeLists.txt
+++ b/src/rp2_common/pico_cyw43_driver/CMakeLists.txt
@@ -89,5 +89,32 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
                 )
     endif()
 
+    # Set an ip address in a compile definition
+    # target name, target type, compile definition name to set then address in a string
+    # This can be used to set the following compile definitions
+    # CYW43_DEFAULT_IP_ADDRESS
+    # CYW43_DEFAULT_IP_MASK
+    # CYW43_DEFAULT_IP_GATEWAY
+    # CYW43_DEFAULT_IP_DNS
+    # CYW43_DEFAULT_AP_IP_ADDRESS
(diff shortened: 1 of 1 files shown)


lurch:
Is it worth clarifying that this is for IPv4 addresses?

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -89,5 +89,32 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
...
+    # Set an ip address in a compile definition
+    # target name, target type, compile definition name to set then address in a string
+    # This can be used to set the following compile definitions
+    # CYW43_DEFAULT_IP_ADDRESS
+    # CYW43_DEFAULT_IP_MASK
+    # CYW43_DEFAULT_IP_GATEWAY
+    # CYW43_DEFAULT_IP_DNS
+    # CYW43_DEFAULT_AP_IP_ADDRESS
+    # e.g. target_set_ip_address(picow_tcpip_server_background PRIVATE CYW43_DEFAULT_IP_ADDRESS "10.3.15.204")
+    function(target_set_ip_address TARGET_LIB TARGET_TYPE DEF_NAME IP_ADDRESS_STR)
```
kilograham, line 101: Function should probably be `pico_set_ip_address`

kilograham (changes requested):

kilograham (changes requested):
I'm a bit on the fence about the name; whilst the other `pico_set_` do generally add compile definitions, they do not offer the chance to do so multiple times...

IDK; `pico_configure_ip4_address`? `pico_add_ip4_address`?

peterharperuk:
pico_configure_ip4_address sounds ok

kilograham (approved):

PR #1426, opened 2023-06-13
Add missing "default defines"

lurch:
To match the documented PICO_CONFIG default values


kilograham (approved):
Superfluous but perhaps clearer, but needed to make doc generation happy, so merging

PR #1427, opened 2023-06-13
Move multicore_lockout victim initialzied tracking to pico_multicore

kilograham:
Added new  multicore_lockout_victim_is_initialzied method, so user initialization of the multicore_lockout independent of pico_flash will work with pico_flash


peterharperuk (approved):
Looks ok. I really need to write a bt example that uses this.

PR #1444, opened 2023-06-30
Fix spelling mistake, add documentation

### LINKED ISSUE #1443: Documentation for gpio_add_raw_irq_handler functions
The documentation for the gpio_add_raw_irq_handler*() functions does not mention that there is a limitation on the number of handlers that can be added. The limitation is due to the call to irq_add_shared_handler(), which has a limit that can be configured via PICO_MAX_SHARED_IRQ_HANDLERS. The default value in the SDK is a maximum of four handlers, of which two or three may be used by USB and the cyw43 driver. I think it would be helpful to mention the limitation since anybody installing more than one or two handlers is likely to exceed the default limit.

hubiscode:
Add documentation to gpio_add_raw_irq_handler*() functions that references irq_add_shared_handler() and notes that the functions will assert if the maximum number of shared handlers configured via PICO_MAX_SHARED_IRQ_HANDLERS would be exceeded.

Fix spelling mistake in static variable in irq.c (hander vs handler).

Fixes #1443


peterharperuk (approved):
Looks good to me

kilograham (approved):

PR #1445, opened 2023-07-01
Fix #1442: Allow adding alternative .gatt file import paths 

### LINKED ISSUE #1442: Please allow custom .gatt #import paths in pico_btstack_make_gatt_header()
I am trying to develop a custom Bluetooth LE GATT service for the Pico W that can be imported into another `.gatt` file using `#import <my_custom.gatt>`. I don't want to try to add this service to the BlueKitchen Bluetooth stack's existing default paths for `.gatt` files.

The file `${PICO_SDK_PATH}/src/rp2_common/pico_btstack/CMakeLists.txt` contains the function `pico_btstack_make_gatt_header()` that is usually invoked like this:
```
pico_btstack_make_gatt_header(my_executable PRIVATE "${CMAKE_CURRENT_LIST_DIR}/my_gap_device.gatt")
```
This causes the Python script `${PICO_SDK_PATH}/lib/btstack/tool/compile_gatt.py` to be invoked with a path to the `.gatt` file and the path to the header file. I need to the ability to add `-I${MY_CUSTOM_GATT_FILE_PATH}` to the end of the argument list so the tool can find my custom `.gatt` file when it reaches the `#import` line that calls for it. That is, I want my project's CMakeLists.txt to be able to call
```
pico_btstack_make_gatt_header(my_executable PRIVATE "${CMAKE_CURRENT_LIST_DIR}/my_gap_device.gatt" -I${MY_CUSTOM_GATT_FILE_PATH})
```

I have a simple fix for this on my local machine and I would like to submit it as a pull request. Should I submit my change as a fork from `master` or `develop` branch?

rppicomidi:
This fix uses the `ARGN` symbol in CMake to extend the number of arguments you can add to `pico_btstack_make_gatt_header()`. The additional arguments should be of the form `-I${Include path}` so the Python script can find your custom `#import <my_gatt.gatt>` files.

### DIFF
--- a/src/rp2_common/pico_btstack/CMakeLists.txt
+++ b/src/rp2_common/pico_btstack/CMakeLists.txt
@@ -232,8 +232,7 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
 
     # Make a GATT header file from a BTstack GATT file
     # Pass the target library name library type and path to the GATT input file
-    # To add additional directories to the gatt #import path, one "[custom_path]"
-    # argument for each custom_path to the end of the argument list.
+    # To add additional directories to the gatt #import path, add them to the end of the argument list.
     function(pico_btstack_make_gatt_header TARGET_LIB TARGET_TYPE GATT_FILE)
             find_package (Python3 REQUIRED COMPONENTS Interpreter)
             get_filename_component(GATT_NAME "${GATT_FILE}" NAME_WE)
--- a/src/rp2_common/pico_btstack/CMakeLists.txt
+++ b/src/rp2_common/pico_btstack/CMakeLists.txt
@@ -238,21 +238,17 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
             find_package (Python3 REQUIRED COMPONENTS Interpreter)
             get_filename_component(GATT_NAME "${GATT_FILE}" NAME_WE)
             get_filename_component(GATT_PATH "${GATT_FILE}" PATH)
-            set(GATT_BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
-            set(GATT_HEADER "${GATT_BINARY_DIR}/${GATT_NAME}.h")
             set(TARGET_GATT "${TARGET_LIB}_gatt_header")
-            set (ExtraGattPaths "")
-            foreach(arg IN LISTS ARGN)
-                string(PREPEND arg "-I")
-                string(APPEND ExtraGattPaths ${arg} " ")
-            endforeach()
+            set(GATT_BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated/${TARGET_GATT}")
+            set(GATT_HEADER "${GATT_BINARY_DIR}/${GATT_NAME}.h")
+            list(TRANSFORM ARGN PREPEND "-I")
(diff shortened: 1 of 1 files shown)


rppicomidi:
In case it is not clear why this change is useful, there are examples of projects that use this change. See the `CMakeLists.txt` file in the projects https://github.com/rppicomidi/pico-w-ble-midi-server-demo or https://github.com/rppicomidi/ble-midi2usbhost

src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -246,7 +248,7 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
                     DEPENDS ${GATT_FILE}
                     WORKING_DIRECTORY ${GATT_PATH}
                     COMMAND ${CMAKE_COMMAND} -E make_directory ${GATT_BINARY_DIR} &&
-                            ${Python3_EXECUTABLE} ${PICO_SDK_PATH}/lib/btstack/tool/compile_gatt.py ${GATT_FILE} ${GATT_HEADER}
+                            ${Python3_EXECUTABLE} ${PICO_SDK_PATH}/lib/btstack/tool/compile_gatt.py ${GATT_FILE} ${GATT_HEADER} ${ARGN}
```
peterharperuk, line 251: It's a bit "clunky" passing in the arguments for the tool. It might be better if ARGN were just treated as a list of directories?

rppicomidi: I am sorry. I don't understand your comment. If there is a better way to achieve the same functionality so I can let the
compile_gatt.py command find my custom .gatt file, then that is the change I am requesting. This change was the
only way I could think of to do it without touching the python code, but I am no expert in CMake.

peterharperuk: I'm just suggesting a small change to not have to pass in the "-I" as part of the argument

rppicomidi: Is my new commit what you meant?

peterharperuk: close, I think you can just replace all that with...
list(TRANSFORM ARGN PREPEND "-I ")

src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -232,21 +232,27 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
 
     # Make a GATT header file from a BTstack GATT file
     # Pass the target library name library type and path to the GATT input file
+    # To add additional directories to the gatt #import path, one "[custom_path]"
+    # argument for each custom_path to the end of the argument list.
```
lurch, line 236: Perhaps this description could be simplified to something like:
"To add additional directories to the gatt #import path, add them to the end of the argument list."
?

src/rp2_common/pico_btstack/CMakeLists.txt
peterharperuk: Hopefully fixed

src/rp2_common/pico_btstack/CMakeLists.txt
peterharperuk: done

peterharperuk (approved):

PR #1462, opened 2023-07-23
Add gpio_set_function_mask function

inventor02:
Similar to gpio_init_mask, gpio_set_function_mask allows setting multiple GPIO functions with one call.


peterharperuk:
LGTM

peterharperuk:
@kilograham Ok to merge?

kilograham (approved):

lurch:
...although _unlike_ the other `_masked` functions, this works by iterating over the GPIOs in a loop, rather than doing a single register-write. I dunno if this is an important-enough distinction to be worth noting in the Doxygen-comment?

PR #1464, opened 2023-07-28
Fix: Pad the hour field with leading zeros in datetime_to_str()

e-ieki:
In the datetime_to_str() function, the hour field was not zero-padded while other time fields (minutes and seconds) were padded with leading zeros. This caused a lack of visual consistency in the output, potentially leading to confusion in interpreting the data.

This commit fixes the issue by adding zero-padding to the hour field, ensuring a consistent and easily readable datetime format.


Reneg973:
Wouldn't it mean that the day also has to be zero-padded?

e-ieki:
@Reneg973 
Thank you. You might be right. Upon second thought, it seems that the original code's date format is also quite common and not as problematic as I initially thought. I would like to withdraw my pull request.

PR #1484, opened 2023-08-27
PioAsm: Add a maximum version to supported CMake versions to avoid deprecation warning

### LINKED ISSUE #1483: CMake deprecation warning displayed when pioasm is built
I lately compiled one of my projects using cmake 3.27.4 and was greeted with this deprecation notice:
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

I tracked this down to the CMakeLists.txt file of pioasm:
https://github.com/raspberrypi/pico-sdk/blob/master/tools/pioasm/CMakeLists.txt#L1

The options I see:
* Either update the minimum version required `VERSION 3.5`. The problem might re-appear in the future
* Add a newer version that it has been also tested with so CMake is aware that the old, compatibility behaviour is not longer required: `VERSION 3.4...3.27` (or some other version between 3.5 and 3.27)

I'll create a PR to implement option 2 since pioasm builds fine for me using cmake 3.27.4. If you want to specify an older version here that's fine for me as well

kripton:
CMake introduced a deprecation warning lately. PioAsm's CMakeLists file triggers it and is easy to fix.

Fixes #1483

### DIFF
--- a/tools/pioasm/CMakeLists.txt
+++ b/tools/pioasm/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.4...3.27)
 project(pioasm CXX)
 
 set(CMAKE_CXX_STANDARD 11)


kripton:
Fixed already by #1477

PR #1495, opened 2023-09-07
Adding I2C Burst Reading/Writing feature

AnhNhan2803:
I realize that the C-SDK of Raspberry Pi Pico does not support the burst mode operation in I2C (only supports normal and repeated start in reading/writing). As burst mode is an important I2C feature that is sometimes required in communication with several I2C slaves. e.g. MAX3010X. Hopefully, we can consider adding this feature to the current SDK.

Thank you

### DIFF
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -131,7 +131,7 @@ void i2c_set_slave_mode(i2c_inst_t *i2c, bool slave, uint8_t addr) {
 }
 
 static int i2c_write_blocking_internal(i2c_inst_t *i2c, uint8_t addr, const uint8_t *src, size_t len, bool nostop,
-                                       check_timeout_fn timeout_check, struct timeout_state *ts, bool burst) {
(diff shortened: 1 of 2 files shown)


peterharperuk:
Any idea how we would test this?

kilograham:
a bit confused at a very qucik glance; can't you just set `nostop` to true in the existing API?

AnhNhan2803:
@kilograham In the original implementation with "nostop" set to true, the start condition can repeat without sending a stop, allowing for repeated starts within a transfer. However, in burst mode, there is only one start and one stop condition, with several commands or data transfers happening continuously in between without any repeated starts or stops.

peterharperuk:
@kilograham Is there a reason this didn't make it into the release?

kilograham:
because it had merge conflicts

peterharperuk:
Actually, it was my mistake.

peterharperuk (approved):

peterharperuk:
Example here https://github.com/raspberrypi/pico-examples/pull/512

peterharperuk:
@kilograham Apologies - could you review this again?

kilograham (approved):

PR #1499, opened 2023-09-08
pico_cyw43_driver/cyw43_bus_pio_spi.c: make SPI_PROGRAM_NAME and CLOC…

### LINKED ISSUE #1498: Not possible to change CYW43 PIO settings in makefile
In its current form there is no (obvious) way to have a project with the pico-sdk as a dependency, where the project would be able to define things like the SPI_PROGRAM_NAME or the CLOCK_DIV - things that are essential to e.g. have a project that is power efficient.

It would be nice if it would be possible to define these things in a makefile. The fix is trivial and a pull request will follow, but the instructions said that an issue should be logged first.

thingsconnected:
Fixes #1498

This patch allows to override the default value for SPI_PROGRAM_NAME and CLOCK_DIV in src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c by defining CYW43_SPI_PROGRAM_NAME and/or CYW43_PIO_CLOCK_DIV.

This is useful when other cpu speeds are involved, e.g. running the Pi Pico W at 12 MHz from xosc.

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -29,13 +29,22 @@
 #define CS_PIN 25u
 #define IRQ_SAMPLE_DELAY_NS 100
 
-#define SPI_PROGRAM_NAME spi_gap01_sample0
+#ifdef CYW43_SPI_PROGRAM_NAME
+#define SPI_PROGRAM_NAME CYW43_SPI_PROGRAM_NAME
+#else
+//#define SPI_PROGRAM_NAME spi_gap0_sample1 // for lower cpu speed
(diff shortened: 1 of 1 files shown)


matsobdev:
It would be nice to have full range of cpu speed at runtime for Pico W.

thingsconnected:
I fully agree @matsobdev, the potential for power efficient applications is huge. But it involves a lot of changes. Running from xosc at 12 MHz with pll_sys disabled for example, is currently not possible using build settings alone. I think the first step in having the full range of cpu speeds available, would be to make the cpu speed and everything depending on it configurable at build time.

ai6aj:
See also #1521 which allows the default clock divisor to be overridden at runtime, which is useful for Micropython etc.  Both PRs look like they should merge without issue.

kilograham (commented):
Would be nice to add a PICO_CONFIG comment so it is added to the docs

kilograham (approved):

PR #1504, opened 2023-09-10
IRQ enabling guard (when no callback set)

### LINKED ISSUE #1721: Interrupt enabling without callback could use an assert
When the user wants to enable the interrupt, Check whether a raw interrupt is added or at least a default one is set. See my problems here: https://forums.raspberrypi.com/viewtopic.php?p=2134008

MrJake222:
When the user wants to enable the interrupt, I added a check whether a raw interrupt is added or at least a default one is set. See my problems here: https://forums.raspberrypi.com/viewtopic.php?p=2134008

edit: fixes #1721


kilograham:
can you please open an issue for this as well; we may want to do a different fix ourselves

MrJake222:
done

kilograham (approved):

Issue #1297, opened 2023-03-02
Bluetooth problems when overclocking


MichaelBell:
When overclocking to 250MHz, I can only connect to bluetooth audio the very first time after powering on the Pico W.  

After a reset (pulling RUN low), I can't reconnect.  At lower clock speeds this works fine.

Changing cyw43_bus_pio_spi.c as follows:
```
-    sm_config_set_clkdiv_int_frac(&config, CLOCK_DIV, CLOCK_DIV_MINOR);
+    const int clock_mhz = clock_get_hz(clk_sys) / MHZ;
+    const int clock_div = (clock_mhz < 170) ? 2 : (1 + clock_mhz / 85);
+    sm_config_set_clkdiv_int_frac(&config, clock_div, 0);
```
fixes it, but given this problem must be related just to some kind of reset, and I don't recall having issues with a 250MHz overclock doing WiFi with a Pico W, this fix seems extreme.

peterharperuk:
> can only connect to bluetooth audio

Which direction? Source or sink or both? i.e. Is the Pico playing audio or is it connecting to a speaker to play audio? The latter I assume?

MichaelBell:
Sink - Pico playing audio from my iPhone.  I'm running https://github.com/Gadgetoid/galactic-bluetooth-audio but with an overclock set in `main()`

peterharperuk:
I was going to mark this as fixed by https://github.com/raspberrypi/pico-sdk/pull/1521
But overclocking to 250 seems to work fine for me without having to adjust the pio divisor. It's somewhat curious that the a2dp_sink_demo now needs to be overclocked to work properly but that's another issue.
I don't now why pulling run low would affect things. I tried this and it reconnected fine.
I'll mark this as fixed as your original issue (having to hack the code to set the divisor) has now been fixed and I can't seem to repro the problem with a2dp_sink_demo. Please reopen if you disagree.

MichaelBell:
As this is in the realms of overclocking maybe it was related to the exact hardware I was running on.
Anyway, now there is the dynamic divider option for the CYW43 PIO that could easily be worked around if necessary so I agree with closing this issue.

Issue #1392, opened 2023-05-25
Communication with cyw43 fails at system clock speeds <= 60MHz


MichaelBell:
It's sometimes useful to run the RP2040 quite slowly to reduce power consumption.  I'm not sure whether it is reasonable to expect Wifi to continue to work at low clock speeds - but I think it is reasonable to at least expect to be able to use the CYW43 GPIO.

However, at around 60MHz and below setting/reading the GPIO from the CYW43 doesn't work.  This means you can't determine whether VBUS is present, and can't control the LED.

A simple repro is to add set_sys_clock_48mhz() to the top of the pico_w blink example.

```diff
index 149b477..600e2a2 100644
--- a/pico_w/wifi/blink/picow_blink.c
+++ b/pico_w/wifi/blink/picow_blink.c
@@ -8,6 +8,7 @@
 #include "pico/cyw43_arch.h"

 int main() {
+    set_sys_clock_48mhz();
     stdio_init_all();
     if (cyw43_arch_init()) {
         printf("Wi-Fi init failed");
```

thingsconnected:
WiFi works flawlessly for us with the Pi Pico running at 12 MHz from xosc. Set clock divider at 1 and spi program spi_gap01_sample0. Running at 48 MHz (to enable USB stdio) clock divider 2 and spi program spi_gap0_sample1 can be used, IIRC.

peterharperuk:
I seem to be able to be able to use cyw43 to flash the led when running from xosc by calling the new function cyw43_set_pio_clock_divisor(1, 0). I didn't seem to have to set CYW43_SPI_PROGRAM_NAME? Anyway - I think this is resolved now as you shouldn't have to hack the code?

PR #1521, opened 2023-10-22
Add cyw43_spi_set_clock_divisor() to allow CYW43 to work with over/underclocked RP2040

ai6aj:
Fixes issues #1297 and #1392

This is a straightforward patch to allow the Pico W to work with the WiFi/BT module when heavily over- or underclocked.  The previous code had the SPI clock divisor hardcoded to a value of 2, this allows the programmer to explicitly set it to any value they'd like.

In both instances, simply set the desired clock divisor PRIOR to initializing the WiFi module:

    set_sys_clock_khz(48000, true);      // cyw43 init will hang or fail, see issue #1392
    cyw43_spi_set_clock_divisor(1,0);   // Speed up the SPI bus so it will work
    if (cyw43_arch_init_with_country(CYW43_COUNTRY_USA)) { ...
    
or

    set_sys_clock_khz(266000, true);      // cyw43 init will fail for reasons similar to #1297
    cyw43_spi_set_clock_divisor(4,0);     // Slow down the SPI bus 
    if (cyw43_arch_init_with_country(CYW43_COUNTRY_USA)) { ...

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -84,6 +84,16 @@ typedef struct {
 
 static bus_data_t bus_data_instance;
 
+int cyw43_spi_clock_div_int = CLOCK_DIV;
+int cyw43_spi_clock_div_frac = CLOCK_DIV_MINOR;
+
+// Allow the user to adjust the SPI clock speed 
+// prior to initialization.
+void cyw43_spi_set_clock_divisor(int clock_div_int, int clock_div_frac) {
+    cyw43_spi_clock_div_int = clock_div_int;
+    cyw43_spi_clock_div_frac = clock_div_frac;
+}
+
 int cyw43_spi_init(cyw43_int_t *self) {
     // Only does something if CYW43_LOGIC_DEBUG=1
     logic_debug_init();
@@ -117,7 +127,8 @@ int cyw43_spi_init(cyw43_int_t *self) {
     bus_data->pio_offset = pio_add_program(bus_data->pio, &SPI_PROGRAM_FUNC);
     pio_sm_config config = SPI_PROGRAM_GET_DEFAULT_CONFIG_FUNC(bus_data->pio_offset);
 
-    sm_config_set_clkdiv_int_frac(&config, CLOCK_DIV, CLOCK_DIV_MINOR);
+    sm_config_set_clkdiv_int_frac(&config, cyw43_spi_clock_div_int, cyw43_spi_clock_div_frac);
+    
     hw_write_masked(&padsbank0_hw->io[CLOCK_PIN],
                     (uint)PADS_DRIVE_STRENGTH << PADS_BANK0_GPIO0_DRIVE_LSB,
                     PADS_BANK0_GPIO0_DRIVE_BITS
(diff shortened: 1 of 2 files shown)


lurch:
I guess this is similar-but-different to #1499 ?

ai6aj:
It's similar to #1499, but having an actual SDK call means that Micropython users (and Lua, JS, etc.) aren't stuck with whatever SPI clock speed was chosen at compile time.  That patch looks like it's entirely compatible with this one and I rather like the option to do it either way.

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -84,6 +84,16 @@ typedef struct {
 
 static bus_data_t bus_data_instance;
 
+int cyw43_spi_clock_div_int = CLOCK_DIV;
```
peterharperuk, line 87: coudn't these be static const?

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -84,6 +84,16 @@ typedef struct {
...
+int cyw43_spi_clock_div_frac = CLOCK_DIV_MINOR;
+
+// Allow the user to adjust the SPI clock speed 
+// prior to initialization.
+void cyw43_spi_set_clock_divisor(int clock_div_int, int clock_div_frac) {
```
peterharperuk, line 92: I wonder if that's the right name for the function. Maybe cyw43_set_pio_clock_divisor?

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
ai6aj: They are set by cyw43_spi_set_clock_divisor() so they can't be const, but they should be static.  I've made the changes in the latest commit.

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
ai6aj: Good call, done

FrankLaterza:
Any updates on this? Has anyone found an alternative I could use so users don't have to edit the SDK when they want to overclock their pi pico w?

peterharperuk:
@FrankLaterza does this change work for you?

FrankLaterza:
After overclocking to 400MHZ this change worked perfect for me.

kilograham (commented):
Is this something that needs to be dynamic? I'd vote for making this a PICO_ define which defaults to the old values

ai6aj:
> Is this something that needs to be dynamic? I'd vote for making this a PICO_ define which defaults to the old values

If the system clock wasn't dynamically adjustable I'd agree that it should be a simple #define, but Micropython and other VMs expose set_system_clock_mhz to the user and need to be able to adjust the SPI clock at runtime.  Otherwise you end up with weird cases like "the Pico W works fine in Micropython when under- or overclocked, unless you want to use WiFi, in which case you'll have to rebuild the entire VM from scratch."

peterharperuk:
Your change has more or less been implemented. You need to define CYW43_PIO_CLOCK_DIV_DYNAMIC and then cyw43_set_pio_clock_divisor is available. Or else at build time you can use CYW43_PIO_CLOCK_DIV_INT and CYW43_PIO_CLOCK_DIV_FRAC. This is all in pico_cyw43_driver. There's an example here https://github.com/raspberrypi/pico-examples/blob/master/pico_w/wifi/blink/picow_blink_slow_clock.c

PR #1530, opened 2023-10-30
Rename configNUMBER_OF_CORES to configNUM_CORES

### LINKED ISSUE #1526: Assertion error when calling cyw43_arch_init() in FreeRTOS
When I try to call `cyw43_arch_init()` in a FreeRTOS task I get the following error printed to the serial:
```
assertion "get_core_num() == async_context_core_num(cyw43_async_context)" failed: file "......../pico-sdk/src/rp2_common/pico_cyw43_driver/cyw43_driver.c", line 53, function: cyw43_irq_init
```
I have the code on my repo [here](https://github.com/ArcaEge/globe-radio/tree/28/10/2023), however it is in a bit of a mess so here's a cleaned up version of the main file:
```
#include "FreeRTOS.h"
#include <task.h>
#include "pico/cyw43_arch.h"
#include <stdio.h>
#include "pico/stdlib.h"

void wifiConnectTask(void *pvParameters) {
    if (cyw43_arch_init()) {
        printf("failed to initialise\n");
        return;
    }
    cyw43_arch_enable_sta_mode();
    printf("Connecting to Wi-Fi...\n");
    if (cyw43_arch_wifi_connect_timeout_ms(WIFI_SSID, WIFI_PASSWORD, CYW43_AUTH_WPA2_AES_PSK, 30000)) {
        printf("failed to connect.\n");
        exit(1);
    } else {
        printf("Connected.\n");
    }
    while (true) {
        vTaskDelay(500);
    }
}

int main() {
    stdio_init_all();

    TaskHandle_t wifiHandle;

    xTaskCreate(wifiConnectTask, "wifiConnect", configMINIMAL_STACK_SIZE, NULL, 1, &wifiHandle);

    // Note: tried with and without this line, no difference
    vTaskCoreAffinitySet(wifiHandle, 1);
    
    vTaskStartScheduler();

    while (true)
        vTaskDelay(500);

    cyw43_arch_deinit();
    return 0;
}
```
Possibly related to #1478?

In addition, I've tried this with both the master and develop branches of the SDK; it is broken with both of them.

peterharperuk:
The Freertos SMP branch has been merged into main with configNUM_CORES renamed to configNUMBER_OF_CORES

Fixes #1526

### DIFF
--- a/src/rp2_common/pico_async_context/async_context_freertos.c
+++ b/src/rp2_common/pico_async_context/async_context_freertos.c
@@ -12,7 +12,7 @@
 
 #include "semphr.h"
 
-#if configNUM_CORES > 1 && !defined(configUSE_CORE_AFFINITY)
+#if configNUMBER_OF_CORES > 1 && !defined(configUSE_CORE_AFFINITY)
 #error async_context_freertos requires configUSE_CORE_AFFINITY under SMP
 #endif
 
@@ -125,7 +125,7 @@ bool async_context_freertos_init(async_context_freertos_t *self, async_context_f
         async_context_deinit(&self->core);
         return false;
     }
-#if configNUM_CORES > 1
+#if configNUMBER_OF_CORES > 1
     UBaseType_t core_id = config->task_core_id;
     if (core_id == (UBaseType_t)-1) {
         core_id = portGET_CORE_ID();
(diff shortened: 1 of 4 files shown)


jaguilar:
It would be really great if this got merged. I just spent a bunch of time trying to get things working, only to figure out that it all wasn't working because the affinity set wasn't being updated since the pico SDK thinks that there is only one core (since it's looking at the wrong #define).

peterharperuk:
Fixed

PR #1539, opened 2023-11-08
rp2_common/pico_standard_link: linker script fixes

### LINKED ISSUE #1538: Linker script bug, generating invalid ELF files that confuse binutils
Due to the order of sections in the linker script, an ELF is generated with a strange layout, that confuses objcopy. This is easily solved by rearranging these sections.

Normally, objcopy is not used on ELF files within the pico-sdk, but there are cases where it is, such as when using the `picowota` firmware loader.

A pull request will follow.

thingsconnected:
The resulting elf binaries contained an unusual section that leads to an error when `objcopy` attempts to update a section (e.g. when using the picowota combined build mechanism).

This is due to the order of sections, where two `RAM` sections were separate by a `RAM AT>FLASH` section. The documentation of `ld` describes that the `AT>FLASH` is effectively "sticky", so that the next `RAM` section behaves as if `AT>FLASH` was specified (which is not intended). By moving the RAM-only sections together, this issue is solved.

Fixes #1538 (see this issue for a detailed analysis)

### DIFF
--- a/src/rp2_common/pico_standard_link/memmap_blocked_ram.ld
+++ b/src/rp2_common/pico_standard_link/memmap_blocked_ram.ld
@@ -117,8 +117,13 @@ SECTIONS
     __binary_info_end = .;
     . = ALIGN(4);
 
-   .ram_vector_table (NOLOAD): {
+    .ram_vector_table (NOLOAD): {
         *(.ram_vector_table)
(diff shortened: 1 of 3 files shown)


kilograham (approved):
Nice catch...

Issue #1453, opened 2023-07-07
printf deadlock usb using two tasks


adri-sass:
Hello, I've a deadlock using the printf in two differents thread with the usb stdio configured.
I'm using:
- FreeRTOS on "smp" branch
- Pico sdk 1.5.1

A simple main that causes the issue is the following:

``` C++
#include <stdio.h>
#include "pico/stdlib.h"
#include "FreeRTOS-Kernel/include/FreeRTOS.h"
#include "FreeRTOS-Kernel/include/task.h"

void TaskA(void*)
{
    int VAL = 0;   
    while(true)
    {
        printf("TaskA VAL = %d\n", VAL++);
    }
}

void TaskB(void*)
{
    int VAL = 0;
    while(true)
    {
        printf("TaskB VAL = %d\n", VAL++);
    }
}


int main() {
    stdio_init_all();
    sleep_ms(1000);
    printf("STARTING\n");
    #ifdef LIB_PICO_MULTICORE
        printf("LIB_PICO_MULTICORE defined\n");
    #else
        printf("LIB_PICO_MULTICORE not defined\n");
    #endif
    xTaskCreate(TaskA, "Bouncing Task1", configMINIMAL_STACK_SIZE, nullptr, 1, nullptr);
    xTaskCreate(TaskB, "Bouncing Task2", configMINIMAL_STACK_SIZE, nullptr, 1, nullptr);
    
    
    vTaskStartScheduler();
    
    printf("Error\n");
    while(true)
    {
        ;;
        sleep_ms(1000);
    }
}
```

The CMake is the following:

``` CMake
cmake_minimum_required(VERSION 3.13)

set(PICO_BOARD pico_w)
set(FREERTOS_KERNEL_PATH ./FreeRTOS-Kernel)
# initialize pico-sdk from submodule
# note: this must happen before project()
include(pico-sdk/pico_sdk_init.cmake)
include(pico_FreeRTOS_Kernel_import.cmake)

project(PICO_OS_TEST)

# initialize the Raspberry Pi Pico SDK
pico_sdk_init()

add_executable(pico_usb_print_deadlock_test main_issue_poc.cpp)

target_include_directories(pico_usb_print_deadlock_test PRIVATE ${CMAKE_CURRENT_LIST_DIR} )
[...]

PR #1549, opened 2023-11-17
ISR versions of mutex_try_enter and mutex_exit

daveythacher:
#1453 Extends mutex API to support holding the spin lock. This can be used to prevent deadlock from competing core.

### DIFF
--- a/src/common/pico_sync/include/pico/mutex.h
+++ b/src/common/pico_sync/include/pico/mutex.h
@@ -47,6 +47,7 @@ extern "C" {
 typedef struct __packed_aligned  {
     lock_core_t core;
     lock_owner_id_t owner;      //! owner id LOCK_INVALID_OWNER_ID for unowned
+    uint32_t save;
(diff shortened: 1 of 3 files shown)


daveythacher:
Note this does cause possible priority inversion, if used incorrectly.

daveythacher:
Does disabling interrupts on core 1, disable them on core 0? This patch does not cover the case of core 0 ISR and core 1 ISR racing against each other. In which case the patch is not worth much. (Even though this is horrible practice.) This would be new behavior and is capable of creating an issue like 1453.

Note the ISR version is capable of creating a slight blocking behavior in non ISR version, which may need to be documented.

daveythacher:
See #1580

daveythacher:
What happened to my changes? spin_lock_blocking is not supposed to be there.

Note a queue makes more sense.

kilograham:
if you want a mutex which holds the spin lock, use critical_section

Issue #1563, opened 2023-11-28
Documentation issues when upgrading to Doxygen 1.9.8


mudge:
The documentation for this SDK builds fine with Doxygen 1.8.17 (as indicated by [the comment in the `DoxygenLayout.xml`](https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/docs/DoxygenLayout.xml#L2)) but has at least two issues when upgrading to the latest version of Doxygen: 1.9.8, released on 25 Aug 2023.

### The "API Documentation" tab is missing from the main menu and no `modules.html` is output

| Doxygen 1.8.17 | Doxygen 1.9.8 |
| -- | -- |
| <img width="1398" alt="Screenshot 2023-11-28 at 10 47 25" src="https://github.com/raspberrypi/pico-sdk/assets/287/243a5fd7-a886-4eb6-ad28-d18dcc3d4b68"> | <img width="1398" alt="Screenshot 2023-11-28 at 10 49 23" src="https://github.com/raspberrypi/pico-sdk/assets/287/f297f321-d903-4397-9d15-dd844ebf27fd"> |

Using the latest default `DoxygenLayout.xml` produced by Doxygen 1.9.8's `doxygen -l` restores the contents of the old `modules.html` but now refers to it as "Topics" instead. Updating [our `DoxygenLayout.xml`](https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/docs/DoxygenLayout.xml) restores the previous page (albeit under a new URL):

```diff
-    <tab type="modules" visible="yes" title="API Documentation" intro="These are the libraries supplied in the Raspberry Pi Pico SDK"/>
+    <tab type="topics" visible="yes" title="API Documentation" intro="These are the libraries supplied in the Raspberry Pi Pico SDK"/>
+    <tab type="modules" visible="no" title="" intro=""/>
```

### Brief descriptions are no longer extracted for most modules by default

| Doxygen 1.8.17 | Doxygen 1.9.8 |
| -- | -- |
[...]

PR #1564, opened 2023-11-28
Fix documentation issues when upgrading Doxygen to 1.9.8

mudge:
See https://github.com/raspberrypi/pico-sdk/issues/1563

Between Doxygen 1.8.17 and Doxygen 1.9.8, the top-level "modules" tab type in the Doxygen layout has changed such that our "API Documentation" section is no longer generated and is instead considered of the type "topics". Update our layout XML to reflect this change and restore this critical section of the documentation.

As of Doxygen 1.8.19 (specifically, https://github.com/doxygen/doxygen/commit/327423e217337bf4e64515aba6cf78b9877bc165), the first line of a comment is no longer automatically extracted as a brief description unless we enable the JAVADOC_AUTOBRIEF setting. See https://www.doxygen.nl/manual/config.html#cfg_javadoc_autobrief:
> If the `JAVADOC_AUTOBRIEF` tag is set to `YES` then doxygen will interpret the first line (until the first dot) of a Javadoc-style comment as the brief description. If set to `NO`, the Javadoc-style will behave just like regular Qt-style comments (thus requiring an explicit @brief command for a brief description.)

### DIFF
--- a/docs/DoxygenLayout.xml
+++ b/docs/DoxygenLayout.xml
@@ -3,7 +3,8 @@
   <!-- Navigation index tabs for HTML output -->
   <navindex>
     <tab type="mainpage" visible="yes" title="Introduction"></tab>
-    <tab type="modules" visible="yes" title="API Documentation" intro="These are the libraries supplied in the Raspberry Pi Pico SDK"/>
+    <tab type="topics" visible="yes" title="API Documentation" intro="These are the libraries supplied in the Raspberry Pi Pico SDK"/>
+    <tab type="modules" visible="no" title="" intro=""/>
     <tab type="user" url="@ref examples_page" visible="yes" title="Examples" intro="Links to SDK examples"/>
     <tab type="usergroup" url="@ref weblinks_page" visible="yes" title="Additional Documentation" intro="Links to datasheets and documentation">
       <tab type="user" url="https://rptl.io/pico-datasheet" visible="yes" title="Raspberry Pi Pico Datasheet" intro=""/>
--- a/docs/Doxyfile.in
+++ b/docs/Doxyfile.in
@@ -52,6 +52,9 @@ EXTRACT_ALL            = NO
 ALWAYS_DETAILED_SEC    = NO
 #REPEAT_BRIEF           = NO
 
+# Needed to automatically extract implicit brief descriptions in recent versions of Doxygen
+JAVADOC_AUTOBRIEF      = YES
+
 # Need these next options to ensure that functions with modifiers do not confuse the Doxygen parser.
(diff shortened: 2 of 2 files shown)


docs/DoxygenLayout.xml
```diff
@@ -3,7 +3,8 @@
   <!-- Navigation index tabs for HTML output -->
   <navindex>
     <tab type="mainpage" visible="yes" title="Introduction"></tab>
-    <tab type="modules" visible="yes" title="API Documentation" intro="These are the libraries supplied in the Raspberry Pi Pico SDK"/>
+    <tab type="topics" visible="yes" title="API Documentation" intro="These are the libraries supplied in the Raspberry Pi Pico SDK"/>
```
mudge, line 6: Frustratingly, this change will break building the docs on Doxygen 1.8.17 as this seems to be a breaking change in Doxygen itself.

aallan: Ah it's a README thing then. 'kay.

mudge:
This no longer seems to work as expected with the very latest version of Doxygen (1.10.0) and I've discovered more subtle bugs with brief description extraction using `JAVADOC_AUTOBRIEF`: e.g. the brief description for https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/hardware_spi/include/hardware/spi.h#L100-L103 is incorrectly extracted as:

```
<para>Initialise SPI instances</para>
<para>Puts the SPI into a known state, and enable it. Must be called before other functions. </para>
```

Note the first line of the detailed description being included here, presumably because there is no blank line after the `\brief` command.

aallan:
@mudge Deprecated in favour of #1660?

mudge:
Yes, I'll close this one though the `DoxygenLayout.xml` changes may still be necessary.

aallan:
@nelliemckesson Do we still need the `DoxygenLayout.xml` changes in this PR?

lurch:
> So, I think `DoxygenLayout.xml` would only affect the original pico-sdk site (can't remember where that lives/lived), but if that still exists, then yeah, those changes are still required.

I think the only place that's still relevant is Appendix E of https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf ? But if this is going to become dependent on the version of Doxygen the user has installed, maybe the Doxygen-version also needs to be mentioned in that Appendix?

aallan:
@nathan-contino Can you throw a quick NOTE block into Appendix E of the Pico C SDK book ahead of the documentation release (which looks to be next week at this point). Thanks!

Issue #1550, opened 2023-11-17
repeating timer not working if set to 8399ms 


ruehlchris:
Hi, 
while I working on a reliable reproduction of  #1531,  I run into this bug:

I set the Timervalue for the repeating timer to 8399ms and the watchdog to (8399 \* 2) the repeating timer isn't firing 
at all and the watchdog is rebooting in a loop.

example code:
[multicore-repeattimer.zip](https://github.com/raspberrypi/pico-sdk/files/13387525/multicore-repeattimer.zip)

namespace {
const int32_t TIMERVALUE = 8399;	/* fire timer every xx mseconds */
const int32_t WATCHDOGTM = TIMERVALUE \* 2;
}

if (add_repeating_timer_ms(TIMERVALUE, timer_callback, nullptr, &_timer))
		watchdog_enable(WATCHDOGTM,false);

timer_callback() is feeding the watchdog.

please note:
  the main() is deadlock with mutex in the while(true) on purpose!

ruehlchris:
I my opinion it is not expected that the Watchdog is triggered because set twice the time of the repeating timer , therefore will be restarted always if the repeating timer is working correct.
But in my finding the repeating timer is not start at all.

boretrk:
IMO the problem here is that if you try to set the watchdog to a too high value it silently cuts it down to 8.3s
```
        load_value = delay_ms * 1000 * 2;

        if (load_value > 0xffffffu)
            load_value = 0xffffffu;
```
If your code typically feeds the watchdog more often but you have a valid situation where 8.3s isn't enough you might end up in a situation where the watchdog doesn't behave as expected.
I think that there at least should be an `assert(load_value <= 0xffffffu);` in there.
Or even better: `assert(delay_ms <= 8389);` to catch possible integer overflows when delay_ms is more than 2147484.
(8389*2000 will be more than 0xffffffu, but only by 785.

ruehlchris:
Peter,
wow - that is an unexpected limitation.
I agree with add an assert() to give a clear warning, it will help a lot to avoid the issue I run into.

I will close this issue now. Thanks for your help!

 Chris

Issue #1505, opened 2023-09-11
Very confusing and wrong description for the watchdog_reboot/watchdog_enable


nicolas-Philips:
The documentation of the watchdog_reboot function is as follows: [pico sdk online docs](https://cec-code-lab.aps.edu/robotics/resources/pico-c-api/group__hardware__watchdog.html#gaa73ccd43d6d755817e71bc3d7eb1fda6)

Or in the source code:

```
/*! \brief Define actions to perform at watchdog timeout
 *  \ingroup hardware_watchdog
 *
 * \note If \ref watchdog_start_tick value does not give a 1MHz clock to the watchdog system, then the \p delay_ms
 * parameter will not be in microseconds. See the datasheet for more details.
 *
 * By default the SDK assumes a 12MHz XOSC and sets the \ref watchdog_start_tick appropriately.
 *
 * \param pc If Zero, a standard boot will be performed, if non-zero this is the program counter to jump to on reset.
 * \param sp If \p pc is non-zero, this will be the stack pointer used.
 * \param delay_ms Initial load value. Maximum value 0x7fffff, approximately 8.3s.
 */
void watchdog_reboot(uint32_t pc, uint32_t sp, uint32_t delay_ms);
```

If the delay is indeed in ms, then the maximum value of 0x7fffff=8.388607e6 would be in microseconds instead of ms.
this cannot be correct. 

If watchdog_start_tick value does not give a 1MHz clock to the watchdog system, then the \p delay_ms parameter will not be in microseconds. Well I think you can see how this is confusing?

What it should probably be:
```
/*! \brief Define actions to perform at watchdog timeout
 *  \ingroup hardware_watchdog
 *
 * \note If \ref watchdog_start_tick value does not give a 1MHz clock to the watchdog system, then the \p delay_ms
 * parameter will not be in milliseconds. See the datasheet for more details.
 * The maximum amount of ticks is 0xffffffu and the watchdog count increases every two ticks. This means that the maximum timeout 
 * will occur at 0xffffffu / 2 = 0x7fffff
 * The timeout is loaded as delay_ms * 1000 * 2
 *
 * By default the SDK assumes a 12MHz XOSC and sets the \ref watchdog_start_tick appropriately.
 *
 * \param pc If Zero, a standard boot will be performed, if non-zero this is the program counter to jump to on reset.
 * \param sp If \p pc is non-zero, this will be the stack pointer used.
[...]

Issue #1409, opened 2023-06-01
watchdog_reboot() has misleading parameter name delay_ms, should be delay_us as the units are microseconds not milliseconds


smcameron:
https://github.com/raspberrypi/pico-sdk/blob/f396d05f8252d4670d4ea05c8b7ac938ef0cd381/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h#L45

ms or msec is the abbreviation for milliseconds.  However, the delay_ms parameter is actually in microseconds.  The abbreviation for microseconds uses the greek letter Mu, however, it is commonly written as "us" in C-code.  The "delay_ms" parameter to watchdog_reboot should be named "delay_us", or "delay_microsecs"

lurch:
Very similar to #1238 and #1241 ?

smcameron:
Fair enough.

Issue #1241, opened 2023-02-11
A question about the code comments of the watchdog module


mbiusxxx:
https://github.com/raspberrypi/pico-sdk/blob/2ccab115de0d42d31d6611cca19ef0cd0d2ccaa7/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h#L75
Maybe there's an error in it: milliseconds -> microsecond

lurch:
Duplicate of #1238

PR #1567, opened 2023-12-02
Add watchdog parameter-validation, and fix up misleading comments

### LINKED ISSUE #1238: watchdog_enable() confuses milliseconds and microseconds
In the file:

https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h

delay_ms is described as:

"delay_ms Number of milliseconds before watchdog will reboot without watchdog_update being called. Maximum of 0x7fffff, which is approximately 8.3 seconds"

But 0x7fffff is 8388607 which is 8.3 seconds only if the unit is microseconds. The text also mentions "then the delay_ms parameter will not be in microseconds".

So I guess it is microseconds? Then I also think the parameter should be named delay_us, not delay_ms.

lurch:
Fixes #1238 and #1550 (and #1241 #1409 #1505 which are duplicates of #1238)

### DIFF
--- a/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
+++ b/src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
@@ -30,17 +30,22 @@
 extern "C" {
 #endif
 
+// PICO_CONFIG: PARAM_ASSERTIONS_ENABLED_WATCHDOG, Enable/disable assertions in the watchdog module, type=bool, default=0, group=hardware_watchdog
+#ifndef PARAM_ASSERTIONS_ENABLED_WATCHDOG
+#define PARAM_ASSERTIONS_ENABLED_WATCHDOG 0
+#endif
+
 /*! \brief Define actions to perform at watchdog timeout
  *  \ingroup hardware_watchdog
  *
  * \note If \ref watchdog_start_tick value does not give a 1MHz clock to the watchdog system, then the \p delay_ms
- * parameter will not be in microseconds. See the datasheet for more details.
+ * parameter will not be in milliseconds. See the datasheet for more details.
  *
  * By default the SDK assumes a 12MHz XOSC and sets the \ref watchdog_start_tick appropriately.
  *
  * \param pc If Zero, a standard boot will be performed, if non-zero this is the program counter to jump to on reset.
  * \param sp If \p pc is non-zero, this will be the stack pointer used.
- * \param delay_ms Initial load value. Maximum value 0x7fffff, approximately 8.3s.
+ * \param delay_ms Initial load value. Maximum value 8388, approximately 8.3s.
  */
 void watchdog_reboot(uint32_t pc, uint32_t sp, uint32_t delay_ms);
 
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
```diff
@@ -72,7 +77,7 @@ void watchdog_update(void);
  * onto the RPI-RP2), then this value will be cleared, and so \ref watchdog_enable_caused_reboot will
  * return false.
  *
- * \param delay_ms Number of milliseconds before watchdog will reboot without watchdog_update being called. Maximum of 0x7fffff, which is approximately 8.3 seconds
+ * \param delay_ms Number of milliseconds before watchdog will reboot without watchdog_update being called. Maximum of 8388, which is approximately 8.3 seconds
```
JamesH65, line 80: 8.3 IIF the source clock is default?

src/rp2_common/hardware_watchdog/include/hardware/watchdog.h
```diff
@@ -30,17 +30,22 @@
...
  * \note If \ref watchdog_start_tick value does not give a 1MHz clock to the watchdog system, then the \p delay_ms
- * parameter will not be in microseconds. See the datasheet for more details.
+ * parameter will not be in milliseconds. See the datasheet for more details.
  *
  * By default the SDK assumes a 12MHz XOSC and sets the \ref watchdog_start_tick appropriately.
  *
  * \param pc If Zero, a standard boot will be performed, if non-zero this is the program counter to jump to on reset.
  * \param sp If \p pc is non-zero, this will be the stack pointer used.
- * \param delay_ms Initial load value. Maximum value 0x7fffff, approximately 8.3s.
+ * \param delay_ms Initial load value. Maximum value 8388, approximately 8.3s.
```
JamesH65, line 48: 8.3 IIF the source lock is default?

lurch: Well, you could argues that the `\note If \ref watchdog_start_tick value does not give a 1MHz clock to the watchdog system, then the \p delay_ms  parameter will not be in milliseconds. See the datasheet for more details.` already covers that ? :man_shrugging:

src/rp2_common/hardware_watchdog/watchdog.c
```diff
@@ -34,6 +35,7 @@ uint32_t watchdog_get_count(void) {
 // tag::watchdog_enable[]
 // Helper function used by both watchdog_enable and watchdog_reboot
 void _watchdog_enable(uint32_t delay_ms, bool pause_on_debug) {
+    valid_params_if(WATCHDOG, delay_ms <= 8388); // i.e. floor(0xffffff / 2000)
```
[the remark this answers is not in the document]
lurch: _Could_ do, but I'm not sure if it's worth it when it only appears once in a debug-level assert?

kilograham (approved):

Issue #1575, opened 2023-12-06
adc_read() not checking for error conditions


ruehlchris:
Hi,

while testing the ADC single read I become aware that the current implementation not check for error conditions, only
look for the ADC_CS_READY_BITS.

I suggest to change the return value from a uint32_t to int32_t and return a negative value if
ADC_CS_ERR_BITS or ADC_CS_ERR_STICKY_BITS had been raised.

Here is what I used in my implementation.

[int32_t_adc_read.patch.gz](https://github.com/raspberrypi/pico-sdk/files/13575534/int32_t_adc_read.patch.gz)

You may consider to change the implementation.

Chris

lurch:
@ruehlchris If you're suggesting a code-change, it's generally better to do that by submitting a Pull Request to the `develop` branch, than to attach a `.patch.gz` file to an Issue.

ruehlchris:
Thanks,

I will have pull request ready soon.

PR #1576, opened 2023-12-11
adc: adc_read() not check for error (#1575)

ruehlchris:
Implementation of adc_read() not check for errors in the ADC_CS register and might hang forever in a while() loop.
    
-  change the return type from uint16_t to int32_t
-  clean start with ADC_CS_START_ONCE_RESET before call ADC_CS_START_ONCE_BITS.
-  check for ADC_CS_READY_BITS and error ADC_CS_ERR_BITS or ADC_CS_ERR_STICKY_BITS
    
- return the ADC value (12bits) or the negative value of ADC_CS_ERR_BITS or ADC_CS_ERR_STICKY_BITS depending on the error

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_driver.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_driver.c
@@ -164,6 +164,10 @@ void cyw43_thread_lock_check(void) {
 #endif
 
 void cyw43_await_background_or_timeout_us(uint32_t timeout_us) {
+    if (__get_current_exception() > 0) {
+        async_context_wait_until(cyw43_async_context, make_timeout_time_us(timeout_us));
+        return;
+    }
     async_context_wait_for_work_until(cyw43_async_context, make_timeout_time_us(timeout_us));
 }
 


Reneg973:
Are the chances that errors occur during arc_read() very high? During all my tests I never had any error. I would prevent error checking if circumstances are unlikely to happen or if high frequency sampling is required.

ruehlchris:
> Are the chances that errors occur during adc_read() very high? During all my tests I never had any error. I would prevent error checking if circumstances are unlikely to happen or if high frequency sampling is required.

What will be the impact on high frequency sampling, the system will be in the tideloop anyway waiting for the ADC to be set the ready bit. Testing the register bits will not have a huge impact. The alternative to use the fifo style implementation which fits much better for what you mentioned.

My  opinion, if the hardware give you a hand and signal a bad ADC then do not use the result.

Reneg973:
> What will be the impact on high frequency sampling, the system will be in the tideloop anyway waiting for the ADC to be set the ready bit. Testing the register bits will not have a huge impact. The alternative to use the fifo style implementation which fits much better for what you mentioned.
> 
> My opinion, if the hardware give you a hand and signal a bad ADC then do not use the result.

Yes, nevertheless setting the ADC_CS_START_ONCE_RESET seems not to be required, as the ADC_CS_START_ONCE_BITS is an auto reset flag. And after compilation the (my) compiler produces an unnecessary jump for the most likely code path.
Also the last condition can be removed as it is the only remaining condition:

```
static inline int32_t adc_read(void) {
     hw_set_bits(&adc_hw->cs, ADC_CS_START_ONCE_BITS);

     while ((adc_hw->cs & (ADC_CS_READY_BITS|ADC_CS_ERR_BITS|ADC_CS_ERR_STICKY_BITS))==0) {
         tight_loop_contents();
     }

     if ((adc_hw->cs & ADC_CS_READY_BITS)>0) {
         return adc_hw->result & 0xfff;
     }
     if ((adc_hw->cs & ADC_CS_ERR_BITS)>0) {
         return -ADC_CS_ERR_BITS;
     }
     /* clear sticky bit */
     hw_set_bits(&adc_hw->cs, ADC_CS_ERR_STICKY_BITS);
     return -ADC_CS_ERR_STICKY_BITS;
 }
```

ruehlchris:
> > What will be the impact on high frequency sampling, the system will be in the tideloop anyway waiting for the ADC to be set the ready bit. Testing the register bits will not have a huge impact. The alternative to use the fifo style implementation which fits much better for what you mentioned.
> > My opinion, if the hardware give you a hand and signal a bad ADC then do not use the result.
> 
> Yes, nevertheless setting the ADC_CS_START_ONCE_RESET seems not to be required, as the ADC_CS_START_ONCE_BITS is an auto reset flag. And after compilation the (my) compiler produces an unnecessary jump for the most likely code path. Also the last condition can be removed as it is the only remaining condition:
> 
> ```
> static inline int32_t adc_read(void) {
>      hw_set_bits(&adc_hw->cs, ADC_CS_START_ONCE_BITS);
> 
>      while ((adc_hw->cs & (ADC_CS_READY_BITS|ADC_CS_ERR_BITS|ADC_CS_ERR_STICKY_BITS))==0) {
>          tight_loop_contents();
>      }
> 
>      if ((adc_hw->cs & ADC_CS_READY_BITS)>0) {
>          return adc_hw->result & 0xfff;
>      }
>      if ((adc_hw->cs & ADC_CS_ERR_BITS)>0) {
>          return -ADC_CS_ERR_BITS;
>      }
>      /* clear sticky bit */
>      hw_set_bits(&adc_hw->cs, ADC_CS_ERR_STICKY_BITS);
>      return -ADC_CS_ERR_STICKY_BITS;
>  }
> ```

Happy New Year,

Yes, I think you nailed it and optimized it out. I will modify the patch accordant to your succession and create a 
new pull request.

PR #1580, opened 2023-12-17
Make fields within mutex volatile, just in case it matters.

daveythacher:
Make fields within mutex structure volatile. This should prevent caching in register or stack. SDK currently manages barriers and has no cache to worry about. This is not likely required under most cases. but for completeness.

### DIFF
--- a/src/common/pico_sync/include/pico/mutex.h
+++ b/src/common/pico_sync/include/pico/mutex.h
@@ -46,10 +46,10 @@ extern "C" {
  */
 typedef struct __packed_aligned  {
     lock_core_t core;
-    lock_owner_id_t owner;      //! owner id LOCK_INVALID_OWNER_ID for unowned
-    uint8_t enter_count;        //! ownership count
+    volatile lock_owner_id_t owner;      //! owner id LOCK_INVALID_OWNER_ID for unowned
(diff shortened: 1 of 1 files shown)


daveythacher:
If the mutex is cached on two threads on the same core undefined behavior may occur. I will just sort this out on my side.

kilograham:
Re-opening as i was unaware that GCC would cache memory values across a compiler memory barrier, which is surprising

kilograham:
Yup; things work as expected...

The compiler will not cache a variable across a compiler memory barrier, so variables are reloaded after entering a spin lock whether volatile or not, and equally a context switch for the correct core cannot occur while IRQs are disabled (as they are during the spin locks used here)

... so i'm not sure what issue you are worried about

kilograham:
not sure what you mean by "destroy the SDK function"... if you mean inline it (which would only happen with LTO) it still doesn't matter, as the compiler memory barriers are in place (in the spin lock functions) to make sure things are reloaded where necessary

daveythacher:
My concern is for decomposing to register in two places. The stacks are not shared address space and the compiler does not know to commit into shared address space.

daveythacher:
Due to #1453, you may want to reconsider this.

PR #1599, opened 2024-01-02
adc: adc_read() not check for error (#1575) v2

ruehlchris:
Implementation of adc_read() not check for errors in the ADC_CS register.
    
-  change the return type from uint16_t to int32_t
-  check for ADC_CS_READY_BITS and error ADC_CS_ERR_BITS or ADC_CS_ERR_STICKY_BITS    
- return the ADC value (12bits) or the negative value of ADC_CS_ERR_BITS or ADC_CS_ERR_STICKY_BITS depending on the error

v2: 
- update the code, optimize the workflow 
- no reset required , so removed it.

### DIFF
--- a/src/rp2_common/hardware_adc/include/hardware/adc.h
+++ b/src/rp2_common/hardware_adc/include/hardware/adc.h
@@ -128,15 +128,24 @@ static inline void adc_set_temp_sensor_enabled(bool enable) {
  *
  *  Performs an ADC conversion, waits for the result, and then returns it.
  *
- * \return Result of the conversion.
+ * \return Result of the conversion or negative value on error.
  */
-static inline uint16_t adc_read(void) {
+static inline int32_t adc_read(void) {
     hw_set_bits(&adc_hw->cs, ADC_CS_START_ONCE_BITS);
+    
+    while ((adc_hw->cs & (ADC_CS_READY_BITS|ADC_CS_ERR_BITS|ADC_CS_ERR_STICKY_BITS))==0) {
+         tight_loop_contents();
+    }
 
-    while (!(adc_hw->cs & ADC_CS_READY_BITS))
-        tight_loop_contents();
-
-    return (uint16_t) adc_hw->result;
+    if ((adc_hw->cs & ADC_CS_READY_BITS)>0) {
+        return adc_hw->result & 0xfff;
+    }
+    if ((adc_hw->cs & ADC_CS_ERR_BITS)>0) {
+        return -ADC_CS_ERR_BITS;
+    }
+    /* clear sticky bit */
+    hw_set_bits(&adc_hw->cs, ADC_CS_ERR_STICKY_BITS);
+    return -ADC_CS_ERR_STICKY_BITS;
 }
 
 /*! \brief Enable or disable free-running sampling mode


src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -128,15 +128,24 @@ static inline void adc_set_temp_sensor_enabled(bool enable) {
...
+    while ((adc_hw->cs & (ADC_CS_READY_BITS|ADC_CS_ERR_BITS|ADC_CS_ERR_STICKY_BITS))==0) {
+         tight_loop_contents();
+    }
 
-    while (!(adc_hw->cs & ADC_CS_READY_BITS))
-        tight_loop_contents();
-
-    return (uint16_t) adc_hw->result;
+    if ((adc_hw->cs & ADC_CS_READY_BITS)>0) {
+        return adc_hw->result & 0xfff;
```
peterharperuk, line 141: Might be better to use ADC_RESULT_BITS

src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -128,15 +128,24 @@ static inline void adc_set_temp_sensor_enabled(bool enable) {
...
+    }
+    if ((adc_hw->cs & ADC_CS_ERR_BITS)>0) {
+        return -ADC_CS_ERR_BITS;
```
peterharperuk, line 144: It's a bit random to return -ADC_CS_ERR_BITS. It would be better to return a "standard" error code? Maybe PICO_ERROR_GENERIC or PICO_ERROR_NO_DATA?

src/rp2_common/hardware_adc/include/hardware/adc.h
```diff
@@ -128,15 +128,24 @@ static inline void adc_set_temp_sensor_enabled(bool enable) {
...
+    }
+    /* clear sticky bit */
+    hw_set_bits(&adc_hw->cs, ADC_CS_ERR_STICKY_BITS);
+    return -ADC_CS_ERR_STICKY_BITS;
```
peterharperuk, line 148: same comment as above

src/rp2_common/hardware_adc/include/hardware/adc.h
ruehlchris: I will have a look if that will have benefit. The original code check for the ADC_CS_READY_BITS.

src/rp2_common/hardware_adc/include/hardware/adc.h
ruehlchris: I see your point, but if we doing this we can't distinguished what caused the error.

Use 
 PICO_ERROR_NO_DATA  
and additional, define a new code:
PICO_ERROR_ADC_STICKY_BIT

src/rp2_common/hardware_adc/include/hardware/adc.h
lurch: I believe Peter meant use `ADC_RESULT_BITS` instead of `0xfff` - see https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_regs/include/hardware/regs/adc.h#L111

ruehlchris: Oh,  :)
Thanks for the eye opener.

peterharperuk (changes requested):
Looks good but see previous comments concerning minor changes

PR #1600, opened 2024-01-04
Add tool for lwip httpd server

peterharperuk:
It would be helpful to be able to use the lwip httpd server, but it generates the content using a tool written in C. This is problematic as it requires a native compiler to build the tools.

Add a python tool to generate the httpd content and a cmake function to make use of it.


src/rp2_common/pico_lwip/tools/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# Compile the http content into a source file "pico_fsdata.inc" in a format suitable for the lwip httpd server
+# Pass the target library name library type and the list of httpd content
+function(pico_lwip_httpd_content TARGET_LIB TARGET_TYPE)
```
kilograham, line 3: probably pico_add_ or pico_set_ depending on whether it can be called multiple times for the same target

peterharperuk: I think it can/should only be used once per target

kilograham (approved):

Issue #1609, opened 2024-01-19
pico_generate_pio_header fails if a custom OUTPUT_DIR is specified but does not exist already


timg236:

kilograham:
merged into develop

PR #1610, opened 2024-01-19
pico_generate_pio_header: Create OUTPUT_DIR

timg236:
Previously, if a custom OUTPUT_DIR was specified but didn't exist then this command would fail.

Resolves: https://github.com/raspberrypi/pico-sdk/issues/1609


peterharperuk (approved):
Looks ok

kilograham (approved):

PR #1613, opened 2024-01-24
corrected issue#1536 - used correct define to create a configuration …

MarcinKlosowskiMobica:
This pull request fixes an issue related to the mask that configure I2C port. 
Current version of software uses `I2C_IC_RAW_INTR_STAT_TX_ABRT_BITS` as a submask. It is misleading. 
If configuration will need to be changed `I2C_IC_RAW_INTR_STAT_TX_ABRT_BITS` might not be changed. 
Better solution is to use `I2C_IC_INTR_MASK_M_TX_ABRT_BITS` that is self explanatory and we are sure that I2C configuration mask will be correct in case of further update.
This pull request solves issue#1536


will-v-pi (approved):
LGTM

PR #1619, opened 2024-01-29
fix harware_alarm_irq_number typo

Poofjunior:
fixes [`harware_alarm_irq_number`](https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/hardware_timer/timer.c#L109) typo.

I found myself using this function and couldn't search for it.

### DIFF
--- a/src/rp2_common/hardware_timer/timer.c
+++ b/src/rp2_common/hardware_timer/timer.c
@@ -106,7 +106,7 @@ void busy_wait_until(absolute_time_t t) {
 }
 /// \end::busy_wait[]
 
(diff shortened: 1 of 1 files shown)


lurch:
I wonder if it might be worth adding a wrapper-function using the old name, for backwards-compatibility? :man_shrugging:

kilograham (approved):
it is not a public function, so no need to worry about backwards compatibility

kilograham (approved):

kilograham (approved):

PR #1620, opened 2024-01-30
Fix PICO_DEOPTIMIZED_DEBUG not updating compiler flags

### LINKED ISSUE #1618: [Possible Fix] PICO_DEOPTIMIZED_DEBUG has no effect after first configuration (causes breakpoint issues)
*Prefacing that I am new to using CMake and new to using debuggers, so take what I suggest with a grain of salt - but I believe I've identified a bug and a solution to it.*

## The problem:

In learning how to use picoprobe and VSCode's debugger, I configured a simple project with the Debug config using a minimal `CMakeLists.txt` that did not explicitly set `PICO_DEOPTIMIZED_DEBUG`.  When I tried setting a breakpoint on the line:

```c
        gpio_put(PIN_LED, true);
``` 

The breakpoint is greyed out in VSCode, delivering the error: 

> "No line 16 in file "...\main.c"

After a while of searching, [it looked like][1] this was a problem with compiler optimizations, so I tried enabling `PICO_DEOPTIMIZED_DEBUG`, but reconfiguring didn't change any behaviour.  To my surprise, however, deleting the `build` directory and configuring the project fresh produced the desired results; which got me digging...

[1]: <https://github.com/Marus/cortex-debug/issues/444>

## Likely cause:

[My understanding][2] is that setting `CMAKE_${LANG}_FLAGS_DEBUG_INIT` specifically only sets the flags for the Debug config *the first time it is **configured***.  This would mean the debug flags get saved to `CMakeCache.txt` when you initially configure the project, and then all subsequent re-configurations that try to set `CMAKE_${LANG}_FLAGS_DEBUG_INIT` have no change to the already-cached value.

What this means in this situation (I think) is that whatever `PICO_DEOPTIMIZED_DEBUG` is set to / defaults to during your first configuration is what will dictate all subsequent debug flags until `CMakeCache.txt` is deleted.  I believe you can see this being the case by observing how the value of `CMAKE_C_FLAGS_DEBUG` in `CMakeCache.txt` does not change after the initial configuration, regardless of changing the value of `PICO_DEOPTIMIZED_DEBUG`.

[2]: <https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG_INIT.html>

## Proposed solution:

I think that this section in `cmake/preload/toolchains/set_flags.cmake`:

```cmake
    if (PICO_DEOPTIMIZED_DEBUG)
        set(CMAKE_${LANG}_FLAGS_DEBUG_INIT "-O0")
    else()
        set(CMAKE_${LANG}_FLAGS_DEBUG_INIT "-Og")
    endif()
```

Needs to change to:

```cmake
    if (PICO_DEOPTIMIZED_DEBUG)
        set(CMAKE_${LANG}_FLAGS_DEBUG "-O0 ${CMAKE_${LANG}_FLAGS_DEBUG}")
    else()
        set(CMAKE_${LANG}_FLAGS_DEBUG "-Og ${CMAKE_${LANG}_FLAGS_DEBUG}")
    endif()
```

These changes remove `_INIT` on both commands and append the default debug flags (which is `-g` in this case, required for proper debugging).  Appending the default flags keeps functional parity with `CMAKE_${LANG}_FLAGS_DEBUG_INIT`'s intended behaviour.

Additionally: What is the reasoning behind defaulting to an optimized debug, if optimizing breaks certain debug functionality?  Perhaps this could be an opportunity to consider making deoptimized debug builds the default for debug builds with `PICO_DEOPTIMIZED_DEBUG` defaulting to `1` when it is declared, in combination with the above proposed change.

I'd appreciate the eyes of someone who has been familiar with CMake for longer than a single week to validate that this is a real bug and a valid/safe solution.  If this does sound reasonable, I'd be happy to submit a pull request with this simple change 🙂

recursivenomad:
Setting `CMAKE_${LANG}_FLAGS_DEBUG_INIT` specifically only sets the flags for the Debug config the first time it is configured by CMake, pulling the initially-configured flags from `CMakeCache.txt` on subsequent configurations.  This causes `PICO_DEOPTIMIZED_DEBUG` to not have any effect after the initial configuration, causing breakpoint issues when debugging certain functions.

Changing the CMake variable to `CMAKE_${LANG}_FLAGS_DEBUG` enables the flags to be updated every configuration, no longer depending on the cache.  We also must append the debug flags which were set prior to this file to maintain functional parity with the behaviour of `CMAKE_${LANG}_FLAGS_DEBUG_INIT`.

See Issue #1618 for further details.

Fixes #1618


recursivenomad:
Setting to draft as I've identified a bug where the initial configuring produces `-Og -Og ` instead of `-Og -g`, as visible in `compile_commands.json`.  Subsequent configurings behave as expected.  This bug was is not present in `develop`.  Working on it...

recursivenomad:
Okay, I understand the problem.  It appears that when CMake is provided `-DCMAKE_BUILD_TYPE:STRING=Debug`, [it sets](https://github.com/Kitware/CMake/blob/master/Modules/Compiler/GNU.cmake#L111):

```cmake
  string(APPEND CMAKE_${lang}_FLAGS_DEBUG_INIT " -g")
```

This gets appended *after* calling:

```cmake
project(project-name)
```

Since `set_flags.cmake` is called *before* creating the project, `CMAKE_${lang}_FLAGS_DEBUG_INIT` has not been set in the cache yet during first configuring.

Looks like unsetting `CMAKE_${LANG}_FLAGS_DEBUG` from the cache right before setting `CMAKE_${LANG}_FLAGS_DEBUG_INIT` (disregarding the first proposed changes) solves the problem and is truly at parity with the initial (and intended) behaviour of `PICO_DEOPTIMIZED_DEBUG`.

recursivenomad:
If accepted, I'd recommend squashing before merge as the first commit contains bugs.

kilograham (approved):

PR #1621, opened 2024-01-30
Sets PICO_DEOPTIMIZED_DEBUG to be enabled by default

recursivenomad:
Optimizing a debug build with PICO_DEOPTIMIZED DEBUG disabled (the current default) results in breakpoints not being possible in certain functions.  This change makes the default behaviour of PICO_DEOPTIMIZED_DEBUG to be enabled, so that debug builds will fully work as expected when no explicit value for the option is passed.

Since the cache stores the default value of PICO_DEOPTIMIZED DEBUG, you will need to delete the CMakeCache.txt previously generated in your project directory for this change to default behaviour to take effect in existing projects.

This commit is related to pull request #1620 in the primary repository.

See Issue #1618 in the primary repository for additional context.

### DIFF
--- a/cmake/preload/toolchains/pico_arm_clang.cmake
+++ b/cmake/preload/toolchains/pico_arm_clang.cmake
@@ -38,7 +38,7 @@ set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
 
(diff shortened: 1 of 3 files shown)


recursivenomad:
(I acknowledge that this change may be controversial as it changes the default optimization of debug builds, so please consider this pull request with caution)

kilograham:
converting to draft as we likely will not doing this (still open for now)

recursivenomad:
Hi @kilograham - Checking in a few months later if you'd like me to update/preserve this PR?  I still like the idea of a debug build providing all debug features by default, but I understand if it is not what the SDK team considers ideal.

If the PR is confidently not needed, I will delete the fork to reduce clutter.

kilograham:
yes; i don't think we will make this the default, so feel free to delete

PR #1623, opened 2024-02-01
Fixed an issue that caused incorrect settings when the SPI clock was …

aizaki77:
…less than 244141Hz.

_Instructions: (please delete)_
 - _please do not submit against `master`, use `develop` instead_
 - _please make sure there is an associated issue for your PR, and reference it via "Fixes #num" in the description_
 - _please enter a detailed description_


lurch:
> * please do not submit against `master`, use `develop` instead



> * please enter a detailed description

Issue #1628, opened 2024-02-06
Add a function that returns the remaining time before an alarm trigger


TheRaf974:
I'ld like to make a PR to add function that returns the remaining time before an alarm trigger.

I'm using an alarm to add a cooldown to an input button that'll be used by people. At the beginning, I was manually checking the difference of time with the last use of the button but it looked better to me to use a timer.

However, if someone press the button while in cooldown, I would like to let him know the time remaining before the button can be pressed effectively. 

Looking in the [RP2040 datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf) chapter 4.6.3, I found we can get the remaining time with something like :
`timer_hw->alarm[alarm_pool] - timer_hw->timerawl`

I think there might be situations where a function like this can be useful. Also, my code may needs somes adjustements.

kilograham:
added

PR #1629, opened 2024-02-06
Add alarm_pool_remaining_time()

TheRaf974:
Add a function that returns the time remaining before triggering for a given pool and alarm id
It uses difference between an alarm register and the timerawl register. (see [RP2040 datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf) chapter 4.6.3)
It also check that the alarm for which we're asking the remaining time is the actual one running on the hardware timer (and not in the alarm pool)

Refer to #1628


kilograham:
fixes #1628

kilograham:
will fix this internally as part of redoing the alarm pool implementation; note the return should be 64 bit.

kilograham:
added

PR #1633, opened 2024-02-11
Fixed an issue that caused incorrect settings when the SPI clock was …

aizaki77:
…less than 244,141Hz.

Fixed an issue that caused incorrect settings when the SPI clock was less than 244,141Hz.

For example, if clk_peri is 125MHz and you specify 150kHz

Before correction
prescale=2, postdiv=256,
125000000/(2*256)=244140.625Hz
Therefore, it exceeds 150kHz.

After correction
prescale=4, postdiv=209,
125000000/(4*209)=149521.5311004785Hz
This is the correct frequency.


lurch:
Unassigning myself, as I don't know anything about the SPI hardware nor its frequency requirements.

will-v-pi (approved):
Looks good to me

PR #1635, opened 2024-02-14
Allow disabling pico_stdio function wrapping

### LINKED ISSUE #1634: No way to disable pico_stdio function wrapping, just pico_printf
I have a personal project using tinyusb to expose a CDC device and an HID device. I wanted to reimplement the `read`, `write`, and `open` newlib syscalls, but looking through the source code of `pico_stdio`, I noticed that I would not be able to just copy their implementation, as `stdio_get_until` and `stdio_put_string` are static and thus private functions to the `stdio.c` source file. So I just reimplemented the syscalls using my own output routines... and started seeing my application crash, as `printf` and `putchar` were being wrapped by the pico-sdk and sending the input data through `pico_stdio`. `pico_set_printf_implementation(${name} compiler)` almost does what I want, but it only manages the functions defined in `pico_printf`, which excludes `putchar`. The current pico-sdk is missing something similar for `pico_stdio`.

I tried using `SKIP_PICO_STDIO`, but that completely avoids the entire library including includes, which blows up other dependencies of `pico_stdlib`.

I have a PR I'll place momentarily that does what I want by borrowing exactly what `pico_printf` does, requiring the use of a cmake property to change the default stdio library.

gemarcano:
This PR copies the approach used by pico_printf and applies it to pico_stdio to allow a developer to opt out of stdio printf/putchar/etc. wrappers. This is done with the use of a new function `pico_set_stdio_implementation` which works identically to `pico_set_printf_implementation` from `pico_printf`.

This enables a developer to opt out of stdio wrappers around the following functions:
  - printf
  - vprintf
  - puts
  - putchar
  - getchar

Additionally, in `pico_stdio`'s `stdio.c`, some additional `#if/#endif` are applied to shut up compiler warnings about unused functions when `pico_printf` compiler implementation is in use.

Fixes #1634

### DIFF
--- a/src/rp2_common/pico_stdio/CMakeLists.txt
+++ b/src/rp2_common/pico_stdio/CMakeLists.txt
@@ -1,17 +1,48 @@
 if (NOT TARGET pico_stdio)
+    # library to be depended on - we make this depend on particular implementations using per target generator expressions
     pico_add_library(pico_stdio)
 
-    target_include_directories(pico_stdio_headers INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
-
-    target_sources(pico_stdio INTERFACE
+    # no custom implementation; falls thru to compiler
+    pico_add_library(pico_stdio_compiler)
+    target_sources(pico_stdio_compiler INTERFACE
             ${CMAKE_CURRENT_LIST_DIR}/stdio.c
     )
 
-    pico_wrap_function(pico_stdio printf)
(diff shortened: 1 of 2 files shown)


peterharperuk:
Looks reasonable as far as I can see

gemarcano:
I noticed a useless newline I added in `stdio.c`, so I just took that out. No other changes with the update.

PR #1643, opened 2024-02-20
    Fix a bug with the enter/exit critical functions missing instruction barrier

### LINKED ISSUE #1644: Bug in save_and_disable_interrupts()
The save_and_disable_interrupts function is buggy for two reasons. Either one is not a bug in itself, but combined they produce the following problem.

The first problem :

The function `save_and_disable_interrupts` is :
`
__force_inline static uint32_t save_and_disable_interrupts(void) {
    uint32_t status;
    pico_default_asm_volatile(
            "mrs %0, PRIMASK\n"
            "cpsid i"
            : "=r" (status) ::);
    return status;
}
`
The inline assembly statement does not include the `: "memory" ` statement at the end. Without this, the compiler is not prevented from reordering and optimizing stuff. 

The second problem is that the function is declared inline. Coupled with problem number one, what it does is that when the body of the function is "copied" in place by the compiler, the compiler can then start optimizing things. Because there is no ~implicit instruction barrier with the "memory", then the compiler can reorder things around. 

I came accross a situation there the following code would not work with optimization : 

`
uint32_t counter = 0;

...

uint32_t irq_save = save_and_disable_interrupts();
{
    counter--;
}
restore_interrupts(irq_save);
`

The problem was that an interrupt modifying `counter` would pop and execute in the middle of the critical zone above like so :

`
uint32_t irq_save = save_and_disable_interrupts();
{
    counter--;         <-------------- Interrupt happening in the middle of the read-modify-write
}
restore_interrupts(irq_save);
`

simonbeaudoin0935:
Fixes #1644

    The save_and_disable_interrupts function was buggy for two reasons. The first one, it is declared
    as a static inline function; once the function's content is inlined, the compiler is free to
    optimize/reorder the content as it wishes with surrounding instructions. This, combined with the
    fact that the inline assembly "cpsid i" does not contain the trailing `: \"memory\" ` that acts
    like an implicit instruction barrier, this allows the compiler to reorder the instructions
    that disable the interrupts with whatever is arround it in the caller function.

    Now, the fix is to use the already existing CMSIS functions __disable_irq() and __enable_irq()
    instead of duplicating the inline assembly. The CMSIS functions DO contain the `: \"memory\" `
    statements that will instruct the compiler it cannot do assumptions about the memory after
    the inline assembly (aka, instruction barrier). "
    error: pathspec 'i does not contain the trailing  that acts
    like an implicit instruction barrier, this allows the compiler to reorder the instructions
    that disable the interrupts with whatever is arround it in the caller function.

    Now, the fix is to use the already existing CMSIS functions __disable_irq() and __enable_irq()
    instead of duplicating the inline assembly. The CMSIS functions DO contain the
    statements that will instruct the compiler it cannot do assumptions about the memory after
    the inline assembly (aka, instruction barrier). '

### DIFF
--- a/src/rp2_common/hardware_sync/include/hardware/sync.h
+++ b/src/rp2_common/hardware_sync/include/hardware/sync.h
@@ -10,6 +10,7 @@
 #include "pico.h"
 #include "hardware/address_mapped.h"
 #include "hardware/regs/sio.h"
+#include "cmsis_compiler.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -212,11 +213,9 @@ __force_inline static void __mem_fence_release(void) {
  * \return The prior interrupt enable status for restoration later via restore_interrupts()
  */
 __force_inline static uint32_t save_and_disable_interrupts(void) {
-    uint32_t status;
-    pico_default_asm_volatile(
-            "mrs %0, PRIMASK\n"
-            "cpsid i"
-            : "=r" (status) ::);
+    uint32_t status = __get_PRIMASK();
+    __disable_irq();
+
     return status;
 }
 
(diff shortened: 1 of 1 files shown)


simonbeaudoin0935:
Oops, yes, let me fix that

kilograham:
Fixes #1644

kilograham:
will fix this by adding `memory`

PR #1645, opened 2024-02-20
Add C11 standard atomic support

### LINKED ISSUE #1642: Runtime for C11 atomics missing
The Pico-SDK does not provide an implementation of standard C11 atomic functions, this makes implementing multi-core algorithms more complicated and error prone while decreasing software potability from Cotrex-M3/4/7 code bases. Unfortunately the Cortex-M0+ does not provide LDREX and STREX instructions for multi-core lock-less atomic operations and thus most tool chains do not provide implementations.

The Pico-SDK should provide an implementation of the C11 atomic functions using interrupt masking for single core atomic read/write/modify operations combined with hardware spin locks to accommodate the RP2040 symmetric dual core design.

sgstreet:
This PR Fixes #1642 Runtime for C11 atomics missing. 

The file pico_atomic.c implements the required functions for the read/modify/write function of atomic variables with sizes of 1, 2, 4 and 8 bytes. Atomic load and store for 8 bytes quantities is also supported.  Support for atomic_flags is also present.  

To eliminate interference with existing hardware spinlock usage while reducing multicore contention on unique atomic variables, we use one of the watchdog scratch registers (WATCHDOG_SCRATCH3) to implement 16, 2 bit, multicore locks, via a variation of [Dekker's Algorithm](https://en.wikipedia.org/wiki/Dekker%27s_algorithm). The lock is selected as a function of the variable address and the stripe width which hashes variables addresses to locks numbers and leverages the RP2040 Atomic Register Access functionality of the WATCHDOG scratch registers.

Clearly the current approach (See __atomic_lock) is slower than using the RP2040 Hardware Spinlocks directly, but I was not sure how to assign spin locks from the reserved range (IDs 0-13) and was concerned about multicore lock contention when the number of hash buckets (hardware spinlocks) is minimized.  Any recommendations concerning the number of spins lock to allocated to this library along with the stripping width (number of addresses hashed to the same lock) would greatly appreciated? My thoughts are four to eight locks with a stripe width of four bytes. Using the hardware spinlocks is an easy update.


sgstreet:
As motivation to add this PR, below is a sample ticket based spin lock implementation similar to the Linux kernel.  This type of implementation removes the limitation of 32 hardware spin locks at the minimal cost of potential hardware spin lock contention.

```
typedef atomic_ulong spinlock_t;

void spin_lock(spinlock_t *spinlock)
{
	uint16_t ticket = atomic_fetch_add(spinlock, 1UL << 16) >> 16;
	while ((*spinlock & 0xffff) != ticket);
}

unsigned int spin_lock_irqsave(spinlock_t *spinlock)
{
	uint32_t state = save_and_disable_interrupts();
	spin_lock(spinlock);
	return state;
}

bool spin_try_lock(spinlock_t *spinlock)
{
	uint32_t value = *spinlock;
	if ((value >> 16) != (value & 0xffff))
		return false;

	return atomic_compare_exchange_strong(spinlock, &value, value + (1UL << 16));
}

bool spin_try_lock_irqsave(spinlock_t *spinlock, unsigned int *state)
{
	uint32_t irq_state = save_and_disable_interrupts();
	bool locked = spin_try_lock(spinlock);
	if (!locked) {
		restore_interrupts(irq_state);
		return false;
	}

	*state = irq_state;
	return true;
}

void spin_unlock(spinlock_t *spinlock)
{
	atomic_fetch_add((uint16_t *)spinlock, 1);
}

void spin_unlock_irqrestore(spinlock_t *spinlock, unsigned int state)
{
	spin_unlock(spinlock);
	restore_interrupts(state);
}
```

src/rp2_common/pico_atomic/include/stdatomic.h
```diff
@@ -0,0 +1,20 @@
+#ifndef __STDATOMIC_H
+#define __STDATOMIC_H
+
+#include_next <stdatomic.h>
```
kilograham, line 4: phew; i was wondering what versions of compilers supported this, but GCC 6 is happy, so OK  i guess!

src/rp2_common/pico_atomic/pico_atomic.c
```diff
@@ -0,0 +1,346 @@
...
+
+#include "hardware/address_mapped.h"
+#include "hardware/regs/watchdog.h"
+#include "hardware/regs/sio.h"
+#include "hardware/sync.h"
+
+#include "pico/config.h"
+
+#ifndef __optimize
+#define __optimize __attribute__((optimize("-Os")))
```
kilograham, line 18: is there a reason for Os by default? note i am merging this now as I want to test it with something else, but may fixup later

kilograham (dismissed):
Thanks very much!

src/rp2_common/pico_atomic/include/stdatomic.h
sgstreet: I could drop a complete replacement for stdatomic.h as different approach?  Seem like that would be worse in the long term.

kilograham: yeah; i am happy with `include_next` i just thought it was a recent add

src/rp2_common/pico_atomic/pico_atomic.c
sgstreet: From my experience, you get better code from -Os on the cortex parts. Less and generally faster code, but our mileage may vary.  I can remove if you wish.  
Sort of a adjacent question is if the atomic code should be forced in to SRAM?

kilograham: good question; in similar bits of code we have a `#define` to pick

No, i think `-Os` is fine, and probably as you say better in many circumstances (our default being -O3 is a historic cmake thing, and i am worried about knock on effects on existing stuff if i change it)

kilograham:
Thanks, I have merged this for now somewhat rapidly as we have something internal dependent on it...

sgstreet:
Glad you and the team found a use for this! Let me know if there is a problem or something else the team would like changed.

kilograham:
cool; i have actually backed it out again temporarily, as it seems certain versions of GCC complain about a few things

sgstreet:
If I could know which compiler versions and what problems cropped up I can see if there is an easy fix?

src/rp2_common/pico_atomic/pico_atomic.c
petrhosek: Note that this makes the code incompatible with Clang which doesn't support the `optimize` attribute which is GCC specific. A correct way to achieve this without relying on non-portable behavior is to set the optimization level for this specific target through the build system.

sgstreet: I tend not to use clang on my projects.  I will remove this momentarily.

sgstreet: I updated the branching underlying this PR, but since the PR has been merged, closed and reverted I'm not sure how to github to see the PR updates. @kilograham @petrhosek can you reopen this PR?

sgstreet:
I create PR #1763 to reopen this PR and remove the GCC specific optimize attribute.

PR #1648, opened 2024-02-26
Make it easier to configure default baud rate

peterharperuk:
To make stdio logging quicker it's probably a good idea to set the default uart speed 921600, but it's only configurable by adding this define to your project. Make it configurable from the cmake command line.

### DIFF
--- a/src/rp2_common/hardware_uart/CMakeLists.txt
+++ b/src/rp2_common/hardware_uart/CMakeLists.txt
@@ -1,3 +1,9 @@
 pico_simple_hardware_target(uart)
 
 pico_mirrored_target_link_libraries(hardware_uart INTERFACE hardware_resets hardware_clocks)
+
+# PICO_CONFIG: PICO_DEFAULT_UART_BAUD_RATE, Define the default UART baudrate, max=921600, default=115200, group=hardware_uart
+if (PICO_DEFAULT_UART_BAUD_RATE)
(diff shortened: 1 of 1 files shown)


kilograham (commented):
should probably add a `PICO_CMAKE_CONFIG` so it appears in the docs (or ought to!)

lurch:
> should probably add a `PICO_CMAKE_CONFIG` so it appears in the docs (or ought to!)

`PICO_DEFAULT_UART_BAUD_RATE` _does_ already appear on page 482 of https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf if that's what you were referring to?
Or am I confusing "Appendix B: SDK configuration" with "Appendix C: CMake build
configuration" ?

peterharperuk:
I think you can only set PICO_DEFAULT_UART_BAUD_RATE if you add it to your CMake file. I was suggesting you can set it by adding PICO_DEFAULT_UART_BAUD_RATE=xyz to the CMake command line

kilograham (changes requested):
pls add PICO_CMAKE_CONFIG

peterharperuk:
I've added PICO_CMAKE_CONFIG to the CMakeLists.txt file. As @lurch points out - it's already in the header file

kilograham (approved):

PR #1649, opened 2024-02-26
Fix stdio_set_chars_available_callback for usb

### LINKED ISSUE #1603: Can't get stdio_set_chars_available_callback to work with USB
I have a project where I want to get a callback for when I receive characters from the stdio USB virtual serial port. It works fine if I call `getchar` in the main loop, but in my main solution I can't easily do that so I was happy to find `stdio_set_chars_available_callback`, but I can't get it to work. The callback is called (I toggle an LED to show that my code is called), but I never receive the character in the callback method.
Here's the callback method:
```
void received_char_callback(void* param)
{
    gpio_put(LED_PIN_INDICATOR, !gpio_get(LED_PIN_INDICATOR));

    int key = getchar_timeout_us(0); // get any pending key press but don't wait
    if (key > 0)
    {
        // We never get here
        receivedChar = (char)key;
    }
}
```

And I set it up like this:
```
stdio_set_chars_available_callback(received_char_callback, NULL);
```

And here's the simple repo: https://github.com/DMXCore/Pico-Examples/tree/main/UsbStdio

Am I doing anything wrong? I'm on Pico SDK 1.5.1. Surely this is supported by USB, not just UART, correct?

peterharperuk:
When you get a callback to tell you a character is available, you should be able to call getchar_timeout_us, but it's not working for USB

Fixes #1603

### DIFF
--- a/src/rp2_common/pico_stdio_usb/stdio_usb.c
+++ b/src/rp2_common/pico_stdio_usb/stdio_usb.c
@@ -57,7 +57,13 @@ static int64_t timer_task(__unused alarm_id_t id, __unused void *user_data) {
 static void low_priority_worker_irq(void) {
     if (mutex_try_enter(&stdio_usb_mutex, NULL)) {
         tud_task();
+#if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
+        uint32_t chars_avail = tud_cdc_available();
+#endif
         mutex_exit(&stdio_usb_mutex);
+#if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
+        if (chars_avail && chars_available_callback) chars_available_callback(chars_available_param);
+#endif
     } else {
         // if the mutex is already owned, then we are in non IRQ code in this file.
         //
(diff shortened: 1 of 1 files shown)


kilograham (commented):
you lost the PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK protection.

peterharperuk:
Fixed. Messy.

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -56,8 +56,17 @@ static int64_t timer_task(__unused alarm_id_t id, __unused void *user_data) {
 
 static void low_priority_worker_irq(void) {
     if (mutex_try_enter(&stdio_usb_mutex, NULL)) {
+#if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
+        uint32_t chars_avail;
+#endif
         tud_task();
+#if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
+        chars_avail = tud_cdc_available();
+#endif
```
lurch, line 65: Could the first two `#ifdef`s be combined into one block just after `tud_task()` ?

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -56,8 +56,17 @@ static int64_t timer_task(__unused alarm_id_t id, __unused void *user_data) {
...
         mutex_exit(&stdio_usb_mutex);
+#if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
+        if (chars_avail) chars_available_callback(chars_available_param);
```
lurch, line 68: The previous code did `if (chars_available_callback) {` before actually calling `chars_available_callback` - does the same thing need to happen here too?

src/rp2_common/pico_stdio_usb/stdio_usb.c
peterharperuk: Depends on the reviewer. I get complaints with some people about not declaring variables at the begining. Lets try it and see

kilograham (approved):

PR #1651, opened 2024-02-26
Add some PIO utility functions

### LINKED ISSUE #1616: pio helper functions
It would be nice to add a pio_init function that picks a free pio, state machine and loads a program, as I keep writing this and it's easy to get wrong.

bool pio_init(const pio_program_t *program, PIO *pio_hw, uint *sm, uint *offset);
static void pio_deinit(const pio_program_t *program, PIO pio_hw, uint sm, uint offset);

and while we're at it, pio_get_instance like
static inline i2c_inst_t *i2c_get_instance(uint instance)

peterharperuk:
Functions for finding PIO, sm and loading a program and then later freeing resources. And a function that returns the IRQ number for a PIO

Fixes #1616


peterharperuk:
Done already

PR #1671, opened 2024-03-18
Bug 1668 - method to prevent SYS_CLK being re-init'ed, for example after it was left running in 'lightsleep'

mungewell:
During 'lightsleep' user may want to keep some clocks running (for example PIOs),
but at present microPython calls 'clock-init()' causing the SYS_CLK to be reinitialised.

This patch allows microPython to provide the previously used 'sleep_en0 & sleep_en1' 
signifying which clocks stayed on. As most are dependant of SYS_CLK it can be 
inferred that this does NOT need to (/should NOT) be re-initilised.

### DIFF
--- a/.github/workflows/macOS.yml
+++ b/.github/workflows/macOS.yml
@@ -23,8 +23,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew install cmake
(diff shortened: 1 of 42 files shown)


mungewell:
No..... not into master. :-O

Issue #1668, opened 2024-03-14
Add ability to specify which clocks are re-initialized in 'clock_init()', after microPython 'lightsleep()'


mungewell:
From thread:
https://github.com/orgs/micropython/discussions/14087

I am looking to keep the PIOs running during 'lightsleep()', as this will help my project save power. The CPU can pre-fill the FIFOs and then sleep, whilst the PIOs are processing the data.

ATM microPython 'lightsleep()' function calls the 'clock_init()' func after the CPU re-awakes, which causes the clocks to be reinitialized. This is 'bad' for me, as it overwrites the prior 'machine.freq(x)' call, but also potentially causes a glitch on the SYS clock - and therefore glitch on the PIO clocks.

This patch allows the passing of previously used `sleep_en0` and `sleep_en1` values, which are used to determine whether SYS_CLOCK should be reinitialized, or skipped:

[20240314_sdk_keep_pios_running.patch.txt](https://github.com/raspberrypi/pico-sdk/files/14602541/20240314_sdk_keep_pios_running.patch.txt)

lurch:
I see that your patch is modifying `src/rp2040/hardware_regs/include/hardware/regs/clocks.h` - FYI everything inside `src/rp2040/hardware_regs` is autogenerated, so you'll have to move the changes you made there to some other file :wink: 

And I also think we generally prefer Pull Requests, rather than patch files attached to Issues.

mungewell:
I presume you mean that they are generated from `pico-sdk/src/rp2040/hardware_regs/rp2040.svd`.

How is that done? Can someone point me at a script? What I had added was a 'collation' of the 'SYS_CLK' dependants, not sure how to auto-gen that....

No problem for a PR, if I can figure it out.

lurch:
> I presume you mean that they are generated from `pico-sdk/src/rp2040/hardware_regs/rp2040.svd`.

Not quite. Both `src/rp2040/hardware_regs/include/hardware/regs/` and `src/rp2040/hardware_regs/rp2040.svd` are generated from the same internal data-files. (which are also used to generate the register tables in https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf ) 

> How is that done? Can someone point me at a script?

This is done with internal (private) tooling.

mungewell:
I am back working on my project for which this is for. I note that the init code has changed substantially in the Pico-SDK and the patch is no longer valid. But....

[...]

PR #1672, opened 2024-03-18
Bug 1668 v2 - Provide method to prevent init of SYS_CLK, when left on during 'lightsleep'

mungewell:
During 'lightsleep' user may want to keep some clocks running (for example PIOs), but at present microPython calls 'clock-init()' causing the SYS_CLK to be reinitialized.

This patch allows microPython to provide the previously used 'sleep_en0 & sleep_en1' signifying which clocks stayed on. As most are dependant of SYS_CLK it can be inferred that this does NOT need to (/should NOT) be re-initialized.

### DIFF
--- a/src/rp2_common/hardware_clocks/clocks.c
+++ b/src/rp2_common/hardware_clocks/clocks.c
@@ -163,11 +163,15 @@ bool clock_configure(enum clock_index clk_index, uint32_t src, uint32_t auxsrc, 
 }
 /// \end::clock_configure[]
 
-void clocks_init() {
+void clocks_init(void) {
     clocks_reinit(0, 0);
 }
 
 void clocks_reinit(uint32_t sleep_en0, uint32_t sleep_en1) {
+	// Determine whether SYS_CLK is already running
+    bool init_sys_clock =  	((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &&
+                                ((sleep_en1 & CLOCKS_SLEEP_EN1_CLK_SYS_DEPENDANTS_BITS) == 0);
+
     // Start tick in watchdog, the argument is in 'cycles per microsecond' i.e. MHz
     watchdog_start_tick(XOSC_KHZ / KHZ);
 
@@ -187,8 +191,7 @@ void clocks_reinit(uint32_t sleep_en0, uint32_t sleep_en1) {
     xosc_init();
 
     // Before we touch PLLs, switch sys and ref cleanly away from their aux sources.
-    if (((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &&
-        ((sleep_en1 & CLOCKS_SLEEP_EN1_CLK_SYS_DEPENDANTS_BITS) == 0)) {
+	if (init_sys_clock) {
 		hw_clear_bits(&clocks_hw->clk[clk_sys].ctrl, CLOCKS_CLK_SYS_CTRL_SRC_BITS);
 		while (clocks_hw->clk[clk_sys].selected != 0x1)
 			tight_loop_contents();
@@ -198,8 +201,7 @@ void clocks_reinit(uint32_t sleep_en0, uint32_t sleep_en1) {
         tight_loop_contents();
 
     /// \tag::pll_init[]
-    if (((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &&
-        ((sleep_en1 & CLOCKS_SLEEP_EN1_CLK_SYS_DEPENDANTS_BITS) == 0))
+	if (init_sys_clock)
         pll_init(pll_sys, PLL_COMMON_REFDIV, PLL_SYS_VCO_FREQ_KHZ * KHZ, PLL_SYS_POSTDIV1, PLL_SYS_POSTDIV2);
     pll_init(pll_usb, PLL_COMMON_REFDIV, PLL_USB_VCO_FREQ_KHZ * KHZ, PLL_USB_POSTDIV1, PLL_USB_POSTDIV2);
     /// \end::pll_init[]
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -138,15 +187,20 @@ void clocks_init(void) {
     xosc_init();
 
     // Before we touch PLLs, switch sys and ref cleanly away from their aux sources.
-    hw_clear_bits(&clocks_hw->clk[clk_sys].ctrl, CLOCKS_CLK_SYS_CTRL_SRC_BITS);
-    while (clocks_hw->clk[clk_sys].selected != 0x1)
-        tight_loop_contents();
+    if (((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &
```
lurch, line 190: Is the `&` at the end of this line supposed to be `&&` ?

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -164,6 +164,7 @@ extern "C" {
  *  Must be called before any other clock function.
  */
 void clocks_init(void);
+void clocks_reinit(uint32_t sleep_en0, uint32_t sleep_en1);
```
lurch, line 167: Some Doxygen function-documentation here would be great :slightly_smiling_face:

mungewell:
Yes, should be a logical "&&".

I'll update PR and add some documentation. Thanks.

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -165,6 +165,15 @@ extern "C" {
  */
 void clocks_init(void);
 
+/*! \brief Re-initialise the clock hardware
+ *  \ingroup hardware_clocks
+ *
+ *  During a 'lightsleep()' some clocks may have been deliberately left running.
```
lurch, line 171: Looks like `lightsleep()` is only a MicroPython function, not a Pico C-SDK function, so perhaps `'lightsleep()'` here should be changed to `"light sleep"` ?

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -118,7 +163,11 @@ bool clock_configure(enum clock_index clk_index, uint32_t src, uint32_t auxsrc,
 }
 /// \end::clock_configure[]
 
-void clocks_init(void) {
+void clocks_init() {
```
lurch, line 166: Should the `void` be left in the parentheses, as it was originally? (IIRC some compilers get grumpy without this)

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -160,11 +214,14 @@ void clocks_init(void) {
 
     /// \tag::configure_clk_sys[]
     // CLK SYS = PLL SYS (usually) 125MHz / 1 = 125MHz
-    clock_configure(clk_sys,
-                    CLOCKS_CLK_SYS_CTRL_SRC_VALUE_CLKSRC_CLK_SYS_AUX,
-                    CLOCKS_CLK_SYS_CTRL_AUXSRC_VALUE_CLKSRC_PLL_SYS,
-                    SYS_CLK_KHZ * KHZ,
-                    SYS_CLK_KHZ * KHZ);
+    if (((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &&
```
lurch, line 217: Hmmm, rather than doing this same logic-test 3 separate times, _perhaps_ it makes sense to do:
```
bool init_sys_clock = ((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &&
                      ((sleep_en1 & CLOCKS_SLEEP_EN1_CLK_SYS_DEPENDANTS_BITS) == 0)
```
just once? :thinking:

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -163,11 +163,15 @@ bool clock_configure(enum clock_index clk_index, uint32_t src, uint32_t auxsrc,
...
 /// \end::clock_configure[]
 
-void clocks_init() {
+void clocks_init(void) {
     clocks_reinit(0, 0);
 }
 
 void clocks_reinit(uint32_t sleep_en0, uint32_t sleep_en1) {
+	// Determine whether SYS_CLK is already running
+    bool init_sys_clock =  	((sleep_en0 & CLOCKS_SLEEP_EN0_CLK_SYS_DEPENDANTS_BITS) == 0) &&
```
lurch, line 172: Can you replace spaces with tabs please, and fix up the indentation? :slightly_smiling_face:

mungewell:
I removed the tabs.

Do you want the whole series resubmitted (_v3?) as a single patch/pull request?

lurch:
That's not my decision to make, so leave it as-is for now unless asked otherwise :slightly_smiling_face:

kilograham:
fixes #1668

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -160,11 +216,13 @@ void clocks_init(void) {
...
-                    SYS_CLK_KHZ * KHZ,
-                    SYS_CLK_KHZ * KHZ);
+    if (init_sys_clock) {
+        clock_configure(clk_sys,
+                        CLOCKS_CLK_SYS_CTRL_SRC_VALUE_CLKSRC_CLK_SYS_AUX,
+                        CLOCKS_CLK_SYS_CTRL_AUXSRC_VALUE_CLKSRC_PLL_SYS,
+                        SYS_CLK_KHZ * KHZ,
+                        SYS_CLK_KHZ * KHZ);
+    }
     /// \end::configure_clk_sys[]
```
lurch, line 226: This "tag" is used to auto-include this code snippet into page 191 of https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf
Screenshot from 2024-05-20 10-55-44

So in order not to clutter that example, it might make sense to move your new `if` check to **outside** of the tag? i.e.
```c
    if (init_sys_clock) {
        /// \tag::configure_clk_sys[]
        // CLK SYS = PLL SYS (usually) 125MHz / 1 = 125MHz
        clock_configure(clk_sys,
                        CLOCKS_CLK_SYS_CTRL_AUXSRC_VALUE_CLKSRC_PLL_SYS,
                        SYS_CLK_KHZ * KHZ,
                        SYS_CLK_KHZ * KHZ);
        /// \end::configure_clk_sys[]
    }
```

projectgus:
If this patch was extended to also check and selectively init PLL_USB then it would also meet the use case in #1746.

kilograham:
cc @will-v-pi to see if handled by new lower power code

will-v-pi:
> cc @will-v-pi to see if handled by new lower power code

The new low power code doesn't have this issue as it doesn't disable any clocks while sleeping, so doesn't need to reinitialise the clocks when it wakes up.

Currently, Micropython disables the unused clocks during sleep, and therefore needs to reinitialise them afterwards, causing these issues. If they switch to the new low power code, this issue will go away, but there may be a slight increase in power consumption during sleep as the unused clocks are just gated during sleep, rather than being disabled.

Issue #1670, opened 2024-03-17
Add static version of queue


daveythacher:
I was looking into the queue API and I noticed that it works off the heap. It should be possible to implement this as static?
https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/common/pico_util/queue.c#L11-L18

It is a little more risky, but for people that do not want to use heap they are probably stuck with that.
```C
void static_queue_init_with_spinlock(queue_t *q, uint element_size, uint element_count, uint spinlock_num, uint8_t *storage) {
    lock_init(&q->core, spinlock_num);
    q->data = storage;
    q->element_count = (uint16_t)element_count;
    q->element_size = (uint16_t)element_size;
    q->wptr = 0;
    q->rptr = 0;
}
```

peterharperuk:
Yes, that would be good

lurch:
In your proposed new function, it might also be worth passing in a `storage_size` parameter, and then asserting that `storage_size >= ((element_count + 1) * element_size)` ?

daveythacher:
Forgot about the assert. I had considered it but did not see the point in changing the return signature.

Note there is no nullptr check.

el-bert0:
Great idea, I find it useful. I would propose to append `_static` instead of prepend `static_` as all the functions of the `queue.h` starts with `queue_`

PR #1673, opened 2024-03-18
#1670 adding static queue methods

daveythacher:
Adding static_queue_init and static_queue_init_spinlock. These accept a pointer to a static storage array. This allows using the queue API without calloc. Added null checks to queue logic also.

No testing performed.

### DIFF
--- a/src/common/pico_util/include/pico/util/queue.h
+++ b/src/common/pico_util/include/pico/util/queue.h
@@ -62,6 +62,31 @@ static inline void queue_init(queue_t *q, uint element_size, uint element_count)
     queue_init_with_spinlock(q, element_size, element_count, next_striped_spin_lock_num());
 }
 
+/*! \brief Initialise a static queue with a specific spinlock for concurrency protection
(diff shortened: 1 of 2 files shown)


peterharperuk:
Looks ok. We might want to protect queue_free so it does nothing if it memory was not allocated on the heap? And maybe add an example, although you can leave that to us.

daveythacher:
Queue free is a good point. Part of me wants to create a new type which forces a duplication of the API. Part of me says we could add a field to the queue structure. Overall I think we could point out in the documentation or example that queue free is not required for static queues. This is ambiguous, which can be problematic.

peterharperuk:
It might be possible to assert the memory is in the heap? But you can leave queue_free as it is for now.

daveythacher:
Any objections to removing the heap completely? (Probably more of a question for version 2.0 of the SDK.)

peterharperuk:
I think as much as we'd like to remove stuff we don't like, we're stuck with it for the time being.

PR #1675, opened 2024-03-20
upgrade btstack to 1.6

emanuellopes:
- Updating btstack to version 1.6
- new features like sending BLE reports for different ID's are ready to be used for example: hids_device_send_input_report_for_id()


emanuellopes:
Thank you, I will close my PR.

PR #1699, opened 2024-05-07
pico_flash stops including certain unneeded headers

### LINKED ISSUE #1698: pico_flash should depend on hardware_exception
Hi there, I think pico_flash should depend on hardware_exception? It unconditionally includes hardware_exception, and if I try to compile a file that depends on pico_flash without editing the SDK, I get this compile error:

```
[build] /home/jaguilar/projects/pico-sdk/src/rp2_common/pico_flash/flash.c:8:10: fatal error: hardware/exception.h: No such file or directory
[build]     8 | #include "hardware/exception.h"
[build]       |          ^~~~~~~~~~~~~~~~~~~~~~
```

If I add hardware_exception to the dependency list of pico_flash then everything works as expected.

jaguilar:
Fixes #1698

### DIFF
--- a/src/rp2_common/pico_flash/flash.c
+++ b/src/rp2_common/pico_flash/flash.c
@@ -5,7 +5,6 @@
  */
 
 #include "pico/flash.h"
-#include "hardware/exception.h"
 #include "hardware/sync.h"
 #if PICO_FLASH_SAFE_EXECUTE_PICO_SUPPORT_MULTICORE_LOCKOUT
 #include "pico/multicore.h"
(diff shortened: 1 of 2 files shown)


kilograham:
Yeah, we have `kitchen_sink` test already which pulls in all libraries to make sure they compile, so pls remove the new test

jaguilar:
@kilograham kitchen_sink_test is already building without my change, however, the test I'm adding is not. I'm still happy to remove it, but I think kitchen_sink_test is hiding the bug by pulling in hardware_exception as a separate dependency.

kilograham:
ah good point, but yeah, i think remove it... we should add a separate (hopefully automated) check of each library

jaguilar:
Updated the state to just remove those three includes,

kilograham (approved):

PR #1710, opened 2024-05-21
Improve Arm Compiler 6 compatibility

will-v-pi:
Add some more `#ifndef`s to improve compatibility with Arm Compiler 6. This branch of the SDK has been used in the [RP2xxx_DFP](https://github.com/raspberrypi/CMSIS-RP2xxx-DFP)

Additionally AC6 doesn't use the same linker scripts, so is missing the preinit_array and init_array definitions, along with some clock_t libraries used in `runtime.c`.


GorgonMeducer:
Please also add the following changes to be compatible with cmsis 6.1.0:

image

will-v-pi:
Have added those changes in - thanks

PR #1731, opened 2024-06-13
Expand bazel build to include configuration options and broader support.

armandomontanez:
Increases coverage of the Bazel build, including:

- Nearly all annotated CMake build configuration options are supported in Bazel.
- Pico W is now supported.
- Host build works.
- Tests build and pass.
- A new script was added to ensure consistency between Bazel and CMake configuration options.
- PIO assembly has Bazel integration.
- ELF2UF2 has Bazel integration.
- Significant cleanups to improve correctness and documentation.


armandomontanez:
@tpudlik

kilograham (approved):

Issue #1553, opened 2023-11-20
`pico_btstack/CMakeLists.txt` makes `PICO_BTSTACK_PATH` unusable


usedbytes:
The `pico_btstack` library `CMakeLists.txt` hard-codes all of the `btstack` source files. This means that it's tightly coupled to a specific `btstack` version.

This is causing problems for my project, `picow_ds4` (see https://github.com/usedbytes/picow_ds4/issues/1) which submodules `btstack` in order to support Playstation 4 controllers (which have HID descriptors which are too big for the upstream default) - and uses `PICO_BTSTACK_PATH` to use that submoduled version. `pico-sdk` has updated its `btstack`, I have not, so now my project doesn't build with an updated `pico-sdk`.

Having all of the source files listed in a `CMakeLists.txt` which lives inside `pico-sdk` seems like a layering violation - it's effectively an unmanaged dependency to a specific version of `btstack`, which is somewhat fragile.

You're welcome to argue that the submodule provides the versioning :-) - if you have the same version of `btstack` as what's submoduled in to `pico-sdk` then there's no issue. However it does make it basically impossible for projects to use `PICO_BTSTACK_PATH` unless they _also_ submodule `pico-sdk`, which I'd like to avoid if possible.

I wonder if it would be better to have the `pico_btstack` build done within `btstack` itself, as a port, rather than managed by the `pico-sdk` build directly?

peterharperuk:
Yes. If there was a cmake file in btstack we would use it.

peterharperuk:
@mringwal Do you have an opinion on this? It would be nice to not hardcode the list of source files in the pico-sdk but just include it from btstack. I think I copied the list from Micropython, so they have the same issue. I imagine I could convert your Makefile.inc files to cmake files with some "magic" but I doubt that would be that robust.

mringwal:
@peterharperuk this is mainly relevant while people try stuff that hasn't been used much in the past ;)
In general, I agree with the observations. I guess there are two main options here: a) is to provide a minimal CMakeLists.txt that contain all files from btstack/src that are currently used for inclusion (that would synchronize the list with the actual files). The fancier version is to fully embrace CMake. We started to consolidate our various CMake builds and have an early proof-of-concept (see global-cmake branch). I'm not sure yet, what path we'll use.

kilograham:
Yeah, note some libraries just provide a CMake file which defines a variable with all the filenames

PR #1736, opened 2024-06-19
Update btstack v1.6.1

peterharperuk:


peterharperuk:
fyi @mringwal This seems to work for develop as well

src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -181,9 +193,11 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
             ${PICO_BTSTACK_PATH}/3rd-party/bluedroid/encoder/srce/sbc_encoder.c
             ${PICO_BTSTACK_PATH}/3rd-party/bluedroid/encoder/srce/sbc_packing.c
             )
-    target_include_directories(pico_btstack_sbc_encoder_headers INTERFACE
-            ${PICO_BTSTACK_PATH}/3rd-party/bluedroid/encoder/include
-            )
+    if (PICO_BTSTACK_VERSION VERSION_GREATER_EQUAL "1.6.0")
```
mringwal, line 196: If you like, you can drop this check (less complexity). The SBC codec has been around for many years in BTstack

peterharperuk: The check for version 1.6.1 is just whether to add the source file btstack_sbc_bluedroid.c which seems to be new.

I haven't been taking much care updating this cmake file in the past and someone was complaining that it can cause problems if they have their own copy of btstack using an older version.

mringwal: Well, it isn't. The SBC codec in 3rd-part/bluedroid was added in 2016 which pre-dates pico-sdk by a few years.
(BTstack v1.1 was tagged in 2019 btw).

So, I'd suggest to remove this check (to avoid clutter) and ignore or discuss with the person that complained about it.

peterharperuk: btstack_sbc_bluedroid.c was added in Nov 2023. We last updated in June 2023. If I don't add it I get a link error for btstack_sbc_encoder_bluedroid_init_instance. If I add it without any of this clutter older versions of btstack won't build. This is the issue I was thinking about https://github.com/raspberrypi/pico-sdk/issues/1553

```
commit 113ff2bc0af39c4454c5c239318755ec65070fce
Author: Matthias Ringwald <matthias@ringwald.ch>
Date:   Wed Nov 8 15:06:21 2023 +0100

    btstack_sbc_bluedroid: implement encoder for Bluedroid SBC

```

mringwal: Oh, sorry, my bad. The Bluedroid codec (incl. headers) has 'always' been there, there was a need to refactor how it's used and that's the new file. Ok, go ahead with the checks (which are not clutter then :)
On the other hand, one could argue that people should not downgrade BTstack within Pico SDK.

peterharperuk:
> people should not downgrade BTstack within Pico SDK

I agree. The issue here is we allow clients to use their own btstack folder in their project. So it's not that they are downgrading, just that they haven't (yet) upgraded.

src/rp2_common/pico_btstack/CMakeLists.txt
peterharperuk: > people should not downgrade BTstack within Pico SDK

I agree. The issue here is we allow clients to use their own btstack folder in their project. So it's not that they are downgrading, just that they haven't (yet) upgraded.

src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ if (DEFINED ENV{PICO_BTSTACK_PATH} AND (NOT PICO_BTSTACK_PATH))
     message("Using PICO_BTSTACK_PATH from environment ('${PICO_BTSTACK_PATH}')")
 endif ()
 
+if (NOT PICO_BTSTACK_VERSION)
```
kilograham, line 6: How is the user supposed to use this; I assume they are supposed to set it if they are using a different version of btstack? It needs have PICO_CMAKE_CONFIG in that case, however it feels a bit odd that we expect them to set it, if it doesn't actual cause anything to happen e.g. choosing a version (if we just care about what version we have, can we determine it?)

peterharperuk: Yes, they'd define this if they want to use this cmake file to build an older version of btstack than the version in the sdk.

> it feels a bit odd that we expect them to set it, if it doesn't actual cause anything to happen
Sorry, I don't follow this comment. It would cause the build to not fail.

peterharperuk: > can we determine it?

Yes, maybe that's better. We can infer it's an older version by seeing if the file is there. I'll try that out.

mringwal: Ideally, all information about BTstack could/should be contained in the BTstack repo and Pico SDK would only refer to it. I'm not too familiar with your build system, but you may make some suggestions on how this could look like. E.g. we could stash Pico related files under port/pico-sdk, or just have CMakeLists.txt files that contains source files and include paths.

peterharperuk: I've pushed a version that just checks if the file exists. 

> Ideally, all information about BTstack could/should be contained in the BTstack repo

The ideal is to just include a file in the btstack repo. But arguably, it's too late now?

kilograham (approved):

PR #1763, opened 2024-07-09
Add C11 standard atomic support

sgstreet:
This new version of the PR #1645 which was merged, closed and reverted.  This version removes all usage of the GCC specific optimize attribute.


sgstreet:
@kilograham @petrhosek Here is some fixes for the MacOS and Windows compilation failures.  I also fixed a problem with the preinit section so it works as intended.  Is is possible to launch the build checks so I can verify the fixes?

sgstreet:
FYI the error:

```
D:/a/pico-sdk/pico-sdk/src/rp2_common/pico_atomic/pico_atomic.c:208:21: error: mismatch in return type of built-in function '__atomic_fetch_sub_4'; expected 'unsigned int' [-Werror=builtin-declaration-mismatch]
  208 | __optimize uint32_t __atomic_fetch_sub_4(volatile void *mem, uint32_t val, __unused int model) {
      |                     ^~~~~~~~~~~~~~~~~~~~
```
Is a GCC bug as `uint32_t` and `unsigned int` are the same type on the ARMv6 arch and interesting the other atomic types do not generate compiler warnings.

src/rp2_common/pico_atomic/pico_atomic.c
```diff
@@ -192,56 +193,56 @@ bool __atomic_compare_exchange_2(volatile void *mem, void *expected, uint16_t de
     return result;
 }
 
-uint32_t __atomic_fetch_add_4(volatile void *mem, uint32_t val, __unused int model) {
-    volatile uint32_t *ptr = mem;
-    uint32_t state = __atomic_lock(mem);
-    uint32_t result = *ptr;
+unsigned int __atomic_fetch_add_4(volatile void *mem, unsigned int val, __unused int model) {
```
lurch, line 196: It looks like this (and the similiar `foo_4` functions) should probably have remained as
```c
uint32_t __atomic_fetch_add_4(volatile void *mem, uint32_t val, __unused int model) {
```
?

Esepcially as the other functions have signatures like:
```c
uint8_t __atomic_fetch_add_1(volatile void *mem, uint8_t val, __unused int model) {
uint16_t __atomic_fetch_add_2(volatile void *mem, uint16_t val, __unused int model) {
uint64_t __atomic_fetch_add_8(volatile void *mem, uint64_t val, __unused int model) {
```

sgstreet:
Is the way the turn on -Wextra for the linux builds so that any further problems are exposed before I push more to the PR.

src/rp2_common/pico_atomic/pico_atomic.c
sgstreet: I agree but there is a GCC build here when -Wextra is enabled.  See https://github.com/raspberrypi/pico-sdk/actions/runs/9848449043/job/27190379443. The current linux cmake configuration does not enable -Wextra but I'm not clear how to enable it as my CMake skill are not great.

lurch: Ah, sorry! I made this comment from just looking at the `diff` view on GitHub, before realising that you'd actually specifically commented on this already!! :facepalm:

sgstreet: No, no I was hacking up a comment at the same time as you were reviewing and commenting.

kilograham:
Ah, sorry, a lot going on, so maybe didn't mention this yesterday: since when we have a rather different implementation from a separate source.

sgstreet:
So no interest in this?

kilograham:
@sgstreet did you actually verify that your functions are called by GCC; I saw that you #undefed a few functions, but I see on GCC that it just uses its own intrinsics by default

sgstreet:
Yes I tested it, the functions 
```
atomic_flag_test_and_set
atomic_flag_test_and_set_explicit
atomic_flag_clear
atomic_flag_clear_explicit
```
are handled by GCC slightly differently than others, thus the overide.  This following is an excerpt from my test map file:

```
 .text.__atomic_init
                0x10003e3c        0xc CMakeFiles/atomic-test.dir/pico-sdk/src/rp2_common/pico_atomic/pico_atomic.c.obj
 .text.__atomic_fetch_add_4
                0x10003e48       0x98 CMakeFiles/atomic-test.dir/pico-sdk/src/rp2_common/pico_atomic/pico_atomic.c.obj
                0x10003e48                __atomic_fetch_add_4
 .text.__atomic_compare_exchange_4
                0x10003ee0       0xb4 CMakeFiles/atomic-test.dir/pico-sdk/src/rp2_common/pico_atomic/pico_atomic.c.obj
                0x10003ee0                __atomic_compare_exchange_4
 .text.__atomic_test_and_set_m0
                0x10003f94       0xa0 CMakeFiles/atomic-test.dir/pico-sdk/src/rp2_common/pico_atomic/pico_atomic.c.obj
                0x10003f94                __atomic_test_and_set_m0
 .text.__atomic_clear_m0
                0x10004034        0xc CMakeFiles/atomic-test.dir/pico-sdk/src/rp2_common/pico_atomic/pico_atomic.c.obj
                0x10004034                __atomic_clear_m0
```
Did you decide to take PR?

sgstreet:
I have used the extensively in a slightly different form. See https://github.com/sgstreet/pico-toolkit/blob/a9172fc66659217560a13445877723b804a71347/test/atomic-test/atomic-test.c#L12.

kilograham:
What functions was your test calling? 

 > Did you decide to take PR?
 
I am not taking the PR as Clang works somewhat differently; i'm trying to take the best parts of both (yours and another), but when i tried calling

```
volatile atomic_uint_fast32_t foo;
uint32_t bar = atomic_load(&foo);
```

it uses its own non multi-core aware code; are you using explcit methods of the right size? also what GCC version?

sgstreet:
Disappointed you did not take the PR nor ask for a clang compatible solution as I would have researched it and built one. 

But to answer you question, please consider how the AHB Lite Bus operates. 1, 2 and 4 byte bus loads and stores are guaranteed atomic with respect to multi-masters. In some circumstances (not so important on ARMv6-m), a memory barrier will be required to ensure memory ordering. I'm sure you examined the generated code for the `atomic_load` above and saw the `dmb ish` or `dmb sys` instructions. The generated load/store code is sufficient and there is no need to implement replacements. This is not the case for 8 byte quantities and atomic read/modify/write functions. 

If you replace `atomic_uint_fast32_t` with `atomic_uint_fast64_t` use should see in the disassembly a call to `__atomic_load_8`. Further, any call to a read/modify/write atomic function such as `atomic_exchange` will generate a function call to one the size matched functions in the library.

How are you implementing the multi-core awareness?  Using the hardware semaphores? If so how many semaphore bits are you allocating?  I have code which does this, but uses all 32 semaphore bits, not a big deal as once you have atomic variables the need for the raw hardware semaphores is reduced. This of course is not compatible with the SDK as current implemented. This could be reduced to 8 or 16 bits with minor impact to performance. Just curious really.

Hope this helps.

sgstreet:
There is not a write buffer in front of the SRAM correct?

kilograham:
> But to answer you question, please consider how the AHB Lite Bus operates. 1, 2 and 4 byte bus loads and stores are guaranteed atomic with respect to multi-masters

Yes, sorry i wasn't thinking thru; CLang still emits calls on m0plus, and I was just comparing that with GCC.

> Disappointed you did not take the PR nor ask for a clang compatible solution as I would have researched it and built one.

I needed it yesterday, and when i first took the PR, there were a bunch of glaring bugs (which you have since fixed)... this interacts with some other stuff we're doing, so I didnt have time to go back and forth and fix the issues on github. thanks for your help, the spirit of your code is intact!!

kilograham:
> There is not a write buffer in front of the SRAM correct?

correct

kilograham:
fixed in SDK 2.0.0 thanks for your PR

PR #1773, opened 2024-07-23
vcocalc.py improvements

### LINKED ISSUE #1451: vcocalc.py should have an option to use the default REFDIV of 2
#952 added support for non-standard REFDIV values in vcocalc.py (though doesn't add command line arguments for them)

REFDIV 1 is now supported, so will be used in preference to REFDIV 2 in many cases (and set_sys_clock_pll by default only uses the default REFDIV value of 2)

perhaps adding the refdiv-min and refdiv-max aeguments and defaulting the former to 2 would be a reasonable way to solve the problem

### LINKED ISSUE #1665: vcocalc.py gives results that set_sys_clock_khz rejects
Don't know if this is an issue or if I am just overlooking something.  I am setting custom clocks in CMakeLists because I am using a non-standard crystal.  Everything seemed to be working except when I would try to switch frequencies during operation(for runtime selectable speed boost).  After set_sys_clock_pll I would lose printf, even if I set the clock before initial stdio_init_all.

Then I decided to take a step back and let set_sys_clock_khz change the frequencies for me but that failed too.  The settings I was supplying to set_sys_clock_khz kept getting rejected, but they were calculated by vcocalc.  I took a quick look at the code and check_sys_clock_khz does a check "!(vco_khz % (postdiv1 * postdiv2))" which is where my values were failing.  For now I just put a similar check in vcocalc to block those results from being given.  Now with the newly calculated values I can change frequencies and everything works.  Am I missing something or is this a bug?

will-v-pi:
Add extra check to match checks in check_sys_clock_khz, and add --lock-refdiv option to set fixed refdiv

Also add blurb to the output when a non-standard refdiv is required

Fixes #1665, and fixes #1451

### DIFF
--- a/src/rp2_common/hardware_clocks/scripts/vcocalc.py
+++ b/src/rp2_common/hardware_clocks/scripts/vcocalc.py
@@ -2,15 +2,6 @@
 
 import argparse
 
-parser = argparse.ArgumentParser(description="PLL parameter calculator")
-parser.add_argument("--input", "-i", default=12, help="Input (reference) frequency. Default 12 MHz", type=float)
-parser.add_argument("--ref-min", default=5, help="Override minimum reference frequency. Default 5 MHz", type=float)
-parser.add_argument("--vco-max", default=1600, help="Override maximum VCO frequency. Default 1600 MHz", type=float)
-parser.add_argument("--vco-min", default=750, help="Override minimum VCO frequency. Default 750 MHz", type=float)
-parser.add_argument("--low-vco", "-l", action="store_true", help="Use a lower VCO frequency when possible. This reduces power consumption, at the cost of increased jitter")
-parser.add_argument("output", help="Output frequency in MHz.", type=float)
-args = parser.parse_args()
-
 # Fixed hardware parameters
 fbdiv_range = range(16, 320 + 1)
 postdiv_range = range(1, 7 + 1)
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_clocks/scripts/vcocalc.py
```diff
@@ -2,23 +2,32 @@
...
 
+# Fixed hardware parameters
+fbdiv_range = range(16, 320 + 1)
+postdiv_range = range(1, 7 + 1)
+ref_min = 5
+refdiv_min = 1
+refdiv_max = 63
+
+def validRefdiv(string):
+    if ((int(string) < refdiv_min) or (int(string) > refdiv_max)):
```
lurch, line 13: Only a minor nit, but I think personally I'd do the `int(string)` once, rather than three separate times?

src/rp2_common/hardware_clocks/scripts/vcocalc.py
```diff
@@ -33,13 +42,33 @@
...
+
+target_compile_definitions(executable_name PRIVATE
+	PLL_SYS_REFDIV={}
+	PLL_SYS_VCO_FREQ_HZ={}
+	PLL_SYS_POSTDIV1={}
+	PLL_SYS_POSTDIV2={}
+)\
+""".format(best[4], int((args.input * 1_000_000) / best[4] * best[1]), best[2], best[3]))
+else:
+	print("UNABLE TO COME UP WITH A SOLUTION....")
```
lurch, line 74: :joy:

PR #1775, opened 2024-07-24
Add extra checks to check_board_header.py

lurch:
The main change here is checking that each "interface" has pins for all of its required "functions".
E.g. if you define `PICO_DEFAULT_I2C` then you also have to define both `PICO_DEFAULT_I2C_SDA_PIN` and `PICO_DEFAULT_I2C_SCL_PIN`, and if you define `PICO_DEFAULT_UART` then you also have to define at least one of`PICO_DEFAULT_UART_TX_PIN` or `PICO_DEFAULT_UART_RX_PIN`.


lurch:
> arguably you don't need to have both UART TX and UART RX

Ahh, I was wondering about that but wasn't sure. Is a similar thing true for SPI where you **have** to have SPI_SCK, but you _might_ have only a SPI_TX or only a SPI_RX ?

lurch:
> arguably you don't need to have both UART TX and UART RX

Fixed with the latest commit :slightly_smiling_face:

kilograham (approved):

Kushal-hub1 (approved):

PR #1784, opened 2024-08-08
Add board definition for Adafruit Feather 2350

tannewt:


lurch (approved):
`tools/check_board_header.py` is happy :+1:

lurch:
Oh, should this have also updated `src/boards/BUILD.bazel` @kilograham ? I'm afraid I know nothing about bazel.

armandomontanez:
> Oh, should this have also updated `src/boards/BUILD.bazel` @kilograham ? I'm afraid I know nothing about bazel.

It's appreciated, but not required. Should be as simple as as adding `"adafruit_feather_rp2350",` to the list of boards in that file. There's a script that will check when new boards aren't added to the Bazel build, so worst case it'll get caught and added later.

PR #1795, opened 2024-08-10
Fixes to picolibc_interface

gneverov:
- Don't include bindings for tinystdio if picolibc was compiled with POSIX_IO.
- Add times() function, which seems to be missed, in the same pattern as settimeofday/gettimeofday.
- GCC complains about taking address of void value of __tls_base.

### DIFF
--- a/src/rp2_common/pico_clib_interface/picolibc_interface.c
+++ b/src/rp2_common/pico_clib_interface/picolibc_interface.c
@@ -29,7 +29,6 @@
 #define weak_raw_vprintf vprintf
 #endif
 
-#ifndef POSIX_IO
 static int picolibc_putc(char c, __unused FILE *file) {
 #if LIB_PICO_STDIO
     stdio_putchar(c);
@@ -41,6 +40,7 @@ static int picolibc_getc(__unused FILE *file) {
 #if LIB_PICO_STDIO
     return stdio_getchar();
 #endif
+    return -1;
 }
 
 static int picolibc_flush(__unused FILE *file) {
@@ -56,7 +56,6 @@ static FILE __stdio = FDEV_SETUP_STREAM(picolibc_putc,
                                         _FDEV_SETUP_RW);
 
 FILE *const stdin = &__stdio; __strong_reference(stdin, stdout); __strong_reference(stdin, stderr);
-#endif
 
 void __weak __assert_func(const char *file, int line, const char *func, const char *failedexpr) {
     weak_raw_printf("assertion \"%s\" failed: file \"%s\", line %d%s%s\n",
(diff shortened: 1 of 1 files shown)


kilograham:
excellent, thank you

kilograham:
Can you explain the upshot of 1.? what uses of picolibc does this affect?

gneverov:
Picolibc has an [option](https://github.com/picolibc/picolibc/blob/main/doc/build.md#options-when-using-tinystdio-bits) to automatically implement stdin etc. using posix functions (read, write, etc.) When this option is on, you don't want Pico SDK to redefine stdin etc.

In the first revision of this PR, I thought the Picolibc option was called POSIX_IO, but it's actually called POSIX_CONSOLE. 🤦‍♂️ I could update this PR with the correct option name, however unlike POSIX_IO, the Picolibc build does not correctly propagate the POSIX_CONSOLE option to the picolibc.h which gets included in the file modified here. This is probably an oversight in Picolibc, and I could submit a PR to Picolibc to fix that, but there's arguably a more general problem in Pico SDK: what if the user wants to define their own implementation of stdin etc.?

There are many ways you could solve that problem. Perhaps the most robust is what Picolibc does [here](https://github.com/picolibc/picolibc/blob/main/newlib/libc/tinystdio/posixiob_stdin.c). So, Pico SDK would define a (strong) symbol `__pico_stdin` and a weak alias `stdin` referring to it. If the user links with Picolibc built without POSIX_CONSOLE then it just works. Otherwise if they link with a Picolibc built _with_ PICO_CONSOLE, they will get a duplicate definition error. Then they can resolve the error by creating their own strong `stdin` alias referring to `__posix_stdin`.

Let me know how you would like me to proceed with this PR. Should I just remove the POSIX_IO part and proceed with the two remaining trivial issues? Are you interested in another PR dealing with the multiple stdin implementation issue?

kilograham:
Thanks, yeah, i think it would be helpful for them to expose all the config settings.

Thus far, our interface with`picolibc` and thus we've been mostly targeting whatever config comes with LLVM ETA. Since both `picolibc` and the SDK try to fix the same problems (both in library implementations, and in interactions with the compiler - e.g. things like thread locals), we should try to make it easy/clear which of the two is in charge in any given area.

Putting the majority of the code in `pico_clib_interface` is a step in the right direction, but we have a ways to go.

Maybe let's open issues for each individual config point where we have problems.

gneverov:
Created issue #1900 for the stdin implementation selection problem.

kilograham (approved):

src/rp2_common/pico_clib_interface/picolibc_interface.c
```diff
@@ -39,6 +40,7 @@ static int picolibc_getc(__unused FILE *file) {
 #if LIB_PICO_STDIO
     return stdio_getchar();
 #endif
+    return -1;
```
lurch, line 43: I guess this _could_ have been
```c
#if LIB_PICO_STDIO
    return stdio_getchar();
#else
    return -1;
#endif
```
but presumably most compilers will make that optimisation automatically?

PR #1816, opened 2024-08-13
Changes for pico2_w

peterharperuk:
Changes for Pico2W


lurch:
The changes I made in #1814 are happy with this new header-file :smiley:

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -330,32 +354,32 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
...
+    gpio_set_dir(CYW43_WL_REG_ON_PIN, GPIO_OUT);
+    gpio_pull_up(CYW43_WL_REG_ON_PIN);
 
     // Setup DO, DI and IRQ (24)
-    gpio_init(DATA_OUT_PIN);
-    gpio_set_dir(DATA_OUT_PIN, GPIO_OUT);
-    gpio_put(DATA_OUT_PIN, false);
+    gpio_init(CYW43_DATA_OUT_PIN);
+    gpio_set_dir(CYW43_DATA_OUT_PIN, GPIO_OUT);
+    gpio_put(CYW43_DATA_OUT_PIN, false);
```
lurch, line 365: I guess this implies that the code would need quite a bit of refactoring if `CYW43_DATA_OUT_PIN`, `CYW43_DATA_IN_PIN` and `CYW43_IRQ_PIN` **weren't** all set to the same value? I wonder if it might be worth enforcing that with a compile-time assertion? :thinking:

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -21,17 +21,41 @@
...
+#define CYW43_DATA_IN_PIN 24u
+#endif
+
+// PICO_CONFIG: CYW43_IRQ_PIN, gpio pin for the irq line from the cyw43 chip, type=int, default=24, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_IRQ_PIN
+#define CYW43_IRQ_PIN 24u
+#endif
+
+// PICO_CONFIG: CYW43_IRQ_PIN, gpio pin for the spi clock line to the cyw43 chip, type=int, default=29, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_CLOCK_PIN
```
lurch, line 46: CYW43_IRQ_PIN -> CYW43_CLOCK_PIN

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -21,17 +21,41 @@
...
+#define CYW43_CLOCK_PIN 29u
+#endif
+
+// PICO_CONFIG: CYW43_IRQ_PIN, gpio pin for the spi chip select to the cyw43 chip, type=int, default=25, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_CS_PIN
```
lurch, line 51: CYW43_IRQ_PIN -> CYW43_CS_PIN

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
peterharperuk: Why would it need "significant" changes? We should allow them to be different. I wouldn't expect it to be too much work?

lurch: I meant that with the code as-currently-written, only `CYW43_DATA_OUT_PIN` is `gpio_init()`ed, and `gpio_init` isn't (yet) called separately for `CYW43_DATA_IN_PIN` and `CYW43_IRQ_PIN`.
(At least that's how it looks from a quick glance, but I've not studied this code in depth, so perhaps it _does_ already cope with the pins being set to different values?)

dpgeorge:
I have a question about board naming, in particular whether there's a space in the "Pico" board names or not.

There are currently the following board definition files in this repo:
```c
include/boards/pico.h:   #define RASPBERRYPI_PICO
include/boards/pico_w.h: #define RASPBERRYPI_PICO_W
include/boards/pico2.h:  #define RASPBERRYPI_PICO2
```
And then this PR adds:
```c
include/boards/pico2_w.h: #define RASPBERRYPI_PICO2_W
```
That seems nice and consistent.  From MicroPython's side we have the corresponding:
- `RPI_PICO` board with name `"Raspberry Pi Pico"`
- `RPI_PICO_W` board with name `"Raspberry Pi Pico W"`
- `RPI_PICO2` board with name `"Raspberry Pi Pico2"`

And then I guess we would add:
- `RPI_PICO2_W` board with name `"Raspberry Pi Pico2 W"`

The board name matters because that's the name of the source directory, eg `ports/rp2/boards/RPI_PICO2`.  And also the name of the firmware, eg `RPI_PICO2-20240816-v1.23.x.uf2`.

---------

My main question is whether there should be a space between "Pico" and "2", because:
- it's referred to as "Raspberry Pi Pico 2" in almost all locations (with a space between "Pico" and "2"), for example in the RP2350 and Pico 2 datasheets
- the datasheet for Pico 2 is called `pico-2-datasheet.pdf`, ie there's a hyphen between the "pico" and "2"
- the URL is https://www.raspberrypi.com/products/raspberry-pi-pico-2/, ie has a hyphen

So that seems a little inconsistent, whether there's a space or not between the "Pico" and the "2", and then that would follow for the "Pico 2 W" (vs "Pico2 W").

I know this is a minor point, but we'd like to get the names correct in MicroPython so the board names, firmware names and download URL are correct and don't need to change in the future.

Should we call our boards:
- `RPI_PICO2` and `RPI_PICO2_W`
- or `RPI_PICO_2` and `RPI_PICO_2_W`?

peterharperuk:
Our board header files have...

```
#ifndef _BOARDS_PICO2_H
#define _BOARDS_PICO2_H

// For board detection
#define RASPBERRYPI_PICO2
```

So I think PICO2_W is consistent. I seriously considered PICO2W but that's inconsistent with PICO_W which doesn't quite work as PICOW. I'll ask around if there's a policy

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -56,6 +56,34 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
             hardware_exception
             )
 
+    if (CYW43_DEFAULT_PIN_WL_REG_ON)
```
peterharperuk, line 59: Need to add PICO_CMAKE_CONFIG stuff for these

src/boards/include/boards/pico2_w.h
```diff
@@ -0,0 +1,116 @@
...
+// pico_cmake_set PICO_CYW43_SUPPORTED = 1
+
+#ifndef _BOARDS_PICO2_W_H
+#define _BOARDS_PICO2_W_H
+
+// For board detection
+#define RASPBERRYPI_PICO2_W
+
+// --- RP2350 VARIANT ---
+#define PICO_RP2350A 1
```
kilograham, line 24: Is it not a B?

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
kilograham: do you want to do that as part of this?

kilograham: Also, why do we want this settable from CMake vs board header?

src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
```diff
@@ -95,16 +99,75 @@ static inline uint32_t cyw43_hal_ticks_ms(void) {
...
+#ifndef CYW43_DEFAULT_PIN_WL_CLOCK
+#define CYW43_DEFAULT_PIN_WL_CLOCK 29u
+#endif
+// PICO_CONFIG: CYW43_DEFAULT_PIN_WL_CS, gpio pin for the spi chip select to the cyw43 chip, type=int, default=25, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_DEFAULT_PIN_WL_CS
+#define CYW43_DEFAULT_PIN_WL_CS 25u
+#endif
+
+// PICO_CONFIG: CYW43_PIN_WL_DYNAMIC, flag to indicate if cyw43 SPI pins can be changed at runtime, type=bool, default=false, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_PIN_WL_DYNAMIC
```
kilograham, line 128: I'm confused why this is useful, or are people trying to write code that runs on multiple boards?

src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
```diff
@@ -95,16 +99,75 @@ static inline uint32_t cyw43_hal_ticks_ms(void) {
...
+#define CYW43_PIN_WL_CS CYW43_DEFAULT_PIN_WL_CS
+#endif // !CYW43_PIN_WL_DYNAMIC
+
 static inline int cyw43_hal_pin_read(cyw43_hal_pin_obj_t pin) {
     return gpio_get(pin);
 }
 
 static inline void cyw43_hal_pin_low(cyw43_hal_pin_obj_t pin) {
-    gpio_clr_mask(1 << pin);
+    gpio_clr_mask64(1 << pin);
```
kilograham, line 166: this is incorrect; should be 1ull << pin, but prefer to use non mask function here

src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
```diff
@@ -95,16 +99,75 @@ static inline uint32_t cyw43_hal_ticks_ms(void) {
...
 }
 
 static inline void cyw43_hal_pin_high(cyw43_hal_pin_obj_t pin) {
-    gpio_set_mask(1 << pin);
+    gpio_set_mask64(1 << pin);
```
kilograham, line 170: this is incorrect; should be 1ull << pin, but prefer to use non mask function here

src/boards/include/boards/pico2_w.h
peterharperuk: Definitely not?

src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
peterharperuk: This "might" be used for Micropython

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
peterharperuk: I have no way of testing different configurations, so I'm closing this as I don't want to make changes which might or might not work.

peterharperuk:
@kilograham I think I resolved all the review comments in this.

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -56,6 +56,43 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
...
+    endif()
+    if (CYW43_DEFAULT_PIN_WL_CS)
+        # PICO_CMAKE_CONFIG: CYW43_DEFAULT_PIN_WL_CS, gpio pin for the spi chip select to the cyw43 chip, type=int, default=25, group=pico_cyw43_driver
+        target_compile_definitions(cyw43_driver_picow INTERFACE CYW43_DEFAULT_PIN_WL_CS=${CYW43_DEFAULT_PIN_WL_CS})
+    endif()
+    if (CYW43_PIO_CLOCK_DIV_INT)
+        # PICO_CMAKE_CONFIG: CYW43_PIO_CLOCK_DIV_INT, integer component of pio clock divider used for cyw43 comms, type=int, default=2, group=pico_cyw43_driver
+        target_compile_definitions(cyw43_driver_picow INTERFACE CYW43_PIO_CLOCK_DIV_INT=${CYW43_PIO_CLOCK_DIV_INT})
+    endif()
+    if (CYW43_PIO_CLOCK_DIV_FRAC)
```
kilograham, line 87: this should probably be _FRAC8 now, or probably we should add _FRAC8 too

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -85,16 +85,10 @@ __force_inline static uint32_t __swap16x2(uint32_t a) {
...
-#endif
-static_assert(CYW43_SPI_PIO_PREFERRED_PIO >=0 && CYW43_SPI_PIO_PREFERRED_PIO < NUM_PIOS, "");
-
 typedef struct {
-    pio_hw_t *pio;
-    uint8_t pio_func_sel;
-    int8_t pio_offset;
-    int8_t pio_sm;
+    PIO pio;
+    uint pio_offset;
```
kilograham, line 90: these should be uint8_t

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
kilograham: actually let's leave this for now; i'll fix with #1926

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
peterharperuk: That would be awkward because pio_claim_free_sm_and_add_program takes uint*

kilograham (approved):

PR #1819, opened 2024-08-14
Support serial flash with 16 byte unique id

### LINKED ISSUE #1641: IS25LP***D flash uID size is wrong in SDK and produce duplicated flash uID
Hi world,

I'm working with custom RP2040 boards with IS25LP***D flash chip (usually IS25LP016D)

I've made custom board file and all works fine.

The trouble is with the uID returned using **_pico_get_unique_board_id( )_** and **_pico_get_unique_board_id_string( )_** functions.

**_pico_unique_board_id_t_** is 8 bytes long but the uID from IS25LP***D is 16 bytes long, so we have a lot of same uID.

Maybe could I override PICO_UNIQUE_BOARD_ID_SIZE_BYTES in my board file ?
... any idea to solve that ?

Thansk for reading.
Sebastien.

peterharperuk:
Flash devices like IS25LP016D have a 16 byte unique id. Reading just the first 8 bytes causes devices to have duplicate unique ids as the numbers only differ in the final 8 bytes.

Support overriding FLASH_UNIQUE_ID_SIZE_BYTES in the board header. Use the final 8 bytes for the unique id.

Fixes #1641

### DIFF
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -46,7 +46,9 @@
 #define FLASH_SECTOR_SIZE (1u << 12)
 #define FLASH_BLOCK_SIZE (1u << 16)
 
+#ifndef FLASH_UNIQUE_ID_SIZE_BYTES
 #define FLASH_UNIQUE_ID_SIZE_BYTES 8
+#endif
 
 // PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes, type=int, default=Usually provided via board header, group=hardware_flash
 
--- a/src/rp2_common/pico_unique_id/unique_id.c
+++ b/src/rp2_common/pico_unique_id/unique_id.c
@@ -8,7 +8,7 @@
 #include "pico/bootrom.h"
 #include "pico/unique_id.h"
 
-static_assert(PICO_UNIQUE_BOARD_ID_SIZE_BYTES == FLASH_UNIQUE_ID_SIZE_BYTES, "Board ID size must match flash ID size");
+static_assert(PICO_UNIQUE_BOARD_ID_SIZE_BYTES <= FLASH_UNIQUE_ID_SIZE_BYTES, "Board ID size must at least be the size of flash ID");
 
 static pico_unique_board_id_t retrieved_id;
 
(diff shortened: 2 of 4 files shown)


andygpz11 (dismissed):
LGTM

src/rp2_common/pico_unique_id/unique_id.c
```diff
@@ -20,8 +20,16 @@ static void __attribute__((constructor)) _retrieve_unique_id_on_boot(void) {
...
+    #elif (PICO_UNIQUE_BOARD_ID_SIZE_BYTES == FLASH_UNIQUE_ID_SIZE_BYTES)
         flash_get_unique_id(retrieved_id.id);
+    #elif (PICO_UNIQUE_BOARD_ID_SIZE_BYTES < FLASH_UNIQUE_ID_SIZE_BYTES)
+        // The flash id is >8 bytes (e.g. IS25LP016D) but we want to keep the
+        // pico unique board id as 8 bytes, just use the last 8 bytes which are likely to change
+        uint8_t flash_id[FLASH_UNIQUE_ID_SIZE_BYTES];
+        flash_get_unique_id(flash_id);
+        memcpy(retrieved_id.id, flash_id + FLASH_UNIQUE_ID_SIZE_BYTES - PICO_UNIQUE_BOARD_ID_SIZE_BYTES, PICO_UNIQUE_BOARD_ID_SIZE_BYTES);
+    #else
+        #error unique board id size is greater than flash unique id size
```
lurch, line 32: Super-minor nitpick, but maybe the "id" in this line should be "ID" ? (to match with line 11)

kilograham (approved):

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -46,7 +46,9 @@
 #define FLASH_SECTOR_SIZE (1u << 12)
 #define FLASH_BLOCK_SIZE (1u << 16)
 
+#ifndef FLASH_UNIQUE_ID_SIZE_BYTES
 #define FLASH_UNIQUE_ID_SIZE_BYTES 8
+#endif
```
lurch, line 51: The description for this PR says "Support overriding FLASH_UNIQUE_ID_SIZE_BYTES in the board header."
Does that mean that in hindsight this should have become a `// PICO_CONFIG:` entry?

peterharperuk: Yes, probably.

PR #1820, opened 2024-08-14
Add PICOTOOL_OVERRIDE_DIR

### LINKED ISSUE #1818: picotool 2.0.0 executable is not found during cmake -G
Environment: Windows 11, SDK 2.0.0, trying to build a project for the Pico.

The issue is that when trying to perform a **cmake -G "NMake Makefiles" ..** or **cmake -G "Ninja" ..** there is a message that picotool is not found and will be built from source.

If the user already has a v2.0.0 picotool executable, it is not detected by the **pico-sdk/tools/Findpicotool.cmake** file.

I cannot use the Pico Setup for Windows installer, because that is version 1.5.

I believe I have a solution (also [described at raspberrypi forum here](https://forums.raspberrypi.com/viewtopic.php?t=375104)).
The solution is to place the prebuilt picotool.exe into any desired folder, and then set a Windows System environment variable called PICOTOOL_OVERRIDE_DIR to that folder path which contains picotool.exe.

Then, I added the following into the pico-sdk/tools/Findpicotool.cmake file, at the top (after the cmake_minimum_required(VERSION 3.17) line:

```
# if PICOTOOL_OVERRIDE_DIR system environment variable is set,
# then use that as the folder for the picotool executable
if (DEFINED ENV{PICOTOOL_OVERRIDE_DIR})
    message("PICOTOOL_OVERRIDE_DIR env var is set to '$ENV{PICOTOOL_OVERRIDE_DIR}'")
    add_executable(picotool IMPORTED GLOBAL)
    set_property(TARGET picotool PROPERTY IMPORTED_LOCATION $ENV{PICOTOOL_OVERRIDE_DIR}/picotool)
    # check the picotool version:
    execute_process(COMMAND $ENV{PICOTOOL_OVERRIDE_DIR}/picotool version
                    OUTPUT_VARIABLE PICOTOOL_VERSION
                    OUTPUT_STRIP_TRAILING_WHITESPACE)
    string(REGEX MATCH "^picotool v${picotool_VERSION_REQUIRED}" PICOTOOL_VERSION_MATCH ${PICOTOOL_VERSION})
    if (NOT PICOTOOL_VERSION_MATCH)
        message("picotool version response was: ${PICOTOOL_VERSION}")
        message(FATAL_ERROR "PICOTOOL_OVERRIDE_DIR is set to '$ENV{PICOTOOL_OVERRIDE_DIR}', but the version of picotool found is not ${picotool_VERSION_REQUIRED}")
    endif()
endif ()
```

What the above does, is if the PICOTOOL_OVERRIDE_DIR system env var is set, it uses the picotool executable in there, tests if it is the correct version, and if so, then simply uses that.

I have tested by doing a build, and can see that the .uf2 file was successfully created. I have attached the output showing that.

In case it is acceptable, I'll create a PR and reference this issue.

[nmake_output.txt](https://github.com/user-attachments/files/16616822/nmake_output.txt)
[cmake_g_nmake_makefile_output.txt](https://github.com/user-attachments/files/16616824/cmake_g_nmake_makefile_output.txt)

shabaz123:
Submitting against 'develop'
Fixes #1818

When using Windows 11 and SDK 2.0.0, when trying to do a **cmake -G "NMake Makefiles" ..** or **cmake -G "Ninja" ..**  then picotool executable is not found, and it gets built within the project. The Findpicotool.cmake file does not seem to find the picotool executable at least with Windows, I have not checked Linux/Mac.

I have a solution which should be benign, since it only kicks in if the user creates a system environment variable in Windows, called PICOTOOL_OVERRIDE_DIR. If that is set, then the picotool.exe file there is used (and the version command is executed to confirm it is the correct version).

The submitted change is in a single file (Findpicotool.cmake), placed near the top of the file, after the cmake_minimum_required(VERSION 3.17) line.

I have tested by performing a build, and I can see that the picotool executable was used, and that the .uf2 file was successfully created.

The attached files evidence that, and at the end there is a dir listing that shows that the .uf2 file was created.

[nmake_output.txt](https://github.com/user-attachments/files/16616889/nmake_output.txt)
[cmake_g_nmake_makefile_output.txt](https://github.com/user-attachments/files/16616890/cmake_g_nmake_makefile_output.txt)

### DIFF
--- a/tools/Findpicotool.cmake
+++ b/tools/Findpicotool.cmake
@@ -6,6 +6,23 @@
 #
 cmake_minimum_required(VERSION 3.17)
 
+# if PICOTOOL_OVERRIDE_DIR system environment variable is set,
+# then use that as the folder for the picotool executable
+if (DEFINED ENV{PICOTOOL_OVERRIDE_DIR})
+    message("PICOTOOL_OVERRIDE_DIR env var is set to '$ENV{PICOTOOL_OVERRIDE_DIR}'")
+    add_executable(picotool IMPORTED GLOBAL)
+    set_property(TARGET picotool PROPERTY IMPORTED_LOCATION $ENV{PICOTOOL_OVERRIDE_DIR}/picotool)
+    # check the picotool version:
+    execute_process(COMMAND $ENV{PICOTOOL_OVERRIDE_DIR}/picotool version
+                    OUTPUT_VARIABLE PICOTOOL_VERSION
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
+    string(REGEX MATCH "^picotool v${picotool_VERSION_REQUIRED}" PICOTOOL_VERSION_MATCH ${PICOTOOL_VERSION})
+    if (NOT PICOTOOL_VERSION_MATCH)
+        message("picotool version response was: ${PICOTOOL_VERSION}")
+        message(FATAL_ERROR "PICOTOOL_OVERRIDE_DIR is set to '$ENV{PICOTOOL_OVERRIDE_DIR}', but the version of picotool found is not ${picotool_VERSION_REQUIRED}")
+    endif()
+endif ()
+
 if (NOT TARGET picotool)
     include(ExternalProject)
 


will-v-pi:
I don't want to add this, as there are already better ways to do this native to CMake. See the [Usage by the SDK](https://github.com/raspberrypi/picotool/blob/develop/README.md#usage-by-the-raspberry-pi-pico-sdk) section of the picotool README for more details.

In short, you can run the following CMake commands from your picotool build directory to install the required files to `$MY_INSTALL_DIR` (replacing `$MY_INSTALL_DIR` with the path you want to use)
```
cmake -DCMAKE_INSTALL_PREFIX=$MY_INSTALL_DIR -DPICOTOOL_FLAT_INSTALL=1 ..
cmake --build .
cmake --install .
```
then set the `picotool_DIR` environment variable to `$MY_INSTALL_DIR/picotool` - so no need to add an additional `PICOTOOL_OVERRIDE_DIR` variable.

bhass1:
Can confirm, after running `cmake --install .` from my cmake build folder of the picotool project, the Findpicotool.cmake finds my precompiled picotool. More details in Issue #1818

shabaz123:
Thanks for the response! 

@will-v-pi does the mixed-case env var picotool_DIR function with Windows? Traditionally env vars are case-insensitive with Windows, and I don't know if Windows CMake would match that with the mixed case (I don't know either way, I've not tried it yet).

I'll close this PR for now, since it's been confirmed working with Linux.

will-v-pi:
I've confirmed it works in Windows too, by running
```
$env:picotool_DIR = '~\.pico-sdk\picotool\2.0.0\picotool'
$env:pioasm_DIR = '~\.pico-sdk\tools\2.0.0\pioasm'
$env:PICO_SDK_PATH = '~\.pico-sdk\sdk\2.0.0'
$env:PICO_TOOLCHAIN_PATH = '~\.pico-sdk\toolchain\13_2_Rel1'
cmake .. -G "NMake Makefiles"
```

PR #1821, opened 2024-08-14
Fix Cmsis irq regression

### LINKED ISSUE #1817: Interrupt handler rename to CMSIS standard broken in 2.0.0
In CircuitPython we use a mix of standard CMSIS irq handler names (like `HardFault_Handler`) and pico sdk specific ones (like `isr_dma_0`).

The rename to standard names doesn't work in 2.0.0 because the IRQ table is now generated with an assembly macro (?) which doesn't appear to happen before the C preprocessor that renames the handlers.

Generator code:
https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/pico_crt0/crt0.S#L55-L64

1.5.1 version:
https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/pico_standard_link/crt0.S#L43-L51

The RP2350 names aren't quite correct either because they end in IRQn:
https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/cmsis/include/cmsis/rename_exceptions.h#L58

kilograham:
Fixes #1817


src/rp2_common/pico_crt0/crt0.S
```diff
@@ -52,17 +52,99 @@ __vectors:
...
 
-.set irq_num, 0
-.rept NUM_IRQS
-isr_irqx %irq_num
-.set irq_num, irq_num + 1
-.endr
-
+// we inlcude a lot of these to allow for different number of IRQs.
+// if_irq_word will only include IRQs that are valid, but we can't
+// use a macro loop becasue irq_irqx MUST appear in the source
```
lurch, line 63: `irq_irqx` -> `isr_irqx` ?

JamesH65: Also, "becasue"

lurch: Also "inlcude" :eyes:

peterharperuk (approved):

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -58,9 +58,9 @@ __vectors:
 .endif
 .endm
 
-// we inlcude a lot of these to allow for different number of IRQs.
+// we include a lot of these to allow for different number of IRQs.
 // if_irq_word will only include IRQs that are valid, but we can't
-// use a macro loop becasue irq_irqx MUST appear in the source
+// use a macro loop becasue isr_irqx MUST appear in the source
```
lurch, line 63: Still says "becasue" instead of "because"

Issue #1812, opened 2024-08-13
Issues with `best_effort_wfe_or_timeout` and WFE in SDK 2.0.0


dpgeorge:
### Background

We are trying to achieve good idle power consumption on RP2350 in MicroPython.  Idle here means sitting at the REPL (with either UART or USB serial) doing nothing.  In this case the CPU should be able to gate its clock via WFI/WFE to reduce power consumption.

Eg on a standard Pico with RP2040, connected to a USB port and a terminal port program open, the Pico draws about 15.2mA (at 5V).  Then running a simple while-True busy loop that increases to about 18mA.  This is expected behaviour.

### Summary of problems

Things seem to have changed in SDK 2.0.0 (the tag 2.0.0).  There are a few issues that seem to be related:
- `best_effort_wfe_or_timeout()` now seems to return immediately, on both RP2040 and RP2350
- `best_effort_wfe_or_timeout()` calls `__sev(); __wfe()` to clear any existing event, which potentially misses events
- on RP2350 spin locks are implemented in software using `ldaexb` and `strexb` exclusive-access instructions, and these seem to set the event flag (equivalent to `__sev()`), meaning that `spin_lock_blocking` sets the event
- due to the above, functions like `hardware_alarm_set_target` set the event flag, making these functions unusable for making a timer to wake from WFE

#### `best_effort_wfe_or_timeout()` now seems to return immediately

Running the following on RP2040:
```c
absolute_time_t timeout_time = make_timeout_time_us(1000000);
while (!best_effort_wfe_or_timeout(timeout_time)) {
}
```
With pico-sdk 1.5.1 that will consume about 15.2mA on a Pico board.  With pico-sdk 2.0.0 that code consumes about 18mA on a Pico board.

Timing how long the `best_effort_wfe_or_timeout` function lasts, on pico-sdk 2.0.0 it always returns pretty much immediately (eg within 5us), so the above loop is effectively a busy loop, hence the higher power consumption.

This looks like a regression with pico-sdk 2.0.0 on RP2040.

#### `best_effort_wfe_or_timeout()` calls `__sev(); __wfe()`

Inspecting the code for `best_effort_wfe_or_timeout` in pico-sdk 2.0.0, there's a new bit:
```c
            // the above alarm add now may force an IRQ which will wake us up,
            // so we want to consume one __wfe.. we do an explicit __sev
            // just to make sure there is one
            __sev(); // make sure there is an event sow ee don't block 
            __wfe();                      
[...]

PR #1822, opened 2024-08-14
Improve best_effort_wfe_or_timeout

kilograham:
Fix some parts of #1812 

The problem with the rewrite of the alarm pool code is that adding an alarm forces an IRQ which cases a SEV. We attempted to fix this in `best_effort_wfe_or_timeout` by consuming the event then doing a WFE again. This is clearly wrong if the event was set prior to the call.

The best we can do (since we cant tell whether an WFE will block) is to only add the alarm if there isn't an IRQ (we only check this alarm pool alarm) already set to happen on or before the target time (which will cause a SEV). This means that the first call will to this method will always return, but when used in a loop, the second will do the right thing.

This is OK as the prescribed behavior is to use the method in a polling loop


src/common/pico_time/time.c
```diff
@@ -439,26 +444,35 @@ bool best_effort_wfe_or_timeout(absolute_time_t timeout_timestamp) {
...
+        // note that as of SDK2.0.0 calling add_alarm_at always causes a SEV. Whwat we really
+        // want to do is cause an IRQ at the specified time in the future if there is not
+        // an IRQ already happening before then. The problem is that the IRQ may be happening on the
+        // other core, so taking an IRQ is the only way to get the state protection.
+        //
+        // Therefore, we make a compromise; we will set the alarm, if we won't wake up before the right time
+        // already. This means that repeated calls to this function with the same timeout will work correctly
+        // after the first one! This is fine, because we ask callers to use a polling loop on another
+        // event variable when using this function
+        if (ta_wakes_up_on_or_before(alarm_pool_get_default()->timer, alarm_pool_get_default()->timer_alarm_num,
```
dpgeorge, line 456: What happens if the other core (or a preemptive IRQ on the current core) modifies this default timer alarm in the middle of this function call?  This call here doesn't look atomic, and the timer may be updated (cancelled, moved further into the future) before the `__wfe()` just below, which would make the `__wfe()` wait longer than the requested timeout?

kilograham: ugh true; that's what i get for trying to write code with COVID!

kilograham: although actually the circumstances where it could not get woken up are really just the other core canceling an alarm (or adding one further out in the future)... i think adding a SEV at the end of the alarm (IRQ) handler should solve that.

dpgeorge: > that's what i get for trying to write code with COVID!

Oh no!  You should just rest.

> although actually the circumstances where it could not get woken up are really just the other core canceling an alarm (or adding the first one further out in the future)... i think adding a SEV at the end of the alarm (IRQ) handler should solve that.

I think the alarm needs to guarantee that, if its IRQ gets enabled then:
1. when the IRQ fires it will do a `__sev()`
2. the alarm time will never be delayed or cancelled, it can only ever be brought forward (earlier) in time

lurch:
This is targeting `master` - should it be targeting `develop`?

src/common/pico_time/time.c
kilograham: Yeah actually I think I misunderstood what you meant when I read it yesterday… when the irq handler sets a new time, it should never make it later than the current time - it should always slow the previous target to happen first

dpgeorge: > when the irq handler sets a new time, it should never make it later than the current time

Yes.

Also, it should never cancel/disable an IRQ if it's already set up.

kilograham: Yeah, sorry, this is in the context of the alarm for the alarm pool which is never disabled any more - it is doomed to fire every 2^32 us

peterharperuk:
I still seem to have sleep issues in Micropython, even with this change

peterharperuk (dismissed):
I wrote some test code to reproduce the scenario I think this is trying to fix. Without this change I can see the problem happening, a sev gets swallowed by best_effort_wfe_or_timeout.
After these changes I see that best_effort_wfe_or_timeout will return early due to the event.

kilograham:
I should probably merge this into develop, and add another issue for additional tests

peterharperuk:
I can add a test

peterharperuk (approved):
I'm very happy with this :)

PR #1824, opened 2024-08-14
.gitignore: Add .cache

### LINKED ISSUE #1823: When using clangd with pico-sdk, .cache folder appears as diffs
Clangd generates a bunch of files in a local directory called .cache. This issue is just documenting that it's a problem for a pull request.

jaguilar:
This is where clangd puts a bunch of junk,
so it's nice to have it ignored for folks who
use clangd.

Fixes #1823

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .vscode
 cmake-*
 .DS_Store
+.cache
 build
 build-*
 


jaguilar:
Argh. I intended to have develop as my baseline. Will submit again in a minute.

PR #1833, opened 2024-08-16
Delete tools/copro_dis.py

lurch:
This functionality has been moved into `picotool`, so remove this Python script to avoid confusion


will-v-pi (approved):
Fine by me - it'll still be in the history, if it ever needs replicating

PR #1835, opened 2024-08-16
Disable interrupts while doing flash_op, explicit_buy, and chain_image

will-v-pi:
For `flash_op` and `explicit_buy`, disable interrupts if running a flash binary. This prevents calling interrupt routines in flash while the flash is being accessed directly (eg when `explicit_buy` does a flash erase), which would cause a bus fault. I haven't added the same to the other flash functions, because those are generally lower-level, whereas `flash_op` is a higher-level function so it's nice to be able to call it without additional configuration.

For `chain_image`, always disable interrupts before chaining, to prevent interrupt routines interrupting the chaining.

### DIFF
--- a/src/rp2_common/pico_bootrom/include/pico/bootrom.h
+++ b/src/rp2_common/pico_bootrom/include/pico/bootrom.h
@@ -549,10 +549,10 @@ typedef struct {
     uint32_t size_bytes;
     uint8_t *buf;
     int *res;
-} flash_op_params_t;
+} rom_helper_flash_op_params_t;
 
-static void flash_op_helper(void *param) {
-    const flash_op_params_t *op = (const flash_op_params_t *)param;
+static inline void rom_helper_flash_op(void *param) {
+    const rom_helper_flash_op_params_t *op = (const rom_helper_flash_op_params_t *)param;
     rom_flash_op_fn func = (rom_flash_op_fn) rom_func_lookup_inline(ROM_FUNC_FLASH_OP);
     *(op->res) = func(op->flags, op->addr, op->size_bytes, op->buf);
 }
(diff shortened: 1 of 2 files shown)


kilograham:
You should use this

```
/**
 * \brief Execute a function with IRQs disabled and with the other core also not executing/reading flash
 * \ingroup pico_flash
 *
 * \param func the function to call
 * \param param the parameter to pass to the function
 * \param enter_exit_timeout_ms the timeout for each of the enter/exit phases when coordinating with the other core
 *
 * \return PICO_OK on success (the function will have been called).
 *         PICO_ERROR_TIMEOUT on timeout (the function may have been called).
 *         PICO_ERROR_NOT_PERMITTED if safe execution is not possible (the function will not have been called).
 *         PICO_ERROR_INSUFFICIENT_RESOURCES if the method fails due to dynamic resource exhaustion (the function will not have been called)
 * \note if \ref PICO_FLASH_ASSERT_ON_UNSAFE is 1, this function will assert in debug mode vs returning
 *       PICO_ERROR_NOT_PERMITTED
 */
int flash_safe_execute(void (*func)(void *), void *param, uint32_t enter_exit_timeout_ms);
```

will-v-pi:
Added - but this does now add a dependency on `pico_flash` for all binaries that use `pico_bootrom`, which I think gets included in `pico_stdlib` (ie all binaries now depend on `pico_flash`)

kilograham:
I think that is probably OK (it doesn't necessarily pull in that much by default - you could check) ... or perhaps it makes more sense to move the bootrom wrapper funcs for this into flash.h - perhaps i'm leaning towards the latter

will-v-pi:
I think that the `explicit_buy` wrapper shouldn't be moved into flash.h - it's just a rom function that happens to touch flash, and therefore needs the flash_safe stuff. Probably fine as is, I checked that `pico_flash` doesn't seem to add much extra, so it's fine to be included

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -542,6 +543,20 @@ static inline void rom_flash_select_xip_read_mode(bootrom_xip_mode_t mode, uint8
...
 
+typedef struct {
+    cflash_flags_t flags;
+    uintptr_t addr;
+    uint32_t size_bytes;
+    uint8_t *buf;
+    int *res;
+} flash_op_params_t;
+
+static void flash_op_helper(void *param) {
```
kilograham, line 554: should be inline - likely to cause compile errors depending on compiler

kilograham:
We either need to 

a) make the struct.typedef and helper function names a bit more scoped or outline this stuff into the C file
b) the problem with C file is that we open ourselves up to problems with executing code from flash - so probably leave them as inline but tack a rom_helper_ or similar on the front.

will-v-pi:
I've inlined the helpers, and renamed them (eg `rom_helper_flash_op` with `rom_helper_flash_op_params_t`)

kilograham (approved):

PR #1841, opened 2024-08-17
Ensure makefsdata.py generates valid variable names

### LINKED ISSUE #1793: pico-examples pico_w httpd build fails on Windows -- makefsdata.py is generating invalid variable name
When trying to do a full build of pico-examples on Windows, it fails on [the httpd example](https://github.com/raspberrypi/pico-examples/tree/master/pico_w/wifi/httpd) with:

```
In file included from C:/Program Files/Raspberry Pi/Pico SDK v2.0.0/pico-sdk/lib/lwip/src/apps/http/fs.c:39:
P:/pico-examples/build/pico_w/wifi/httpd/generated/pico_fsdata.inc:199:36: error: stray '\' in program
  199 | static const unsigned char data_img\rpi_png[] = {
      |                                    ^
```

So there's a `\` in the variable name.

The Python script that's generating this file is [assuming UNIX-style path separators](https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/pico_lwip/tools/makefsdata.py#L107).

The fix would be to simply use `os.sep` instead of `"/"`. Let me know if this sounds reasonable and I can open a PR for it.

ndabas:
This PR:
- Updates makefsdata.py to strip out all characters that would be invalid in a C variable name.
- Adds a minor fix when getting the MIME type from the map to throw the exception we intend to, instead of `KeyError`.
- Fixes a minor indentation inconsistency which would bug me forever if not fixed.

There are a couple more issues in the script which can potentially be fixed, please let me know if you think it's worthwhile to do so and I will add these changes too:
- We could end up with duplicate variable names in some cases, if two or more file names differ only in special characters.
- It might be a good idea to use Python's built-in [mimetypes](https://docs.python.org/3/library/mimetypes.html) module to figure out the MIME type rather than our own map.

Fixes #1793.

### DIFF
--- a/src/rp2_common/pico_lwip/tools/makefsdata.py
+++ b/src/rp2_common/pico_lwip/tools/makefsdata.py
@@ -1,31 +1,8 @@
 #!/usr/bin/env python3
 import argparse
+import mimetypes
 from pathlib import Path
 import re
-
-file_types = {
-    "html": "text/html",
-    "htm":  "text/html",
-    "shtml": "text/html",
-    "shtm": "text/html",
-    "ssi":  "text/html",
-    "gif":  "image/gif",
-    "png":  "image/png",
-    "jpg":  "image/jpeg",
-    "bmp":  "image/bmp",
-    "ico":  "image/x-icon",
-    "class": "application/octet-stream",
-    "cls":  "application/octet-stream",
-    "js":   "application/javascript",
-    "ram":  "application/javascript",
-    "css":  "text/css",
-    "swf":  "application/x-shockwave-flash",
-    "xml":  "text/xml",
-    "xsl":  "application/pdf",
-    "pdf":  "text/xml",
-    "json": "application/json",
-    "svg":  "image/svg+xml"
-}
 
 response_types = {
   200: "HTTP/1.0 200 OK",
@@ -42,9 +19,9 @@ def process_file(input_dir, file):
     results = []
 
     # Check content type
-    content_type = file_types.get(file.suffix[1:].lower())
+    content_type, _ = mimetypes.guess_type(file)
     if content_type is None:
-        raise RuntimeError(f"Unsupported file type {file.suffix}")
+        content_type = "application/octet-stream"
 
     # file name
     data = f"/{file.relative_to(input_dir)}\x00"
@@ -165,6 +142,11 @@ def run_tool():
     )
     args = parser.parse_args()
     print(args.input)
+
+    mimetypes.init()
+    for ext in [".shtml", ".shtm", ".ssi"]:
+        mimetypes.add_type("text/html", ext)
+
     with open(args.output, "w") as fd:
         process_file_list(fd, args.input)
 
--- a/src/rp2_common/pico_lwip/tools/makefsdata.py
+++ b/src/rp2_common/pico_lwip/tools/makefsdata.py
@@ -82,6 +82,11 @@ def process_file_list(fd, input):
         # make a variable name
         var_name = str(file.relative_to(input_dir))
         var_name = re.sub(r"\W+", "_", var_name, flags=re.ASCII)
+
+        # Add a suffix if the variable name is used already
+        if any(d["data_var"] == f"data_{var_name}" for d in data):
+            var_name += f"_{len(data)}"
+
         data_var = f"data_{var_name}"
         file_var = f"file_{var_name}"
 
--- a/src/rp2_common/pico_lwip/tools/makefsdata.py
+++ b/src/rp2_common/pico_lwip/tools/makefsdata.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import argparse
 from pathlib import Path
+import re
 
 file_types = {
     "html": "text/html",
@@ -41,7 +42,7 @@ def process_file(input_dir, file):
     results = []
 
     # Check content type
-    content_type = file_types[file.suffix[1:].lower()]
+    content_type = file_types.get(file.suffix[1:].lower())
     if content_type is None:
         raise RuntimeError(f"Unsupported file type {file.suffix}")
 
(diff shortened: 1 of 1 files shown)


lurch (commented):
LGTM

lurch:
> * We could end up with duplicate variable names in some cases, if two or more file names differ only in special characters.
> * It might be a good idea to use Python's built-in [mimetypes](https://docs.python.org/3/library/mimetypes.html) module to figure out the MIME type rather than our own map.

Sounds sensible, if you're happy to put in the additional effort. The documentation for the mimetypes module suggests it relies on an external system-installed `mime.types` file - are there any situations in which that file might not exist?

ndabas:
Sure, would be happy to add in those bits.

The mimetypes module has a [default](https://github.com/python/cpython/blob/e9d1bf353c3ccafc0d9b61b1b3688051bc976604/Lib/mimetypes.py#L402), hard-coded map as well, and the system-installed files are only used to augment the map:
> mimetypes.init(files=None)
Initialize the internal data structures. If given, files must be a sequence of file names which should be used to augment the default type map. 

So we should be fine even if no system-wide `mime.types` files are found.

I wrote a quick little script to check that all existing mappings exist in the defaults:

```py
import mimetypes

# Copied in from makefsdata.py
file_types = {
    "html": "text/html",
    "htm":  "text/html",
    "shtml": "text/html",
    "shtm": "text/html",
    "ssi":  "text/html",
    "gif":  "image/gif",
    "png":  "image/png",
    "jpg":  "image/jpeg",
    "bmp":  "image/bmp",
    "ico":  "image/x-icon",
    "class": "application/octet-stream",
    "cls":  "application/octet-stream",
    "js":   "application/javascript",
    "ram":  "application/javascript",
    "css":  "text/css",
    "swf":  "application/x-shockwave-flash",
    "xml":  "text/xml",
    "xsl":  "application/pdf",
    "pdf":  "text/xml",
    "json": "application/json",
    "svg":  "image/svg+xml"
}

mimetypes.init(files=None)

for ext, mime in file_types.items():
    guessed, _ = mimetypes.guess_type("abc." + ext)
    if  guessed != mime:
        print(f"Diff: {ext}, expected: {mime}, got: {guessed}")

```

I got this:

```
Diff: shtml, expected: text/html, got: None
Diff: shtm, expected: text/html, got: None
Diff: ssi, expected: text/html, got: None
Diff: class, expected: application/octet-stream, got: None
Diff: cls, expected: application/octet-stream, got: None
Diff: js, expected: application/javascript, got: text/javascript
Diff: ram, expected: application/javascript, got: application/x-pn-realaudio
Diff: xsl, expected: application/pdf, got: text/xml
Diff: pdf, expected: text/xml, got: application/pdf
```

We can add in just these missing ones, but what's up with the current mappings for `.ram`, `.xsl`, and `.pdf`? I guess the `xsl` and `pdf` ones got swapped but what was `ram` supposed to be?

ndabas:
I have modified the code to use the mimetypes library and fixed another bug in the course of things. I don't know how important backwards compatibility is going to be for this, but here's how I've reconciled the differences between the hard-coded mimetypes and the library call:
- `shtml`, `shtm`, `ssi`: added a static mapping to support these extensions.
- `class`, `cls`: added a default of `application/octet-stream` when the mapping isn't found, which is how most web servers behave.
- `js`: was `application/javascript`, but will now return `text/javascript`, which I think is fine.
- `ram` was `application/javascript`, but will now be `application/x-pn-realaudio` -- hopefully this doesn't break anybody's workflow!
- `xsl` and `pdf` were swapped, and will now be fixed.

peterharperuk (approved):
Looks good. Thanks!

ndabas:
Thanks, I've [verified](https://github.com/ndabas/pico-setup-windows/actions/runs/10454227003) that this fixes the issue on Windows too.

peterharperuk:
@kilograham Ok to merge this?

lurch:
@ndabas Were you still planning to add the "We could end up with duplicate variable names in some cases, if two or more file names differ only in special characters." feature?

ndabas:
@lurch I wasn't planning to but I can if you think it's significant. Probably just a couple of extra lines of code?

ndabas:
@lurch I gave it a shot and it was easier than expected. So that takes care of all of the issues I mentioned.

peterharperuk (approved):

src/rp2_common/pico_lwip/tools/makefsdata.py
```diff
@@ -103,8 +81,12 @@ def process_file_list(fd, input):
 
         # make a variable name
         var_name = str(file.relative_to(input_dir))
-        var_name = var_name.replace(".", "_")
-        var_name = var_name.replace("/", "_")
+        var_name = re.sub(r"\W+", "_", var_name, flags=re.ASCII)
+
+        # Add a suffix if the variable name is used already
+        if any(d["data_var"] == f"data_{var_name}" for d in data):
+            var_name += f"_{len(data)}"
```
lurch, line 88: I guess using `len(data)` as the suffix means that if the first and 6th items are named identically, then the "first duplicate name" would get a suffix of `_5`, which may be unexpected?

An alternative approach might be to add a function like:
```python3
unique_names = set()
def get_unique_name(name):
    if name in unique_names:
        counter = 1
        new_name = f"{name}_{counter}"
        while new_name in unique_names:
            counter += 1
            new_name = f"{name}_{counter}"
        name = new_name
    unique_names.add(name)
    return name
```
and then you could just do `var_name = get_unique_name(var_name)` and then any duplicates would just be numbered sequentially (starting at 1).

But of course this is your PR, so it's up to you / @peterharperuk if you decide to incorporate this suggestion :slightly_smiling_face:

ndabas: I expected this comment 😄 

It just seemed simpler to add those two lines instead of a loop like this. Yes, the numbers will not be sequential, but there _is_ a method to the madness; and the generated code will likely not be scrutinized by users. I couldn't find a technical reason to not do it this way personally; @peterharperuk let me know your thoughts too.

lurch: > and the generated code will likely not be scrutinized by users

Yeah, I'm possibly getting carried away and losing sight of where the output of all of this **actually** ends up getting used :laughing:

peterharperuk: Let's keep it as it is.

Issue #1842, opened 2024-08-18
boot2_name missing in UF2 when compiled with sdk 2.0


codetiger:
I've a custom PCB based on RP2040 and tried upgrading the SDK to the latest version. What is observed is the device won't start when reset, but works perfectly when dropping UF2. This problem sounds familiar and I went through this when I first tried to run code on the custom PCB. I fixed the issue with the below lines 

```
#define PICO_XOSC_STARTUP_DELAY_MULTIPLIER 64
#define PICO_BOOT_STAGE2_CHOOSE_W25Q080 1
#define PICO_FLASH_SPI_CLKDIV 2

```

After compiling the code using SDK 2.0, the issue started again. Tried setting the same config in the CMakeLists.txt

```
add_compile_definitions(PICO_XOSC_STARTUP_DELAY_MULTIPLIER=64)
add_compile_definitions(PICO_BOOT_STAGE2_CHOOSE_W25Q080=1)
add_compile_definitions(PICO_FLASH_SPI_CLKDIV=2)

```

But does not solve the issue. 

Using picotool to see what's the difference in the uf2 below is what I found

```
codetiger@Mac-mini build % picotool info -a ~/Downloads/GameTiger.uf2
File /Users/codetiger/Downloads/GameTiger.uf2:

Program Information
 name:          GameTiger
 features:      USB stdin / stdout
 binary start:  0x10000000
 binary end:    0x10098d3c

Fixed Pin Information
 none

Build Information
 sdk version:       1.5.1
 pico_board:        gametiger_rp2040
 boot2_name:        boot2_w25q080
 build date:        Apr  3 2024
 build attributes:  Release
```

```
codetiger@Mac-mini build % picotool info -a GameTiger.uf2            
File GameTiger.uf2:

Program Information
 name:          GameTiger
 features:      USB stdin / stdout
 binary start:  0x10000000
 binary end:    0x10098d38

Fixed Pin Information
 none

Build Information
 sdk version:       2.0.0
 pico_board:        gametiger_rp2040
 build date:        Aug 18 2024
 build attributes:  Release

```
 
If you see closely the parameter `boot2_name ` is missing in the UF2 file.

codetiger:
Not sure why, but with SDK 2.0, it works only with 
`#define PICO_FLASH_SPI_CLKDIV 4`

kilograham:
> Not sure why, but with SDK 2.0, it works only with
[...]

PR #1844, opened 2024-08-18
Fix missing Binary Info for boot_stage2

kilograham:
Regression in SDK2.0.0

### DIFF
--- a/src/rp2350/boot_stage2/CMakeLists.txt
+++ b/src/rp2350/boot_stage2/CMakeLists.txt
@@ -68,7 +68,7 @@ function(pico_define_boot_stage2 NAME SOURCES)
 
     add_custom_target(${NAME}_padded_checksummed_asm DEPENDS ${PADDED_CHECKSUMMED_ASM})
     add_custom_command(OUTPUT ${PADDED_CHECKSUMMED_ASM} DEPENDS ${ORIGINAL_BIN}
-            COMMAND ${Python3_EXECUTABLE} ${PICO_BOOT_STAGE2_DIR}/pad_checksum -s 0xffffffff ${ORIGINAL_BIN} ${PADDED_CHECKSUMMED_ASM}
+            COMMAND ${Python3_EXECUTABLE} ${PICO_BOOT_STAGE2_DIR}/pad_checksum -s 0xffffffff -a $<IF:${PICO_RISCV},riscv,arm> ${ORIGINAL_BIN} ${PADDED_CHECKSUMMED_ASM}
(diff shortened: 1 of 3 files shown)


kilograham:
Fixes #1842

src/rp2350/boot_stage2/CMakeLists.txt
```diff
@@ -63,12 +63,13 @@ function(pico_define_boot_stage2 NAME SOURCES)
     find_package (Python3 REQUIRED COMPONENTS Interpreter)
 
     add_custom_target(${NAME}_bin DEPENDS ${ORIGINAL_BIN})
-    add_custom_command(OUTPUT ${ORIGINAL_BIN} DEPENDS ${NAME} COMMAND ${CMAKE_OBJCOPY} -Obinary $<TARGET_FILE:${NAME}> ${ORIGINAL_BIN})
+    add_custom_command(OUTPUT ${ORIGINAL_BIN} DEPENDS ${NAME} COMMAND ${CMAKE_OBJCOPY} -Obinary $<TARGET_FILE:${NAME}> ${ORIGINAL_BIN}
+        VERBATIM)
 
     add_custom_target(${NAME}_padded_checksummed_asm DEPENDS ${PADDED_CHECKSUMMED_ASM})
     add_custom_command(OUTPUT ${PADDED_CHECKSUMMED_ASM} DEPENDS ${ORIGINAL_BIN}
-            COMMAND ${Python3_EXECUTABLE} ${PICO_BOOT_STAGE2_DIR}/pad_checksum -s 0xffffffff -a $<IF:${PICO_RISCV},riscv,arm> ${ORIGINAL_BIN} ${PADDED_CHECKSUMMED_ASM}
```
lurch, line 70: Is the removal of the `-a` flag here deliberate? It looks like https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/boot_stage2/pad_checksum _does_ use that flag? (but I dunno enough of the the low-level details to know whether that's significant or not!)

will-v-pi: Yes, it shouldn't be removed here, as it is necessary on RP2350

will-v-pi (commented):
Other than @lurch's comment about the `-a` flag, this all looks and works fine

src/rp2350/boot_stage2/CMakeLists.txt
kilograham: good catch - COVID development again!

PR #1850, opened 2024-08-20
Add TBYB safe rom_pick_ab_partition function

will-v-pi:
This PR adds an extra wrapper around `rom_pick_ab_partition`, which makes it safer when called during a TBYB boot, and includes a call to get the flash_update_base

Currently, when doing a flash update boot which requires a flash erase, if you call `pick_ab_partition` before calling `explicit_buy` (eg to pick a firmware partition or similar) then it will clear `always->zero_init.version_downgrade_erase_flash_addr`. This means the subsequent `explicit_buy` call will not perform the required flash erase, which means the flash update boot will not succeed.

This adds saving and restoring the `erase_flash_addr` to prevent this happening, while still allowing the `erase_flash_addr` to be set if the firmware partition is being updated and requires a flash erase

This also adds setting up `always->zero_init.tbyb_flag_flash_addr` to allow for TBYB executables in the firmware partition to have the address set up correctly

The function returns an error if too many things are updating at once - either if both the boot and firmware images are TBYB, or if the boot and firmware images both require a flash erase

It also returns an error if the chosen image was not verified, which acts as a signature check on the firmware image

### DIFF
--- a/src/rp2_common/pico_bootrom/bootrom.c
+++ b/src/rp2_common/pico_bootrom/bootrom.c
@@ -75,7 +75,7 @@ void __attribute__((noreturn)) rom_reset_usb_boot(uint32_t usb_activity_gpio_pin
 #define VERSION_DOWNGRADE_ERASE_ADDR *(uint32_t*)0x400e0338
 #define TBYB_FLAG_ADDR *(uint32_t*)0x400e0348
 #define IMAGE_DEF_VERIFIED(scan_workarea) *(uint32_t*)(0x64 + (uint32_t)scan_workarea)
-#define IMAGE_DEF_TBYB_FLAGGED(scan_workarea) *(uint32_t*)(0x4c + (uint32_t)scan_workarea)
+#define IMAGE_DEF_TBYB_FLAGGED(scan_workarea) *(bool*)(0x4c + (uint32_t)scan_workarea)
 #define IMAGE_DEF_BASE(scan_workarea) *(uint32_t*)(0x54 + (uint32_t)scan_workarea)
 #define IMAGE_DEF_REL_BLOCK_OFFSET(scan_workarea) *(uint32_t*)(0x5c + (uint32_t)scan_workarea)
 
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_bootrom/bootrom.c
```diff
@@ -7,6 +7,9 @@
 #include "pico/bootrom.h"
 #include "boot/picoboot.h"
 #include "boot/picobin.h"
+#if !PICO_RP2040
```
kilograham, line 10: strictly more idiomatic to use `#if HAS_RP2040_RTC`

src/rp2_common/pico_bootrom/include/pico/bootrom_structs.h
```diff
@@ -0,0 +1,211 @@
+#include "boot/picobin.h"
```
kilograham, line 1: I'm really not a fan of exposing this header of structs. It implies that this is a public API, which it isn't

Can we try with just using memory offsets, and then we need to guard with the rom_version also.

src/rp2_common/pico_bootrom/bootrom.c
will-v-pi: This is the RCP not the RTC - given the functions in this file are also guarded by `!PICO_RP2040` I think this if fine to use here too?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1021,6 +1021,22 @@ static inline int rom_get_last_boot_type(void) {
  */
 int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32_t permissions);
 
+/*! \brief  Pick A/B partition for a separate partition
+ * \ingroup pico_bootrom
+ *
+ * This will perform extra checks to prevent disrupting a main image TBYB, and return errors
```
lurch, line 1027: Might want to reword this slightly? "This will ... return errors" :wink:

will-v-pi:
Superseded by #1969 - that adds this function plus a use of it

lurch:
Just to keep the dots joined up (for the benefit of any future readers) : this function actually got added in #2182

PR #1854, opened 2024-08-20
Bazel: disable link_extra_lib when building bootloader

### LINKED ISSUE #1853: bazel: including bootloader targets in `link_extra_lib` results in a dependency cycle
The bazel bootloader targets (e.g. @@pico-sdk~//src/rp2040/boot_stage2:boot_stage2_elf_actual) get built with `link_extra_lib` in their deps, so including them in the `link_extra_lib` list for other binaries causes a dependency cycle. This cycle should be broken by removing `link_extra_lib` dependencies when building bootloader targets.

PR incoming.

cramertj:
Fixes #1853

The bazel bootloader targets (e.g. @@pico-sdk~//src/rp2040/boot_stage2:boot_stage2_elf_actual) get built with link_extra_lib in their deps, so including them in the link_extra_lib list for other binaries causes a dependency cycle. This cycle should be broken by removing link_extra_lib dependencies when building bootloader targets.

### DIFF
--- a/bazel/util/transition.bzl
+++ b/bazel/util/transition.bzl
@@ -87,11 +87,18 @@ rp2040_bootloader_binary = declare_transtion(
         "_allowlist_function_transition": attr.label(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
+	"_link_extra_libs": attr.label(default = "//bazel:empty_cc_lib"),
     },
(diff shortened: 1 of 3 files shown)


armandomontanez (approved):
LGTM, @kilograham this is ready to go.

kilograham (approved):

PR #1857, opened 2024-08-21
Fix stdio_usb with reset interface disabled on Windows

### LINKED ISSUE #1802: sdk2.0.0: PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE=0 incompatible with CDC
As per the title, My application was using this parameter in previous releases (1.5, 1.6) with no issues. On rebuilding with sdk2.0.0 the application didn't work (see https://forums.raspberrypi.com/viewtopic.php?p=2243782#p2243782) and this was tracked down to this setting. Adding the parameter to the build of hello-usb proved the problem

will-v-pi:
Only use bcdUSB = 0x0210 when reset interface and MS_OS_20 descriptor are enabled, otherwise Windows doesn't recognise the USB device

I can confirm this fixes #1802 on Windows


kilograham:
can you please add a comment on why the difference

will-v-pi:
I've added an explanatory comment

kilograham (approved):

PR #1858, opened 2024-08-21
Reorder pins in bi_pins_with_names to be correct

will-v-pi:
Alternative to #1791 - instead of asserting when the order is incorrect, reorder the names so they are in the correct order for the pin mask

Requires more pre-processor stuff, but doesn't break existing code with the pins in the wrong order, so [pico-examples/524](https://github.com/raspberrypi/pico-examples/pull/524) would no longer be required

Has been tested using the following python code to generate all permutations of `bi_decls` which are then copied and compiled into a binary, and then checked with `picotool info` to ensure pins are all named correctly

```python
import itertools

pins = [
    'PICO_DEFAULT_SPI_RX_PIN, "SPI RX"',
    'PICO_DEFAULT_SPI_TX_PIN, "SPI TX"',
    'PICO_DEFAULT_SPI_SCK_PIN, "SPI SCK"',
    'PICO_DEFAULT_SPI_CSN_PIN, "SPI CS"'
]

pinperms = itertools.permutations(pins, 2)
for p in pinperms:
    print(f"bi_decl(bi_2pins_with_names({p[0]}, {p[1]}));")

pinperms = itertools.permutations(pins, 3)
for p in pinperms:
    print(f"bi_decl(bi_3pins_with_names({p[0]}, {p[1]}, {p[2]}));")

pinperms = itertools.permutations(pins, 4)
for p in pinperms:
    print(f"bi_decl(bi_4pins_with_names({p[0]}, {p[1]}, {p[2]}, {p[3]}));")
```

### DIFF
--- a/src/common/pico_binary_info/include/pico/binary_info/code.h
+++ b/src/common/pico_binary_info/include/pico/binary_info/code.h
@@ -208,8 +208,62 @@ static const struct _binary_info_named_group __bi_lineno_var_name = { \
 #define bi_7pins_with_func(p0, p1, p2, p3, p4, p5, p6,func) __bi_encoded_pins_64_with_func(BI_PINS_ENCODING_MULTI | ((func << 3)) | ((p0) << 8) | ((p1) << 16) | ((p2) << 24) | ((uint64_t)(p3) << 32) | ((uint64_t)(p4) << 40) | ((uint64_t)(p5) << 48) | ((uint64_t)(p6) << 56))
 
 #define bi_1pin_with_name(p0, name)                   bi_pin_mask_with_name(1ull << (p0), name)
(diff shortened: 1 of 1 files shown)


lurch:
ZOMG! :exploding_head:  As long as you've tested this, I'll assume it's okay :wink: 

Although I guess your tests could be simplified to use e.g. `pinperms = itertools.permutations(pins[:2], 2)` instead?

kilograham:
Yeah, it is a bit scary; i was wondering if we should do a recursive implementation, but

1. that means splitting out a separate macro for the string part
2. I don't know that it is significantly simpler

So unless anyone has any fantastic ideas i say we take this as is ;-)

lurch:
> So unless anyone has any fantastic ideas i say we take this as is ;-)

https://github.com/svpv/qsort :rofl:

kilograham (approved):
I'm gonna believe you!

PR #1870, opened 2024-08-26
Fix issue with conversion to uf2 using picotool

lukstep:
There is a problem converting the binary file to uf2. The problem seems to be the lack of a file extension when generating the converting command in `tools/CMakeLists.txt`

Compilation output:
```
ERROR: filename '/workspaces/pico-2-template-project/build/sample' does not have a recognized file type (extension)
make[2]: *** [CMakeFiles/sample.dir/build.make:1205: sample] Error 157
make[2]: *** Deleting file 'sample'
make[1]: *** [CMakeFiles/Makefile2:1853: CMakeFiles/sample.dir/all] Error 2
```

and the snipet of makefile with wrong picotool command:
```
	/usr/local/bin/picotool uf2 convert --quiet /workspaces/pico-2-template-project/build/sample sample.uf2 --family rp2350-riscv --abs-block
```

The fix creates a correct uf2 file that can be uploaded to the microcontroller and run.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -410,7 +410,7 @@ function(pico_add_uf2_output TARGET)
             ARGS uf2 convert
                 --quiet
                 ${uf2_package_args}
-                $<TARGET_FILE:${TARGET}>
+                $<TARGET_FILE:${TARGET}>.bin
                 ${output_path}$<IF:$<BOOL:$<TARGET_PROPERTY:${TARGET},OUTPUT_NAME>>,$<TARGET_PROPERTY:${TARGET},OUTPUT_NAME>,$<TARGET_PROPERTY:${TARGET},NAME>>.uf2
                 --family ${picotool_family}
                 ${extra_uf2_args}


lurch:
@will-v-pi Is it also worth ensuring that this `pico_add_uf2_output` function is actually exercised somewhere, either in `pico-sdk` or `pico-examples` ? (as I didn't come across this error in all the recent testing I've been doing of pico-sdk, pico-examples, pico-extras and pico-playground)

will-v-pi:
The target file should already contain a .elf extension - I think you must have a custom build setup if your file doesn’t have an extension?

The .elf extension is set in the PICO platform file (cmake/Platform/PICO.cmake). If you want to set your own file extension on a custom target, you’ll need to set the `SUFFIX` target property to something else

@lurch it is already exercised in all the tests that generate a UF2 output - all passing fine

lukstep:
Here is the project I used, https://github.com/lukstep/pico-2-template-project.
I building it in docker container with Ubuntu 22.04

kilograham:
the import of `pico_sdk_init.cmake` should precede the project - i.e. move your project statement to line 19

This will probably also allow you not to have to hack up CMAKE_ compiler paths.

lukstep:
It fix a lot of my problems, thanks!

PR #1879, opened 2024-08-28
Specify minimum cmake version for generic_board.cmake

lurch:
`cmake/generic_board.cmake` uses `POP_FRONT`, which https://cmake.org/cmake/help/latest/command/list.html#pop-front says was added in CMake version 3.15.
(this change means that people still using an ancient version of cmake will get a nicer error-message)

### DIFF
--- a/cmake/generic_board.cmake
+++ b/cmake/generic_board.cmake
@@ -1,4 +1,5 @@
 # For boards without their own cmake file, we look for a header file
+cmake_minimum_required(VERSION 3.15)
 
 # PICO_CMAKE_CONFIG: PICO_BOARD_HEADER_DIRS, List of directories to look for <PICO_BOARD>.h in. This may be specified the user environment, type=list, group=build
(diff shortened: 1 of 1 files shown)


kilograham:
This should probably go in the top-level `CMakeLists.txt` - i am loathe to change minimum versions, but then again, this code is _always_ used an no one has complained, so most people must be >= 3.15

kilograham:
@will-v-pi thoughts?

will-v-pi:
`FindPicotool.cmake` is only used when the SDK needs to compile picotool (contrary to its name), so if you have installed picotool then that file won’t be used and a lower CMake version should work. I think it makes sense to do the same for `generic_board.cmake` as you could set your own simplified board cmake file if you need to use an older version of CMake.

I’d agree that for any CMakeLists.txt files we should update it in the top level project, but for these script-esque files which are not always included it’s probably better to put the minimum version in the file?

kilograham (approved):

PR #1880, opened 2024-08-28
fix bug in sm_config_set_in_pin_count

### LINKED ISSUE #1878: sm_config_set_in_pin_count() appears incorrect
This function is newly added for RP2350, and the implementation appears inconsistent with the datasheet.

Datasheet says:

> Set the number of pins which are not masked to 0 when read by an
> IN PINS, WAIT PIN or MOV x, PINS instruction.
> For example, an IN_COUNT of 5 means that the 5 LSBs of the IN pin group are
> visible (bits 4:0), but the remaining 27 MSBs are masked to 0. A count of 32 is
> encoded with a field value of 0, so the default behaviour is to not perform any
> masking.

ie. the number in the register is the number of pins you are actually using, except that zero represents  32.

However, the code puts `(in_count - 1)`.

https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/hardware_pio/include/hardware/pio.h#L393

Looks like it should be `(in_count & 0x1f)`

kilograham:
fixes #1878

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -382,10 +382,10 @@ static inline void sm_config_set_in_pins(pio_sm_config *c, uint in_base) {
     sm_config_set_in_pin_base(c, in_base);
 }
 
-/*! \brief Set the number of significant consecutive 'in' pins in a state machine configuration
+/*! \brief Set the count of 'in' pins in a state machine configuration
  *  \ingroup sm_config
  *
- * When reading pins using the IN pin mapping, this many (low) bits will be read, with the reset taking
+ * When reading pins using the IN pin mapping, this many (low) bits will be read, with the rest taking
  * the value zero.
  *
  * \if rp2040_specific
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -382,6 +382,19 @@ static inline void sm_config_set_in_pins(pio_sm_config *c, uint in_base) {
     sm_config_set_in_pin_base(c, in_base);
 }
 
+/*! \brief Set the number of significant consecutive 'in' pins in a state machine configuration
```
lurch, line 385: Should "significant" here be "least-significant" ?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -382,6 +382,19 @@ static inline void sm_config_set_in_pins(pio_sm_config *c, uint in_base) {
...
+ *  \ingroup sm_config
+ *
+ * When reading pins using the IN pin mapping, this many (low) bits will be read, with the reset taking
```
lurch, line 388: "with the reset" -> "with the rest" ?

peterharperuk:
Looks good. I think Andrews comments are correct

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: well i meant which are significant. perhaps better wording required

peterharperuk (approved):

PR #1881, opened 2024-08-29
separate actual shared bootrom header into boot_bootrom_headers

kilograham:
... as pico_bootrom which is focused on calling the bootrom now has non trivial dependencies like pico_flash


src/rp2_common/boot_bootrom_headers/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "boot_bootrom_headers",
```
armandomontanez, line 4: I suspect there may be a circular dependency if you try to add this to the deps of [//src/rp2_common/pico_bootrom:pico_bootrom_headers](https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/pico_bootrom/BUILD.bazel#L17), so I'm game to sort out the complexity here in a follow-up.

lurch:
Well the CI isn't happy :confused: 
```
Scanning dependencies of target picotool
In file included from /opt/actions-runner/_work/pico-sdk/pico-sdk/build/_deps/picotool-src/main.cpp:45:
/opt/actions-runner/_work/pico-sdk/pico-sdk/src/rp2_common/pico_bootrom/include/pico/bootrom_constants.h:8:10: fatal error: boot/bootrom_constants.h: No such file or directory
    8 | #include "boot/bootrom_constants.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

src/rp2_common/boot_bootrom_headers/BUILD.bazel
kilograham: thanks - yes figured this would break u guys (though I didn't know about the circular dependency - i just copied the dependency on pico_platform_internal - not sure if it is needed

kilograham:
merging this due to circular dependency with picotool

PR #1885, opened 2024-08-29
[Bazel] Update version string to 2.0.1-develop

armandomontanez:
Updates the Bazel module version string to match the current CMake version string for the develop branch.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "pico-sdk",
-    version = "2.0.0",
+    version = "2.0.1-develop",
 )
 
 bazel_dep(name = "platforms", version = "0.0.9")


kilograham (approved):

lurch:
This may be a stupid question, but could it be worth adding a check for this to your `tools/compare_build_systems.py` (or similar) ?

armandomontanez:
> This may be a stupid question, but could it be worth adding a check for this to your `tools/compare_build_systems.py` (or similar) ?

Yeah, that's a good idea. I had [a script that could extract the version from CMake](https://github.com/armandomontanez/pico-sdk/blob/7ef45fad2fee9df44c71e387ccb2972b557824c3/bazel/parse_version.py) a while back. Could re-introduce that to check for a discrepancies.

PR #1890, opened 2024-08-31
Fix a compile error by using clang-x86_64-pc-windows-msvc

kkitayam:
Fix compile errors when using clang-x86_64-pc-windows-msvc

Please see also https://github.com/raspberrypi/picotool/pull/129

- Environment

```console
$ clang --version
clang version 18.1.8
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin
Error messages
```

- An error message

```console
C:/hoge/pico-sdk/src/host/pico_platform/include\pico/platform.h:99:24: error: definition of builtin function '__builtin_unreachable'
   99 | static __noreturn void __builtin_unreachable() {
      |                        ^
1 error generated.
```

### DIFF
--- a/src/host/pico_platform/include/pico/platform.h
+++ b/src/host/pico_platform/include/pico/platform.h
@@ -47,7 +47,7 @@ extern void tight_loop_contents();
 #define __STRING(x) #x
 #endif
 
-#ifndef _MSC_VER
+#if !defined(_MSC_VER) || defined(__clang__)
 #ifndef __noreturn
 #define __noreturn __attribute((noreturn))
 #endif


src/host/pico_platform/include/pico/platform.h
```diff
@@ -47,7 +47,7 @@ extern void tight_loop_contents();
 #define __STRING(x) #x
 #endif
 
-#ifndef _MSC_VER
+#if !defined(_MSC_VER) || defined(__clang__)
```
kilograham, line 50: did you mean `if !defined(_MSC_VER) && !defined(__clang__)`?

kkitayam: No, it's not. The changes were intended to disable LL.74-103 when using clang-x86_64-pc-windows-msvc.

clang-x86_64-pc-windows-msvc has built-in functions like gcc syntax. Specifically there are `__builtin_unreachable` and `__builtin_clz`. Also, clang-x86_64-pc-windows-msvc has a macro _MSC_VER already defined. So, LL.74-103 were enabled. As a result, the compilation fails.

kilograham (approved):

PR #1913, opened 2024-09-07
Fixed string trimming issue at line 48

### LINKED ISSUE #1912: pico_get_unique_board_id_string prints an extra character at the start
Hello,
I noticed in my code that for some reason `pico_get_unique_board_id_string()` is trimming the cstring incorrectly:

If I do this:
``` c++
#define BOARD_ID_LEN 17
char board_id[BOARD_ID_LEN];
pico_get_unique_board_id_string(board_id, BOARD_ID_LEN);
```

I get the following string:
E6625CA563857138

but if I do this: (if I change `BOARD_ID_LEN` to 16)
``` c++
#define BOARD_ID_LEN 16
char board_id[BOARD_ID_LEN];
pico_get_unique_board_id_string(board_id, BOARD_ID_LEN);
```

I get this string:
E6625CA56385713

weirdly enough, this fixes the problem:

``` c++
#define BOARD_ID_LEN 16
char board_id[BOARD_ID_LEN + 1];
pico_get_unique_board_id_string(board_id, BOARD_ID_LEN);
```

I looked in [https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_unique_id/unique_id.c](the source code) for this and found that there's an issue here:

```c++
    for (i = 0; (i < len - 1) && (i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES * 2); i++) {
```

I think it should be `len` instead of  `len - 1` because we are using the less than operator, and not the LE operator. 

I can make a PR later today if needed!

Corn

LordOfDeadbush:
Fixed an issue where in pico_get_unique_board_id_string, the limit for the for loop was set too low, so I incremented it and now it should work for strings with length 16.

Fixes #1912

### DIFF
--- a/src/rp2_common/pico_unique_id/unique_id.c
+++ b/src/rp2_common/pico_unique_id/unique_id.c
@@ -45,7 +45,7 @@ void pico_get_unique_board_id_string(char *id_out, uint len) {
     assert(len > 0);
     size_t i;
     // Generate hex one nibble at a time
-    for (i = 0; (i < len - 1) && (i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES * 2); i++) {
+    for (i = 0; (i < len) && (i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES * 2); i++) {
         int nibble = (retrieved_id.id[i/2] >> (4 - 4 * (i&1))) & 0xf;
         id_out[i] = (char)(nibble < 10 ? nibble + '0' : nibble + 'A' - 10);
     }


peterharperuk:
I think the function is currently written to assume that you are passing a buffer with space for a zero terminator. Which is non-obvious. So I would say this change would improve things.

```
 * \param id_out a pointer to a char buffer of size len, to which the identifier will be written
 * \param len the size of id_out. For full serial, len >= 2 * PICO_UNIQUE_BOARD_ID_SIZE_BYTES + 1
```

peterharperuk (approved):

peterharperuk:
Actually, ignore that. I missed the zero terminator at the end.

peterharperuk (changes requested):
Missing the zero termiantor at the end

peterharperuk:
Closing as the code is correct.

PR #1921, opened 2024-09-09
Enable building with Bazel on Raspberry PI OS machines

### LINKED ISSUE #1920: Feature Request: Allow bazel building on Raspberry PI OS
The bazel toolchain is missing entries for building on Raspberry PI OS machines like the Raspberry PI 5. It would be nice to enable bazel builds also on the Raspberry PI itself.

seesturm:
Fixes #1920


kilograham:
cc @armandomontanez

armandomontanez (approved):
LGTM. I think clang has aarch64 binaries too, I can dig those up and add those in a follow-up.

PR #1942, opened 2024-09-23
Clarify that RP2350 gets its unique id from OTP

### LINKED ISSUE #1915: pico_unique_id header needs some updates
From reading the comments in [unique_id.h](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_unique_id/include/pico/unique_id.h) (which also appear [here](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_unique_id) ) you might think that `pico_get_unique_board_id` **always** returns the ID of the flash chip. However looking at the [implementation](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_unique_id/unique_id.c#L15) it looks like that on RP2350 it actually gives you the `SYS_INFO_CHIP_INFO` (which I think is described in section 5.4.8.17. of https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf ?).
And I guess this also means that you _won't_ get a unique_id of all-0xEE on PICO_NO_FLASH builds for RP2350? :wink:

peterharperuk:
Fixes #1915

### DIFF
--- a/src/rp2_common/pico_unique_id/include/pico/unique_id.h
+++ b/src/rp2_common/pico_unique_id/include/pico/unique_id.h
@@ -31,6 +31,8 @@ extern "C" {
  *
  * This avoids some pitfalls of the hardware_flash API, which requires any
  * flash-resident interrupt routines to be disabled when called into.
+ *
+ * On boards using RP2350, the unique identifier is read from OTP memory.
  */
 
 #define PICO_UNIQUE_BOARD_ID_SIZE_BYTES 8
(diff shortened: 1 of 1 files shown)


lurch:
Looks good, but I wonder if this ought to use the `\if` mechanism that can be seen in e.g. https://github.com/raspberrypi/pico-sdk/pull/1945/files ?

peterharperuk:
yes, looks like that might be a good idea.

src/rp2_common/pico_unique_id/include/pico/unique_id.h
```diff
@@ -64,11 +83,16 @@ void pico_get_unique_board_id(pico_unique_board_id_t *id_out);
 /*! \brief Get unique ID in string format
  *  \ingroup pico_unique_id
  *
- * Get the unique 64-bit device identifier which was retrieved from the
- * external NOR flash device or OTP memory at boot, formatted as an ASCII hex string.
- * Will always 0-terminate.
+ * Get the unique 64-bit device identifier formatted as a 0-terminated ASCII hex string.
```
lurch, line 86: Would "null-terminated" be better than "0-terminated" ?

lurch:
One small quibble but otherwise looks great! :+1:

kilograham (approved):

PR #1954, opened 2024-09-25
Avoid setting an alarm in the past

### LINKED ISSUE #1953: alarm pool can appear to stop working
We have an issue with a program on RP2040 that we communicate with via a USB, it was occasionally unresponsive. After some investigation a repeated alarm callback that is used to call tud_task_ext was not happening. This only seems to be a problem on sdk 2.0. Further investigation suggests it's possible for the alarm pool to register an alarm in the past, so it'll only go off in ~71 minutes. The attached program seems to reproduce the problem.

The suggested fix to alarm_pool_irq_handler is to update "now" after the call to "ta_time_us_64"

```
diff --git a/src/common/pico_time/time.c b/src/common/pico_time/time.c
index 118547d8..2f5409a4 100644
--- a/src/common/pico_time/time.c
+++ b/src/common/pico_time/time.c
@@ -259,13 +259,13 @@ static void alarm_pool_irq_handler(void) {
                 index = next;
             }
         }
-        now = (int64_t) ta_time_us_64(timer);
         earliest_index = pool->ordered_head;
         if (earliest_index < 0) break;
         // need to wait
         alarm_pool_entry_t *earliest_entry = &pool->entries[earliest_index];
         earliest_target = earliest_entry->target;
         ta_set_timeout(timer, timer_alarm_num, earliest_target);
+        now = (int64_t) ta_time_us_64(timer);
         // check we haven't now past the target time; if not we don't want to loop again
     } while ((earliest_target - now) <= 0);
 }
```

[alarm_test.zip](https://github.com/user-attachments/files/17134509/alarm_test.zip)

peterharperuk:
Fixes #1953

### DIFF
--- a/src/common/pico_time/time.c
+++ b/src/common/pico_time/time.c
@@ -259,13 +259,13 @@ static void alarm_pool_irq_handler(void) {
                 index = next;
             }
         }
-        now = (int64_t) ta_time_us_64(timer);
         earliest_index = pool->ordered_head;
         if (earliest_index < 0) break;
         // need to wait
         alarm_pool_entry_t *earliest_entry = &pool->entries[earliest_index];
         earliest_target = earliest_entry->target;
         ta_set_timeout(timer, timer_alarm_num, earliest_target);
+        now = (int64_t) ta_time_us_64(timer);
         // check we haven't now past the target time; if not we don't want to loop again
     } while ((earliest_target - now) <= 0);
 }
--- a/src/common/pico_time/time.c
+++ b/src/common/pico_time/time.c
@@ -142,7 +142,6 @@ static void alarm_pool_irq_handler(void) {
     uint timer_num = ta_timer_num(timer);
     alarm_pool_t *pool = pools[timer_num][timer_alarm_num];
     assert(pool->timer_alarm_num == timer_alarm_num);
-    int64_t now = (int64_t) ta_time_us_64(timer);
     int64_t earliest_target;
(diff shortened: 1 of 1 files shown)


kilograham (approved):

src/common/pico_time/time.c
```diff
@@ -259,13 +259,13 @@ static void alarm_pool_irq_handler(void) {
...
             }
         }
-        now = (int64_t) ta_time_us_64(timer);
         earliest_index = pool->ordered_head;
         if (earliest_index < 0) break;
         // need to wait
         alarm_pool_entry_t *earliest_entry = &pool->entries[earliest_index];
         earliest_target = earliest_entry->target;
         ta_set_timeout(timer, timer_alarm_num, earliest_target);
+        now = (int64_t) ta_time_us_64(timer);
```
peterharperuk, line 268: Maybe get rid of "now" altogether and just call ta_time_us_64 in the two places where its needed?

kilograham: sure; it avoids now getting stale ;-)

andygpz11 (approved):
This looks good and makes sense to me.

@kilograham please can you review and merge to `develop` if happy?

It would be useful to be able to pull this.

Thanks

lurch:
Perhaps Graham will merge it 71 minutes in the future :laughing:

kilograham (approved):

PR #1969, opened 2024-10-03
Add Wi-Fi firmware partition support for Pico 2 W

will-v-pi:
This adds the ability to store and load the Wi-Fi firmware for Pico 2 W in a partition. It can be enabled by adding `pico_use_wifi_firmware_partition(<exe_name>)` to your CMakeLists.txt, which will embed a compatible partition table in the binary, and output two firmware UF2s to use (`<exe_name>_firmware.uf2` for regular, and `<exe_name>_firmware_tbyb.uf2` for TBYB). You can also create your own partition table and use that.

A Wi-Fi firmware partition is detected as having the ID `0x776966696669726d`, and the UF2 family ID for Wi-Fi firmware blobs is `0xe48bff55`. The default firmware partition starts at 3500K into the flash - should this be changed to depend on PICO_FLASH_SIZE_BYTES? The default firmware partition is also duplicated with A/B partitions in the same location in flash - this is required to ensure a signature check is performed before loading the Wi-Fi firmware, as there's no way to call the bootrom to check the signature of a single partition (unless chaining into it), you can only call `pick_ab_partition`.

The Wi-Fi firmware blob is marked in it's image_def as an RP2350 Risc-V executable, and the partition is marked as `ignored_during_riscv_boot` - this ensures that it can work with TBYB (as TBYB only works for executable image_defs), and that signature checks are performed before loading the firmware when Secure Boot is enabled (because signature checks are performed for all executable image_defs in a partition that is not marked `ignored_during_arm_boot`). This is slightly clunky, but seems to work robustly.


lurch:
pinging @peterharperuk as he did a lot of the Wifi-related stuff for Pico 1.

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -155,4 +155,104 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
     endfunction()
 
     pico_promote_common_scope_vars()
+
+    # Set env var, so it can be accessed in the function
+    set(ENV{PICO_CYW43_DRIVER_CURRENT_PATH} ${CMAKE_CURRENT_LIST_DIR})
```
peterharperuk, line 160: Why does this have to be an "ENV" var rather than just a normal CMake variable?

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -155,4 +155,104 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
...
+    function(pico_use_partition_firmware TARGET)
+        target_compile_definitions(${TARGET} PRIVATE CYW43_USE_PARTITION_FIRMWARE=1)
+        pico_embed_pt_in_binary(${TARGET} $ENV{PICO_CYW43_DRIVER_CURRENT_PATH}/wifi_pt.json)
+
+        find_package (Python3 REQUIRED COMPONENTS Interpreter)
+
+        # Wifi firmware blob
+        add_custom_target(${TARGET}_firmware_w_blob DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/firmware_w_blob.S)
```
peterharperuk, line 168: It might be worth dropping the "w" only version for simplicity. I only added this in case there was an issue with the new "wb" firmware.

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
will-v-pi: It needs to be accessible from where the function is called (which is in the user's CMakeLists.txt) so a normal CMake variable wouldn't work as it doesn't get propogate upwards - but actually a pico_common_scope_var would probably be better than an environment variable so I'll change it to that

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
will-v-pi: Thanks - done

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -104,6 +114,73 @@ static void cyw43_sleep_timeout_reached(async_context_t *context, __unused async
 }
 
 bool cyw43_driver_init(async_context_t *context) {
+#if CYW43_USE_PARTITION_FIRMWARE
+    const int buf_words = (16 * 4) + 1;   // maximum of 16 partitions, each with maximum of 4 words returned, plus 1
+    uint32_t* buffer = malloc(buf_words * 4);
```
peterharperuk, line 119: malloc is best avoided for Micropython (we use some for BT but we need to fix that). You could use cyw43_state as a temp buffer it's not needed yet?

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -104,6 +114,73 @@ static void cyw43_sleep_timeout_reached(async_context_t *context, __unused async
...
+                if (id == 0x123456789abcdef0) {
+                    picked_p = p;
+                }
+            }
+
+            p++;
+        }
+
+        if (picked_p >= 0) {
+            uint32_t* workarea = malloc(0x1000);
```
peterharperuk, line 145: Ah. That's too big for cyw43_state

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -104,6 +114,73 @@ static void cyw43_sleep_timeout_reached(async_context_t *context, __unused async
...
+                return false;
+            }
+
+            CYW43_DEBUG("Chosen CYW43 firmware in partition %d\n", picked_p);
+            int ret = rom_get_partition_table_info(buffer, buf_words, PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION | (picked_p << 24));
+            assert(buffer[0] == (PT_INFO_PARTITION_LOCATION_AND_FLAGS | PT_INFO_SINGLE_PARTITION));
+            assert(ret == 3);
+
+            uint32_t location_and_permissions = buffer[1];
+            uint32_t saddr = ((location_and_permissions >> PICOBIN_PARTITION_LOCATION_FIRST_SECTOR_LSB) & 0x1fffu) * 4096;
```
peterharperuk, line 164: 4096 == FLASH_SECTOR_SIZE?

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
will-v-pi: Should I switch this one to use cyw43_state and leave the other one with malloc, or just leave them both as malloc?

peterharperuk: Leave it as it is for now. I can't immediately spot the code in Micropython which controls the heap space we have available.

peterharperuk:
Works nicely. I think we need to fix the Pico W (rp2040) build errors in src/rp2_common/pico_cyw43_driver/cyw43_driver.c
It would be nice if it gave an error if you're a numpty like me and put pico_use_partition_firmware after pico_add_extra_outputs, but that could be improved later?

will-v-pi:
> It would be nice if it gave an error if you're a numpty like me and put pico_use_partition_firmware after pico_add_extra_outputs, but that could be improved later?

Have added in separate PR #2054

will-v-pi:
> I think we need to fix the Pico W (rp2040) build errors in src/rp2_common/pico_cyw43_driver/cyw43_driver.c

I've added a check that it's not RP2040 in the CMake function, so it'll throw a fatal error at that point if you try to build a binary with `pico_use_partition_firmware` for RP2040

armandomontanez:
If the Bazel checks are getting in your way, you can add the two new files to a `filegroup` and file an issue+leave a TODO to support this in Bazel and assign it to me.

peterharperuk (dismissed):

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1071,6 +1071,22 @@ static inline int rom_get_last_boot_type(void) {
...
 
+/*! \brief  Pick A/B partition for a separate partition
+ * \ingroup pico_bootrom
+ *
+ * This will perform extra checks to prevent disrupting a main image TBYB, and return errors
+ * 
+ * Also checks that the chosen partition contained a valid image
+ *
+ * \param workarea_base base address of work area
+ * \param workarea_size size of work area
```
lurch, line 1082: Presumably there's some kind of minimum size that the workarea needs to have?

will-v-pi:
> If the Bazel checks are getting in your way, you can add the two new files to a `filegroup` (which will silence the error) and file an issue+leave a TODO to support this in Bazel and assign it to me.

Thanks, have done - I couldn't actually assign the issue to you, but have tagged you in it

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: It's documented as 3064 in the datasheet and other doxygen (this is the workarea for the `rom_pick_ab_partition_function`) and recommends the use of 3k (3072), but I believe that's too small - I always just use 4k (4096) which works

lurch: Just in case I was being too subtle - I was trying to ask if this minimum size should be mentioned in _this_ doxygen?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1071,6 +1071,22 @@ static inline int rom_get_last_boot_type(void) {
  */
 int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32_t permissions);
 
+/*! \brief  Pick A/B partition for a separate partition
+ * \ingroup pico_bootrom
+ *
+ * This will perform extra checks to prevent disrupting a main image TBYB, and return errors
+ * 
+ * Also checks that the chosen partition contained a valid image
```
lurch, line 1079: I still think it's extremely vague what this function actually **does**. Any chance you could add a more detailed description, or would the purpose of this function be more obvious if I actually understood all the partitions stuff better?

lurch:
> A Wi-Fi firmware partition is detected as having the ID 0x123456789abcdef0, and the UF2 family_is for Wi-Fi firmware blobs is 0x12345678 - these should probably both be changed to something else before merging?

Is that still going to be happening? I guess choosing randomly-generated IDs is much less likely to produce an accidental collision than a "nice" ID like 0x123456789abcdef0 ? And should those IDs be added to a header-file, to allow other UF2s to make use of this Wi-Fi firmware partition?

will-v-pi:
> > A Wi-Fi firmware partition is detected as having the ID 0x123456789abcdef0, and the UF2 family_is for Wi-Fi firmware blobs is 0x12345678 - these should probably both be changed to something else before merging?
> 
> Is that still going to be happening? I guess choosing randomly-generated IDs is much less likely to produce an accidental collision than a "nice" ID like 0x123456789abcdef0 ? And should those IDs be added to a header-file, to allow other UF2s to make use of this Wi-Fi firmware partition?

Yes - I was looking for ideas. We could go with `0x776966696669726d` (hex for `wififirm`) for the partition ID, and just use the `data` family_id we already have?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: Yeah, I didn't want to add an incorrect size to my new definition - double checking with @kilograham but I think the correct size for the workarea (for `rom_pick_ab_partition` and `rom_get_uf2_target_partition`) is 3264, obtained by calling `rom_pick_ab_partition` with increasing size until it returns a non-negative value. So it would probably be best to recommend a size of 3.25k (0xD00, 3328)?

will-v-pi: I've added a note that it requires the same minimum workarea size as `rom_pick_ab_partition` - the doxygen for that and `rom_get_uf2_target_partition` should still be updated to say the correct minimum workarea size, but I'll pop that in a separate PR

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: I've added a more detailed descriptions, with an example of when this function is necessary. I didn't want to delve too into the weeds (there are comments in the function for that), but hopefully it's a bit clearer

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1074,9 +1074,15 @@ int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32
 /*! \brief  Pick A/B partition for a separate partition
  * \ingroup pico_bootrom
  *
- * This will perform extra checks to prevent disrupting a main image TBYB, and return errors
+ * This will call `rom_pick_ab_partition` with the current `flash_update_boot_window_base`, while performing extra checks to prevent disrupting a main image TBYB.
+ * It requires the same minimum workarea size as `rom_pick_ab_partition`.
  * 
- * Also checks that the chosen partition contained a valid image
+ * For example, if an `explicit_buy` is pending then calling `pick_ab_partition` would normally clear the saved `flash_erase_addr` so the required erase would not
```
lurch, line 1080: This doxygen mentions both `rom_pick_ab_partition` and `pick_ab_partition` - are they referring to the same function or are they different?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1071,6 +1071,28 @@ static inline int rom_get_last_boot_type(void) {
...
+ * occur when `explicit_buy` is called - this function saves and restores that address to prevent this issue, and returns `BOOTROM_ERROR_NOT_PERMITTED` if the
+ * partition chosen by `pick_ab_partition` also requires a flash erase version downgrade (as you can't erase 2 partitions with one `explicit_buy` call).
+ * 
+ * It also checks that the chosen partition contained a valid image (eg a signed image when using secure boot), and returns `BOOTROM_ERROR_NOT_FOUND`
+ * if it does not.
+ *
+ * \param workarea_base base address of work area
+ * \param workarea_size size of work area
+ * @param partition_a_num the A partition of the pair
```
lurch, line 1089: nitpick: Might be worth sticking to `@param` rather than mixing `@param` and `\param` :wink:

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1071,6 +1071,28 @@ static inline int rom_get_last_boot_type(void) {
  */
 int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32_t permissions);
 
+/*! \brief  Pick A/B partition for a separate partition
+ * \ingroup pico_bootrom
+ *
+ * This will call `rom_pick_ab_partition` with the current `flash_update_boot_window_base`, while performing extra checks to prevent disrupting a main image TBYB.
```
lurch, line 1077: Thanks for adding the extra wording. Should the user always prefer using `rom_pick_ab_update_partition` over `rom_pick_ab_partition`, or do they serve different purposes?
Probably worth adding a `\see rom_pick_ab_partition` here? (and vice-versa in the other function's doxygen)

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1071,6 +1071,28 @@ static inline int rom_get_last_boot_type(void) {
  */
 int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32_t permissions);
 
+/*! \brief  Pick A/B partition for a separate partition
```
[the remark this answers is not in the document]
will-v-pi: `rom_…` is the SDK function wrapper for the `pick_ab_partition` function - so I was using that to refer to it in the SDK (ie what this calls, and where more docs are), but using `pick_ab_partition` to refer to the behaviour of the underlying bootrom function

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: I think that if you’re calling it in a TBYB binary before calling `explicit_buy` you should prefer this - for anything else the normal bootrom function is fine

Where does the \see go in the doxygen - is it just below \ingroup?

lurch:
> Yes - I was looking for ideas. We could go with `0x776966696669726d` (hex for `wififirm`) for the partition ID, and just use the `data` family_id we already have?

I've got no opinions on this so it's probably something that you and @peterharperuk ought to decide. Would MicroPython be able to use the same Wifi firmware partition as C-SDK code? (or is that another one of my stupidly naive questions? :laughing: )

src/rp2_common/pico_bootrom/include/pico/bootrom.h
lurch: > Where does the \see go in the doxygen - is it just below \ingroup?

I think it can go pretty much anywhere? https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_irq/include/hardware/irq.h has lots of examples.

src/rp2_common/pico_bootrom/include/pico/bootrom.h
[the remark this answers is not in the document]
will-v-pi: I'm trying to mean "Pick A/B partition for an A/B partition pair that is not the one which the current binary has booted from" - ie if you have partitions `0 (A), 1 (B w/ 0), 2 (A), 3 (B w/ 2)` and have booted the binary in partition 0/1, that binary might call this to pick between the separate partitions 2/3. I wasn't sure how to word it succinctly.

Maybe "Pick A/B partition with TBYB guards" would be better?

will-v-pi:
> Would MicroPython be able to use the same Wifi firmware partition as C-SDK code? (or is that another one of my stupidly naive questions? 😆 )

Yes, they definitely could - there's no C-SDK specific code in the Wifi firmware partition, it's just a raw binary blob, so should work with MicroPython too.

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: Thanks - added

src/rp2_common/pico_bootrom/include/pico/bootrom.h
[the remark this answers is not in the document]
lurch: "Pick A/B partition from a different partition-pair" ? :man_shrugging:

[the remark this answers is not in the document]
will-v-pi: After rethinking, I think that mentioning the different partition pair doesn't make sense - because you'll usually be using the normal `pick_ab_partition` on a different partition pair too.

I think the main distinction of this function from the raw bootrom one is the TBYB guards, so I think "Pick A/B partition with TBYB guards" is probably best?

lurch:
On a vaguely-related note, the doxygen at https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_bootrom/include/pico/bootrom.h#L834 talks about an "explicit buy flag" - is that the same thing as the TBYB flag referred to elsewhere? (If so, it might be worth editing that doxygen for consistency?)

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -104,6 +120,73 @@ static void cyw43_sleep_timeout_reached(async_context_t *context, __unused async
...
+        while (i < ret) {
+            i++;
+            uint32_t flags_and_permissions = buffer[i++];
+            bool has_id = (flags_and_permissions & PICOBIN_PARTITION_FLAGS_HAS_ID_BITS);
+            if (has_id) {
+                uint64_t id = 0;
+                id |= buffer[i++];
+                id |= ((uint64_t)(buffer[i++]) << 32ull);
+                if (id == CYW43_WIFI_FW_PARTITION_ID) {
+                    picked_p = p;
```
kilograham, line 143: is it deliberate not to break here?

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -104,6 +120,73 @@ static void cyw43_sleep_timeout_reached(async_context_t *context, __unused async
...
+        if (picked_p >= 0) {
+            uint32_t* workarea = malloc(0x1000);
+            picked_p = rom_pick_ab_update_partition(workarea, 0x1000, picked_p);
+            free(workarea);
+
+            if (picked_p < 0) {
+                if (picked_p == BOOTROM_ERROR_NOT_FOUND) {
+                    CYW43_DEBUG("Chosen CYW43 firmware partition was not verified\n");
+                } else if (picked_p == BOOTROM_ERROR_NOT_PERMITTED) {
+                    CYW43_DEBUG("Too many update boots going on at once\n");
```
kilograham, line 159: not sure what this error message means!

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
kilograham: i would just put this on the stack, it is only 17 words

kilograham (commented):
new source files should have copyright/license

src/rp2_common/pico_bootrom/bootrom.c
```diff
@@ -59,6 +62,23 @@ void __attribute__((noreturn)) rom_reset_usb_boot(uint32_t usb_activity_gpio_pin
...
+
+// Generated from adding the following code into the bootrom
+// scan_workarea_t* scan_workarea = (scan_workarea_t*)workarea;
+// printf("VERSION_DOWNGRADE_ERASE_ADDR %08x\n", &(always->zero_init.version_downgrade_erase_flash_addr));
+// printf("TBYB_FLAG_ADDR %08x\n", &(always->zero_init.tbyb_flag_flash_addr));
+// printf("IMAGE_DEF_VERIFIED %08x\n", (uint32_t)&(scan_workarea->parsed_block_loops[0].image_def.core.verified) - (uint32_t)scan_workarea);
+// printf("IMAGE_DEF_TBYB_FLAGGED %08x\n", (uint32_t)&(scan_workarea->parsed_block_loops[0].image_def.core.tbyb_flagged) - (uint32_t)scan_workarea);
+// printf("IMAGE_DEF_BASE %08x\n", (uint32_t)&(scan_workarea->parsed_block_loops[0].image_def.core.enclosing_window.base) - (uint32_t)scan_workarea);
+// printf("IMAGE_DEF_REL_BLOCK_OFFSET %08x\n", (uint32_t)&(scan_workarea->parsed_block_loops[0].image_def.core.window_rel_block_offset) - (uint32_t)scan_workarea);
+#define VERSION_DOWNGRADE_ERASE_ADDR *(uint32_t*)0x400e0338
```
kilograham, line 75: arguably these are not addresses. i would leave the leading * off and use in code references

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
will-v-pi: No - I will add a break

src/rp2_common/pico_bootrom/bootrom.c
will-v-pi: Technically they are addresses, eg the version downgrade erase address (ie the address in flash to erase) is stored at 0x400e0338

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
will-v-pi: It means there are multiple partitions being updated at once, so if you update the boot partition with a TBYB image and the firmware partition in the same boot you get this error.

I can't actually think of a scenario when this might happen, because you can only have one `flash_update_base`, but I left in the error handling for it in case it does (as I did get it to happen during my early testing, although the code has changed a lot since then)

src/rp2_common/pico_bootrom/bootrom.c
lurch: Page 31 of the RP2350 datasheet says `BOOTRAM_BASE 0x400e0000`

kilograham: yeah, more of a stylistic thing I guess; it's weird to see a CONSTANT as a lhs

kilograham: but i'm ok

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1071,6 +1071,29 @@ static inline int rom_get_last_boot_type(void) {
  */
 int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32_t permissions);
 
+/*! \brief  Pick A/B partition with TBYB guards
```
kilograham, line 1074: I still find the function description a bit confusing:
* maybe something along the lines of "Pick A/B partition for an arbitrary pair of partitions without disturbing any in process update A/B or TBYB boot" (needs improvement still)...
* not sure people know what a flash_erase_addr is maybe describe it in words (e.g. the erase of xxx because yyy"
* I'd say "using the flash_update_boot_window_base from the current boot"

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -19,7 +19,23 @@
...
-static async_context_t *cyw43_async_context;
+static async_context_t *cyw43_async_context = NULL;
+
+#if CYW43_USE_PARTITION_FIRMWARE
+    #include "pico/bootrom.h"
+    #include "hardware/flash.h"
+    #include "boot/picobin.h"
+    #include <stdlib.h>
+
+    // PICO_CONFIG: CYW43_WIFI_FW_PARTITION_ID, ID of Wi-Fi firmware partition, type=int, default=0x776966696669726d (wififirm), group=pico_cyw43_driver
```
kilograham, line 30: I would put this in the header, it seems like other people might need to know the value

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -154,5 +154,73 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
                     )
     endfunction()
 
+    set(PICO_CYW43_DRIVER_CURRENT_PATH ${CMAKE_CURRENT_LIST_DIR})
+    pico_register_common_scope_var(PICO_CYW43_DRIVER_CURRENT_PATH)
+
     pico_promote_common_scope_vars()
+
+    function(pico_use_partition_firmware TARGET)
```
kilograham, line 162: I think in the future as mentioned elsewhere we might want to allow more params to configure how we do things (partition size/loc, versioninfo, partition ID) but this is fine too add via CMake args later

That said, the name in the global namespace is a bit non-specific. perhaps `pico_configure_wifi_firmware_partition' (or select instead of configure

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -19,7 +19,23 @@
...
+    #ifndef CYW43_WIFI_FW_PARTITION_ID
+    #define CYW43_WIFI_FW_PARTITION_ID 0x776966696669726d // wififirm
```
kilograham, line 32: I was humming an hah-ing about this, as i don't know whether we want

a) this to be generic (so that different WiFi chips can use the same ID)
b) non-specific with version info, so people can search for the right partition

I think for now this is OK (even though the code is cyw43_ specific) as we can always change it in the future

kilograham: I suspect that in the future we're likely to want to pass the `CYW43_WIFI_FW_PARTITION_ID` from the CMake in which case the `CYW43_USE_PARTITION_FIRMWARE` is perhaps superfluous, but i'm fine with it. (note i guess we won't pass ID from CMake for now as it would require dynamic JSON)

Please however rename `CYW43_WIFI_FW_PARTITION_ID` to `CYW43_WIFI_FIRMWARE_PARTITION_ID` as FW_ is not used as a contraction in any other defines... and I could go with `CYW43_USE_FIRMWARE_PARTITION`

kilograham: Finally, is this WiFi only; in which case either one or the other define should gain/lose `_WIFI`

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: I think "Pick A/B partition without disturbing any in progress update or TBYB boot" works for the description - pick_ab_partition is always for an arbitrary pair of partitions, so that probably doesn't need including in the description?

src/rp2_common/pico_cyw43_driver/cyw43_driver.c
will-v-pi: Will do - so CYW43_FIRMWARE_PARTITION_IF and CYW43_USE_FIRMWARE_PARTITION (I was just using WIFI_FW as CYW43_WIFI_FW_LEN is the define I was replacing from cyw43-driver)

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
will-v-pi: Maybe `pico_use_wifi_firmware_partition`?

will-v-pi:
I've noticed an issue with the Risc-V build of the firmware UF2 - for some reason the compilation is putting a programme header in the ELF with offset of `0x0000` in the file loading to `0x1ffff000`, when the Arm compiler correctly puts this at offset `0x1000` in the file loading to `0x20000000`. @Wren6991 or @kilograham any ideas why the Arm and Risc-V compilers might be compiling the `wifi_firmware.S` file differently?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
kilograham: Ah, I was (wrongly) thinking this new API existed because pick_ab_partition was for the regular boot partitions, rather than because that API is somewhat broken. Perhaps just add a warning to the other API referencing this one

will-v-pi:
I've now got it working for Risc-V by storing copies of the firmware ELFs in the SDK, and then copying those into the build when necessary. It also needs increased bootrom stack size. This is tested and working fine.

However, I'm not sure if this is actually a solution that we want to put in the 2.1.0 SDK release, so it might be best if we don't merge this for 2.1.0 and I can investigate further to fix the Risc-V firmware build?

richard9999999999:
I see that default firmware partition ID is defined as follows:
#define CYW43_FIRMWARE_PARTITION_ID 0x776966696669726d // wififirm
SDK uses this value as default to find the partition

..Also in pt.json, this long value is used
     "id": "0x776966696669726d",

However when I check by picotool, I see following number set as partition ID:
```
 partition 1 (A):       0036b000->003a7000 S(rw) NSBOOT(rw) NS(rw), id=000000006669726d, "Wi-Fi Firmware", uf2 { data },
                        arm_boot 1, riscv_boot 0
```
So upper 32 bits are zeros..
And I need to set this truncated value to CYW43_FIRMWARE_PARTITION_ID to get it working.

Did I miss something?

richard9999999999:
One more note:
```
function(pico_use_wifi_firmware_partition TARGET)
        if (PICO_PLATFORM STREQUAL "rp2040")
                message(FATAL_ERROR "RP2040 does not support storing wi-fi firmware in partitions")
        endif()
        target_compile_definitions(${TARGET} PRIVATE CYW43_USE_FIRMWARE_PARTITION=1)
        pico_embed_pt_in_binary(${TARGET} ${PICO_CYW43_DRIVER_CURRENT_PATH}/wifi_pt.json)
```
 seems like that partitioning of binary is always done according wifi_pt.json file. In many cases user prefers to have its own partitioning (when e.g. custom bootloader is used). Could'n this be done via preprocessor macro define  e.g. something like "CYW43_EMBED_FIRMWARE_PARTITION=0"?

will-v-pi:
Rebased onto develop, which contains the `rom_pick_ab_update_partition` function

will-v-pi:
> However when I check by picotool, I see following number set as partition ID:

This looks like a bug in your build of picotool - for me it shows the correct full ID, and the ID in the actual partition table in flash should also be correct. If you're still seeing this with the latest picotool develop branch, could you raise an issue against that repo with details of how you're building picotool?

> In many cases user prefers to have his own partitioning (when e.g. custom bootloader is used)

I will push a commit to only set the partition table if it has not already been defined, and if the user defines it afterwards then that will override this definition

richard9999999999:
> This looks like a bug in your build of picotool 

It doesn't seems so. I needed to define partition id as follows:
CYW43_FIRMWARE_PARTITION_ID=0x000000006669726d
in CMakeLists.txt to make it working. Picotool I've just used to decide which id I need to set there..

richard9999999999:
,while in pt.json I have following id:

richard9999999999:
The
picotool version
shows
"picotool v2.1.2-develop (Windows, GNU-8.1.0, Release)"

richard9999999999:
`I will push a commit to only set the partition table if it has not already been defined, and if the user defines it afterwards then that will override this definition`
This solution does not fit to my case: I use encrypted bootloader and all partitioning (including wifi firmware partition) is defined in the bootloader binary (so in the different binary). Thus in this case we cannot decide wifi partition has been included or not using your
get_target_property(picotool_embed_pt ${TARGET} PICOTOOL_EMBED_PT)

richard9999999999:
One more note (possibly bug), which I found while debugging my problem with truncated partition ID:

file cyw43_driver.c

```
    } else {
        CYW43_DEBUG("No partition table, so cannot get firmware from partition - get_partition_table_info returned %d\n", ret);
        return false;
    }
```
above else is for "no partition table found" case. However the error case when we have some partitions but not with correct ID is not handled and it continues in initialization.
I mean missing else for:
`if (picked_p >= 0) {`

richard9999999999:
> That use case will work fine - the embedded partition table in these binaries will be ignored when they’re booted from partitions by the bootloader (as opposed to when they’re booted from slots, see 5.1.12-15 in the datasheet for more details)

Unfortunately in the encrypted bootloader sample like scenario (like my project) the wifi partition in the executable binary seems causes malfunction. So I have to exclude the wifi partition from the executable (by hack of your branch) and add it into the bootloader only.

will-v-pi:
> > That use case will work fine - the embedded partition table in these binaries will be ignored when they’re booted from partitions by the bootloader (as opposed to when they’re booted from slots, see 5.1.12-15 in the datasheet for more details)
> 
> Unfortunately in the encrypted bootloader sample like scenario (like my project) the wifi partition in the executable binary seems causes malfunction. So I have to exclude the wifi partition from the executable (by hack of your branch) and add it into the bootloader only.

I've fixed this, so you can now pass `NO_EMBEDDED_PT` to the function and it will not embed the partition table

will-v-pi:
Would we want to use the next family ID for this then (`0xe48bff5c`), and add a define into the SDK and picotool for it (`WIFI_FIRMWARE_FAMILY_ID`)?

peterharperuk:
ah right. I think CYW43_FIRMWARE_FAMILY_ID might be better - ie. have a reference to cyw in there. I wouldn't mention bluetooth

peterharperuk (dismissed):
Spent some time playing with this. Seems to work nicely

src/common/boot_uf2_headers/include/boot/uf2.h
```diff
@@ -26,6 +26,8 @@
 #define UF2_FLAG_MD5_PRESENT             0x00004000u
 #define UF2_FLAG_EXTENSION_FLAGS_PRESENT 0x00008000u
 
+#define CYW43_FIRMWARE_FAMILY_ID    0xe48bff55u
```
lurch, line 29: Just out of curiosity, why has this been added "below" `RP2040_FAMILY_ID` rather than "above" `RP2350_ARM_NS_FAMILY_ID` ?

And tangentially, I wonder if it'd be better to do
```c
#define RP2040_FAMILY_ID 0xe48bff56u
#define DEFAULT_FAMILY_ID_MIN RP2040_FAMILY_ID

#define RP2350_ARM_NS_FAMILY_ID 0xe48bff5bu
#define FAMILY_ID_MAX RP2350_ARM_NS_FAMILY_ID
```
in order to eliminate the duplicated hard-coded values? :thinking:

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -165,5 +165,91 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
...
 
+    set(PICO_CYW43_DRIVER_CURRENT_PATH ${CMAKE_CURRENT_LIST_DIR})
+    pico_register_common_scope_var(PICO_CYW43_DRIVER_CURRENT_PATH)
+
     pico_promote_common_scope_vars()
+
+    # pico_use_wifi_firmware_partition(TARGET [NO_EMBEDDED_PT])
+    # \brief\ Use a partition for the Wi-Fi firmware
+    #
+    # This will read the CYW43 firmware from a partition with the ID 0x776966696669726d,
```
lurch, line 176: Does the `ID 0x776966696669726d` here need to be updated to `ID of CYW43_FIRMWARE_FAMILY_ID` ?

src/rp2_common/pico_cyw43_driver/include/pico/cyw43_driver.h
```diff
@@ -20,6 +20,13 @@
 #include "cyw43_configport.h"
 #endif
 
+#if CYW43_USE_FIRMWARE_PARTITION
+// PICO_CONFIG: CYW43_FIRMWARE_PARTITION_ID, ID of Wi-Fi firmware partition, type=int, default=0x776966696669726d (wififirm), group=pico_cyw43_driver
+#ifndef CYW43_FIRMWARE_PARTITION_ID
+#define CYW43_FIRMWARE_PARTITION_ID 0x776966696669726d // wififirm
```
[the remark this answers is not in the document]
lurch: ...and if I **have** misunderstood, perhaps there ought to be a comment here telling the user that if they change this value, they'll also need to change the "Wi-Fi Firmware" `id` value in `wifi_pt.json` ?

src/rp2_common/pico_cyw43_driver/wifi_pt.json
```diff
@@ -0,0 +1,52 @@
...
+        "secure": "rw",
+        "nonsecure": "rw",
+        "bootloader": "rw"
+      },
+      "ignored_during_riscv_boot": true,
+      "no_reboot_on_uf2_download": true
+    },
+    {
+      "start": "3500k",
+      "size": "256k",
```
lurch, line 40: Another stupid question: Why does this partition have the same `start` and `size` as the previous partition?

src/rp2_common/pico_cyw43_driver/cyw43_firmware.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2024 Raspberry Pi (Trading) Ltd.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import sys
+
+with open(sys.argv[1], "r") as f:
```
lurch, line 9: Perhaps it's worth checking that `len(sys.argv) >= 3` before doing anything else?

src/rp2_common/pico_cyw43_driver/cyw43_firmware.py
```diff
@@ -0,0 +1,38 @@
...
+    data = f.read()
+    lines = data.split(";")
+    for line in lines[1].split("\n"):
```
lurch, line 12: I guess you could `break` from this `for` loop once both `cyw43_wifi_fw_len` and cyw43_clm_len` have been read?

src/rp2_common/pico_cyw43_driver/include/pico/cyw43_driver.h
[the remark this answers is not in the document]
will-v-pi: I'll add a note to the config that if you use a different ID you need to change it in the partition table JSON, although that use case is probably more for people using custom partition tables with a custom ID.

src/rp2_common/pico_cyw43_driver/CMakeLists.txt
```diff
@@ -165,5 +165,89 @@ if (EXISTS ${PICO_CYW43_DRIVER_PATH}/${CYW43_DRIVER_TEST_FILE})
...
+
+        target_compile_definitions(${TARGET}_firmware PRIVATE
+                NO_PICO_PLATFORM=1
+                )
+        target_compile_definitions(${TARGET}_firmware_tbyb PRIVATE
+                NO_PICO_PLATFORM=1
+                PICO_CRT0_IMAGE_TYPE_TBYB=1
+                )
+
+        target_link_options(${TARGET}_firmware PRIVATE -nostartfiles -nodefaultlibs -N LINKER:--script=${PICO_CYW43_DRIVER_CURRENT_PATH}/wifi_firmware.ld)
```
kilograham, line 226: does this work on non GCC?

src/common/boot_uf2_headers/include/boot/uf2.h
```diff
@@ -26,13 +26,18 @@
...
+
+// Bootrom supported family IDs
 #define RP2040_FAMILY_ID            0xe48bff56u
 #define ABSOLUTE_FAMILY_ID          0xe48bff57u
 #define DATA_FAMILY_ID              0xe48bff58u
 #define RP2350_ARM_S_FAMILY_ID      0xe48bff59u
 #define RP2350_RISCV_FAMILY_ID      0xe48bff5au
 #define RP2350_ARM_NS_FAMILY_ID     0xe48bff5bu
-#define FAMILY_ID_MAX               0xe48bff5bu
+#define DEFAULT_FAMILY_ID_MIN       RP2040_FAMILY_ID
```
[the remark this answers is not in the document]
will-v-pi: It's used in picotool to generate partition tables, to distinguish between bootrom flag family IDs and extra family IDs. So `BOOTROM_FAMILY_ID_MIN` and `_MAX` sound sensible, I will switch to those

kilograham (approved):

richard9999999999:
I tried to use the develop branch however getting following error:
```
[build] [  0%] Linking ASM executable efis_firmware_tbyb.elf
[build] 
[build] ERROR: cyw43-firmware is not a valid family ID
[build] 
[build] SYNOPSIS:
[build]     picotool uf2 convert [--quiet] [--verbose] <infile> [-t <type>] <outfile> [-t <type>] [-o <offset>] [--family
[build]                 <family_id>] [[--abs-block] [<abs_block_loc>]]
[build] 
[build] Use "picotool help uf2 convert" for more info
```
The build folder was deleted before and the picotool is loaded and build automatically. Does it use correct picotool version? Or did I miss something?

will-v-pi:
The picotool fix was only merged this morning, so you may need to delete your build folder and try again - https://github.com/raspberrypi/picotool/pull/247

richard9999999999:
Yes, it works now. Sorry I was too anxious to try the new changes.. :-)

Issue #1971, opened 2024-10-03
Update pico_clib_interface function signature 


Prabhuk:
LLVM libc's function signature has changed upstream to inclue __NOEXCEPT for mktime function.
The redeclaration in pico_clib_interface must change accordingly to continue to work with the latest revisions of clang toolchain.

lurch:
Fixed in `develop` branch.

PR #1972, opened 2024-10-03
Update mktime header declaration to match upstream LLVM libc.

Prabhuk:
Fixes:#1971

### DIFF
--- a/src/rp2_common/pico_clib_interface/include/llvm_libc/time.h
+++ b/src/rp2_common/pico_clib_interface/include/llvm_libc/time.h
@@ -14,7 +14,7 @@
 
 __BEGIN_C_DECLS
 
-time_t mktime(struct tm* arg);
+time_t mktime(struct tm* arg) __NOEXCEPT;
 struct tm* localtime_r(const time_t* timer, struct tm* buf);
 
 __END_C_DECLS


armandomontanez:
I'm guessing this isn't backwards-compatible with older toolchain distributions?

kilograham:
.. which it needs to be

Prabhuk:
Thanks for bringing up the compatibility question. I should've added the testing info to the PR description. 
The change is compatible with the old toolchain revision.

I was able to use the current Clang toolchain revision - e894df6392beea3723627329009f3e6d51d16f47 against pico-sdk master branch with this PR's commit cherry-picked. The bazel build was completed successfully.

petrhosek:
`mktime` is now included in [LLVM libc `time.h` header](https://github.com/llvm/llvm-project/blob/98a15c7b0c6ec129d371f0c121dbe9396c4f5609/libc/config/baremetal/arm/entrypoints.txt#L211), so the correct solution here is to delete that declaration altogether, but we also need to bump the toolchain version in [`MODULE.bazel`](https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/MODULE.bazel#L61-L92).

petrhosek:
I opened https://github.com/raspberrypi/pico-sdk/pull/1976 which updates the toolchain.

Prabhuk:
> `mktime` is now included in [LLVM libc `time.h` header](https://github.com/llvm/llvm-project/blob/98a15c7b0c6ec129d371f0c121dbe9396c4f5609/libc/config/baremetal/arm/entrypoints.txt#L211), so the correct solution here is to delete that declaration altogether, but we also need to bump the toolchain version in [`MODULE.bazel`](https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/MODULE.bazel#L61-L92).

That makes sense. @armandomontanez Will there ever be a case in Pigweed presubmits where the patches landed in the "develop" branch here be built with the toolchain revision mentioned in Pigweed's bazel source files?

armandomontanez:
Not automatically. For now, we'll be manually revving the required version of the Pico SDK. There may come a day where we do builds against HEAD of `develop` to learn about breakages sooner, but that's a ways out.

Prabhuk:
> Not automatically. For now, we'll be manually revving the required version of the Pico SDK. There may come a day where we do builds against HEAD of `develop` to learn about breakages sooner, but that's a ways out.

Great! I'll close this PR. I'm assuming when we update the toolchain revision in Pigweed, we will update the PicoSDK revision which includes #1976 to get the presubmits to pass.

PR #1973, opened 2024-10-04
[Bazel] Fix bazel build, add presubmit

armandomontanez:
* Fixes a missing dep in the Bazel build breaking the host build.
* Automagically finds all board headers.
* Improves presubmit script polish for GH Action readiness.
* Adds a GitHub action workflow for the Bazel build.


kilograham (approved):
thx

PR #1982, opened 2024-10-10
Move cyw32 pin definitions

peterharperuk:
It was probably not a good idea to put the default pin definitions in cyw43_configport.h as they're specific to pico_w and Micropython needs these but it has it's own version of cyw43_configport.h

### DIFF
--- a/src/boards/include/boards/pico2_w.h
+++ b/src/boards/include/boards/pico2_w.h
@@ -113,4 +113,39 @@
 #define PICO_RP2350_A2_SUPPORTED 1
 #endif
 
+// PICO_CONFIG: CYW43_PIN_WL_DYNAMIC, flag to indicate if cyw43 SPI pins can be changed at runtime, type=bool, default=false, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_PIN_WL_DYNAMIC
+#define CYW43_PIN_WL_DYNAMIC 0
+#endif
+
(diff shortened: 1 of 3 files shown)


src/boards/include/boards/pico2_w.h
```diff
@@ -113,4 +113,39 @@
...
+
+// PICO_CONFIG: CYW43_DEFAULT_PIN_WL_REG_ON, gpio pin to power up the cyw43 chip, type=int, default=23, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_DEFAULT_PIN_WL_REG_ON
+#define CYW43_DEFAULT_PIN_WL_REG_ON 23
+#endif
+
+// PICO_CONFIG: CYW43_DEFAULT_PIN_WL_DATA_OUT, gpio pin for spi data out to the cyw43 chip, type=int, default=24, advanced=true, group=pico_cyw43_driver
+#ifndef CYW43_DEFAULT_PIN_WL_DATA_OUT
+#define CYW43_DEFAULT_PIN_WL_DATA_OUT 24u
+#endif
```
lurch, line 129: Only asking out of curiosity: Why is there a `u` after `24` but not after `23`?

peterharperuk: mistake

lurch:
...and yet `#define PICO_DEFAULT_SPI_SCK_PIN 18` (and others) doesn't have a `u` after it?

kilograham (approved):

PR #1995, opened 2024-10-23
Remove Segger RTT from pico-btstack

peterharperuk:
We now have this in the SDK so adding it again in pico-btstack leads to link errors.

### DIFF
--- a/src/rp2_common/pico_btstack/CMakeLists.txt
+++ b/src/rp2_common/pico_btstack/CMakeLists.txt
@@ -28,7 +28,6 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
     target_sources(pico_btstack_base INTERFACE
         ${PICO_BTSTACK_PATH}/3rd-party/micro-ecc/uECC.c
         ${PICO_BTSTACK_PATH}/3rd-party/rijndael/rijndael.c
-        ${PICO_BTSTACK_PATH}/3rd-party/segger-rtt/SEGGER_RTT.c
         ${PICO_BTSTACK_PATH}/3rd-party/segger-rtt/SEGGER_RTT_printf.c
         ${PICO_BTSTACK_PATH}/platform/embedded/btstack_tlv_flash_bank.c
         ${PICO_BTSTACK_PATH}/platform/embedded/hci_dump_embedded_stdout.c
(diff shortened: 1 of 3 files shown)


src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -28,7 +28,6 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
     target_sources(pico_btstack_base INTERFACE
         ${PICO_BTSTACK_PATH}/3rd-party/micro-ecc/uECC.c
         ${PICO_BTSTACK_PATH}/3rd-party/rijndael/rijndael.c
-        ${PICO_BTSTACK_PATH}/3rd-party/segger-rtt/SEGGER_RTT.c
         ${PICO_BTSTACK_PATH}/3rd-party/segger-rtt/SEGGER_RTT_printf.c
```
lurch, line 31: Does this other segger-rtt line not need removing too?

lurch:
Ahhh, looks like this is a consequence of #1411

src/rp2_common/pico_btstack/CMakeLists.txt
peterharperuk: We don't have SEGGER_RTT_printf.c in the sdk so presumably not

lurch:
I don't know much about cmake / compilers / linkers / etc. but does this change potentially mean that a project using `pico_btstack` might fail to build if `pico_stdio_rtt` isn't also included / enabled? :man_shrugging:

peterharperuk:
> pico_btstack might fail to build if pico_stdio_rtt isn't also included / enabled?

Good point. Let me check.

src/rp2_common/pico_stdio_rtt/stdio_rtt.c
```diff
@@ -26,11 +26,11 @@ void stdio_rtt_deinit(void) {
 }
 
 static void stdio_rtt_out_chars(const char *buf, int length) {
-    SEGGER_RTT_Write(0, buf, length);
+    SEGGER_RTT_Write(0, buf, (unsigned)length);
 }
```
lurch, line 30: Presumably `length` can never be negative (?), so perhaps it makes sense to just change the function-prototype to `stdio_rtt_out_chars(const char *buf, unsigned int length)` ?
(Apologies if this is a stupid suggestion)

peterharperuk: Unfortunately the prototype of this function is defined by stdio_driver_t 

```
stdio_driver_t stdio_rtt = {
    .out_chars = stdio_rtt_out_chars,
```

lurch: Ahhh, so it _was_ a stupid suggestion :laughing:  Fair enough.

PR #1998, opened 2024-10-24
Abilty to limit iterations in hci processing loop

### LINKED ISSUE #1455: Deadlock problem when connecting to Bluetooth devices with high frequency output
My test setup is using single core, reading data from Bluetooth and write to USB 
When connecting to Bluetooth devices with high frequency output, the HCI layer always has data to read.
The Pico driver has a while loop to read data, it causes the loop will never end.
So that other data processing functions will never be called.
My solution is, killing the while loop, it may generate delay for reading but it works OK. 
Is there any other possible solution?

Here are my changes
btstack_hci_transport_cyw43.c
```
static void hci_transport_cyw43_process(void) {
    CYW43_THREAD_LOCK_CHECK
    uint32_t len = 0;
    //bool has_work;
    //do {
        int err = cyw43_bluetooth_hci_read(hci_packet_with_pre_buffer, sizeof(hci_packet_with_pre_buffer), &len);
        BT_DEBUG("bt in len=%lu err=%d\n", len, err);
        if (err == 0 && len > 0) {
            hci_transport_cyw43_packet_handler(hci_packet_with_pre_buffer[3], hci_packet_with_pre_buffer + 4, len - 4);
            //has_work = true;
        } else {
            //has_work = false;
        }
    //} while (has_work);
}
```

peterharperuk:
Fixes #1455

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
+++ b/src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
@@ -139,6 +139,9 @@ static void hci_transport_cyw43_process(void) {
     CYW43_THREAD_LOCK_CHECK
     uint32_t len = 0;
     bool has_work;
+#ifdef PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT
+    uint32_t loop_count = 0;
+#endif
     do {
         int err = cyw43_bluetooth_hci_read(hci_packet_with_pre_buffer, sizeof(hci_packet_with_pre_buffer), &len);
         BT_DEBUG("bt in len=%lu err=%d\n", len, err);
@@ -148,6 +151,12 @@ static void hci_transport_cyw43_process(void) {
         } else {
             has_work = false;
         }
+// PICO_CONFIG: PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT, limit the max number of iterations of the hci processing loop, type=int, advanced=true, group=pico_btstack
+#ifdef PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT
+        if (++loop_count >= PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT) {
+            break;
+        }
+#endif
     } while (has_work);
 }
 


src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -148,6 +151,13 @@ static void hci_transport_cyw43_process(void) {
         } else {
             has_work = false;
         }
+// PICO_CONFIG: PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT, max number of iterations of the hci processing loop, type=int, default=infinite, advanced=true, group=pico_btstack
```
lurch, line 154: `tools/extract_configs.py` doesn't support `default=infinite`. Perhaps it's better to just remove the `default` and change the description to "limit the max number..." ? (which hopefully makes it obvious that if `PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT` isn't set, there's no limit?)

kilograham: waiting on this

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -148,6 +151,13 @@ static void hci_transport_cyw43_process(void) {
         } else {
             has_work = false;
         }
+// PICO_CONFIG: PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT, limit the max number of iterations of the hci processing loop, type=int, advanced=true, group=pico_btstack
+#ifdef PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT
+        loop_count++;
+        if (loop_count > PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT) {
```
kilograham, line 157: seems like this should be >=

lurch: Or I guess both lines could be squashed to:
```c
if (loop_count++ > PICO_BTSTACK_CYW43_MAX_HCI_PROCESS_LOOP_COUNT) {
```
?

kilograham (approved):

PR #2002, opened 2024-10-27
Move CYW43-related PICO_CONFIG lines to cyw43_driver.h

lurch:
This is a follow-up to #1982 - I'm afraid that I didn't think to check at the time, but the `tools/extract_configs.py` script was getting upset that there were duplicate PICO_CONFIG entries in different files. So this PR fixes that.

### DIFF
--- a/src/boards/include/boards/pico2_w.h
+++ b/src/boards/include/boards/pico2_w.h
@@ -113,37 +113,37 @@
 #define PICO_RP2350_A2_SUPPORTED 1
 #endif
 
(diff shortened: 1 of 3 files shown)


kilograham (approved):

peterharperuk:
I don't really like these defines in cyw43_driver.h. I'd rather remove the PICO_CONFIG lines completely if they're causing a problem. Although this sounds like a tooling defect to me.

lurch:
I've just force-pushed an updated commit, which removes the defines from `cyw43_driver.h`

peterharperuk (approved):

peterharperuk:
> I've just force-pushed an updated commit, which removes the defines from `cyw43_driver.h`

ta

PR #2013, opened 2024-11-04
Add XIP cache maintenance API (fixes #2005)

### LINKED ISSUE #2005: Add XIP cache management routines
The XIP cache works really well except in certain cases involving [PSRAM and flash updates](https://forums.raspberrypi.com/viewtopic.php?t=378249) with a really [non-obvious (IMHO) workaround](https://forums.raspberrypi.com/viewtopic.php?t=378249#p2263677).

Because cache cleaning needs some special magic to work with PSRAM as discussed in the links above, it may make sense to add  some XIP cache management operations to the SDK (especially "XIP Clean").  Right now we have `arduino-pico`, `CircuitPython` and `MicroPython` all implementing the same XIP cache management code separately.  It would be cleaner and safer to factor that out up into the SDK.

I imagine something as simple as
* `xip_cache_clean`
* `xip_cache_invalidate`
* `xip_cache_invalidate_range` (?)

might be all that's needed, but maybe others have additional requirements?

Kind-of related to #1983, but at a higher level.

Wren6991:
Also add a cache clean to all `hardware_flash` implementations which flush the cache, to avoid losing cached PSRAM writes.

This adds five new functions:

* `xip_cache_invalidate_all()`
* `xip_cache_invalidate_range()`
* `xip_cache_clean_all()` (no-op on RP2040)
* `xip_cache_clean_range()` (no-op on RP2040)
* `xip_cache_pin_range()` (unavailable on RP2040)

It uses offsets (`- XIP_BASE`) instead of addresses, to be consistent with the `hardware_flash` APIs. If anyone strongly disagrees then we should change it now before we're stuck with it.

The library is unfortunately dependent on `hardware_sync`, just to get the barrier macros.

### DIFF
--- a/src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
+++ b/src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
@@ -17,7 +17,7 @@
  *
  * These functions apply some maintenance operation to either the entire cache contents, or a range
  * of offsets within the downstream address space. Offsets start from 0 (indicating the first byte
- * pf flash), so pointers should have XIP_BASE subtracted before passing into one of these
+ * of flash), so pointers should have XIP_BASE subtracted before passing into one of these
  * functions.
  *
  * \if rp2040-specific
@@ -55,13 +55,9 @@
  *
  */
 
-// PICO_CONFIG: PARAM_ASSERTIONS_ENABLED_HARDWARE_XIP_CACHE, Enable/disable assertions in the hardwdare_xip_cache module, type=bool, default=0, group=hardwdare_xip_cache
+// PICO_CONFIG: PARAM_ASSERTIONS_ENABLED_HARDWARE_XIP_CACHE, Enable/disable assertions in the hardware_xip_cache module, type=bool, default=0, group=hardware_xip_cache
 #ifndef PARAM_ASSERTIONS_ENABLED_HARDWARE_XIP_CACHE
-#ifdef PARAM_ASSERTIONS_ENABLED_FLASH // backwards compatibility with SDK < 2.0.0
-#define PARAM_ASSERTIONS_ENABLED_HARDWARE_XIP_CACHE PARAM_ASSERTIONS_ENABLED_FLASH
-#else
 #define PARAM_ASSERTIONS_ENABLED_HARDWARE_XIP_CACHE 0
-#endif
 #endif
 
 #define XIP_CACHE_LINE_SIZE _u(8)
@@ -152,7 +148,6 @@ void xip_cache_invalidate_range(uintptr_t start_offset, uintptr_t size_bytes);
  * \endif
  *
  */
-
 void xip_cache_clean_all(void);
 
 /*! \brief Clean a range of offsets within the XIP address space
--- a/src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
+++ b/src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
@@ -1,0 +1,215 @@
+/*
+ * Copyright (c) 2024 Raspberry Pi Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _HARDWARE_XIP_CACHE_H
+#define _HARDWARE_XIP_CACHE_H
+
+#include "pico.h"
+#include "hardware/regs/addressmap.h"
+
+/** \file xip_cache.h
+ *  \defgroup hardware_xip_cache hardware_xip_cache
+ *
+ * \brief Low-level cache maintenance operations for the XIP cache
+ *
+ * These functions apply some maintenance operation to either the entire cache contents, or a range
+ * of offsets within the downstream address space. Offsets start from 0 (indicating the first byte
+ * pf flash), so pointers should have XIP_BASE subtracted before passing into one of these
+ * functions.
+ *
+ * \if rp2040-specific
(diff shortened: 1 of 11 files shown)


src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
```diff
@@ -0,0 +1,215 @@
...
+ * 3. Invalidate ("flush") the entire cache (e.g. using xip_cache_invalidate_all())
+ *
+ * The invalidate ensures the programming is visible to subsequent reads. The clean ensures that the
+ * invalidate does not discard any cached PSRAM write data.
+ *
+ * \endif
+ *
+ */
+
+// PICO_CONFIG: PARAM_ASSERTIONS_ENABLED_HARDWARE_XIP_CACHE, Enable/disable assertions in the hardwdare_xip_cache module, type=bool, default=0, group=hardwdare_xip_cache
```
lurch, line 58: A couple of `hardwdare_xip_cache` typos here which ought to be `hardware_xip_cache`

src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
```diff
@@ -0,0 +1,215 @@
...
+ *  \ingroup hardware_xip_cache
+ *
+ * This causes the cache to write out all pending write data to the downstream memory. For example,
+ * when suspending the system with state retained in external PSRAM, this ensures all data has made
+ * it out to external PSRAM before powering down.
+ *
+ * This function is faster than calling xip_cache_clean_range() for the entire address space,
+ * because it iterates over cachelines instead of addresses.
+ *
+ * \if rp2040-specific
```
lurch, line 143: Small nitpick, but IMHO it looks a bit weird having rp2040-specific doxygen inside a `#if !XIP_CACHE_IS_READ_ONLY` block, when `XIP_CACHE_IS_READ_ONLY` is 1 on RP2040.

I wonder if it might look better as e.g. :
```c
/*! Doxygen for the xip_cache_clean_all function
*/
#if !XIP_CACHE_IS_READ_ONLY
void xip_cache_clean_all(void);
#else
static inline void xip_cache_clean_all(void) {}
#endif
```
as that way the Doxygen remains "outisde" of the `#if` block. Or does my suggested approach of having the `#if` between doxygen and the function-declaration screw up the doxygen processing? (I'm aware that it can be a bit "delicate"!)

src/rp2_common/hardware_xip_cache/xip_cache.c
```diff
@@ -0,0 +1,108 @@
...
+// (which makes flash execution momentarily unsafe)
+
+__always_inline static void check_xip_offset_range(uintptr_t start_offset, uintptr_t size_bytes) {
+    // We use offsets, not addresses, for consistency with the flash API. This means the range of
+    // valid inputs starts at 0.
+    (void)start_offset;
+    (void)size_bytes;
+    valid_params_if(HARDWARE_XIP_CACHE, start_offset <= XIP_CACHE_ADDRESS_SPACE_SIZE);
+    valid_params_if(HARDWARE_XIP_CACHE, start_offset + size_bytes <= XIP_CACHE_ADDRESS_SPACE_SIZE);
+    valid_params_if(HARDWARE_XIP_CACHE, start_offset + size_bytes <= start_offset);
```
lurch, line 34: Huh, presumably `start_offset + size_bytes <= start_offset` will only be true if `size_bytes` is `0`?

kilograham:
pls add new library to `docs/index.h`

src/rp2_common/hardware_xip_cache/xip_cache.c
Wren6991: It's a test for unsigned wrapping...

src/rp2_common/hardware_xip_cache/include/hardware/xip_cache.h
Wren6991: > Or does my suggested approach of having the #if between doxygen and the function-declaration screw up the doxygen processing

I don't know either, and the fact I would have to check means I don't think this change is worth considering

src/rp2_common/hardware_xip_cache/xip_cache.c
Wren6991: Btw you usually get a warning these days if you accidentally write a simple vacuously true/false comparison

Wren6991: Oh I'm an idiot, the check is inverted. Thanks

will-v-pi (approved):
LGTM for RP2350 - I've tested all the functions and they work as expected with PSRAM and Flash

kilograham (approved):

lurch:
@kilograham @Wren6991 If I try building the `kitchen_sink` in the SDK with arm-none-eabi-gcc 6.3.1 (which is admittedly pretty ancient) it complains at me that:
```
...../pico-sdk/src/rp2_common/hardware_xip_cache/xip_cache.c:27:29: error: always_inline function might not be inlinable [-Werror=attributes]
 __always_inline static void check_xip_offset_range(uintptr_t start_offset, uintptr_t size_bytes) {
                             ^~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

If I replace the `__always_inline` there with `__force_inline` then the `kitchen_sink` builds successfully to completion.

Let me know if this was the right fix, and I'll submit a PR (or of course you can fix it directly yourself if that's easier).

kilograham:
correct; we have `__force_inline` to work around GCC variations/bugs

PR #2016, opened 2024-11-04
Remove code bloat in public SDK for 64K roms

kilograham:


Wren6991:
Yep, this is cleaner. As much fun as it was to always deliberately take a bus fault during startup

Wren6991 (approved):

PR #2018, opened 2024-11-04
Add missing instruction to vestigial copro_dis.py

kilograham:
@will-v-pi can you make sure `picotool` has this too


will-v-pi (approved):
LGTM

PR #2021, opened 2024-11-05
Fix flash_safe_execute timeout under FreeRTOS

### LINKED ISSUE #2007: flash_safe_execute timeout under FreeRTOS
There appears to be a bug in `flash_safe_execute` which causes it to timeout on FreeRTOS after being called repeatedly, no matter how large the enter/exit timeout is set to. Once it has started timing out, it never recovers (i.e. all future calls time out).

From some debugging, it seems like we're somehow ending up with a flash lockout task with the wrong "other core number" - maybe it's left over from a previous `flash_safe_execute` call?

```
(gdb) info thread
  Id   Target Id                                                          Frame
* 1    Thread 1 "rp2040.core0" (Name: rp2040.core0, state: debug-request) 0x10006b2e in default_enter_safe_zone_timeout_ms (timeout_ms=<optimized out>)
    at pico-sdk/src/rp2_common/pico_flash/flash.c:157
  2    Thread 2 "rp2040.core1" (Name: rp2040.core1, state: debug-request) 0x2000020e in flash_lockout_task (vother_core_num=0x1)
    at pico-sdk/src/rp2_common/pico_flash/flash.c:126
```

It's worth noting that this isn't fully consistent - it seems to depend on how FreeRTOS chooses to schedule the tasks. The following example code reproduces after a short while when dropped into `hello_freertos` from `pico-examples`:

```c
#include "FreeRTOS.h"
#include "task.h"

#include "pico/flash.h"

void vFlash( void * pvArg )
{
}

void vTask( void * pvArg )
{
    for( ; ; )
    {
        const int xResult = flash_safe_execute( vFlash, NULL, 1000 );
        configASSERT( xResult == PICO_OK );
    }
}

int main( void )
{
    const BaseType_t xResult = xTaskCreate( vTask, "vTask",
                                            configMINIMAL_STACK_SIZE, NULL, 0,
                                            NULL );
    configASSERT( xResult == pdPASS );

    vTaskStartScheduler();

    return 0;
}
```

jackwilsdon:
In some cases, the flash lockout task can start on the wrong core and not have time to move to the correct core before exit is called. This causes a timeout as the exit function is looking at the wrong core when checking for the lockout task.

Fixes #2007.

### DIFF
--- a/src/rp2_common/pico_flash/flash.c
+++ b/src/rp2_common/pico_flash/flash.c
@@ -142,16 +142,14 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         // it only prevents the other core from also entering a critical section.
         // Therefore, we must do our own handshake which starts a task on the other core and have it disable interrupts
         uint core_num = get_core_num();
-        // create at low priority
+        // create at low priority on other core
         TaskHandle_t task_handle;
-        if (pdPASS != xTaskCreate(flash_lockout_task, "flash lockout", configMINIMAL_STACK_SIZE, (void *)core_num, 0, &task_handle)) {
+        if (pdPASS != xTaskCreateAffinitySet(flash_lockout_task, "flash lockout", configMINIMAL_STACK_SIZE, (void *)core_num, 0, 1u << (core_num ^ 1), &task_handle)) {
             return PICO_ERROR_INSUFFICIENT_RESOURCES;
         }
         lockout_state[core_num] = FREERTOS_LOCKOUT_LOCKER_WAITING;
         __sev();
-        // bind to other core
-        vTaskCoreAffinitySet(task_handle, 1u << (core_num ^ 1));
-        // and make it super high priority
+        // make it super high priority
         vTaskPrioritySet(task_handle, configMAX_PRIORITIES -1);
         absolute_time_t until = make_timeout_time_ms(timeout_ms);
         while (lockout_state[core_num] != FREERTOS_LOCKOUT_LOCKEE_READY && !time_reached(until)) {


src/rp2_common/pico_flash/flash.c
```diff
@@ -142,16 +142,14 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
...
         // Therefore, we must do our own handshake which starts a task on the other core and have it disable interrupts
         uint core_num = get_core_num();
-        // create at low priority
+        // create at low priority on other core
         TaskHandle_t task_handle;
-        if (pdPASS != xTaskCreate(flash_lockout_task, "flash lockout", configMINIMAL_STACK_SIZE, (void *)core_num, 0, &task_handle)) {
+        if (pdPASS != xTaskCreateAffinitySet(flash_lockout_task, "flash lockout", configMINIMAL_STACK_SIZE, (void *)core_num, 0, 1u << (core_num ^ 1), &task_handle)) {
             return PICO_ERROR_INSUFFICIENT_RESOURCES;
         }
         lockout_state[core_num] = FREERTOS_LOCKOUT_LOCKER_WAITING;
```
peterharperuk, line 150: I wonder if this line should be moved before the call to xTaskCreateAffinitySet

jackwilsdon: Hmm yes it probably should, as otherwise it'll use the previous lockout state.

peterharperuk: and then there's no point creating it with a low priority? May as well make it super high immediately?

jackwilsdon: All sounds reasonable! I'll change it to do that and give it a test tomorrow :+1:

peterharperuk: Ignore my comments. On reflection, I prefer it the way it is.

peterharperuk (approved):

peterharperuk:
@kilograham This seems like a harmless change and seems to improve things for the (albeit artificial) example in the report.

kilograham (approved):

PR #2026, opened 2024-11-07
Clear i2c abort reason less often.

### LINKED ISSUE #2025: i2c lockup
If you run the i2c bus_scan example on Pico 2 with or without anything connected and ONLY stdout USB enabled then the device invariably locks up. It seems to get stuck in the loop waiting for an abort which never seems to happen! It works fine if sdio uart is enabled?! And Pico 1 seems ok.

```
        i2c->hw->data_cmd =
                bool_to_bit(first && i2c->restart_on_next) << I2C_IC_DATA_CMD_RESTART_LSB |
                bool_to_bit(last && !nostop) << I2C_IC_DATA_CMD_STOP_LSB |
                I2C_IC_DATA_CMD_CMD_BITS; // -> 1 for read

        do {
            abort_reason = i2c->hw->tx_abrt_source;
            abort = (bool) i2c->hw->clr_tx_abrt;
            if (timeout_check) {
                timeout = timeout_check(ts, false);
                abort |= timeout;
            }
        } while (!abort && !i2c_get_read_available(i2c)); <-- NEVER EXITS
```

Usually the abort seems to happen after around 130 iterations. If I repeat the command in the loop after 200 iterations or so, it seems to fix the problem.

peterharperuk:
It seems to be possible to get stuck in the loop which is checking for abort. It can take 100s of iterations before an abort happens and on each iteration we're clearing the abort interrupt even when it's not required. If we only clear the abort when needed the lockup doesn't seem to be reproducible.

Fixes #2025

### DIFF
--- a/src/rp2_common/hardware_i2c/i2c.c
+++ b/src/rp2_common/hardware_i2c/i2c.c
@@ -298,7 +298,10 @@ static int i2c_read_blocking_internal(i2c_inst_t *i2c, uint8_t addr, uint8_t *ds
 
         do {
             abort_reason = i2c->hw->tx_abrt_source;
-            abort = (bool) i2c->hw->clr_tx_abrt;
+            if (i2c->hw->raw_intr_stat & I2C_IC_RAW_INTR_STAT_TX_ABRT_BITS) {
+                abort = true;
+                i2c->hw->clr_tx_abrt;
+            }
             if (timeout_check) {
                 timeout = timeout_check(ts, false);
                 abort |= timeout;


src/rp2_common/hardware_i2c/i2c.c
```diff
@@ -298,7 +298,10 @@ static int i2c_read_blocking_internal(i2c_inst_t *i2c, uint8_t addr, uint8_t *ds
 
         do {
             abort_reason = i2c->hw->tx_abrt_source;
-            abort = (bool) i2c->hw->clr_tx_abrt;
+            if (i2c->hw->raw_intr_stat & I2C_IC_INTR_MASK_M_TX_ABRT_BITS) {
```
kilograham, line 301: Seems like this would be clearer with I2C_IC_RAW_INTR_STAT_TX_ABRT_BITS

kilograham (approved):

PR #2033, opened 2024-11-11
More CYW43 config tweaks

lurch:
- Re-synchronise the PICO_CMAKE_CONFIG: entries with the corresponding PICO_CONFIG: entries
 - Fix a silly typo I made in #2002
 - Enhance config-extraction scripts to catch similar typos in future

This is a follow-up to the previous #2002 (I spotted the typo, and ended up "fixing" a few other things in the process of fixing that!).

PR #2036, opened 2024-11-12
Fix mislabeled PICO_CMAKE_CONFIG annotation

armandomontanez:
Fixes the CYW43_DEFAULT_PIN_WL_CLOCK configuration option's documentation annotation to use the correct name.


lurch:
I believe #2033 already contains this fix?

armandomontanez:
Indeed! closing.

PR #2045, opened 2024-11-15
Remove ARM assembly from RISC-V CYW43 SPI PIO

### LINKED ISSUE #2044: CYW43 PIO SPI driver has hardcoded ARM assembly, breaks for RISC-V builds
There's hardcoded ARM assembly in the WiFi software SPI driver which causes RISC-V builds to fail:
https://github.com/raspberrypi/pico-sdk/blob/1b64cac8d697cdfda50527dfcff58c083f55429b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c#L81-L85

earlephilhower:
Replace the ARM assembly for dual 16b swap with generic GCC builtin macros to allow building CYW43 SPI PIO driver for RISC-V

Fixes #2044

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -80,7 +80,15 @@ static uint32_t counter = 0;
 
 //#define SWAP32(A) ((((A) & 0xff000000U) >> 8) | (((A) & 0xff0000U) << 8) | (((A) & 0xff00U) >> 8) | (((A) & 0xffU) << 8))
 __force_inline static uint32_t __swap16x2(uint32_t a) {
+#ifndef __riscv
     pico_default_asm ("rev16 %0, %0" : "+l" (a) : : );
+#else
+    uint32_t tmp;
+    pico_default_asm (
+        "rev8 %1, %0\n"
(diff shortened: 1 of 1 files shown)


kilograham:
This might be more compact

```
__force_inline static uint32_t __swap16x2(uint32_t a) {
#ifndef __riscv
    pico_default_asm ("rev16 %0, %0" : "+l" (a) : : );
#else
    uint32_t tmp;
    pico_default_asm (
        "rev8 %1, %0\n"
        "rori %0, %1, 16\n"
        : "+l" (a), "=l" (tmp));
#endif
    return a;
}
```

earlephilhower:
Agreed.  Quick check of generated assembly using `__builtin` gives
````
  5c:	6987d713          	rev8	a4,a5
  60:	83c1                	srli	a5,a5,0x10
  62:	8341                	srli	a4,a4,0x10
  64:	6987d793          	rev8	a5,a5
  68:	8fd9                	or	a5,a5,a4
````
22 bytes vs only 8 for the hand ASM.
````
  4a:	6987d713          	rev8	a4,a5
  4e:	61075793          	rori	a5,a4,0x10
````

I'll update and squash accordingly.

earlephilhower:
Updated with hand-asm patch.  Verified the assembly output does match the ARM and GCC builtins.

kilograham (approved):

PR #2048, opened 2024-11-18
PPP2 and Plasma2350: SPICE to SPCE.

Gadgetoid:
As per discussion on https://github.com/adafruit/circuitpython/pull/9803

Change "SPICE" to "SPCE" for consistency.

This may affect user code, but it's early days so the impact should be pretty low. 😬


lurch (approved):
Pretty small change, so LGTM

PR #2050, opened 2024-11-19
Remove CMake from choco_packages.config

lurch:
This will hopefully fix the Windows GitHub CI as https://github.com/actions/runner-images/blob/win22/20241113.3/images/windows/Windows2022-Readme.md#tools says that CMake is already installed by default.

GitHub CI for pico-sdk was failing on Windows with:
```
A newer version of cmake.install (v3.31.0) is already installed.
 Use --allow-downgrade or --force to attempt to install older versions.
Failed to install cmake because a previous dependency failed.
```

### DIFF
--- a/.github/workflows/choco_packages.config
+++ b/.github/workflows/choco_packages.config
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="gcc-arm-embedded" version="10.2.1" />
-  <package id="cmake" version="3.25.2" installArguments="ADD_CMAKE_TO_PATH=System" />
   <package id="mingw" version="12.2.0" />
   <package id="ninja" version="1.11.1" />
 </packages>


lurch:
@liamfraser I guess #2075 means that this PR is redundant now?

kilograham:
closed as obsolete per @liamfraser

lurch:
This problem has popped up again in https://github.com/gneverov/pico-sdk/actions/runs/12738602275/job/35501170042 except now the error-message is:
```
Failures
 - cmake - Failed to install cmake because a previous dependency failed.
 - cmake.install - A newer version of cmake.install (v3.31.3) is already installed.
 Use --allow-downgrade or --force to attempt to install older versions.
Error: Process completed with exit code 1.
```
And https://github.com/actions/runner-images/blob/win22/20250105.1/images/windows/Windows2022-Readme.md#tools confirms that the `20250105.1.0` Windows GitHub runner image (which is the one that happened to be used by that CI run) now has `CMake 3.31.3` pre-installed.

PR #2054, opened 2024-11-19
Enforce ordering of pico_add_extra_outputs and picotool functions

will-v-pi:
This throws a fatal error if a picotool post-processing function (eg `pico_sign_binary`) comes after `pico_add_extra_outputs` has been called for that target. Without this any picotool post-processing functions after `pico_add_extra_outputs` will have no effect.


peterharperuk (dismissed):
Tested with pico_use_partition_firmware and works nicely. Thanks

tools/CMakeLists.txt
```diff
@@ -422,6 +444,9 @@ endfunction()
 # Run picotool post-processing on the binary - must be called after
 # all required properties have been set
 function(picotool_postprocess_binary TARGET)
+    set_target_properties(${TARGET} PROPERTIES
+        PICOTOOL_PROCESSING_DONE true
```
lurch, line 448: It seems _slightly_ misleading to set `PICOTOOL_PROCESSING_DONE true` as the very first thing inside the `picotool_postprocess_binary` function, _before_ any post-processing has actually been done?
I know I'm only arguing semantics, but I wonder if naming this `PICOTOOL_POSTPROCESSING_STARTED` would be slightly less misleading / confusing? (with `picotool_check_not_done` similarly renamed to `picotool_check_not_started`)

will-v-pi: It doesn't really start at this point either, because this is all happening during the CMake Configure which doesn't actually do anything other than write some make/ninja files. It only starts when you actually run the CMake Build step.

I can change it to `PICOTOOL_PROCESSING_CONFIGURED` instead if that's clearer, and then `picotool_check_not_configured`?

will-v-pi: I've gone with `PICOTOOL_PROCESSING_CONFIGURED` and `picotool_check_configurable` - as it's checking if the picotool processing is configurable at that point, and the configuration has been finalised when `PICOTOOL_PROCESSING_CONFIGURED` gets set to `true`

kilograham (dismissed):

kilograham (approved):

PR #2058, opened 2024-11-19
allow PICO_CORE1_STACK_SIZE=0

### LINKED ISSUE #1803: Compilation warnings in  multicore.c: when PICO_CORE1_STACK_SIZE=0x0
When explicitly creating a memory region to use multicore_launch_core1_with_stack it is reasonable to then set the size of the default core1 stack to zero in the cmake file. This creates multiple compilation warnings in multicore.c (-Wall). Not a big deal but is would be nice to have the compilation run with no warnings.

kilograham:
fixes #1803

### DIFF
--- a/src/rp2_common/pico_multicore/include/pico/multicore.h
+++ b/src/rp2_common/pico_multicore/include/pico/multicore.h
@@ -29,7 +29,7 @@ extern "C" {
  * \include multicore.c
 */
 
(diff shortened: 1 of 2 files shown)


will-v-pi (approved):
LGTM, and have double checked it works. Although the use case of setting `PICO_CORE1_STACK_SIZE` to 0 and then allocating the stack on the fly throws no warnings with this, but didn't seem to throw warnings for me without this fix either even with `-Wall`

PR #1926, opened 2024-09-11
Change div_int_frac methods to be suffixed by the number of bits of fraction e.g. div_int_frac8

kilograham:
fixes #1924

PR #2061, opened 2024-11-20
Fix the name of a PICO_CMAKE_CONFIG

lurch:
Looks like this change accidentally got omitted from #1926


kilograham (approved):
will merge for completeness; though i think i'm going to remove all the CYW43_ related CMake vars, since they really belong just in the board header IMHO

lurch:
> though i think i'm going to remove all the CYW43_ related CMake vars

I'm not gonna complain (I'm still not sure why some things are header defines and others are CMake vars), but pulling in @peterharperuk as IIRC he added these CYW43 CMake vars...

Is it so that you can add e.g. `-DCYW43_PIO_CLOCK_DIV_FRAC8=4` to the `cmake` command-line to easily override the value set in the board header file?

PR #2062, opened 2024-11-20
make panic loop back to breakpoint FWIW, and remove spurious header

kilograham:


Wren6991 (approved):
Seems fine. `bkpt` and `ebreak` do not "return" except when under debugger control anyway, so this just avoids the case where you continue and the debug prompt goes quiet.

kilograham:
yup; i just noticed it in passing

PR #2063, opened 2024-11-20
making pio_calculate_clkdiv8_from_float round to the neareset 1/256 (not lower 1/256)

### LINKED ISSUE #1931: pio_calculate_clkdiv_from_float is not rounding
The `pio_calculate_clkdiv_from_float()` function is truncating the fractional result:

```c
*div_frac = (uint8_t)((div - (float)*div_int) * (1u << 8u));
```

It should be rounding. One way to do this is to add half of 1/256 up front, as the very first line of the function, before the range check:

```c
// Bump div by 1/512 so that we're rounding instead of truncating.
div += 0.001953125f;
```

This error caused an NTSC color burst to be far enough off in frequency to not be recognized by two different monitors.

kilograham:
fixes #1931


lurch:
I'm afraid I'm not familiar enough with bit-manipulation / fixed-point arithmetic to know if this fix is correct, but it appears to match the suggestion in #1931 . Would a similar problem affect any of the other int_frac functions?

kilograham:
this was superceded by #2065

PR #2066, opened 2024-11-20
use configured uart_instance rather than default instance for flushing STDIO uart

### LINKED ISSUE #1936: stdio_uart and stdio_uart_init_full don't play nice with flush
If using stdio_uart_init_full to initialize stdio without using the macros (PICO_DEFAULT_UART_INSTANCE is unset) then 
stdio_uart_out_flush will get stuck in assert loop from printf

Suggest that in stdio_uart.c the function stdio_uart_out_flush is implemented using the uart_instance variable

static void stdio_uart_out_flush(void) {
    uart_tx_wait_blocking(uart_instance);
}

kilograham:
fixes #1936


will-v-pi (approved):
LGTM

peterharperuk (approved):

Issue #1961, opened 2024-09-29
PIOasm disasm comment wrong


NollKollTroll:
When compiling the following test-code:
`.program test
.fifo txput
.wrap_target
    mov  rxfifo[0], ISR
    mov  rxfifo[1], ISR
    mov  rxfifo[2], ISR
    mov  rxfifo[3], ISR
    mov  rxfifo[Y], ISR
.wrap`

I seem to get a correct binary according to the documentation. Sadly the disasm comments are wrong, swapping indexed and Y-mode:
`static const uint16_t test_program_instructions[] = {
            //     .wrap_target
    0x8018, //  0: mov    rxfifo[y], isr             
    0x8019, //  1: mov    rxfifo[y], isr             
    0x801a, //  2: mov    rxfifo[y], isr             
    0x801b, //  3: mov    rxfifo[y], isr             
    0x8010, //  4: mov    rxfifo[0], isr             
            //     .wrap
};`

I have tried very hard to get the line endings correct in the preview for the code sections, but have failed despite all my efforts. Sorry.

kilograham:
merged into develop

PR #2067, opened 2024-11-20
fix incorrect disassembly of MOV rx_fifo instructions

kilograham:


kilograham:
fixes #1961

will-v-pi (approved):
LGTM (after reading up on the assembler syntax)

PR #2068, opened 2024-11-20
prevent early macro expansion in PICOBIN_..._AS_BITS() macros

### LINKED ISSUE #1974: Adding a -DRP2350 causes this error src/rp2_common/pico_crt0/embedded_start_block.inc.S:51: Error: invalid operands (*UND* and *ABS* sections) for `<<'
Simply adding a compile time definition  -DRP2350 to a pico2 CMake project and compiling the crt0.S with the GNU assembler causes the following error from the included file embedded_start_block.inc.S
 "pico-sdk/src/rp2_common/pico_crt0/embedded_start_block.inc.S:51: Error: invalid operands (*UND* and *ABS* sections) for `<<'"

A simple #undef RP2350 placed at the beginning of "embedded_start_block.inc.S" allows it to compile/assemble properly

kilograham:
fixes #1974


will-v-pi (approved):
LGTM - tested with -DRP2350 compile option and it works fine

PR #2072, opened 2024-11-20
Update documented max PIN values, to accomodate RP2350B

lurch:


peterharperuk (dismissed):

kilograham:
I know the old max was incorrect in some places, but I'm not sure in terms of helping the user that using `NUM_BANK0_GPIOS-1` is useful. maybe we should use a pseudo value here like (highest GPIO) or (GPIO count) depending on the context...

Still on the fence though so open to thoughts (cc @liamfraser @will-v-pi @peterharperuk)

peterharperuk:
Yes, it's a bit techy. What's a bank anyway?

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -1403,12 +1403,12 @@ extern void gpio_debug_pins_init(void);
 #endif
 
 
-// PICO_CONFIG: PICO_DEBUG_PIN_BASE, First pin to use for debug output (if enabled), min=0, max=28, default=19, group=hardware_gpio
+// PICO_CONFIG: PICO_DEBUG_PIN_BASE, First pin to use for debug output (if enabled), min=0, max=NUM_BANK0_GPIOS - 1, default=19, group=hardware_gpio
```
kilograham, line 1406: this is actually 0-31 still

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -1403,12 +1403,12 @@ extern void gpio_debug_pins_init(void);
...
 #ifndef PICO_DEBUG_PIN_BASE
 #define PICO_DEBUG_PIN_BASE 19u
 #endif
 
-// PICO_CONFIG: PICO_DEBUG_PIN_COUNT, Number of pins to use for debug output (if enabled), min=1, max=28, default=3, group=hardware_gpio
+// PICO_CONFIG: PICO_DEBUG_PIN_COUNT, Number of pins to use for debug output (if enabled), min=1, max=NUM_BANK0_GPIOS, default=3, group=hardware_gpio
```
kilograham, line 1411: this is 32 still (though actually really BASE + COUNT <=32), but i don't think we care atm

kilograham:
> > Yes, it's a bit techy.
> 
> I don't know if what I've added here is any more "techy" than anything else in section 5.1 of https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf ? 🤔

well most of those actually refer to other configuration values whereas NUM_BANK0_GPIOS is a bit of an internal thing.

we are a bit more conversational on others, e.g "1 on RP2350", so I'd say either "highest GPIO number", or "47 on RP2350B, 31 otherwise"

kilograham:
29 if you are being pedantic

src/rp2_common/hardware_gpio/include/hardware/gpio.h
lurch: Ah, thanks!

lurch:
Thanks for the suggestion - I'm much happier with something specific like "47 on RP2350B, 31 otherwise" than with something vague like "highest GPIO". (That change has been pushed)

src/rp2_common/hardware_gpio/include/hardware/gpio.h
lurch: Actually, I'm not able to test it, but would what I've done in the latest commit actually work? :thinking:

src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -1428,8 +1428,8 @@ extern void gpio_debug_pins_init(void);
 #define CU_SELECT_DEBUG_PINS(x) template<> inline bool __debug_pin_settings<x>::enabled() { return true; };
 #define DEBUG_PINS_ENABLED(p) (__debug_pin_settings<p>::enabled())
 #endif
-#define DEBUG_PINS_SET(p, v) if (DEBUG_PINS_ENABLED(p)) gpio_set_mask((unsigned)(v)<<PICO_DEBUG_PIN_BASE)
-#define DEBUG_PINS_CLR(p, v) if (DEBUG_PINS_ENABLED(p)) gpio_clr_mask((unsigned)(v)<<PICO_DEBUG_PIN_BASE)
-#define DEBUG_PINS_XOR(p, v) if (DEBUG_PINS_ENABLED(p)) gpio_xor_mask((unsigned)(v)<<PICO_DEBUG_PIN_BASE)
+#define DEBUG_PINS_SET(p, v) if (DEBUG_PINS_ENABLED(p)) gpio_set_mask64((unit64_t)(v)<<PICO_DEBUG_PIN_BASE)
+#define DEBUG_PINS_CLR(p, v) if (DEBUG_PINS_ENABLED(p)) gpio_clr_mask64((unit64_t)(v)<<PICO_DEBUG_PIN_BASE)
+#define DEBUG_PINS_XOR(p, v) if (DEBUG_PINS_ENABLED(p)) gpio_xor_mask64((unit64_t)(v)<<PICO_DEBUG_PIN_BASE)
```
lurch, line 1433: Happy to move this change into a separate commit or separate PR if you'd prefer.

kilograham: yeah, I'm not sure i want to make that change - DEBUG_PINS is not really used much anyway; not sure anyone yet cares about using pins > 32

kilograham:
presumably we can't have a comma in the middle of the max value though (oops)

kilograham:
maybe / instead of , or "47/29 depending on RP2350B/other"

src/rp2_common/hardware_gpio/include/hardware/gpio.h
lurch: Fair. I've removed this change, squashed and force-pushed.

kilograham (approved):

PR #2073, opened 2024-11-20
Update work area size from 3064 to 3264

will-v-pi:
Currently the descriptions for `rom_load_partition_table`, `rom_pick_ab_partition`, `rom_get_uf2_target_partition` and `rom_chain_image` document the minimum workarea size as 3064, with 3K recommended. This is too small, as the actual size required is 3264, so 3.25K should be recommended instead.

The minimum size was obtained with the following for loop in a binary:
```
for (int i=0xc00; i <= 0x1000; i++) {
    uint8_t* workarea = malloc(i);
    int ret = rom_load_partition_table(workarea, i, true);
    free(workarea);
    if (ret != -13) {
        printf("PT Load Ret %d, size %d\n", ret, i);
        break;
    }
}
```


kilograham (approved):

lurch:
> Currently the descriptions for `rom_load_partition_table`, `rom_pick_ab_partition`, `rom_get_uf2_target_partition` and `rom_chain_image` document the minimum workarea size

I forgot to ask earlier - do each of these four functions definitely have the _same_ minimum workarea size?

Issue #1810, opened 2024-08-12
pico_aon_timer pulls in a lot of C Lib code on RP2040


kilograham:
See https://github.com/micropython/micropython/pull/15619

kilograham:
We should offer both time and date based libraries perhaps, and/or at least let people plug in their own conversion routines if they already have them

kilograham:
Since this is worked-around in MP i'm gonna push this to 2.2.0 as I don't have a magic bullet for the general case

kilograham:
actually changed my mind again

kilograham:
merged into develop

PR #2079, opened 2024-11-21
support both struct timespec and struct tm variants of the aon timer APIs

kilograham:
… since use of one type on RP2040 and the other on RP2350 require pulling in C library code. 

Also provide weak `pico_` wrappers for `localtime_r` and `mktime` so that the user can override the conversion functions

### DIFF
--- a/src/rp2_common/pico_aon_timer/aon_timer.c
+++ b/src/rp2_common/pico_aon_timer/aon_timer.c
@@ -14,9 +14,6 @@ static aon_timer_alarm_handler_t aon_timer_alarm_handler;
 #include "hardware/rtc.h"
 #include "pico/util/datetime.h"
 
-static __force_inline bool ts_to_tm(const struct timespec *ts, struct tm *tm) {
-    return pico_localtime_r(&ts->tv_sec, tm) != NULL;
-}
 #elif HAS_POWMAN_TIMER
 #include "hardware/powman.h"
 
@@ -28,6 +25,9 @@ static void powman_timer_irq_handler(void) {
     if (aon_timer_alarm_handler) aon_timer_alarm_handler();
 }
 
+static bool ts_to_tm(const struct timespec *ts, struct tm *tm) {
+    return pico_localtime_r(&ts->tv_sec, tm) != NULL;
+}
 #endif
 
 static bool tm_to_ts(const struct tm *tm, struct timespec *ts) {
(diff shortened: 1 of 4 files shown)


kilograham:
fixes #1810

src/rp2_common/pico_aon_timer/include/pico/aon_timer.h
```diff
@@ -23,6 +23,23 @@
  * \if rp2350_specific
  * This library uses the Powman Timer on RP2350.
  * \endif
+ *
+ * This library supports both `aon_timer_xxx_calendar()` methods which take a calendar date/time (as struct tm),
+ * and `aon_timer_xxx()` methods which take a time since epoch (as struct timespec)
+ *
+ * \if rp2040_specific
+ * NOTE: On RP2040 the non date/time methods must convert the struct timespec to a date/time value which is handled via
```
peterharperuk, line 31: I don't really understand what it means by "non date/time methods" here

peterharperuk: maybe a list of the methods that are affected by the clib overhead in 2040 and 2350 (a different set?) would be useful?

kilograham: can you suggest a clarification - this refers very much to the previous paragraph (i could add "calendar" here)

I suppose we could list them all

src/rp2_common/pico_aon_timer/aon_timer.c
```diff
@@ -23,56 +27,89 @@ static void powman_timer_irq_handler(void) {
...
 #if HAS_RP2040_RTC
-    ((void)wakeup_from_low_power); // don't have a choice
     datetime_t dt;
+    rtc_get_datetime(&dt);
+    datetime_to_tm(&dt, tm);
+    return true;
+#elif HAS_POWMAN_TIMER
+    struct timespec ts;
+    bool ok = tm_to_ts(tm, &ts);
+    return ok && aon_timer_get_time(&ts);
```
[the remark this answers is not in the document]
kilograham: ha oops

peterharperuk (approved):
Seems to be working ok and I can't spot any issues. Docs are much clearer now!

peterharperuk (approved):

PR #2082, opened 2024-11-21
Save/restore QMI window 1 registers across calls to flash_exit_xip().

Wren6991:
This works around the RP2350 ROM reinitialising window 1 even when it does not actually issue an XIP exit sequence to that QSPI device. If no exit sequence is issued, then the original configuration still applies.

This is not valid in the case where the ROM *does* issue an XIP exit sequence to the second chip select, because the original mode is no longer valid once the device is in the serial command state. The ROM does this when FLASH_DEVINFO (loaded from OTP, stored in boot RAM) has a nonzero size for chip select 1. To distinguish the two cases, this patch adds getters/setters for FLASH_DEVINFO.

If chip select 1 has a nonzero size in FLASH_DEVINFO then the ROM's register changes for window 1 are not reverted, and instead the hardware_flash code makes additional changes to ensure writes also work (as the ROM only reinitialises for reads).

This means that, in the case you have opted into ROM support for the second chip select by setting a nonzero CS1 size in FLASH_DEVINFO, the device on chip select 1 will continue to function correctly, but full performance won't be restored until it is reinitialised. The best way to handle this is initialising both chip select devices in your XIP setup function (your boot2).

### DIFF
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -123,31 +123,31 @@ void flash_flush_cache(void);
 
 #if !PICO_RP2040
 typedef enum {
-	FLASH_DEVINFO_SIZE_NONE = 0x0,
-	FLASH_DEVINFO_SIZE_8K = 0x1,
-	FLASH_DEVINFO_SIZE_16K = 0x2,
-	FLASH_DEVINFO_SIZE_32K = 0x3,
-	FLASH_DEVINFO_SIZE_64K = 0x4,
-	FLASH_DEVINFO_SIZE_128K = 0x5,
-	FLASH_DEVINFO_SIZE_256K = 0x6,
-	FLASH_DEVINFO_SIZE_512K = 0x7,
-	FLASH_DEVINFO_SIZE_1M = 0x8,
-	FLASH_DEVINFO_SIZE_2M = 0x9,
-	FLASH_DEVINFO_SIZE_4M = 0xa,
-	FLASH_DEVINFO_SIZE_8M = 0xb,
-	FLASH_DEVINFO_SIZE_16M = 0xc,
-	FLASH_DEVINFO_SIZE_MAX = 0xc
+    FLASH_DEVINFO_SIZE_NONE = 0x0,
+    FLASH_DEVINFO_SIZE_8K = 0x1,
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -121,6 +121,132 @@ void flash_do_cmd(const uint8_t *txbuf, uint8_t *rxbuf, size_t count);
 
 void flash_flush_cache(void);
 
+#if !PICO_RP2040
```
kilograham, line 124: ughh.. this file is using TABs not 4 spaces

kilograham (changes requested):

src/rp2_common/hardware_flash/include/hardware/flash.h
Wren6991: oh, yeah. My editor is set up to guess from the file, but there were no indented lines in here.

kilograham (approved):
sweet

will-v-pi:
I was having a play with powering down flash, and this PR appears to have broken that on RP2350? I have the following function, which works on RP2040 with 2.0.0 and 2.1.0, but on RP2350 only works on 2.0.0, or 2.1.0 with this PR reverted. On 2.1.0 with this PR included it hits a hardfault during the first `flash_do_cmd`.

From looking at the `.dis`, it appears that when `flash_devinfo_get_cs_size` uses `flash_devinfo_ptr`, it calls the `rom_data_lookup` stored in flash which is currently powered down, hence the hardfault.

I guess the fix would be to add a `rom_data_lookup_inline` function and use that instead?

```c
void __no_inline_not_in_flash_func(test_flash_powercycle)() {
    // Disable Flash
    uint8_t txbuf[1] = {0xB9};
    uint8_t rxbuf[1] = {0x00};

    uint32_t intrs = save_and_disable_interrupts();
    flash_do_cmd(txbuf, rxbuf, 1);

    busy_wait_at_least_cycles(125000000);

    // Won't do anything as flash is powered down
    flash_range_erase(0, 4096);

    // Enable flash
    txbuf[0] = 0xAB;

    flash_do_cmd(txbuf, rxbuf, 1);

    busy_wait_at_least_cycles(125000000);
    restore_interrupts_from_disabled(intrs);
}
```

PR #2084, opened 2024-11-21
add rom_reset_usb_boot_extra which supports >32 pins and ACTIVE_LOW

kilograham:
I noticed this in passing - this change also adds relevant PICO_CONFIG to specify the ACTIVE_LOW default (note if the ACTIVITY_LED is not specified, the OTP defaults will be used on RP2350)

### DIFF
--- a/src/rp2_common/pico_bootrom/include/pico/bootrom.h
+++ b/src/rp2_common/pico_bootrom/include/pico/bootrom.h
@@ -238,6 +238,25 @@ static inline void __attribute__((noreturn)) reset_usb_boot(uint32_t usb_activit
 }
 
 /*!
+ * \brief Reboot the device into BOOTSEL mode
+ * \ingroup pico_bootrom
+ *
+ * This function reboots the device into the BOOTSEL mode ('usb boot").
+ *
+ * Facilities are provided to enable an "activity light" via GPIO attached LED for the USB Mass Storage Device,
+ * and to limit the USB interfaces exposed.
+ *
+ * \param usb_activity_gpio_pin  GPIO pin to be used as an activitiy pin, or -1 for none
+ *                               from the host.
+ * \param disable_interface_mask value to control exposed interfaces
+ *  - 0 To enable both interfaces (as per a cold boot)
+ *  - 1 To disable the USB Mass Storage Interface
+ *  - 2 To disable the USB PICOBOOT Interface
+ * \param usb_activity_gpio_pin_active_low Activity GPIO is active low (ignored on RP2040)
+ */
+void __attribute__((noreturn)) rom_reset_usb_boot_extra(int usb_activity_gpio_pin, uint32_t disable_interface_mask, bool usb_activity_gpio_pin_active_low);
+
+/*!
  * \brief Connect the SSI/QMI to the QSPI pads
  * \ingroup pico_bootrom
  * 
--- a/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
+++ b/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
@@ -73,6 +73,11 @@
 
 // PICO_CONFIG: PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED, Optionally define a pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE), type=int, min=0, max=47 on RP2350B, 29 otherwise, group=pico_stdio_usb
 
+// PICO_CONFIG: PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED_ACTIVE_LOW, Whether pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE) is active low, type=bool, default=0, group=pico_stdio_usb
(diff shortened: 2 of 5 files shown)


peterharperuk (approved):
Seems to work!

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -237,6 +237,25 @@ static inline void __attribute__((noreturn)) reset_usb_boot(uint32_t usb_activit
...
 }
 
+/*!
+ * \brief Reboot the device into BOOTSEL mode
+ * \ingroup pico_bootrom
+ *
+ * This function reboots the device into the BOOTSEL mode ('usb boot").
+ *
+ * Facilities are provided to enable an "activity light" via GPIO attached LED for the USB Mass Storage Device,
+ * and to limit the USB interfaces exposed.
```
lurch, line 247: This chunk of doxygen is **identical** to that for `rom_reset_usb_boot`, which will probably look pretty confusing in the rendered Doxygen output? (both https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf and https://www.raspberrypi.com/documentation/pico-sdk/ )
Is the addition of active_low the only difference between `rom_reset_usb_boot` and `rom_reset_usb_boot_extra` ?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -237,6 +237,25 @@ static inline void __attribute__((noreturn)) reset_usb_boot(uint32_t usb_activit
...
+ *
+ * \param usb_activity_gpio_pin  GPIO pin to be used as an activitiy pin, or -1 for none
+ *                               from the host.
+ * \param disable_interface_mask value to control exposed interfaces
+ *  - 0 To enable both interfaces (as per a cold boot)
+ *  - 1 To disable the USB Mass Storage Interface
+ *  - 2 To disable the USB PICOBOOT Interface
+ * \param usb_activity_gpio_pin_active_low Activity GPIO is active low (ignored on RP2040)
+ */
+void __attribute__((noreturn)) rom_reset_usb_boot_extra(int usb_activity_gpio_pin, uint32_t disable_interface_mask, bool usb_activity_gpio_pin_active_low);
```
lurch, line 257: `rom_reset_usb_boot` has `uint32_t usb_activity_gpio_pin_mask`, whereas `rom_reset_usb_boot_extra` has `int usb_activity_gpio_pin`. This might be a stupid question, but given the similarity of these two functions, would it make sense to modify `rom_reset_usb_boot_extra` to accept a `pin_mask` value too?

src/rp2_common/pico_bootrom/include/pico/bootrom.h
peterharperuk: It doesn't use a mask

lurch:
> (note if the ACTIVITY_LED is not specified, the OTP defaults will be used on RP2350)

Given that the OTP `BOOTSEL_LED_CFG` has both `PIN` and `ACTIVELOW` fields, should we make `PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED_ACTIVE_LOW` and `PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED_ACTIVE_LOW` optional-and-unset-by-default (just like their `_ACTIVITY_LED` counterparts), rather than forcing them to default to 0 ?

kilograham:
> > (note if the ACTIVITY_LED is not specified, the OTP defaults will be used on RP2350)
> 
> Given that the OTP `BOOTSEL_LED_CFG` has both `PIN` and `ACTIVELOW` fields, should we make `PICO_BOOTSEL_VIA_DOUBLE_RESET_ACTIVITY_LED_ACTIVE_LOW` and `PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED_ACTIVE_LOW` optional-and-unset-by-default (just like their `_ACTIVITY_LED` counterparts), rather than forcing them to default to 0 ?

they are only used when the corresponding PIN define is set

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -73,6 +73,11 @@
 
 // PICO_CONFIG: PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED, Optionally define a pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE), type=int, min=0, max=47 on RP2350B, 29 otherwise, group=pico_stdio_usb
 
+// PICO_CONFIG: PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED_ACTIVE_LOW, Whether pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE) is active low, type=bool, default=0, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED_ACTIVE_LOW
+#define PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED_ACTIVE_LOW 0
```
lurch, line 78: This should have been `PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED_ACTIVE_LOW` (i.e. no `FIXED`) - #2096 fixes that.

PR #2086, opened 2024-11-23
RISC-V __mulsf3: handle exponent increase due to rounding

Wren6991:
Originally reported by @Panda381. I thought this case was unreachable for multiply, but it's unfortunately just rare. This costs one cycle.


Wren6991:
Ah, this needs to also check overflow after exponent adjustment. I'll add another testcase

Wren6991:
...and do that without making underflow slower. Ready for review now

kilograham (approved):

PR #2090, opened 2024-11-23
Comment out CMake cyw43 vars which aren't really that useful in geneal i don't think

kilograham:
These can be specified via board header or `target_compile_definitions`


peterharperuk (commented):
Hmm, ok. If you're removing them, wouldn't it be better to delete them?

peterharperuk (approved):

kilograham:
yeah, i decided to leave them for now in case we want them back

PR #2095, opened 2024-11-25
Update armv8m_mpu.h - adding void

blckmn:
Removes error: 
error: old-style function definition [-Werror=old-style-definition]


kilograham:
worth seeing if we just need to update the CMSIS headers

will-v-pi:
The headers upstream haven’t had a new release since 6.1.0 in May (which we have), and it doesn’t look like this is fixed upstream. I think our stance on the other CMSIS headers issues we’ve had is that we don’t want to change them downstream, so please submit this PR to upstream instead (https://github.com/ARM-software/CMSIS_6) and we’ll merge them into this once they do a new release.

blckmn:
I will close now that it is PR'd upstream.

https://github.com/ARM-software/CMSIS_6/pull/211

PR #2113, opened 2024-11-28
Parse included board files for pico_cmake_set declarations

### LINKED ISSUE #2112: "// pico_cmake_set" not seen in included header files.
If a board header file includes other header files, then CMake does not see any **pico_cmake_set** directives in the secondary header file.

This particularly affects the **vgaboard.h** header file, which includes **pico.h**. None of the pico settings are seen by CMake. This does not matter in this case as all the settings are default values anyway.

However, it is not possible to simply change **pico.h** to **pico2.h** for a VGA board with pico2 installed, as CMake will still think it is compiling for RP2040 rather than RP2350.

will-v-pi:
Currently if you have a `#include` within a board header file (eg in [vgaboard.h](https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/vgaboard.h)) any `pico_cmake_set` and `pico_cmake_set_default` declarations in the included file aren't parsed (#2112)

This fixes #2112 by parsing those files for these comments. It only parses includes which start with `boards/`, as it needs to extract the `.h` file directly for use with `pico_find_in_paths`.

The inclusion order matches what would be expected (I think) - ie:
```c
// pico_cmake_set PICO_PLATFORM = rp2350

#include "boards/pico.h"
```
gives `PICO_PLATFORM = rp2040` from `pico.h`,
```c
#include "boards/pico.h"

// pico_cmake_set PICO_PLATFORM = rp2350
```
gives `PICO_PLATFORM = rp2350`,
```c
// pico_cmake_set_default PICO_FLASH_SIZE_BYTES = (8 * 1024 * 1024)
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (8 * 1024 * 1024)
#endif

#include "boards/pico.h"
```
gives `PICO_FLASH_SIZE_BYTES = (8 * 1024 * 1024)` (matching the pre-processor defines), and
```c
#include "boards/pico.h"

// pico_cmake_set_default PICO_FLASH_SIZE_BYTES = (8 * 1024 * 1024)
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (8 * 1024 * 1024)
#endif
```
gives `PICO_FLASH_SIZE_BYTES = (2 * 1024 * 1024)` from `pico.h` (again matching the pre-processor defines)


lurch:
If the behaviour matches what the pre-processor does, then I think this sounds great! Well done @will-v-pi :smiley:

PR #2127, opened 2024-12-05
ta_set_timeout can fail to set an alarm

### LINKED ISSUE #2118: Problem with repeating timer in SDK2.1.0
Hello,
I am experiencing a problem with repeating timer in SDK 2.1.0 (I haven't seen it in SDK 1.x.x before). The problem seems happen only when timer callback function is "light" (I mean doesn't contain too much of processing):

```
repeating_timer_t my_sampling_timer;

bool my_timer_callback(repeating_timer_t *rt)
{
	//volatile int tmp = 0;
	//for(int i=0; i <10000; i++)
	//{
	//	tmp++;
	//}
	
	return true;
}

int main()
{	
	alarm_pool_t *pool = alarm_pool_create(2, 1);	
	alarm_pool_add_repeating_timer_ms(pool, -20, my_timer_callback, NULL, &my_sampling_timer);

	while(1)
	{
		printf("looping\n");
	}
}
```
If the code is as above, the callback is called only few times. If I uncomment commented part in the callback (to make functionality to spend some more time in callback), the callback is called still during main looping (as desired). 
The problem happens in the RELEASE build only.
I use GCC 10.3.1 (not sure if related, probably not)

peterharperuk:
If alarm_pool_irq_handler takes <1us between handling an alarm and calling ta_set_timeout then no alarms will be set as it will appear as if an earlier alarm is already armedi before the target time.

Make sure ta_set_timeout always leaves with an alarm set by checking the armed status.

Fixes #2118

### DIFF
--- a/src/rp2_common/pico_time_adapter/include/pico/time_adapter.h
+++ b/src/rp2_common/pico_time_adapter/include/pico/time_adapter.h
@@ -48,7 +48,8 @@ static inline void ta_set_timeout(alarm_pool_timer_t *timer, uint alarm_num, int
     // 2. If we just passed the target time, then time_til_target will be high, meaning we will
     //    likely not do the update, but this is OK since the caller who has the full 64 bits
     //    must check if the target time has passed when we return anyway to avoid races.
-    if (time_til_target < time_til_alarm) {
+    // 3. We should never leave here without an alarm being armed
+    if (time_til_target < time_til_alarm || (timer_hw_from_timer(timer)->armed & (1 << alarm_num)) == 0) {
         timer_hw_from_timer(timer)->alarm[alarm_num] = (uint32_t) target;
     }
 }
--- a/src/rp2_common/pico_time_adapter/include/pico/time_adapter.h
+++ b/src/rp2_common/pico_time_adapter/include/pico/time_adapter.h
@@ -76,6 +76,7 @@ static inline void ta_enable_irq_handler(alarm_pool_timer_t *timer, uint alarm_n
 
 static inline void ta_disable_irq_handler(alarm_pool_timer_t *timer, uint alarm_num, irq_handler_t irq_handler) {
     uint irq_num = timer_hardware_alarm_get_irq_num(timer, alarm_num);
+    timer_hw_from_timer(timer)->armed = 1u << alarm_num; // disarm the timer
     hw_clear_bits(&timer_hw_from_timer(timer)->inte, 1u << alarm_num);
     irq_set_enabled(irq_num, true);
     irq_remove_handler(irq_num, irq_handler);
--- a/test/pico_time_test/pico_time_test.c
+++ b/test/pico_time_test/pico_time_test.c
@@ -8,6 +8,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <hardware/sync.h>
+#include "hardware/clocks.h"
 #include "pico/stdlib.h"
 #include "pico/test.h"
 // Include sys/types.h before inttypes.h to work around issue with
@@ -74,6 +75,7 @@ static bool repeating_timer_callback(struct repeating_timer *t) {
 int issue_195_test(void);
 int issue_1812_test(void);
 int issue_1953_test(void);
+int issue_2118_test(void);
 
 int main() {
     setup_default_uart();
@@ -246,6 +248,8 @@ int main() {
 
     issue_1953_test();
 
+    issue_2118_test();
+
     PICOTEST_END_TEST();
 }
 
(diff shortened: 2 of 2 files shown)


kilograham:
From #2118:

```


        if (earliest_target != -1) { // cancelled alarm has target of -1
            ta_set_timeout(timer, timer_alarm_num, earliest_target);
        }
        // check we haven't now passed the target time; if not we don't want to loop again
    } while ((earliest_target - (int64_t)ta_time_us_64(timer)) <= 0);
```

why does this not loop?

src/rp2_common/pico_time_adapter/include/pico/time_adapter.h
```diff
@@ -48,7 +48,8 @@ static inline void ta_set_timeout(alarm_pool_timer_t *timer, uint alarm_num, int
     // 2. If we just passed the target time, then time_til_target will be high, meaning we will
     //    likely not do the update, but this is OK since the caller who has the full 64 bits
     //    must check if the target time has passed when we return anyway to avoid races.
-    if (time_til_target < time_til_alarm) {
+    // 3. We should never leave here without an alarm being armed
+    if (time_til_target < time_til_alarm || (timer_hw_from_timer(timer)->armed & (1 << alarm_num)) == 0) {
```
kilograham, line 52: why is `time_til_target <= time_til_alarm` not sufficient?

peterharperuk: time_til_alarm is zero. time_til_target is the time the callback has asked for (e.g. 20000us). I considered just checking for time_til_alarm == 0, but it seems dangerous to me to assume that a timer value implies the timer is armed?

peterharperuk:
> why does this not loop?

earliest_target is correct and > the current time so the loop will exit.

We've just called ta_set_timeout with earliest_target, but that's no guarantee that the alarm was set to this time IF it thinks there's an earlier alarm. It will wrongly think there's an earlier alarm set if the current alarm time == current time.

peterharperuk:
> ok, this is really difficult to follow;

```
// Everything happens in alarm_pool_irq_handler
// ta_time_us_64=10000: IRQ goes off for the only alarm
static void alarm_pool_irq_handler(void) {
    <snip>
    int64_t earliest_target;
    do {
        <snip>
        if (earliest_index >= 0) {
            alarm_pool_entry_t *earliest_entry = &pool->entries[earliest_index];
            earliest_target = earliest_entry->target;
// ta_time_us_64=10000, earliest_target=10000: time to call the callback
            if (((int64_t)ta_time_us_64(timer) - earliest_target) >= 0) {
                <snip>
                int64_t delta;
                if (earliest_target >= 0) {
// Now we call the callback which executes quickly and returns the time it wants to be called next, e.g. 20000us
                    if (earliest_entry->callback == repeating_timer_marker) {
                        repeating_timer_t *rpt = (repeating_timer_t *)earliest_entry->user_data;
                        delta = rpt->callback(rpt) ? rpt->delay_us : 0;
                    } else {
                        alarm_id_t id = make_alarm_id(pool->ordered_head, earliest_entry->sequence);
                        delta = earliest_entry->callback(id, earliest_entry->user_data);
                    }
// delta=20000 (or -20000 for a repeating alarm, it doesn't matter)
                    <snip>
                    if (delta) {
                        int64_t next_time;
                        if (delta < 0) {
                            next_time = earliest_target - delta;
                        } else {
                            next_time = (int64_t) ta_time_us_64(timer) + delta;
                        }
// next_time=30000 the time we want the next alarm to go off
                        earliest_entry->target = next_time;
                        <snip>
                }
            }
        }
        <snip>
        earliest_target = earliest_entry->target;
        if (earliest_target != -1) { // cancelled alarm has target of -1
// earliest_target=30000, ta_time_us_64=10000: we've got to this point in <1us
// we call ta_set_timeout asking for an alarm at earliest_target (30000), this is the code expanded...
            uint32_t current = timer_time_us_32(timer_hw_from_timer(timer));
// current=10000: we've got to this point in <1us
            uint32_t time_til_target = (uint32_t) target - current;
// time_til_target=20000: what we asked for - we still haven't taken > 1us
            uint32_t time_til_alarm = timer_hw_from_timer(timer)->alarm[alarm_num] - current;
// time_til_alarm=0: This is the time for the alarm that's just gone off
// time_til_target [20000] < time_til_alarm [0] is FALSE so the alarm target is NOT set
            if (time_til_target < time_til_alarm) {
// We don't get here
                timer_hw_from_timer(timer)->alarm[alarm_num] = (uint32_t) target;
            }
        }
// earliest_target=30000
// ta_time_us_64=10010: Let's pretend that time has advanced a bit, it doesn't matter as earliest_target is still a long way in the future
    } while ((earliest_target - (int64_t)ta_time_us_64(timer)) <= 0);
// The IRQ exits but it hasn't actually set the alarm that was requested
    <snip>
}
```

kilograham:
OK, i agree this makes sense....

Also while looking, i think we should clear the arming in `ta_disable_irq_handler`

can you please add that to this PR

peterharperuk:
> Also while looking, i think we should clear the arming in ta_disable_irq_handler

Good spot. I've added that change.

kilograham (approved):

kilograham:
nice!

PR #2130, opened 2024-12-08
Send release power-down (0xAB) instruction, to revive flash, that might be in power-down state

### LINKED ISSUE #2111: Flash power-up in boot_stage2
User `gmx` suggested on the [forum](https://forums.raspberrypi.com/viewtopic.php?t=378659&start=25#p2272217), that powering flash down can further lower power consumption. And it works indeed - about 10 uA (typically). Using `powman` with `BOOT0..3` registers set to 0, as I understand it boots back from flash, especially when sleeping in `P1.7` with no SRAM retention, like in `powman` example [here](https://github.com/peterharperuk/pico-examples/tree/powman/powman). Powered down flash responds to `0xAB` instruction only, and any `boot2` doesn't seem to revive it that way.
So my request here is, that maybe it will be useful to have conditional (ex. using `SCRATCH0..7` register value) sending of an `0xAB` instruction in case, flash was powered down before.

matsobdev:
It fixes #2111 of RP2350, by adding `0xAB` instruction before communicating with flash device, ex. when restarting from flash by `powman`. Flash might be put in power-down state before and will only respond to `0xAB`.

### DIFF
--- a/src/rp2350/boot_stage2/boot2_generic_03h.S
+++ b/src/rp2350/boot_stage2/boot2_generic_03h.S
@@ -46,6 +46,7 @@
 #error "RX delay greater than maximum"
 #endif
 
+#define CMD_RELEASE_POWERDOWN 0xAB
 #define CMD_READ 0x03
 
 // ----------------------------------------------------------------------------
@@ -59,6 +60,15 @@
 // Setup timing parameters: short sequential-access cooldown, configured
 // CLKDIV and RXDELAY, and no constraints on CS max assertion, CS min
 // deassertion, or page boundary burst breaks.
+
+// Need to use direct serial mode to send SR commands. Choose a
+// conservative direct-mode divisor (5 MHz at 150 MHz clk_sys)
+// since the XIP-mode divisor may be unsafe without an RX delay.
+#define INIT_DIRECT_CSR (\
+    30 << QMI_DIRECT_CSR_CLKDIV_LSB | \
+    QMI_DIRECT_CSR_EN_BITS | \
+    QMI_DIRECT_CSR_AUTO_CS0N_BITS | \
+0)
 
 #define INIT_M0_TIMING (\
     1                      << QMI_M0_TIMING_COOLDOWN_LSB |\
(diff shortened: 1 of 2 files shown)


lurch:
Should this be gated behind a `#ifdef` setting, since I guess this will only be needed by executables that have explicitly powered-down the flash?

matsobdev:
Sounds like a right direction. Is there a naming scheme to follow?

lurch:
> Sounds like a right direction. Is there a naming scheme to follow?

Naming scheme is generally `PICO_SOME_THING` - see section 5.1 of https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf for a big list of examples :slightly_smiling_face:

src/rp2040/boot_stage2/boot2_at25sf128a.S
```diff
@@ -75,6 +75,9 @@
 // A better solution is to use a volatile SR write if your device supports it.
 #define PROGRAM_STATUS_REG
 
+#if PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN
+#define CMD_RELEASE_POWERDOWN 0xAB
+#endif
```
lurch, line 80: It's probably fine to just add this `#define` directly (to the end of the `#define CMD_...` lines below), without wrapping it in an `#if` guard?

matsobdev:
I've updated it, extended to RP2040, but now it is not picking up definition :D I'll work on it.

src/rp2040/boot_stage2/boot2_at25sf128a.S
matsobdev: I'd agree. I guess, it will be safe.

src/rp2040/boot_stage2/include/boot_stage2/config.h
```diff
@@ -11,6 +11,11 @@
 
 #include "pico.h"
 
+// PICO_CONFIG: PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN, Enable/disable releasing flash device from the power-down state during boot stage 2, type=bool, default=0, group=boot_stage2
```
lurch, line 14: Given that this is a boolean value, perhaps the "enable/disable" is implicit, and so "Release the flash device from power-down state during boot stage 2" would be an easier-to-read description?

matsobdev:
Now I know, why `PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN` was unreachable. Since there are two artefacts (if I wrap my head around it right way) built and boot stage 2 is the one of them, then it requires its own `target_compile_definitions`:
```
target_compile_definitions(bs2_default PRIVATE
	PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN=1
	)
target_compile_definitions(my_program PRIVATE
	PICO_EMBED_XIP_SETUP=1
	)
```
Maybe very beginning of chapter 5. of https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf requires then a word about `bs2_default`?

src/rp2040/boot_stage2/include/boot_stage2/config.h
matsobdev: I agree.

lurch:
It doesn't have much [documentation](https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf) yet, but `PICO_EMBED_XIP_SETUP` is also mentioned in a few other [issues](https://github.com/search?q=repo%3Araspberrypi%2Fpico-sdk+PICO_EMBED_XIP_SETUP&type=issues) and [PRs](https://github.com/search?q=repo%3Araspberrypi%2Fpico-sdk+PICO_EMBED_XIP_SETUP&type=pullrequests).

matsobdev:
Yes, `PICO_EMBED_XIP_SETUP` is a key to get boos stag 2 on RP2350 where it might benefit the most from that 10 uA, but I was thinking about:
```
target_compile_definitions(bs2_default PRIVATE
	PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN=1
	)
```
since `PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN` will be accessible only, when `target(ing)_compile_definitions` of `bs2_default`. Root `my_program`, for basicly user's code won't reach that and `PICO_BOOT_STAGE2_FLASH_RELEASE_POWERDOWN` is useless then.

PR #2137, opened 2024-12-13
Add pico_set_modified_binary_type function

will-v-pi:
This adds a function to create binaries with modified RAM/SCRATCH addresses

For example, to only use SRAM1 so you can power down SRAM0 in your binary you could use
```cmake
pico_set_modified_binary_type(<my_exe> flash RAM "0x20040000" "256k")
```

To create a XIP_SRAM only binary you could use
```cmake
pico_set_modified_binary_type(<my_exe> sram
    RAM "0x13ffc000" "12k"
    SCRATCH_X "0x13fff000" "2k"
    SCRATCH_Y "0x13fff800" "2k")
```

The existing `memmap_blocked_ram.ld` could be created using
```cmake
pico_set_modified_binary_type(<my_exe> flash RAM "0x21000000" "256k")
```

This adds some tests to kitchen_sink to ensure that these generated linker scripts match the default ones, to prevent them diverging

### DIFF
--- a/src/rp2_common/pico_crt0/rp2040/memmap_flash.template.ld
+++ b/src/rp2_common/pico_crt0/rp2040/memmap_flash.template.ld
@@ -1,0 +1,286 @@
+/* Based on GCC ARM embedded samples.
+   Defines the following symbols for use by code:
+    __exidx_start
+    __exidx_end
+    __etext
+    __data_start__
+    __preinit_array_start
+    __preinit_array_end
+    __init_array_start
+    __init_array_end
+    __fini_array_start
+    __fini_array_end
+    __data_end__
+    __bss_start__
+    __bss_end__
+    __end__
+    end
+    __HeapLimit
+    __StackLimit
+    __StackTop
+    __stack (== StackTop)
+*/
+
+MEMORY
+{
+    INCLUDE "pico_flash_region.ld"
+    RAM(rwx) : ORIGIN =  @RAM_ORIGIN@, LENGTH = @RAM_LENGTH@
+    SCRATCH_X(rwx) : ORIGIN = @SCRATCH_X_ORIGIN@, LENGTH = @SCRATCH_X_LENGTH@
+    SCRATCH_Y(rwx) : ORIGIN = @SCRATCH_Y_ORIGIN@, LENGTH = @SCRATCH_Y_LENGTH@
+}
+
+ENTRY(_entry_point)
+
+SECTIONS
+{
+    /* Second stage bootloader is prepended to the image. It must be 256 bytes big
+       and checksummed. It is usually built by the boot_stage2 target
+       in the Raspberry Pi Pico SDK
+    */
+
+    .flash_begin : {
+        __flash_binary_start = .;
+    } > FLASH
+
+    .boot2 : {
+        __boot2_start__ = .;
+        KEEP (*(.boot2))
+        __boot2_end__ = .;
+    } > FLASH
+
+    ASSERT(__boot2_end__ - __boot2_start__ == 256,
+        "ERROR: Pico second stage bootloader must be 256 bytes in size")
+
+    /* The second stage will always enter the image at the start of .text.
+       The debugger will use the ELF entry point, which is the _entry_point
+       symbol if present, otherwise defaults to start of .text.
+       This can be used to transfer control back to the bootrom on debugger
+       launches only, to perform proper flash setup.
+    */
+
+    .text : {
+        __logical_binary_start = .;
+        KEEP (*(.vectors))
+        KEEP (*(.binary_info_header))
+        __binary_info_header_end = .;
+        KEEP (*(.embedded_block))
+        __embedded_block_end = .;
+        KEEP (*(.reset))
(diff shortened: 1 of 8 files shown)


kilograham:
this is good; we need to think about whether this is best as modifying the linker script or injecting as per FLASH_SIZE (either way it should - and i think it is for FLASH_SIZE be modifiable via linker script)

lurch:
> To create a XIP_SRAM only binary you could use
> 
> ```cmake
> pico_set_modified_binary_type(<my_exe> no_flash
>     RAM "0x13ffc000" "12k"
>     SCRATCH_X "0x13fff000" "2k"
>     SCRATCH_Y "0x13fff800" "2k")
> ```

Perhaps it would be nice if you could use symbolic aliases here, rather than having to use hardcoded addresses? :shrug:

lurch:
Maybe this is a first step towards #1547 ?

will-v-pi:
> this is useful functionality; we need to think about whether this is best as modifying the linker script or injecting as per FLASH_SIZE (either way it should - and i think it is for FLASH_SIZE be modifiable via linker script)

I've thought it through a bit more - injecting like FLASH_SIZE wouldn't work, because you might need different injected bits for each binary in your project, whereas FLASH_SIZE is set project-wide as it's fixed for a given board

Instead, I've added separate template linker scripts `memmap_sram.template.ld` and `memmap_flash.template.ld`, which are copies of `memmap_no_flash.ld` and `memmap_default.ld` respectively with the RAM and SCRATCH regions set to be templated. I've also added tests at compile time to `kitchen_sink` to ensure these files stay in sync.

These template linker scripts could then be extended further to support #1547 - but I think that's probably best for a second PR, I'd like to just get this functionality in first so you can make SRAM1 or XIP_SRAM binaries

src/rp2_common/pico_crt0/rp2040/memmap_flash.template.ld
```diff
@@ -0,0 +1,286 @@
+/* Based on GCC ARM embedded samples.
+   Defines the following symbols for use by code:
```
lurch, line 2: There are a load of double-underscore identifiers (e.g. `__logical_binary_start`,  `__embedded_block_end`, `__binary_info_start`) which aren't included in this list. Should they be? :shrug: (I know almost nothing about linker scripts)

src/rp2_common/pico_crt0/rp2040/memmap_flash.template.ld
```diff
@@ -0,0 +1,286 @@
...
+        __logical_binary_start = .;
+        KEEP (*(.vectors))
+        KEEP (*(.binary_info_header))
+        __binary_info_header_end = .;
+        KEEP (*(.embedded_block))
+        __embedded_block_end = .;
+        KEEP (*(.reset))
+        /* TODO revisit this now memset/memcpy/float in ROM */
+        /* bit of a hack right now to exclude all floating point and time critical (e.g. memset, memcpy) code from
+         * FLASH ... we will include any thing excluded here in .data below by default */
```
lurch, line 71:
```suggestion
         * FLASH ... we will include anything excluded here in .data below by default */
```

src/rp2_common/pico_crt0/rp2040/memmap_flash.template.ld
will-v-pi: This is an identical copy of `memmap_default.ld`, so changing anything other than the memory regions is outside the scope of this PR - maybe worth pinging @kilograham about tidying up the linker scripts?

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -78,6 +78,99 @@ if (NOT TARGET pico_standard_link)
...
+            set(PICO_DEFAULT_SCRATCH_X_LENGTH 4k)
+            set(PICO_DEFAULT_SCRATCH_Y_ORIGIN 0x20081000)
+            set(PICO_DEFAULT_SCRATCH_Y_LENGTH 4k)
+        endif()
+
+        # Scripts that will be created by this function
+        set(LINKER_CMAKE_SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/memmap_${TARGET}.cmake")
+        set(LINKER_LD_SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/memmap_${TARGET}.ld")
+
+        # Configure memmap_${TARGET}.cmake script file to create memmap_${TARGET}.ld
```
lurch, line 122: Is it worth having this function checking that none of the LENGTH values are too big for the given chip? (i.e. I guess you could set a 300KB RAM_LENGTH for RP2350, but not for RP2040?)

will-v-pi: That check would only be useful on RP2040 to ensure the length is <264k, as on RP2350 you could be using PSRAM so can have a length of 16M - but that should be caught anyway because if you set a RAM region >256k on RP2040 but have the correct SCRATCH regions, then they'll overlap and throw a linker error. So I think I'll probably not add this check, and leave it to the linker

will-v-pi:
This has been superceded by other linker script changes

PR #2145, opened 2024-12-21
Added some Waveshare board support

waveshare:
Added some weveshare board support,PTAL.


lurch:
@waveshare Apologies for the delay in getting round to reviewing this.
My [check_board_header.py](https://github.com/raspberrypi/pico-sdk/blob/develop/tools/check_board_header.py) script doesn't report any problems with any of these new header files :+1:  

However if I compare the `waveshare_rp2350_one.h` in this PR to the one in #2174 , I can see that your PR says that RP2350-One has:
```
// no PICO_DEFAULT_WS2812_PIN
```
but the version in #2174 says:
```
// --- WS2812 ---
#define PICO_DEFAULT_WS2812_PIN 16
#endif
```
and https://www.waveshare.com/wiki/RP2350-One#Pinout_Definition also says that there's a WS2812 connected to `GP16` ? :thinking: 
Also, the version in this PR says
```
#define PICO_DEFAULT_SPI_RX_PIN 16
```
but your pinout diagram seems to show that GP16 isn't available on any of the user-accessible pins?

waveshare:
Sorry, this is our mistake, #2174 is right, we will fix it soon
GP16 is only used for WS2812

waveshare:
Hello,
After a long wait, we have corrected the corresponding pin definitions.Please take a look.

src/boards/include/boards/waveshare_rp2350_geek.h
```diff
@@ -0,0 +1,107 @@
...
+#endif
+#ifndef PICO_DEFAULT_SPI_RX_PIN
+#define PICO_DEFAULT_SPI_RX_PIN 20
+#endif
+#ifndef PICO_DEFAULT_SPI_CSN_PIN
+#define PICO_DEFAULT_SPI_CSN_PIN 17
+#endif
+
+// --- LCD ---
+#ifndef WAVESHARE_RP2040_TOUCH_LCD_SPI
```
lurch, line 65: This is the header-file for `waveshare_rp2350_geek.h` so it looks a bit weird that it has `#define`s mentioning `WAVESHARE_RP2040_...`. Was this deliberate or is it a copy'n'paste mistake?

waveshare: It was a copy'n'paste mistake , please forgive my carelessness.

lurch:
Thanks for the updates, my `check_board_header.py` now doesn't report any problems :+1: 

However I see that in various different files you've defined:
```
WAVESHARE_RP2040_TOUCH_LCD_SPI
WAVESHARE_RP2350_TOUCH_LCD_SPI
WAVESHARE_RP2350_LCD_SPI
```
(along with the associated pin-defines for each of those SPI interfaces).
You obviously know your own hardware much better than I do, but is there a reason for all these different names, or _might_ it makes sense to just name them all `WAVESHARE_LCD_SPI` ?

Similarly, perhaps it would make sense for both
```
WAVESHARE_RP2040_BAT_ADC_PIN
WAVESHARE_RP2350_BAT_ADC_PIN
```
to be named just `WAVESHARE_BAT_ADC_PIN` ?

These suggestions are entirely optional; if you disagree I'm happy to approve this PR as-is.

waveshare:
I think your suggestion is very good. Give me some time and I will solve it.

waveshare:
Please take a look.

src/boards/include/boards/waveshare_rp2350_touch_lcd_1.28.h
```diff
@@ -0,0 +1,112 @@
...
+#endif
+#ifndef WAVESHARE_LCD_CS_PIN
+#define WAVESHARE_LCD_CS_PIN 9
+#endif
+#ifndef WAVESHARE_LCD_SCLK_PIN
+#define WAVESHARE_LCD_SCLK_PIN 10
+#endif
+#ifndef WAVESHARE_LCD_TX_PIN
+#define WAVESHARE_TX_PIN 11
+#endif
```
lurch, line 79: Slight typo here - presumably `WAVESHARE_TX_PIN` was supposed to be `WAVESHARE_LCD_TX_PIN` ?

EngineerWill: please forgive my colleague
For some reasons, he had to use the command line editor, which he was not used to, which led to him often making mistakes.
hope you understand

lurch: No apologies needed!

lurch (commented):
Fantastic :+1: 

Thank you for taking my comments into account, and thank you **so** much for adding all these board headers! :heart:

lurch (approved):

PR #2156, opened 2025-01-02
Add a delay after rtc_set_datetime

### LINKED ISSUE #2148: pico_aon_timer time setting functions should add a delay on the RP2040
We have code running on RP2040 and RP2350 that uses `pico_aon_timer` as a RTC and found that these don't work correctly on the RP2040 when a call to `aon_timer_start` is followed quickly by a call to `aon_timer_get_time`. Since `pico_aon_timer` uses the hardware RTC on the RP2040 there should be a delay after setting and before reading it.

This could be fixed by adding a delay in `aon_timer_set_time_calendar` after the `rtc_set_datetime` call, or amending the SDK documentation that one should add a delay after time setting `pico_aon_timer` function calls on the RP2040.

peterharperuk:
If you call rtc_get_datetime immediately after rtc_set_datetime you get junk back. According to the datasheet "Writing to the RTC will take 2 clk_rtc clock periods to arrive". So add a delay after calling rtc_set_datetime in aon_timer_set_time_calendar.

Fixes #2148

### DIFF
--- a/src/rp2_common/pico_aon_timer/CMakeLists.txt
+++ b/src/rp2_common/pico_aon_timer/CMakeLists.txt
@@ -12,6 +12,7 @@ pico_mirrored_target_link_libraries(pico_aon_timer INTERFACE pico_util)
 
 if (TARGET hardware_rtc)
     pico_mirrored_target_link_libraries(pico_aon_timer INTERFACE hardware_rtc)
+    target_link_libraries(pico_aon_timer_headers INTERFACE pico_time_headers)
 endif()
 
 if (TARGET hardware_powman)
--- a/src/rp2_common/pico_aon_timer/aon_timer.c
+++ b/src/rp2_common/pico_aon_timer/aon_timer.c
@@ -13,6 +13,8 @@ static aon_timer_alarm_handler_t aon_timer_alarm_handler;
 #if HAS_RP2040_RTC
 #include "hardware/rtc.h"
 #include "pico/util/datetime.h"
+#include "pico/time.h"
+#include "hardware/clocks.h"
 
 #elif HAS_POWMAN_TIMER
 #include "hardware/powman.h"
@@ -56,6 +58,10 @@ bool aon_timer_set_time_calendar(const struct tm *tm) {
     datetime_t dt;
     tm_to_datetime(tm, &dt);
     rtc_set_datetime(&dt);
+
+    // Writing to the RTC will take 2 clk_rtc clock periods to arrive
+    uint rtc_freq = clock_get_hz(clk_rtc);
+    busy_wait_us(((1000000 + rtc_freq - 1) / rtc_freq) * 2);
     return true;
 #elif HAS_POWMAN_TIMER
     struct timespec ts;
--- a/test/pico_time_test/pico_time_test.c
+++ b/test/pico_time_test/pico_time_test.c
@@ -13,6 +13,11 @@
 #endif
 #include "pico/stdlib.h"
 #include "pico/test.h"
+
+#if LIB_PICO_AON_TIMER
+#include "pico/aon_timer.h"
+#endif
+
 // Include sys/types.h before inttypes.h to work around issue with
 // certain versions of GCC and newlib which causes omission of PRIi64
 #include <sys/types.h>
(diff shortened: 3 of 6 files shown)


test/pico_time_test/pico_time_test.c
```diff
@@ -364,3 +368,33 @@ int issue_2118_test(void) {
...
     return 0;
 }
+
+static int issue_2148_test(void) {
+#if HAS_RP2040_RTC
+    PICOTEST_START_SECTION("Issue #2148 defect - get time after rtc start");
+    struct tm tm = { 0 };
+    struct tm tm_check = { 0 };
+
+    tm.tm_sec = 55;
```
lurch, line 378: I know that this test will probably take _much_ less than 5 seconds to run, but given that the checking-code below is checking for `tm_sec` rollover but **isn't** checking for `tm_min` rollover, perhaps it makes sense to set the `tm_sec` to something like `15` rather than `55` ? :wink:

lurch:
ping @armandomontanez - any idea why the CMake build is passing but the Bazel build is failing?

armandomontanez:
Looks like you need to add `//src/rp2_common/pico_aon_timer` to [the `deps` of the test](https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/test/pico_time_test/BUILD.bazel#L13). You also might need to add `"//src/common/pico_time` to [the `deps` of pico_aon_timer](https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/pico_aon_timer/BUILD.bazel#L11) since it's pulling in a new header. If you hit another snag let me know and I'll cook up a diff to get it working.

test/pico_time_test/pico_time_test.c
peterharperuk: What about 42?

src/rp2_common/pico_aon_timer/aon_timer.c
```diff
@@ -56,6 +58,10 @@ bool aon_timer_set_time_calendar(const struct tm *tm) {
     datetime_t dt;
     tm_to_datetime(tm, &dt);
     rtc_set_datetime(&dt);
+
+    // Writing to the RTC will take 2 clk_rtc clock periods to arrive
+    uint rtc_freq = clock_get_hz(clk_rtc);
+    busy_wait_us((1000000 / rtc_freq) * 2);
```
kilograham, line 64: I presume you probably want `(((1000000 + rtc_freq - 1) / rtc_freq) * 2)` for rounding safety

kilograham (dismissed):

kilograham:
@peterharperuk can you fix merge conflicts and i'll re-review

peterharperuk:
Apologies. Merge conflicts fixed. And I squashed everything together.

kilograham:
Ah, not sure if this is fixed in different issue but you have added `#include "pico/aon_timer.h"` to `pico_time_test.c` which will break host build; it (and test which use it) should be guarded by `#if LIB_PICO_AON_TIMER` probably

`pico_aon_timer` now depends on `hardware_clocks` AND `pico_time`.. probably should be explicit about dependencies (note just `pico_time_headers` is likely fine. Note in the CMakeLists.txt I'd also make the dependency (and the includes) RP2040 only

peterharperuk:
I've tried to make the changes you suggested. You'll have to tell me how to test aon timer in a host build.

kilograham:
`cmake -DPICO_PLATFORM=host ..` then build everything

src/rp2_common/pico_aon_timer/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ pico_mirrored_target_link_libraries(pico_aon_timer INTERFACE pico_util)
 
 if (TARGET hardware_rtc)
     pico_mirrored_target_link_libraries(pico_aon_timer INTERFACE hardware_rtc)
```
kilograham, line 14: note i think you still need `hardware_clocks` here

peterharperuk: hardware_rtc depends on hardware_clocks? Do I need to repeat it here?

PR #2165, opened 2025-01-06
Provide incoming HCI buffer for BTstack

mringwal:
The current driver reserves a larger buffer for the BTstack Pre-Buffer in memory, but does not use it when receiving an HCI packet from CYW43 driver. As some BTstack events will use this buffer, this causes an out-of-bound write / crash with BTstack v1.6.2 or higher.

Note: the 4-byte CYW43 header implicitly provided a 3-byte pre-buffer, which worked for BTstack v1.6.1 and earlier. In BTstack v1.6.2, the pre-buffer was increased which causes this bug to unhide. 

This replaces https://github.com/raspberrypi/pico-sdk/pull/2157#issuecomment-2569614331

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
+++ b/src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
@@ -6,20 +6,34 @@
 
 #include "pico.h"
 #include "cyw43.h"
-#include "hci_transport.h"
+#include "btstack_config.h"
 #include "hci.h"
+#include "hci_transport.h"
 #include "pico/btstack_hci_transport_cyw43.h"
 #include "pico/btstack_chipset_cyw43.h"
 
+// cyw43_bluetooth_hci_write and cyw43_bluetooth_hci_read require a custom 4-byte packet header in front of the actual HCI packet
+// the HCI packet type is stored in the fourth byte of the packet header
+#define CYW43_PACKET_HEADER_SIZE 4
+
 // assert outgoing pre-buffer for cyw43 header is available
-#if !defined(HCI_OUTGOING_PRE_BUFFER_SIZE) || (HCI_OUTGOING_PRE_BUFFER_SIZE < 4)
-#error HCI_OUTGOING_PRE_BUFFER_SIZE not defined or smaller than 4. Please update btstack_config.h
+#if !defined(HCI_OUTGOING_PRE_BUFFER_SIZE) || (HCI_OUTGOING_PRE_BUFFER_SIZE < CYW43_PACKET_HEADER_SIZE)
+#error HCI_OUTGOING_PRE_BUFFER_SIZE not defined or smaller than 4 (CYW43_PACKET_HEADER_SIZE) bytes. Please update btstack_config.h
 #endif
 
 // assert outgoing packet fragments are word aligned
 #if !defined(HCI_ACL_CHUNK_SIZE_ALIGNMENT) || ((HCI_ACL_CHUNK_SIZE_ALIGNMENT & 3) != 0)
-#error HCI_ACL_CHUNK_SIZE_ALIGNMENT not defined or not a multiply of 4. Please update btstack_config.h
+#error HCI_ACL_CHUNK_SIZE_ALIGNMENT not defined or not a multiple of 4. Please update btstack_config.h
 #endif
+
+// ensure incoming pre-buffer for cyw43 header is available (defaults from btstack/src/hci.h)
+#if HCI_INCOMING_PRE_BUFFER_SIZE < CYW43_PACKET_HEADER_SIZE
+#undef HCI_INCOMING_PRE_BUFFER_SIZE
+#define HCI_INCOMING_PRE_BUFFER_SIZE CYW43_PACKET_HEADER_SIZE
+#endif
+
+// ensure buffer for cyw43_bluetooth_hci_read starts word aligned (word align pre buffer)
+#define HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED ((HCI_INCOMING_PRE_BUFFER_SIZE + 3) & ~3)
 
 #define BT_DEBUG_ENABLED 0
 #if BT_DEBUG_ENABLED
@@ -31,9 +45,12 @@
 // Callback when we have data
 static void (*hci_transport_cyw43_packet_handler)(uint8_t packet_type, uint8_t *packet, uint16_t size) = NULL;
 
-// Incoming packet buffer - cyw43 packet header (incl packet type) + incoming pre buffer + max(acl header + acl payload, event header + event data)
+// The incoming packet buffer consist of a pre-buffer and the actual HCI packet
+// For the call to cyw43_bluetooth_hci_read, the last 4 bytes (CY43_PACKET_HEADER_SIZE) of the pre-buffer is used for the CYW43 packet header
+// After that, only the actual HCI packet is forwarded to BTstack, which expects HCI_INCOMING_PACKET_BUFFER_SIZE of pre-buffer bytes for its own use.
 __attribute__((aligned(4)))
-static uint8_t hci_packet_with_pre_buffer[4 + HCI_INCOMING_PRE_BUFFER_SIZE + HCI_INCOMING_PACKET_BUFFER_SIZE ];
+static uint8_t hci_packet_with_pre_buffer[HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED + HCI_INCOMING_PACKET_BUFFER_SIZE ];
+static uint8_t * cyw43_receive_buffer = &hci_packet_with_pre_buffer[HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED - CYW43_PACKET_HEADER_SIZE];
 
 static btstack_data_source_t transport_data_source;
 static bool hci_transport_ready;
@@ -97,8 +114,8 @@ static int hci_transport_cyw43_can_send_now(uint8_t packet_type) {
 static int hci_transport_cyw43_send_packet(uint8_t packet_type, uint8_t *packet, int size) {
     // store packet type before actual data and increase size
     // This relies on HCI_OUTGOING_PRE_BUFFER_SIZE being set
-    uint8_t *buffer = &packet[-4];
-    uint32_t buffer_size = size + 4;
+    uint8_t *buffer = &packet[-CYW43_PACKET_HEADER_SIZE];
+    uint32_t buffer_size = size + CYW43_PACKET_HEADER_SIZE;
     buffer[3] = packet_type;
 
     CYW43_THREAD_ENTER
@@ -143,10 +160,10 @@ static void hci_transport_cyw43_process(void) {
     uint32_t loop_count = 0;
 #endif
     do {
-        int err = cyw43_bluetooth_hci_read(hci_packet_with_pre_buffer, sizeof(hci_packet_with_pre_buffer), &len);
+        int err = cyw43_bluetooth_hci_read(cyw43_receive_buffer, CYW43_PACKET_HEADER_SIZE + HCI_INCOMING_PACKET_BUFFER_SIZE , &len);
         BT_DEBUG("bt in len=%lu err=%d\n", len, err);
         if (err == 0 && len > 0) {
-            hci_transport_cyw43_packet_handler(hci_packet_with_pre_buffer[3], hci_packet_with_pre_buffer + 4, len - 4);
+            hci_transport_cyw43_packet_handler(cyw43_receive_buffer[3], &cyw43_receive_buffer[CYW43_PACKET_HEADER_SIZE], len - CYW43_PACKET_HEADER_SIZE);
             has_work = true;
         } else {
             has_work = false;


peterharperuk:
I changed the branch to develop

lurch:
I've got **no** idea why this was assigned to me (I have no BTstack knowledge), so assigning to Peter instead.

peterharperuk:
Oops - I should get this in. Will just run a quick test.

peterharperuk (approved):

peterharperuk:
@kilograham Can we merge this?

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -21,6 +21,9 @@
 #error HCI_ACL_CHUNK_SIZE_ALIGNMENT not defined or not a multiply of 4. Please update btstack_config.h
 #endif
 
+// assert buffer for cyw43_bluetooth_hci_read starts word aligned (word align pre buffer)
+#define HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED ((HCI_INCOMING_PRE_BUFFER_SIZE + 3) & 0xfffc)
```
kilograham, line 25: this is mildly confusing - why 0xfffc rather than ~3 - does this say something about this always being <64K?

kilograham:
Also i don't understand this at all - maybe a comment about what the various ranges of the buffer are used for (particularly the 4 vs the PREBUFFER - why are they not one and the same, and where alignment comes into play)

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
peterharperuk: I think ~3 would do the job. It's the cyw43 chip that needs the buffers to be aligned I believe.

mringwal: Yes, ~3 would be fine. The HCI Buffer matches the largest packet over the air, which is 1023. For reasons, BNEP (tethering) puts Ethernet frames (up to 1651 or something linke that) into an L2CAP packet that we then put into an large HCI buffer. So 1659 is sufficient for all use cases.

I was told that the cyw43 driver requires its read/write buffer to be word aligned.
 
I'll try to add some more comments on the data. The 4 bytes for the cyw43 specific header could be subsumed in the general HCI_INCOMING_PRE_BUFFER_SIZE, but that would increase the logic even more (but save a full 4 bytes :).

kilograham: yeah; i don't think the 4 bytes is critical, just wanted to reason as to whether it was correct ;-)

mringwal:
I have improved the documentation and merged the CYW43 packet header buffer with BTstack's pre-buffer, by asserting that HCI_INCOMING_PRE_BUFFER_SIZE is >= CYW43_PACKET_HEADER_SIZE (4).

It's better to read now. However, the build fails for me with "#error HCI_INCOMING_PRE_BUFFER_SIZE not defined or smaller than 4 (CYW43_PACKET_HEADER_SIZE) bytes. Please update btstack_config.h".

@peterharperuk Could you try to find out, where HCI_INCOMING_PRE_BUFFER_SIZE is set? 
I get the feeling that different btstack_config.h are used to compile btstack_hci_transport_cyw43.c and the rest of BTstack. Both need to see the same value of HCI_INCOMING_PRE_BUFFER_SIZE to avoid memory issues. E.g. the GATT Client has some requirements on the pre-buffer size, which needs to be provided by  btstack_hci_transport_cyw43.c. Feel free to ask for specific details.

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -6,21 +6,34 @@
...
 // assert outgoing pre-buffer for cyw43 header is available
-#if !defined(HCI_OUTGOING_PRE_BUFFER_SIZE) || (HCI_OUTGOING_PRE_BUFFER_SIZE < 4)
-#error HCI_OUTGOING_PRE_BUFFER_SIZE not defined or smaller than 4. Please update btstack_config.h
+#if !defined(HCI_OUTGOING_PRE_BUFFER_SIZE) || (HCI_OUTGOING_PRE_BUFFER_SIZE < CYW43_PACKET_HEADER_SIZE)
+#error HCI_OUTGOING_PRE_BUFFER_SIZE not defined or smaller than 4 (CYW43_PACKET_HEADER_SIZE) bytes. Please update btstack_config.h
 #endif
 
 // assert outgoing packet fragments are word aligned
 #if !defined(HCI_ACL_CHUNK_SIZE_ALIGNMENT) || ((HCI_ACL_CHUNK_SIZE_ALIGNMENT & 3) != 0)
 #error HCI_ACL_CHUNK_SIZE_ALIGNMENT not defined or not a multiply of 4. Please update btstack_config.h
```
lurch, line 26: Presumably the "not a multiply of 4" here was supposed to be "not a multiple of 4" ?

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -6,21 +6,34 @@
...
 #endif
 
+// assert incoming pre-buffer for cyw43 header is available
+#if !defined(HCI_INCOMING_PRE_BUFFER_SIZE) || (HCI_INCOMING_PRE_BUFFER_SIZE < CYW43_PACKET_HEADER_SIZE)
+#error HCI_INCOMING_PRE_BUFFER_SIZE not defined or smaller than 4 (CYW43_PACKET_HEADER_SIZE) bytes. Please update btstack_config.h
+#endif
+
+// assert buffer for cyw43_bluetooth_hci_read starts word aligned (word align pre buffer)
+#define HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED ((HCI_INCOMING_PRE_BUFFER_SIZE + 3) & ~3)
```
lurch, line 35: There's no "assert" here, perhaps "ensure" would be a more appropriate description?

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -31,9 +44,12 @@
...
 static void (*hci_transport_cyw43_packet_handler)(uint8_t packet_type, uint8_t *packet, uint16_t size) = NULL;
 
-// Incoming packet buffer - cyw43 packet header (incl packet type) + incoming pre buffer + max(acl header + acl payload, event header + event data)
+// The incoming packet buffer consist of a pre-buffer and the actual HCI packet
+// For the call to cyw43_bluetooth_hci_read, the last 4 bytes (CY43_PACKET_HEADER_SIZE) of the pre-buffer is used for the CYW43 packet header
+// After that, only the actual HCI packet is forwarded to BTstack, which expects HCI_INCOMING_PACKET_BUFFER_SIZE of pre-buffer bytes for its own use.
 __attribute__((aligned(4)))
-static uint8_t hci_packet_with_pre_buffer[4 + HCI_INCOMING_PRE_BUFFER_SIZE + HCI_INCOMING_PACKET_BUFFER_SIZE ];
+static uint8_t hci_packet_with_pre_buffer[HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED + HCI_INCOMING_PACKET_BUFFER_SIZE ];
+static uint8_t * cyw43_receive_buffer = &hci_packet_with_pre_buffer[HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED - CYW43_PACKET_HEADER_SIZE];
```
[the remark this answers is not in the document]
mringwal: Nope as HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED is >= HCI_INCOMING_PRE_BUFFER_SIZE, 
HCI_INCOMING_PRE_BUFFER_SIZE_ALIGNED is also >= CYW43_PACKET_HEADER_SIZE

peterharperuk:
> Could you try to find out, where HCI_INCOMING_PRE_BUFFER_SIZE is set?

As far as I can see, it's coming from btstack/src/hci.h

```
// BNEP may uncompress the IP Header by 16 bytes, GATT Client requires six additional bytes for long characteristic reads
// wih service_id + connection_id
#ifndef HCI_INCOMING_PRE_BUFFER_SIZE
#ifdef ENABLE_CLASSIC
#define HCI_INCOMING_PRE_BUFFER_SIZE (16 - HCI_ACL_HEADER_SIZE - 4)
#else
#define HCI_INCOMING_PRE_BUFFER_SIZE 6
#endif
#endif

```

mringwal:
@peterharperuk Does it compile for you? Also: where is support for Classic and/or LE defined? If Classic would be defined, then HCI_INCOMING_PRE_BUFFER_SIZE should be (16 - HCI_ACL_HEADER_SIZE - 4) = 8, which is larger than 4. I wasn't able to compile this as it seems to have picked up the value 2 from the #else branch - without Classic.

peterharperuk:
If you link to pico_btstack_classic it defines ENABLE_CLASSIC and if you link to pico_btstack_ble it defines ENABLE_BLE. Most of the examples are linking to both. https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_btstack/CMakeLists.txt#L107

Let me do some more testing. We haven't updated btstack to v1.6.2, we're still using v1.6.1. Should we update?

mringwal:
Interesting. 

So, how does btstack_hci_transport_cyw43.c  know which HCI_INCOMING_PRE_BUFFER_SIZE to use? Is it part of pico_btstack_classic as well as pico_btstack_ble.

If in doubt, we can just use a rather large HCI_INCOMING_PRE_BUFFER_SIZE, e.g. 16, for . btstack_hci_transport_cyw43. We don't plan on increasing the pre-buffer and just did a larger rewrite of the GATT Client logic, which is a big user of HCI_INCOMING_PRE_BUFFER_SIZE.

Yes, you should update to master (1.6.2 + one additional fix).

peterharperuk:
There's only on version of btstack_hci_transport_cyw43.c. So if you link to classic you get 8. If you ONLY link to ble you get 2. If you link to both you get 8? Is that wrong? I'm fine with us adding a set number to be safe. I thought your change was ok though?

mringwal:
btstack_hci_transport_cyw43.c should use whatever hci.h choses based on ENABLE_CLASSIC or not. 
If that holds true, everything is fine. 

(I don't understand the pico-sdk build system. I generally assume that a particular .c file is only compiled once into an .o file (by this having a fixed value for HCI_INCOMING_PRE_BUFFER_SIZE)  and then I don't understand how an example can link to pico_btstack_classic as well as pico_btstack_ble at the same time).

peterharperuk:
Everything is a cmake interface library in the sdk https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

mringwal:
@peterharperuk Thanks. Interesting, I need to revisit my concept of "library". Didn't know it could set defines for the main project (which totally makes sense if you start to thing about it).

Anyway. There's a new force-pushed version. It validates the size of the outgoing buffer and throws an error if needed, as the outgoing buffer is owned by hci.c. For the incoming buffer, which we need to provide, it increases it if needed. 

By this, pico-sdk should compile with the current (older) BTstack version as well as newer ones (v1.6.2 will require a larger pre-buffer for LE examples).

src/rp2_common/pico_cyw43_driver/btstack_hci_transport_cyw43.c
```diff
@@ -6,21 +6,35 @@
...
 // assert outgoing packet fragments are word aligned
 #if !defined(HCI_ACL_CHUNK_SIZE_ALIGNMENT) || ((HCI_ACL_CHUNK_SIZE_ALIGNMENT & 3) != 0)
-#error HCI_ACL_CHUNK_SIZE_ALIGNMENT not defined or not a multiply of 4. Please update btstack_config.h
+#error HCI_ACL_CHUNK_SIZE_ALIGNMENT not defined or not a multiple of 4. Please update btstack_config.h
 #endif
 
+// ensure incoming pre-buffer for cyw43 header is available (defaults from btstack/src/hci.h)
+#if HCI_INCOMING_PRE_BUFFER_SIZE < CYW43_PACKET_HEADER_SIZE
+#undef HCI_INCOMING_PRE_BUFFER_SIZE
+#define HCI_INCOMING_PRE_BUFFER_SIZE CYW43_PACKET_HEADER_SIZE
```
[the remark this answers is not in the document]
lurch: Also, should this code handle the case of `#if !defined(HCI_INCOMING_PRE_BUFFER_SIZE)` ? :man_shrugging:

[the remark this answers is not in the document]
peterharperuk: I think no on both counts as we expect this to redefined for ble. HCI_INCOMING_PRE_BUFFER_SIZE is defined by hci.h and always has been.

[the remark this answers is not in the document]
lurch: Ahh, fair enough! :+1:

peterharperuk:
Tested with 1.6.1 and 1.6.2 https://github.com/raspberrypi/pico-sdk/pull/2202

peterharperuk (approved):

PR #2169, opened 2025-01-07
Remove deprecated uses of FetchContent_Populate

### LINKED ISSUE #2124: CMake recommends FetchContent_MakeAvailable over FetchContent_Populate
According to [the CMake docs](https://cmake.org/cmake/help/git-stage/module/FetchContent.html#command:fetchcontent_populate), FetchContent_Populate is deprecated as of 3.30. When syncing the examples, the following is printed by CMake 3.30:

```
  Calling FetchContent_Populate(pico_sdk) is deprecated, call
  FetchContent_MakeAvailable(pico_sdk) instead.  Policy CMP0169 can be set to
  OLD to allow FetchContent_Populate(pico_sdk) to be called directly for now,
  but the ability to call it with declared details will be removed completely
  in a future version.
```

FetchContent_MakeAvailable was added in 3.14, so to maintain compatibility with 3.13, a version check would be needed.

will-v-pi:
This removes the deprecated uses of FetchContent_Populate (see #2124)

The SDK cannot just use MakeAvailable as the files shouldn't be added to the build, so instead this uses the full signature for FetchContent_Populate which has not been deprecated. The FetchContent_Declare call is kept in `pico_sdk_import.cmake` else it will re-clone the SDK for every new build directory when `PICO_SDK_FETCH_FROM_GIT_PATH` is set, rather than re-using the cloned SDK in `PICO_SDK_FETCH_FROM_GIT_PATH`.

Fixes #2124

### DIFF
--- a/external/pico_sdk_import.cmake
+++ b/external/pico_sdk_import.cmake
@@ -40,25 +40,26 @@ if (NOT PICO_SDK_PATH)
         if (PICO_SDK_FETCH_FROM_GIT_PATH)
             get_filename_component(FETCHCONTENT_BASE_DIR "${PICO_SDK_FETCH_FROM_GIT_PATH}" REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
         endif ()
-        # GIT_SUBMODULES_RECURSE was added in 3.17
-        if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.17.0")
-            FetchContent_Declare(
-                    pico_sdk
-                    GIT_REPOSITORY https://github.com/raspberrypi/pico-sdk
-                    GIT_TAG ${PICO_SDK_FETCH_FROM_GIT_TAG}
-                    GIT_SUBMODULES_RECURSE FALSE
-            )
-        else ()
-            FetchContent_Declare(
-                    pico_sdk
-                    GIT_REPOSITORY https://github.com/raspberrypi/pico-sdk
-                    GIT_TAG ${PICO_SDK_FETCH_FROM_GIT_TAG}
-            )
-        endif ()
+        FetchContent_Declare(
+                pico_sdk
+                GIT_REPOSITORY https://github.com/raspberrypi/pico-sdk
+                GIT_TAG ${PICO_SDK_FETCH_FROM_GIT_TAG}
+        )
 
         if (NOT pico_sdk)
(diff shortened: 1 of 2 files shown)


external/pico_sdk_import.cmake
```diff
@@ -40,25 +40,26 @@ if (NOT PICO_SDK_PATH)
         if (PICO_SDK_FETCH_FROM_GIT_PATH)
             get_filename_component(FETCHCONTENT_BASE_DIR "${PICO_SDK_FETCH_FROM_GIT_PATH}" REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
         endif ()
-        # GIT_SUBMODULES_RECURSE was added in 3.17
```
lurch, line 43: Does the new version of the code still need to do something different if we're using an older version of CMake which doesn't support `GIT_SUBMODULES_RECURSE` ?

lurch: > now that TinyUSB doesn't have any submodules

Good point, I hadn't considered that! But might be better to err on the side of caution though, just in case we add any other submodules in future which _do_ have submodules of their own? :thinking:

external/pico_sdk_import.cmake
```diff
@@ -40,25 +40,40 @@ if (NOT PICO_SDK_PATH)
...
+                FetchContent_Populate(
+                        pico_sdk
+                        QUIET
+                        GIT_REPOSITORY https://github.com/raspberrypi/pico-sdk
+                        GIT_TAG ${PICO_SDK_FETCH_FROM_GIT_TAG}
+                        GIT_SUBMODULES_RECURSE FALSE
+
+                        SOURCE_DIR ${FETCHCONTENT_BASE_DIR}/pico_sdk-src
+                        BINARY_DIR ${FETCHCONTENT_BASE_DIR}/pico_sdk-build
+                        SUBBUILD_DIR ${FETCHCONTENT_BASE_DIR}/pico_sdk-subbuild
```
lurch, line 62: I don't think we ever use this "subbuild" anywhere do we? Do we need to specify it?

will-v-pi: It's used by CMake when populating the other directories. We don't use this "build" anywhere though, but I think it's best to specify it to be safe, and to match the current behaviour.

kilograham (approved):

Issue #2186, opened 2025-01-16
Pico W hangs connecting to WiFi in station mode


stfnmllr:
Pico W hangs calling cyw43_arch_wifi_connect_timeout_ms.

SDK Version: 2.1.0
Mode: pico_cyw43_arch_lwip_poll

Example: https://github.com/stfnmllr/pico-w-issue

In the example cyw43_arch_wifi_connect_timeout_ms is used to establish a connection and independently from success reboots the Pico W. Executing the program eventually stops progressing and hangs in cyw43_arch_wifi_connect_timeout_ms.
Changing the program not to reboot but calling cyw43_arch_wifi_connect_timeout_ms in a loop shows the same effect. I just used the watchdog in the example as this is closer to my real use case.

In case the Pico W is connected to a Serial Monitor (via USB), sending data is going to resume work and prints
[CYW43] do_ioctl(i, j, z): timeout

I couldn't reproduce the issue in pico_cyw43_arch_lwip_threadsafe_background mode on Pico W. I also couldn't reproduce the issue on Pico2 W whether in pico_cyw43_arch_lwip_poll nor in pico_cyw43_arch_lwip_threadsafe_background mode.
With Pico W the issue is reproducible using different Pico W and / or connecting it via USB to different USB ports and different machines.

As the warning is raised in cyw43_ll.c / cyw43_do_ioctl I did increase the timeout in
  
#define CYW43_DO_IOCTL_WAIT cyw43_await_background_or_timeout_us(1000);

from 1000 to 10000 us with no effect, but after adding a sleep (see below) the example program runs 'stable' on the Pico W continuously since several days.

```
diff --git a/src/cyw43_ll.c b/src/cyw43_ll.c
index 033eec2..9ce10ae 100644
--- a/src/cyw43_ll.c
+++ b/src/cyw43_ll.c
@@ -1186,6 +1186,7 @@ static int cyw43_do_ioctl(cyw43_int_t *self, uint32_t kind, uint32_t cmd, size_t
     while (cyw43_hal_ticks_us() - start < CYW43_IOCTL_TIMEOUT_US) {
         size_t res_len;
         uint8_t *res_buf;
+        sleep_ms(10);
         ret = cyw43_ll_sdpcm_poll_device(self, &res_len, &res_buf);
         if (ret == CONTROL_HEADER) {
             #if CYW43_USE_STATS
```

peterharperuk:
[...]

PR #2190, opened 2025-01-17
Fix ta_wakes_up_on_or_before

peterharperuk:
Handle target in the past and running for > 2^32 us.

Fixes https://github.com/raspberrypi/pico-sdk/issues/2186

### DIFF
--- a/test/pico_time_test/pico_time_test.c
+++ b/test/pico_time_test/pico_time_test.c
@@ -76,6 +76,7 @@ int issue_195_test(void);
 int issue_1812_test(void);
 int issue_1953_test(void);
 int issue_2118_test(void);
+int issue_2186_test(void);
 
 int main() {
     setup_default_uart();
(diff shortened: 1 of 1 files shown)


kilograham:
Hmm; i think the bug lies elsewhere...

does this help?

```
static inline bool ta_wakes_up_on_or_before(alarm_pool_timer_t *timer, uint alarm_num, int64_t target) {
    int64_t current = (int64_t)timer_time_us_64(timer_hw_from_timer(timer));
    int64_t time_til_target = target - current;
    uint32_t time_til_alarm = timer_hw_from_timer(timer)->alarm[alarm_num] - (uint32_t)current;
    return time_til_alarm <= time_til_target;
}
```

(it also should fix another fairly glaring error if you'd been running for > 2^32 us

peterharperuk:
@kilograham Your fix seems to work. I think my analysis was wrong? It was actually seems stuck on WFE? I'll use your fix instead.

kilograham (approved):

PR #2207, opened 2025-01-23
fix pico_time_test.c compile on native

kilograham:
was broken in a recent commit


peterharperuk (approved):
I guess this was my fault

Issue #2152, opened 2024-12-29
CYW43 core not starting on Pico2 W


Memotech-Bill:
A call to **cyw43_arch_init_with_country** is failing.

My program is [PicoBB](https://github.com/Memotech-Bill/PicoBB). I am finding that for the Pico2 GUI build, which is using the ScanVideo library to produce VGA output on Core 1, and Core 0 running the BBC BASIC interpreter attempting to open a WiFi connection I am obtaining the following diagnostic messages:

````
[CYW43] core not in reset
[CYW43] core not in reset
[CYW43] core not in reset
[CYW43] core not in reset
[CYW43] core not up
[CYW43] HT not ready
````

and the WiFi is failing to connect.

Enabling all the diagnostics in the CYW43 driver, and comparing (diff -y) the failing case with a case in which Core 1 is idle gives the attached, where the failing case is on the left and the successful case is on the right.

[diff.txt](https://github.com/user-attachments/files/18270941/diff.txt)

Looking at the source file **pico-sdk/lib/cyw43-driver/src/cyw43_ll.c** I see the function [device_core_is_up()](https://github.com/georgerobotics/cyw43-driver/blob/cf924bb04c8984675ca0fc2178f082e404e048c3/src/cyw43_ll.c#L1260):

````
static void device_core_is_up(cyw43_int_t *self, int core_id) {
    uint32_t base = get_core_address(core_id);
    uint32_t reg = cyw43_read_backplane(self, base + AI_IOCTRL_OFFSET, 1);
    if ((reg & (SICF_FGC | SICF_CLOCK_EN)) != SICF_CLOCK_EN) {
        CYW43_WARN("core not up\n");
    }
    reg = cyw43_read_backplane(self, base + AI_RESETCTRL_OFFSET, 1);
    if (reg & AIRC_RESET) {
        CYW43_WARN("core not up\n");
    }
    // if we get here then the core is up
    CYW43_VDEBUG("core %d IS up\n", core_id);
}
````

This function seems to be testing for two possible failure conditions, but then does nothing about them other than issuing diagnostic messages. Surely it should be returning a true/false status so that subsequent code can handle any failures.

In the diagnostics attached, the following output can be seen:

````
cyw43_write_reg_u8 BACKPLANE_FUNCTION 0x1000b=0x1810		cyw43_write_reg_u8 BACKPLANE_FUNCTION 0x1000b=0x1810
[...]

Issue #2123, opened 2024-12-02
Bus Priority on the RP2350 vs RP2040


XZCE:
In my Pico W project (RP2040) I use the following code to make sure the 2nd CPU doesn't get stalled (it's serving the hardware bus of an old retro computer):

bus_ctrl_hw->priority = BUSCTRL_BUS_PRIORITY_PROC1_BITS;

I've obtained some Pico 2Ws (RP2350), compiled my project for them, and I've tracked down CPU0 being so sluggish that WIFI isn't working reliably (LED isn't reliable, connections aren't either) and SD Card reads are very slow.. to this statement.

My question is, is this no longer an appropriate thing to use on the RP2350 when expecting WIFI to work?  My thoughts are that the M0+ didn't have enough grunt to soak up resources to a point of failure, but maybe the M33's do?

Thanks for any advice.

peterharperuk:
I didn't even know this was a thing. @kilograham @Wren6991 Can you fiddle with bus priorities on RP2350 "to make sure the 2nd CPU doesn't get stalled"

peterharperuk:
@XZCE for my own curiosity, how could the 2nd CPU get stalled? What is it doing with the hardware bus of the retro computer?

kilograham:
I mean i can imagine that core 1 could stall out core 0, since the M33 can access memory every cycle, but you'd have to be doing a lot of memory access; I guess SIO is bus accessed on RP2350 which makes a difference in the example above, though I'm still surprised you get complete starvation.

Might be worth looking at the bus performance counters

XZCE:
@peterharperuk one of the CPUs stalls when they both go to access the same resource at the same time.  The timing on my Pico W project is very tight though, and I can't afford for CPU1 to stall (at least not by much) otherwise there will be memory errors (I have a forum thread full of memory error photos from other users - which are now all fixed).

To describe what it's doing on the bus of the retro computer, it's spinning on sio_hw->gpio_in looking for various signals to change, so I can change state in my finite state machine.  Each state then spins again the same way, looking for the same or maybe a different signal to change, to enter the next state.  So it is doing something similar to the example above, but it's not complete starvation - CPU0 moves along, just slowly.

[...]

PR #2209, opened 2025-01-24
Fix unreliable writes to cyw43.

### LINKED ISSUE #2206: cyw43: dma_channel_abort interacting weirdly with chained DMA
I've been chasing an issue with wireless in our MicroPython builds, and after confirming that it's related to our DMA/PIO drivers for LED matrix displays came up with a reproduction: https://github.com/micropython/micropython/issues/16627

This reproduction reliably shows that setting up chained DMA (yes it's in MicroPython but my original code was a C module) upsets the cyw43 driver. This only happens on Pico 2 W/RP2350 (possibly related to DMA errata E5?) and the same (original C) code has worked on RP2040 since the original Pico W launched.

@MichaelBell rolled with this and found that adding `asm volatile (" nop\n nop\n nop " );` after calls to `dma_channel_abort` in `cyw43_bus_pio_spi.c` would somehow fix this, coming up with the following patch:

```diff
diff --git a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
index bcc7284..1c6bd84 100644
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -260,6 +260,7 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
         pio_sm_exec(bus_data->pio, bus_data->pio_sm, pio_encode_jmp(bus_data->pio_offset));
         dma_channel_abort(bus_data->dma_out);
         dma_channel_abort(bus_data->dma_in);
+        asm volatile (" nop\n nop\n nop " );
 
         dma_channel_config out_config = dma_channel_get_default_config(bus_data->dma_out);
         channel_config_set_bswap(&out_config, true);
@@ -301,6 +302,7 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
         pio_sm_exec(bus_data->pio, bus_data->pio_sm, pio_encode_out(pio_y, 32));
         pio_sm_exec(bus_data->pio, bus_data->pio_sm, pio_encode_jmp(bus_data->pio_offset));
         dma_channel_abort(bus_data->dma_out);
+        asm volatile (" nop\n nop\n nop " );
 
         dma_channel_config out_config = dma_channel_get_default_config(bus_data->dma_out);
         channel_config_set_bswap(&out_config, true);
```

We don't have a good sense for exactly what is happening, and it seems to vary between compilers (breaks as expected with GCC 14.2.1), but it's clear that it's a uniquely RP2350 problem and that a mitigation should be possible in Pico SDK.

peterharperuk:
We use a pio and dma to write to the cyw43 chip using spi. Normally you write an address and then read the data from that address, so the pio program does does a write then read.

If you just want to write data in the case of uploading firmware we use the fdebug_tx_stall flag to work out if the pio has stalled waiting to read data which will never arrive.

The theory is that this flag will also get set if the bus is busy. So we mistakenly think a write to cyw43 has completed.

Add a check for the dma irq as well.

Fixes #2206

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -308,11 +308,13 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
 
         dma_channel_configure(bus_data->dma_out, &out_config, &bus_data->pio->txf[bus_data->pio_sm], tx, tx_length / 4, true);
 
+        pio_sm_set_enabled(bus_data->pio, bus_data->pio_sm, true);
+        dma_channel_wait_for_finish_blocking(bus_data->dma_out);
+
         uint32_t fdebug_tx_stall = 1u << (PIO_FDEBUG_TXSTALL_LSB + bus_data->pio_sm);
         bus_data->pio->fdebug = fdebug_tx_stall;
-        pio_sm_set_enabled(bus_data->pio, bus_data->pio_sm, true);
         while (!(bus_data->pio->fdebug & fdebug_tx_stall)) {
-            tight_loop_contents(); // todo timeout
+            tight_loop_contents();
         }
         __compiler_memory_barrier();
         pio_sm_set_enabled(bus_data->pio, bus_data->pio_sm, false);


src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -315,6 +316,12 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
             tight_loop_contents(); // todo timeout
         }
         __compiler_memory_barrier();
+
+        while (!(dma_hw->intr & 1u << bus_data->dma_out)) {
```
lurch, line 320: Will this have to re-calculate `1u << bus_data->dma_out` each time around the loop? Perhaps it's worth calculating it once, before the `while` loop?

peterharperuk, line 320: Should be able to wait for DMA to finish before waiting for stall?

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
```diff
@@ -231,7 +231,7 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
         return CYW43_FAIL_FAST_CHECK(-CYW43_EINVAL);
     }
 
-    bus_data_t *bus_data = (bus_data_t *)self->bus_data;
+    volatile bus_data_t *bus_data = (bus_data_t *)self->bus_data;
```
peterharperuk, line 234: Why is this needed?

lurch:
Fixes #2206 #2152 #2123

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
Wren6991: The compiler can hoist simple loop-invariant code like this, unless you have done something weird like mark `bus_data` as `volatile`.

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
lurch: In some of the linked issues, it's mentioned that different compiler optimisation levels and/or different compiler versions can cause the bug to appear / not appear.

> That's odd for a stack variable.

Perhaps it means that this bug goes a bit deeper?

peterharperuk: Yes, it fails without it. Something fishy is going on in release

peterharperuk: If I add __attribute((noinline)) to the function it's ok (volatile not needed).

Wren6991:
So if I understand the code correctly, these lines:

```c
        dma_channel_configure(bus_data->dma_out, &out_config, &bus_data->pio->txf[bus_data->pio_sm], tx, tx_length / 4, true);

        uint32_t fdebug_tx_stall = 1u << (PIO_FDEBUG_TXSTALL_LSB + bus_data->pio_sm);
        bus_data->pio->fdebug = fdebug_tx_stall;
        pio_sm_set_enabled(bus_data->pio, bus_data->pio_sm, true);
        while (!(bus_data->pio->fdebug & fdebug_tx_stall)) {
            tight_loop_contents(); // todo timeout
        }
```

...are attempting to block until the state machine has consumed all of the TX data, before deconfiguring the state machine. The problem is that this code *also* detects the case where the TX FIFO has bottomed out for reasons other than all of the data having been transferred. In particular, it immediately falls through if no DMA write has yet taken place between the `dma_channel_configure()` call and the write to `bus_data->pio->fdebug`, which is really no time at all. We probably got away with this on RP2040 because the processors were slower relative to the DMA.

If we think that is the failure mode, then the most direct fix would be this one:

```patch
diff --git a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
index bcc7284..a9f5970 100644
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -309,8 +309,10 @@ int cyw43_spi_transfer(cyw43_int_t *self, const uint8_t *tx, size_t tx_length, u
         dma_channel_configure(bus_data->dma_out, &out_config, &bus_data->pio->txf[bus_data->pio_sm], tx, tx_length / 4, true);
 
         uint32_t fdebug_tx_stall = 1u << (PIO_FDEBUG_TXSTALL_LSB + bus_data->pio_sm);
-        bus_data->pio->fdebug = fdebug_tx_stall;
         pio_sm_set_enabled(bus_data->pio, bus_data->pio_sm, true);
+
+        dma_channel_wait_for_finish_blocking(bus_data->dma_out);
+        bus_data->pio->fdebug = fdebug_tx_stall;
         while (!(bus_data->pio->fdebug & fdebug_tx_stall)) {
             tight_loop_contents(); // todo timeout
         }
```

So the steps are:

* wait for all data to be enqueued in TX FIFO
* **then** clear the TX stall flag, because we know it can no longer be caused by data not having been transferred
* then poll the TX stall flag to ensure the SM has completed

For me this also fixes the reproducer from #2123.

This also raises the question of why we are messing about with the DMA so much given this code is actually 100% blocking.

src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
Wren6991: I think this patch is still unsound because it checks things in the wrong order (see this comment https://github.com/raspberrypi/pico-sdk/pull/2209#issuecomment-2615466462).

It's likely that pushing the code around like this is adding delays and hiding symptoms.

peterharperuk:
Yes, I agree. dma_channel_wait_for_finish_blocking in the right place works. Still something funny going on with the bus data but I think that's a separate problem.

Wren6991:
> Still something funny going on with the bus data but I think that's a separate problem.

The fields you are accessing here are read-only after initialisation, so IMO this more likely to be a timing issue than anything to do with those particular accesses being volatile.

peterharperuk:
> Still something funny going on with the bus data but I think that's a separate problem

False alarm I think. The code could do with a refactor. But I should probably avoid doing that here?

lurch:
Minor nitpick, but given Luke's diagnosis above, should the "The theory is that this flag will also get set if the bus is busy. So we mistakenly think a write to cyw43 has completed." part be removed from the commit-message?

peterharperuk:
> if the PIO is divided significantly

You mean this flag could be set by another state machine?

kilograham:
i meant the clock, but either way, i think checking the address is better but i haven't dived deeply into this

peterharperuk:
Ah, ok. I'll try checking the address.

kilograham (approved):

matsobdev:
I'm just a simple man, don't much of that stuff, but will the last else if statement be reachable here: https://github.com/raspberrypi/pico-sdk/blob/3d746b3fa7ea7aa5ed80f66e47d54850e96da1f5/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c#L236-L328
given the fact that:
https://github.com/raspberrypi/pico-sdk/blob/3d746b3fa7ea7aa5ed80f66e47d54850e96da1f5/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c#L236
and:
https://github.com/raspberrypi/pico-sdk/blob/3d746b3fa7ea7aa5ed80f66e47d54850e96da1f5/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c#L322
Just asking, maybe it is suppose to be that way.

peterharperuk:
No it won't.

PR #2218, opened 2025-01-29
Declare platform.h functions with C linkage

### LINKED ISSUE #2217: Cannot call rp2040_chip_version/rp2350_chip_version from C++ code
The platform.h headers for both RP2040 and RP2350 need `extern "C"` declarations when included from C++ code or `rp2040_chip_version`
and `rp2350_chip_version` won't be found by the linker.

rhulme:
The platform.h headers for both RP2040 and RP2350 need `extern "C"` declarations when included from C++ code or `rp2040_chip_version` and `rp2350_chip_version` won't be found by the linker.

Fixes #2217

### DIFF
--- a/src/rp2040/pico_platform/include/pico/platform.h
+++ b/src/rp2040/pico_platform/include/pico/platform.h
@@ -77,6 +77,10 @@
 
 #ifndef __ASSEMBLER__
 
+#ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


peterharperuk:
Looks ok I guess

kilograham (approved):

PR #2221, opened 2025-02-01
Eat PATH argument to pico_generate_pio_header which was ignored before supporting multi-target on pico_generate_pio_header

kilograham:
see #2188


will-v-pi (approved):
LGTM

PR #2223, opened 2025-02-02
add multicore_lockout_victim_deinit. clear lockedout_flag for core1 during core1 reset

### LINKED ISSUE #2201: multicore_reset_core1 does not reset lockout_victim_initialized
@MichaelBell and I have been chasing a bug in MicroPython whereupon an attempt to write to the filesystem - after initialising core1 and performing a soft reset - would cause an indefinite lockup.

It looks like soft reset calls `multicore_reset_core1` which resets the core, but *does not* indicate that multicore lockout is no longer configured. This results in the following code  (in MicroPython's `rp2_flash.c`) blocking forever:

```
    if (multicore_lockout_victim_is_initialized(1 - get_core_num())) {
        multicore_lockout_start_blocking();
    }
```

I think `multicore_reset_core1` should include `lockout_victim_initialized[1] = false`.

*However* this runs counter to the comment:

https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/pico_multicore/multicore.c#L26-L32

Could we at least get a `multicore_lockout_victim_deinit`?

Or, y'know, just unbreak this even if it might be broken in other cases? 😭

kilograham:
fixes #2201

### DIFF
--- a/src/rp2_common/pico_multicore/include/pico/multicore.h
+++ b/src/rp2_common/pico_multicore/include/pico/multicore.h
@@ -452,7 +452,7 @@ void multicore_lockout_victim_init(void);
 /*! \brief Stop the current core being able to be a "victim" of lockout (i.e. forced to pause in a known state by the other core)
  *  \ingroup multicore_lockout
  *
- * This code unhooks the intercore FIFO IRQ, and the FIFO mayt be used for any other purpose after this.
+ * This code unhooks the intercore FIFO IRQ, and the FIFO may be used for any other purpose after this.
  */
 void multicore_lockout_victim_deinit(void);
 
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_multicore/include/pico/multicore.h
```diff
@@ -449,6 +449,13 @@ static inline uint multicore_doorbell_irq_num(uint doorbell_num) {
  */
 void multicore_lockout_victim_init(void);
 
+/*! \brief Stop the current core being able to be a "victim" of lockout (i.e. forced to pause in a known state by the other core)
+ *  \ingroup multicore_lockout
+ *
+ * This code unhooks the intercore FIFO IRQ, and the FIFO mayt be used for any other purpose after this.
```
peterharperuk, line 455: np mayt = may

peterharperuk (dismissed):

Issue #2166, opened 2025-01-07
pioasm: python output for RP2350 mov pindirs incorrect


magy00:
`word()` is prepended with part of the `mov()` that it replaces:

```
$ cat test.pio
.program test
.pio_version 1
_:  mov pindirs, null
jmp _
```
```python
$ pioasm -o python test.pio
# -------------------------------------------------- #
# This file is autogenerated by pioasm; do not edit! #
# -------------------------------------------------- #

import rp2
from machine import Pin
# ---- #
# test #
# ---- #

@rp2.asm_pio()
def test():
    wrap_target()
    label("0")
    mov(, null)             word(41059)                           # 0
    jmp("0")                              # 1
    wrap()
```

Ideally, this could look something like:
```python
import rp2
from machine import Pin
# ---- #
# test #
# ---- #

@rp2.asm_pio()
def test():
    wrap_target()
    label(0)
    word(0xa063)  # mov(pindirs, null)  # 0
    jmp(0)                              # 1
    wrap()
```

lurch:
ping @dpgeorge  Just out of curiosity, does the [PIO support](https://docs.micropython.org/en/latest/rp2/tutorial/pio.html) in MicroPython include the new instructions / options added for RP2350? (which I guess would be "nicer" than just emitting `word()` functions?)

dpgeorge:
> does the [PIO support](https://docs.micropython.org/en/latest/rp2/tutorial/pio.html) in MicroPython include the new instructions / options added for RP2350?

No, we don't support the new RP2350 PIO options yet.

lurch:
I suspect that we need to keep the string-quoting around labels? See the example at https://docs.micropython.org/en/latest/rp2/tutorial/pio.html#an-example

lurch:
Fixed by #2224 ?

kilograham:
correct

Issue #2100, opened 2024-11-26
Wrong exception message in pio_assembler.cpp


hlide:
https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/tools/pioasm/pio_assembler.cpp#L66

Should be 65536, not 65546.

lurch:
Haha whoops. Feel free to create a PR against the `develop` branch.

kilograham:
merged into develop

Issue #2204, opened 2025-01-22
pioasm: incomplete wait jmppin disassembly/micropython code generation


magy00:
```
.program test
.pio_version 1
wait 0 jmppin
```
results in
```
    0x2060, //  0: wait   0
```

Maybe something like this (not tested), and trimmed trailing whitespace and `~` instead of `!` could also be nice:

```diff
diff --git tools/pioasm/pio_disassembler.cpp tools/pioasm/pio_disassembler.cpp
index 2a34796..f62e7b3 100644
--- tools/pioasm/pio_disassembler.cpp
+++ tools/pioasm/pio_disassembler.cpp
@@ -56,7 +56,9 @@ std::string disassemble(uint inst, uint sideset_bits_including_opt, bool sideset
                     if (arg2 & 0x1cu) {
                         invalid = true;
                     } else if (arg2) {
-                        guts = "jmppin " + std::to_string(arg2 & 3u);
+                        guts = "jmppin + " + std::to_string(arg2 & 3u);
+                    } else {
+                        guts = "jmppin";
                     }
                     break;
             }
@@ -131,7 +133,7 @@ std::string disassemble(uint inst, uint sideset_bits_including_opt, bool sideset
                 op("mov");
                 std::string guts = dest + ", ";
                 if (operation == 1) {
-                    guts += "!";
+                    guts += "~";
                 } else if (operation == 2) {
                     guts += "::";
                 }
@@ -193,6 +195,7 @@ std::string disassemble(uint inst, uint sideset_bits_including_opt, bool sideset
     }
     delay &= ((1u << (5 - sideset_bits_including_opt)) - 1u);
     ss << std::left << std::setw(4) << (delay ? ("[" + std::to_string(delay) + "]") : "");
-    return ss.str();
+    auto str = ss.str();
+    return str.substr(0, str.find_last_not_of(" "));
 }
```

lurch:
I guess this is vaguely similar to #2166

lurch:
Fixed by #2224 ?

kilograham:
yes, so closing - we still produce `word(x)` since MP does not have support yet

PR #2224, opened 2025-02-02
small pioasm improvements

kilograham:
- fix disassembly of `wait jmppin`
- fix incorrect error message
- make python output emit `word(x)` for all unsupported instructions


kilograham:
somewhat fixes #2166  and #2204 ... we now emit `word(x)` which is better than garbage. can't do much more until MP defines the format

peterharperuk (approved):
Looks ok to me

kilograham:
fixes #2100

PR #2225, opened 2025-02-02
clarify clock_configure and make it return the correct value achieved for bad clock inputs

### LINKED ISSUE #2189: Bug with clock division for CLK_SYS
This almost certainly a hardware bug, not an SDK bug. 
I don't know where to submit hardware bugs.
I am testing with SDK 2.0.0. 

The field CLK_SYS_DIV_FRAC is not used  when CLK_SYS_DIV_INT is 1.
It is used for other values of CLK_SYS_DIV_INT.
So, it is impossible to divide the input clock by 1.5 but it is possible to divide by 2.5.

I work in micropython.
I made changes to the interpreter in C and also have micropython code I used for my testing.
If needed, I could reproduce the problem soley in C, but that might take awhile.

I am working to reduce power consumption for my application which runs on a PICO W.
Amongst other things, I want to disable pll_sys and clocking everything from pll_usb.
From my testing the cyw43 drivers don't like CPU frequency less than 64 MHZ.
I need to set pll_usb to a value that is an integer multiple of 48 MHZ.
I set pll_usb to run at 96 MHZ dividing by 2 for clk_usb, and clk_adc.
I want clk_sys to run at 64 MHZ the lowest frequency the cyw43 drivers tolerate.
I need to divide pll_usb by 1.5 to get 64 MHZ.

I configured pll_usb and clk_sys as follows:

```
  pll_init(pll_usb, 64*XOSC_HZ, 4, 2);
  clock_configure(clk_sys,
		  CLOCKS_CLK_SYS_CTRL_AUXSRC_VALUE_CLKSRC_PLL_USB,
		  96*1000*1000,
		  64*1000*1000);
```
The clock_configure() work correctly. When I dump the clk_sys registers I get:
```
clk_sys:
   CTRL(0x00000021) src=AUXSRC(1) auxsrc=PLL_USB(1)
   DIV(0x00000180) int=0x000001(1) frac=0x80(128)
   SELECTED value=0x2
```

I used the FC0 frequency counter to measure the clock frequencies and I get:
```
src:    PLL_SYS  1 freq:       0.000 Khz
src:    PLL_USB  2 freq:   96000.000 Khz
src:    CLK_SYS  9 freq:   96000.000 Khz
```

Note CLK_SYS is 96 MHZ same as PLL_USB.

To further understand the problem I tried other values in CLK_SYS_DIV.
I tried CLK_SYS_DIV of 0x1c0 (divide by 1.75), 0x1f0(divide by 1.88 approximately) and the frequency counter still reports 96 MHZ for CLK_SYS.

I then tried 0x280(divide by 2.5) and it worked.
```
clk_sys:
   CTRL(0x00000021) src=AUXSRC(1) auxsrc=PLL_USB(1)
   DIV(0x00000280) int=0x000002(2) frac=0x80(128)
   SELECTED value=0x2
```

```
src:    PLL_SYS  1 freq:       0.000 Khz
src:    PLL_USB  2 freq:   96000.000 Khz
src:    CLK_SYS  9 freq:   38400.000 Khz
```

kilograham:
fixes #2189

### DIFF
--- a/src/rp2_common/hardware_clocks/include/hardware/clocks.h
+++ b/src/rp2_common/hardware_clocks/include/hardware/clocks.h
@@ -262,10 +262,17 @@ extern "C" {
 
 typedef clock_num_t clock_handle_t;
 
-/*! \brief Configure the specified clock
+/*! \brief Configure the specified clock with automatic clock divisor setup
  *  \ingroup hardware_clocks
  *
+ * This method allows both the src_frequency of the input clock source AND the desired
+ * frequency to be specified, and will set the clock divider to achieve the exact or higher frequency
+ * achievable, with the maximum being the src_freq.
+ *
+ * Note: That the clock hardware only support divisors of exactly 1 or 2.0->65535.0
+ *
  * See the tables in the description for details on the possible values for clock sources.
+ *
  *
  * \param clock The clock to configure
  * \param src The main clock source, can be 0.
@@ -275,7 +282,7 @@ typedef clock_num_t clock_handle_t;
  */
 bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32_t src_freq, uint32_t freq);
 
-/*! \brief Configure the specified clock to use the undividded input source
+/*! \brief Configure the specified clock to use the undivided input source
  *  \ingroup hardware_clocks
  *
  * See the tables in the description for details on the possible values for clock sources.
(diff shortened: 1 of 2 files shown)


peterharperuk (approved):

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -287,7 +294,7 @@ bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32
  */
 void clock_configure_undivided(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32_t src_freq);
 
-/*! \brief Configure the specified clock to use the undividded input source
+/*! \brief Configure the specified clock to use the undivided input source
```
lurch, line 297: Hmmm, same 'brief' description for both `clock_configure_undivided` and `clock_configure_int_divider` - is that correct, or was it a copy'n'paste error?

kilograham:
Actually it turns out the RP2350 datasheet is incorrect and 1.x IS supported on RP2350… the datasheet will be updated and I’ll make a new PR to correct the function documentation

PR #2229, opened 2025-02-03
pico_flash: Support FreeRTOS Static Allocation

### LINKED ISSUE #2120: FreeRTOS: pico_flash/flash.c calls xTaskCreateAffinitySet which is not available using static memory allocation
pico_flash/flash.c uses xTaskCreateAffinitySet, which is only available in FreeRTOS if confiugred to support dynamic allocation.
https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/pico_flash/flash.c#L148

It's defined here:
https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/974351fe4aec19779f5be774007810c6ab36a8ab/tasks.c#L1758

If one was trying to use pure static FreeRTOS
```
#define configSUPPORT_DYNAMIC_ALLOCATION    0
#define configKERNEL_PROVIDED_STATIC_MEMORY     1
```
you'll get the following compilation warning:
pico_flash/flash.c:148:23: warning: implicit declaration of function 'xTaskCreateAffinitySet'; did you mean 'vTaskCoreAffinitySet'? [-Wimplicit-function-declaration]

For static allocation, it looks like you'd first have to create the static task then use ```vTaskCoreAffinitySet```.

Perhaps it doesn't need to change (as pico sdk does other dynamic allocation, I believe), so maybe just outputting a warning during compilation  that says "pico sdk requires configSUPPORT_DYNAMIC_ALLOCATION to be 1" would be enough

SConaway:
Fixes #2120

Allows `pico_flash` to use FreeRTOS static allocation as currently static-only builds fail without this change.

### DIFF
--- a/src/rp2_common/pico_flash/flash.c
+++ b/src/rp2_common/pico_flash/flash.c
@@ -147,7 +147,7 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         TaskHandle_t task_handle;
 
         // when FreeRTOS dynamic allocation is disabled (configSUPPORT_DYNAMIC_ALLOCATION == 0), the following instruction fails
-#ifndef configSUPPORT_DYNAMIC_ALLOCATION
+#if configSUPPORT_DYNAMIC_ALLOCATION
         if (pdPASS != xTaskCreateAffinitySet(flash_lockout_task, "flash lockout", configMINIMAL_STACK_SIZE, (void *)core_num, 0, 1u << (core_num ^ 1), &task_handle)) {
 #else
         static StackType_t flash_lockout_stack[configMINIMAL_STACK_SIZE];
--- a/src/rp2_common/pico_flash/flash.c
+++ b/src/rp2_common/pico_flash/flash.c
@@ -146,16 +146,14 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         // create at low priority on other core
         TaskHandle_t task_handle;
 
-        // when FreeRTOS dynamic allocation is disabled (configSUPPORT_DYNAMIC_ALLOCATION == 0), the following instruction fails
+        // when FreeRTOS dynamic allocation is disabled (configSUPPORT_DYNAMIC_ALLOCATION == 0), the following fails
 #if configSUPPORT_DYNAMIC_ALLOCATION
         if (pdPASS != xTaskCreateAffinitySet(flash_lockout_task, "flash lockout", configMINIMAL_STACK_SIZE, (void *)core_num, 0, 1u << (core_num ^ 1), &task_handle)) {
 #else
         static StackType_t flash_lockout_stack[configMINIMAL_STACK_SIZE];
         static StaticTask_t flash_lockout_task_tcb;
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_flash/flash.c
```diff
@@ -145,7 +145,18 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         uint core_num = get_core_num();
         // create at low priority on other core
         TaskHandle_t task_handle;
+
+        // when FreeRTOS dynamic allocation is disabled (configSUPPORT_DYNAMIC_ALLOCATION == 0), the following instruction fails
+#ifndef configSUPPORT_DYNAMIC_ALLOCATION
```
kilograham, line 150: should be `if !configSUPPORT_DYNAMIC_ALLOCATION`

kilograham: it was also backwards

kilograham:
Seems like this might be a nicer alternative

```
    TaskHandle_t xTaskCreateStaticAffinitySet( TaskFunction_t pxTaskCode,
                                               const char * const pcName, /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
                                               const uint32_t ulStackDepth,
                                               void * const pvParameters,
                                               UBaseType_t uxPriority,
                                               StackType_t * const puxStackBuffer,
                                               StaticTask_t * const pxTaskBuffer,
                                               UBaseType_t uxCoreAffinityMask ) PRIVILEGED_FUNCTION;
```

src/rp2_common/pico_flash/flash.c
SConaway: How did I miss that? good catch!

SConaway:
One thing I want to get your thoughts on (and sorry for the couple days of delay here): would it be better to default to static allocation and only switch to dynamic if static allocation isn't enabled?

SConaway:
> Seems like this might be a nicer alternative
> 
> ```
>     TaskHandle_t xTaskCreateStaticAffinitySet( TaskFunction_t pxTaskCode,
>                                                const char * const pcName, /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
>                                                const uint32_t ulStackDepth,
>                                                void * const pvParameters,
>                                                UBaseType_t uxPriority,
>                                                StackType_t * const puxStackBuffer,
>                                                StaticTask_t * const pxTaskBuffer,
>                                                UBaseType_t uxCoreAffinityMask ) PRIVILEGED_FUNCTION;
> ```

For some reason I skipped that — I can switch to it and let you know if it compiles. Unfortunately I don't really have any way to make sure that things still work besides compiling.

kilograham:
compiling is better than me as I don't have a config with this setup, so would have to make one

kilograham (approved):
OK, so this isn't ideal as it will likely crash if `flash_safe_execute()` is called from both cores concurrently in this case.. I am approving anyway, because

a) This does compile and likely works correctly when not doing that
b) That use case is probably unlikely
c) Arguably the existing implementation is not very robust in that regard, so I will open a separate issue

PR #2230, opened 2025-02-03
add a bunch of verbosity around shared vtables

### LINKED ISSUE #2164: irq_set_exclusive_handler description is misleading regarding cores
The description in the pico-sdk documentation describes the `irq_set_exclusive_handler` function as:
> Set an exclusive interrupt handler for an interrupt on the executing core.

It was clarified to me on the forums that the current SDK only uses a single interrupt vector table (VTABLE) for both cores, so as a result this function _effectively_ sets the exclusive interrupt handler for _**both**_ cores.  Until/unless the SDK is updated to support separate VTABLEs per core (https://github.com/raspberrypi/pico-sdk/issues/2158), IMO this function description is a bit misleading as written.

kilograham:
fixes #2164

### DIFF
--- a/src/rp2_common/hardware_irq/include/hardware/irq.h
+++ b/src/rp2_common/hardware_irq/include/hardware/irq.h
@@ -286,6 +286,33 @@ void irq_set_mask_n_enabled(uint n, uint32_t mask, bool enabled);
  * This method will assert if there is already any sort of interrupt handler installed
  * for the specified irq number.
  *
+ * NOTE: By default, the SDK uses a single shared vector table per core, and the current installed
+ * IRQ handlers are effectively a linked list starting a vector table entry for a particular IRQ numbers.
+ * Therefore, this method (when using the same vector table for both cores) sets the same interrupt handler
+ * for both cores.
+ *
+ * On RP2040 this was never really a cause of any confusion, because it rarely made sense to enable
+ * the same interrupt number in the NVIC on both cores (see \ref irq_set_enabled()), because the interrupt
+ * would then fire on both cores, and the interrupt handlers would race.
+ *
+ * The problem *does* exist however when dealing with interrupts which are independent on the two cores.
+ *
+ * This includes:
+ *
+ * * the core local "spare" IRQs
+ * * on RP2350 the SIO FIFO IRQ which is now the same irq number for both cores (vs RP2040 where it was two)
+ *
+ * In the cases where you want to enable the same IRQ on both cores, and both cores are sharing the same vector
+ * table, you should install the IRQ handler once - it will be used on both cores - and check the core
+ * number (via \ref get_core_num()) on each core.
+ *
+ * NOTE: It is not thread safe to add/remove/handle IRQs for the same irq number in the same vector table
+ * from both cores concurrently.
+ *
+ * NOTE: The SDK defines a PICO_VTABLE_PER_CORE variable indicating whether the two vector tables are separate,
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -286,6 +286,33 @@ void irq_set_mask_n_enabled(uint n, uint32_t mask, bool enabled);
  * This method will assert if there is already any sort of interrupt handler installed
  * for the specified irq number.
  *
+ * NOTE: By default, the SDK uses a single shared vector table per core, and the current installed
+ * IRQ handlers are effectively a linked list starting a vector table entry for a particular IRQ numbers.
```
peterharperuk, line 290: "for a particular IRQ numbers" should lose the s? "for a particular IRQ number"

src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -316,6 +343,33 @@ irq_handler_t irq_get_exclusive_handler(uint num);
  * the (total across all IRQs on both cores) maximum (configurable via PICO_MAX_SHARED_IRQ_HANDLERS) number of shared handlers
  * would be exceeded.
  *
+ * NOTE: By default, the SDK uses a single shared vector table per core, and the current installed
+ * IRQ handlers are effectively a linked list starting a vector table entry for a particular IRQ numbers.
```
peterharperuk, line 347: same comment as above

peterharperuk (approved):

src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -286,6 +286,33 @@ void irq_set_mask_n_enabled(uint n, uint32_t mask, bool enabled);
  * This method will assert if there is already any sort of interrupt handler installed
  * for the specified irq number.
  *
+ * NOTE: By default, the SDK uses a single shared vector table per core, and the current installed
```
lurch, line 289: "single shared... per core" seems a bit contradictory? Would "a single shared vector table for both cores" make more sense?
Also, "current installed" should probably be "currently installed" ?

src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -286,6 +286,33 @@ void irq_set_mask_n_enabled(uint n, uint32_t mask, bool enabled);
...
+ * * on RP2350 the SIO FIFO IRQ which is now the same irq number for both cores (vs RP2040 where it was two)
+ *
+ * In the cases where you want to enable the same IRQ on both cores, and both cores are sharing the same vector
+ * table, you should install the IRQ handler once - it will be used on both cores - and check the core
+ * number (via \ref get_core_num()) on each core.
+ *
+ * NOTE: It is not thread safe to add/remove/handle IRQs for the same irq number in the same vector table
+ * from both cores concurrently.
+ *
+ * NOTE: The SDK defines a PICO_VTABLE_PER_CORE variable indicating whether the two vector tables are separate,
```
lurch, line 312: 'variable' -> 'setting' ?

src/rp2_common/hardware_irq/include/hardware/irq.h
kilograham: pushed directly to develop

src/rp2_common/hardware_irq/include/hardware/irq.h
kilograham: pushed directly to develop

Issue #2107, opened 2024-11-27
SDK 2.1.0: clang 19.1.1 picolibc_interface compile error


geurtv:
pico-sdk 2.1.0 compiles fine with clang 18.1.3 ([LVM-ET-Arm-18.1.3](https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/releases/tag/release-18.1.3)) but with clang 19.1.1 ([LLVM-ET-Arm-19.1.1](https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/releases/tag/release-19.1.1)) I get the following compile error:

```
pico-sdk/2.1.0/src/rp2_common/pico_clib_interface/picolibc_interface.c:102:83: error: use of undeclared identifier 'CLOCKS_PER_SEC'
  102 |     tms->tms_utime = (clock_t)(to_us_since_boot(get_absolute_time()) / (1000000 / CLOCKS_PER_SEC));
      |                                                                                   ^
```

I don't know what a correct fix is. For example, if I add the following code to `picolibc_interface.c`:

```c
#include <time.h>
```

It compiles and `CLOCKS_PER_SEC == 1000000`.

clang 18's `time.h` includes `machine/time.h`, clang 19's does not. So if instead I add:

```c
#include <machine/time.h>
#include <time.h>
```

It again compiles, but now `CLOCKS_PER_SEC == 100` (same as clang 18).

lurch:
I've not used clang myself, but is https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_clib_interface/include/llvm_libc/sys/times.h#L12 relevant here?

geurtv:
I assume `llvm_libc/sys/times.h` will only be included for `clang+llvm_libc` (and not `clang+picolibc`), but I guess it does indicate the value for `CLOCKS_PER_SEC` is expected to be `100`. So that leaves the question what the correct include is to get `CLOCKS_PER_SEC == 100` in picolibc_interface.c.

theofabilous:
Related commit: d6e3fa06f197a8641ccd4fec4fad2d64f00e33cd
I'm almost certain the correct include is `<machine/time.h>` followed by `<time.h>`. In both LLVM and GCC embedded toolchains, `machine/time.h` contains the following definition:
```c
// #if __arm__ ...
#define _CLOCKS_PER_SEC_ 100
```
[...]

PR #2231, opened 2025-02-03
Clang related fixes

kilograham:
* Fix compilation on llvm 19 with picolibc
* FIx disassembly of Arm floating point instructions on RP2350


kilograham:
Note @will-v-pi i did a quick smoke test, but just worth checking if this trips up (or simplifies) any of the pioasm dissembly (dcp_ seemed fine)

kilograham:
fixes #2107

PR #2238, opened 2025-02-06
Fix semicolon in makefsdata.py

HugoTro:
Simple semicolon that breaks the file otherwise (it used to break my project but it doesn't anymore for some reason)
There is no associated issue for this, but no semi colon should be at the end of an instruction line in Python.


peterharperuk (approved):

lurch:
Easy mistake to make, when you're writing Python code to generate C code :laughing:

lurch (approved):

Issue #2071, opened 2024-11-20
Add explicit license and copyright headers to external/pico_sdk_import.cmake


P33M:
This came up in a review of debugprobe - as this file gets copied around a lot, and potentially into other projects which may have difference licencing arrangements, this file needs to be specifically attributed to the original authors.

lurch:
According to https://github.com/raspberrypi/pico-sdk/commits/develop/external/pico_sdk_import.cmake those authors are @kilograham , @willeccles and @ahmetkizilay .

kilograham:
@willeccles @ahmetkizilay are you fine for this to be (c) Raspberry Pi only, or do you want your names!

willeccles:
Just Raspberry Pi is fine, thanks.

kilograham:
merged into develop

PR #2239, opened 2025-02-06
Add license to pico_sdk_import.cmake

kilograham:
This file is copied elsewhere, so makes sense to have the license embedded

### DIFF
--- a/external/pico_sdk_import.cmake
+++ b/external/pico_sdk_import.cmake
@@ -3,6 +3,28 @@
 # This can be dropped into an external project to help locate this SDK
 # It should be include()ed prior to project()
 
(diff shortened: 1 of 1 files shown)


peterharperuk (approved):

P33M:
Can the text be compressed with a SPDX short-form licence identifier, or is the boilerplate customised?

kilograham:
it is not customized... I wasn't sure if we needed the long form if the LICENSE.TXT isn't present (which it wouldn't be) - if not, lets shorten it

P33M:
Actually I think the complete text has to be included - the notice must be reproduced in derivative works.

kilograham:
fixes #2071

PR #2241, opened 2025-02-06
Fix Windows CI

will-v-pi:
Don't install CMake, as it comes pre-installed on GitHub runners. We already rely on other pre-installed software like Python, so this should be fine.

This fixes the CMake version incompatibilities that keep appearing with the Windows CI build.


will-v-pi:
Tested successfully in this manual actions run https://github.com/raspberrypi/pico-sdk/actions/runs/13183560292

kilograham (approved):

lurch:
LOL :wink:  #2050

PR #2244, opened 2025-02-07
Add clock_configure_mhz function

will-v-pi:
This adds a `clock_configure_mhz` function, which trades off accuracy compared to `clock_configure` (only gets to +/- 1MHz) for reduced code size as it doesn't need 64 bit division.

It takes the source and target frequencies in MHz as 16-bit numbers, to ensure they fit within the 32-bit division.

This is useful for example if you want to calibrate the ROSC using an XOSC and then run off the ROSC, in which case you don't need a very accurate frequency as the ROSC frequency isn't accurate.


kilograham:
As per new code in #2225 we should catch cases where the divider between 1 and 2.

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -282,6 +282,22 @@ typedef clock_num_t clock_handle_t;
  */
 bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32_t src_freq, uint32_t freq);
 
+/*! \brief Configure the specified clock to +/- 1MHz
+ *  \ingroup hardware_clocks
+ *
+ * This function differs from clock_configure in that it does not configure the clocks as accurately,
+ * but therefore doesn't need to bring in 64-bit division functions, reducing the code size.
```
kilograham, line 289: reducing code size if 64-bit division is not otherwise used by the application

lurch:
> It takes the source and target frequencies in MHz as 16-bit numbers

I guess that means you can't overclock the Pico to 66 GHz? :rofl:

lurch:
See also #2255

will-v-pi:
> > It takes the source and target frequencies in MHz as 16-bit numbers
> 
> I guess that means you can't overclock the Pico to 66 GHz? 🤣

That's better than the normal clock_configure, which only lets you overclock to 4.3GHz 🤣

will-v-pi:
@kilograham do we want this function, or should I close this PR?

kilograham:
no, it is good; i didn't look in detail at the accuracy; does it at least get you near to the right answer, vs 1Mhz away ;-)

will-v-pi:
> no, it is good; i didn't look in detail at the accuracy; does it at least get you near to the right answer, vs 1Mhz away ;-)

If the divider is an integer (ie `(src_freq << CLOCKS_CLK_GPOUT0_DIV_INT_LSB) / freq` is an integer) then it'll always get the exact frequency, and from having a play I can't find any frequency combinations where that isn't an integer, so I think it'll generally always be correct? If there is a case where that divider is a non-integer then it's rounded down, so you'll get a slightly higher frequency than requested

kilograham (approved):

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -318,6 +318,31 @@ typedef clock_num_t clock_handle_t;
  */
 bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32_t src_freq, uint32_t freq);
 
+/*! \brief Configure the specified clock to +/- 1MHz
```
kilograham: Actually could we clarify what this means per our discussion

kilograham (changes requested):

src/rp2_common/hardware_clocks/include/hardware/clocks.h
will-v-pi: I think this is correct

```suggestion
/*! \brief Configure the specified clock with 1MHz accuracy
```

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -318,6 +318,31 @@ typedef clock_num_t clock_handle_t;
...
+ * Note: The RP2040 clock hardware only supports divisors of exactly 1.0 or 2.0->16777216.0 in steps of 1/256
+ * \endif
+ * 
+ * See the tables in the description for details on the possible values for clock sources.
+ *
+ * \param clock The clock to configure
+ * \param src The main clock source, can be 0.
+ * \param auxsrc The auxiliary clock source, which depends on which clock is being set. Can be 0
+ * \param src_freq_mhz Frequency of the input clock source in MHz
+ * \param freq_mhz Requested frequency in MHz
```
lurch, line 342: Looks like the doxygen is missing a `\return` ?

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -123,6 +123,26 @@ bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32
...
+#if PICO_RP2040
+    // on RP2040 only clock divider of 1, or  >= 2 are supported
+    if (div < (2u << CLOCKS_CLK_GPOUT0_DIV_INT_LSB)) {
+        div = (1u << CLOCKS_CLK_GPOUT0_DIV_INT_LSB);
+    }
+#endif
+    uint32_t actual_freq = (uint32_t) ((((uint32_t) src_freq_mhz) << CLOCKS_CLK_GPOUT0_DIV_INT_LSB) / div) * MHZ;
+
+    clock_configure_internal(clock, src, auxsrc, actual_freq, div);
+    // Store the configured frequency
```
lurch, line 142: What does this comment mean? (and is it referring to the line below or the line above the comment?)

will-v-pi: I’m not entirely sure (it was copied from the function above) - it’s related to the line above it as that’s what stores the configured frequency

src/rp2_common/hardware_clocks/include/hardware/clocks.h
will-v-pi:
```suggestion
 * \param freq_mhz Requested frequency in MHz
 * \return true if the clock is updated, false if freq > src_freq
```

src/rp2_common/hardware_clocks/clocks.c
lurch: Perhaps it'd be less confusing if the comment was moved _above_ the line it's referring to, as is done conventionally? :slightly_smiling_face:

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -318,6 +318,32 @@ typedef clock_num_t clock_handle_t;
...
+ * \return true if the clock is updated, false if freq > src_freq
+ */
+bool clock_configure_mhz(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint16_t src_freq_mhz, uint16_t freq_mhz);
```
kilograham: hard for me to eyeball whether clock_configure_mhz(N) always configures the same clock as clock_config(N*1000000)... (and returns the same value) ... based on that we may want to rename to clock_configure_approx_mhz() and/or update comments (the current comment implies it doesn't, but with the current name i;d expect it to - we probide _hz, _khz_ _mhz functions elsewhere which do exactly the same thing)

kilograham (changes requested):

src/rp2_common/hardware_clocks/include/hardware/clocks.h
will-v-pi: I can confirm for N in range 1-150, they both configure to the same frequency

kilograham (approved):

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -123,6 +123,26 @@ bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32
     return true;
 }
 
+bool clock_configure_mhz(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint16_t src_freq_mhz, uint16_t freq_mhz) {
```
kilograham: actually i think i'd rather these be uint32_t - 0-65535 is not a magic range, and at first glance we don't rely on these being 16 bit anywhere.

If we want a sensible maximum we'd enforce that with an IF

(note we generally don't use smaller fields unless it explicitly matches some hardware field)

kilograham (changes requested):

src/rp2_common/hardware_clocks/clocks.c
will-v-pi: It has to fit in 16-bits for this line, as it gets shifted by 16:
```
uint32_t div = (uint32_t)((((uint32_t) src_freq_mhz) << CLOCKS_CLK_GPOUT0_DIV_INT_LSB) / freq_mhz);
```

lurch: `CLOCKS_CLK_GPOUT0_DIV_INT_LSB` is 16 for RP2350, but only 8 for RP2040.
(clock dividers are 24.8 on RP2040 but 16.16 on RP2350)

src/rp2_common/hardware_clocks/include/hardware/clocks.h
lurch: Is it also true for frequencies up to 200MHz, which is the "official" max frequency for RP2040?

src/rp2_common/hardware_clocks/clocks.c
will-v-pi: Ok, I've changed it to take a `uint32_t`, and added an assert and return if it doesn't fit in `CLOCKS_CLK_GPOUT0_DIV_INT_BITS`

src/rp2_common/hardware_clocks/clocks.c
```diff
@@ -123,12 +123,17 @@ bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32
...
 }
 
-bool clock_configure_mhz(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint16_t src_freq_mhz, uint16_t freq_mhz) {
+bool clock_configure_mhz(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32_t src_freq_mhz, uint32_t freq_mhz) {
     assert(src_freq_mhz >= freq_mhz);
 
     if (freq_mhz > src_freq_mhz)
         return false;
 
+    assert(src_freq_mhz <= (uint32_t)(CLOCKS_CLK_GPOUT0_DIV_INT_BITS >> CLOCKS_CLK_GPOUT0_DIV_INT_LSB));
```
lurch, line 132: Should this be using `invalid_params_if` rather than a manual `assert` ?

src/rp2_common/hardware_clocks/clocks.c
will-v-pi: I've put it in the correct place now, which is inside the `clock_configure_internal` function

src/rp2_common/hardware_clocks/clocks.c
will-v-pi: `clock_configure` which this was based on also uses `assert` and a return to validate parameters, so I'll stick with that in case there is a reason for it - if we want to change that behaviour that could come in a separate PR (eg to `invalid_params_if_and_return`)

src/rp2_common/hardware_clocks/include/hardware/clocks.h
will-v-pi: Yes, true for those too

src/rp2_common/hardware_clocks/clocks.c
lurch: Okay :+1:

kilograham (approved):

PR #2255, opened 2025-02-11
update previous fix as 1.0 < clkdiv < 2.0 IS supported on RP2350

### LINKED ISSUE #2189: Bug with clock division for CLK_SYS

kilograham:
fixes #2189 more accurately

### DIFF
--- a/src/rp2_common/hardware_clocks/include/hardware/clocks.h
+++ b/src/rp2_common/hardware_clocks/include/hardware/clocks.h
@@ -273,7 +273,7 @@ typedef clock_num_t clock_handle_t;
  * Note: The RP2350 clock hardware supports divisors from 1.0->65536.0 in steps of 1/65536
  *
  * \endif
- * \if rp2030_specific
+ * \if rp2040_specific
  * Note: The RP2040 clock hardware only supports divisors of exactly 1.0 or 2.0->16777216.0 in steps of 1/256
  * \endif
  *
--- a/src/rp2_common/hardware_clocks/include/hardware/clocks.h
+++ b/src/rp2_common/hardware_clocks/include/hardware/clocks.h
@@ -269,16 +269,22 @@ typedef clock_num_t clock_handle_t;
  * frequency to be specified, and will set the clock divider to achieve the exact or higher frequency
  * achievable, with the maximum being the src_freq.
  *
- * Note: That the clock hardware only support divisors of exactly 1 or 2.0->65535.0
+ * \if rp2350_specific
+ * Note: That the RP2350 clock hardware supports divisors from 1.0->65535.0 in steps of 1/65536
+ *
+ * \endif
+ * \if rp2030_specific
+ * Note: That the RP2040 clock hardware only support divisors of exactly 1.0 or 2.0->65535.0 in steps of 1/256
+ * \endif
  *
  * See the tables in the description for details on the possible values for clock sources.
- *
  *
  * \param clock The clock to configure
  * \param src The main clock source, can be 0.
  * \param auxsrc The auxiliary clock source, which depends on which clock is being set. Can be 0
  * \param src_freq Frequency of the input clock source
  * \param freq Requested frequency
+ * \return true if the clock is updated, false if freq > src_freq
  */
 bool clock_configure(clock_handle_t clock, uint32_t src, uint32_t auxsrc, uint32_t src_freq, uint32_t freq);
 
--- a/src/rp2_common/hardware_clocks/include/hardware/clocks.h
+++ b/src/rp2_common/hardware_clocks/include/hardware/clocks.h
@@ -270,11 +270,11 @@ typedef clock_num_t clock_handle_t;
  * achievable, with the maximum being the src_freq.
  *
  * \if rp2350_specific
- * Note: That the RP2350 clock hardware supports divisors from 1.0->65535.0 in steps of 1/65536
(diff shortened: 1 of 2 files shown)


src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -269,16 +269,22 @@ typedef clock_num_t clock_handle_t;
  * frequency to be specified, and will set the clock divider to achieve the exact or higher frequency
  * achievable, with the maximum being the src_freq.
  *
- * Note: That the clock hardware only support divisors of exactly 1 or 2.0->65535.0
+ * \if rp2350_specific
+ * Note: That the RP2350 clock hardware supports divisors from 1.0->65535.0 in steps of 1/65536
```
lurch, line 273: Looking at e.g. https://www.raspberrypi.com/documentation/pico-sdk/hardware.html it seems that "Note: That the..." should just be "Note: The..." ?

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -269,16 +269,22 @@ typedef clock_num_t clock_handle_t;
...
  *
- * See the tables in the description for details on the possible values for clock sources.
+ * \endif
+ * \if rp2030_specific
+ * Note: That the RP2040 clock hardware only support divisors of exactly 1.0 or 2.0->65535.0 in steps of 1/256
```
lurch, line 277: Section 2.15.3.3. in https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf says "A fully featured divider divides by 1 or a fractional number in the range 2.0 to 2^24-0.01" ?

lurch: Also, "only support" -> "only supports"

src/rp2_common/hardware_clocks/include/hardware/clocks.h
```diff
@@ -269,16 +269,22 @@ typedef clock_num_t clock_handle_t;
  * frequency to be specified, and will set the clock divider to achieve the exact or higher frequency
  * achievable, with the maximum being the src_freq.
  *
- * Note: That the clock hardware only support divisors of exactly 1 or 2.0->65535.0
+ * \if rp2350_specific
+ * Note: That the RP2350 clock hardware supports divisors from 1.0->65535.0 in steps of 1/65536
  *
- * See the tables in the description for details on the possible values for clock sources.
+ * \endif
+ * \if rp2030_specific
```
will-v-pi, line 276:
```suggestion
 * \if rp2040_specific
```

will-v-pi (commented):
LGTM, other than @lurch’s comments and a small typo

peterharperuk (approved):
after staring at this for some time. I think it makes sense.

PR #2264, opened 2025-02-13
Add additional Bazel config tinyUSB and pico_stdio_usb

mbrase:
This change the makes the following defines setable in Bazel:
- `LIB_TINYUSB_HOST`
- `LIB_TINYUSB_DEVICE`
- `PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE`
- `PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE`

This also adds `//bazel/config:PICO_TINYUSB_CONFIG`, which allows overriding the "tusb_config.h" header used to build tinyUSB. This might be used to add additional USB interfaces while preserving the functionality from pico_stdio_usb.

### DIFF
--- a/bazel/config/BUILD.bazel
+++ b/bazel/config/BUILD.bazel
@@ -186,6 +186,12 @@ int_flag(
     build_setting_default = 0,
 )
 
+# PICO_BAZEL_CONFIG: PICO_TINYUSB_CONFIG, [Bazel only] The library that provides TinyUSB config header (e.g. tusb_config.h), default=//src/rp2_common/pico_stdio_usb:tusb_config, group=build
+label_flag(
+    name = "PICO_TINYUSB_CONFIG",
+    build_setting_default = "//src/rp2_common/pico_stdio_usb:tusb_config",
+)
+
 # PICO_BAZEL_CONFIG: PICO_TINYUSB_LIB, [Bazel only] The library that provides TinyUSB, default=@tinyusb//:tinyusb, group=build
 label_flag(
     name = "PICO_TINYUSB_LIB",
--- a/src/rp2_common/pico_stdio_usb/BUILD.bazel
+++ b/src/rp2_common/pico_stdio_usb/BUILD.bazel
@@ -24,6 +24,9 @@ cc_library(
     hdrs = ["include/tusb_config.h"],
     includes = ["include"],
     target_compatible_with = compatible_with_rp2(),
+    deps = [
+        ":pico_stdio_usb_headers",
+    ],
 )
 
 pico_sdk_define(
@@ -43,10 +46,6 @@ cc_library(
     hdrs = ["include/pico/stdio_usb.h"],
     includes = ["include"],
     target_compatible_with = compatible_with_rp2(),
-    visibility = [
-        ":__pkg__",
-        "//src/rp2_common/tinyusb:__pkg__",
-    ],
     deps = [
         ":LIB_PICO_STDIO_USB",
         ":PICO_STDIO_USB_CONNECT_WAIT_TIMEOUT_MS",
--- a/src/rp2_common/tinyusb/BUILD.bazel
+++ b/src/rp2_common/tinyusb/BUILD.bazel
@@ -15,6 +15,7 @@ cc_library(
     includes = ["include"],
     target_compatible_with = compatible_with_rp2(),
     deps = [
+        "//bazel/config:PICO_TINYUSB_CONFIG",
         "//src/common/pico_binary_info",
         "//src/common/pico_stdlib_headers",
         "//src/common/pico_sync",
@@ -30,7 +31,6 @@ cc_library(
         "//src/rp2_common/pico_stdio_semihosting",
         "//src/rp2_common/pico_stdio_uart",
         "//src/rp2_common/pico_stdio_usb:pico_stdio_usb_headers",
-        "//src/rp2_common/pico_stdio_usb:tusb_config",
         "//src/rp2_common/pico_unique_id",
     ],
 )


kilograham:
cc @armandomontanez

bazel/util/sdk_define.bzl
```diff
@@ -11,7 +11,9 @@ def _pico_sdk_define_impl(ctx):
         val = 1 if val else 0
     cc_ctx = cc_common.create_compilation_context(
         defines = depset(
-            direct = ["{}={}".format(ctx.attr.define_name, val)],
+            direct = [
+                "{}={}".format(ctx.attr.define_name, val)
+            ] if val or not ctx.attr.undef_if_falsy else [],
```
kilograham, line 16: why `falsy` not `false`

bazel/config/BUILD.bazel
```diff
@@ -186,6 +186,36 @@ int_flag(
     build_setting_default = 0,
 )
 
+# PICO_BAZEL_CONFIG: PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE, Enable/disable resetting into BOOTSEL mode if the host sets the baud rate to a magic value (PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE), type=bool, default=1, group=pico_stdio_usb
+bool_flag(
+    name = "PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE",
```
armandomontanez, line 191: All the flags that currently live here are flags that affect more than just a `#define`: all of them impact the build graph in some way. For scalability/maintainability reasons, I'd prefer we look at these defines wholistically if we're going to start the ones that just add a `-DPICO_FOO=BAR` to the compilation.

Right now, the expected ways to configure the defines are:

* Set `--copt=-DPICO_FOO=BAR` in your platform's flags.
* Set the `PICO_CONFIG_*` flags to point to your customized Pico platform libraries.
* Add `-D` to your toolchain flags.

I'll take an action item to update the Bazel-specific docs to clarify this.

src/rp2_common/pico_stdio_usb/BUILD.bazel
```diff
@@ -38,18 +43,44 @@ pico_sdk_define(
...
+
+pico_sdk_define(
+    name = "PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE",
+    define_name = "PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE",
+    from_flag = "//bazel/config:PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE",
+)
+
+pico_sdk_define(
+    name = "LIB_TINYUSB_HOST",
+    define_name = "LIB_TINYUSB_HOST",
```
armandomontanez, line 60: All the `LIB_` defines are intended to come from a library, signaling whether or not a library is linked in. In this case, enabling `LIB_TINYUSB_HOST` would set the define but not link in any of the USB host sources. Let's not add this unless we're properly hooking up host support.

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -48,6 +48,8 @@
 #ifndef PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE
 #if !defined(LIB_TINYUSB_HOST) && !defined(LIB_TINYUSB_DEVICE)
```
armandomontanez, line 49: These are the only instances in the Pico SDK that use `#if defined(LIB_*)`, I suspect they were intended to be `#if LIB_*`, which would address the need for `undef_if_falsy` altogether.

armandomontanez: Well, in this case it'd be
```diff
- #if !defined(LIB_TINYUSB_HOST) && !defined(LIB_TINYUSB_DEVICE)
+ #if !LIB_TINYUSB_HOST && !LIB_TINYUSB_DEVICE
```

bazel/util/sdk_define.bzl
mbrase: I had named it `falsy` because it is similar to a "falsy" value in python (i.e. "False", "0" and empty string would all trigger this). But based on @armandomontanez's feedback, I don't think I'm going to need this.

mbrase:
I chatted with @armandomontanez on Friday and I'm going to upload a narrower patch based on his feedback.

bazel/config/BUILD.bazel
mbrase: I've uploaded a new patch that does not include the new bool_flag() targets

src/rp2_common/pico_stdio_usb/BUILD.bazel
mbrase: I've uploaded a new patch that doesn't define LIB_TINYUSB_HOST

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
mbrase: Will address this in a separate PR

mbrase:
I've uploaded a new patch which is a subset of the previous patch

Add Bazel label flag for overriding tusb_config.h
    
This adds `//bazel/config:PICO_TINYUSB_CONFIG`, which is used to select
which tusb_config.h gets linked into the tinyUSB port. This defaults to
the tusb_config.h from the pico_stdio_usb (same as before), but allows
Bazel users to override this config with their own.

mbrase:
Hi, I'm ready for this to be reviewed

src/rp2_common/pico_stdio_usb/BUILD.bazel
```diff
@@ -43,10 +46,6 @@ cc_library(
     hdrs = ["include/pico/stdio_usb.h"],
     includes = ["include"],
     target_compatible_with = compatible_with_rp2(),
-    visibility = [
```
armandomontanez, line 46: Publicly exposing this is a little unfortunate, but it's probably the only path forward anyway.

armandomontanez (approved):

armandomontanez:
Bazel changes look good to me, sorry for the dealy.

src/rp2_common/pico_stdio_usb/BUILD.bazel
mbrase: Yeah, I wanted to include the stdio_usb.h in my tusb_config.h, much like the default tusb_config.h does, and I need the header only target to avoid a circular dependency.

Issue #2141, opened 2024-12-18
pioasm & sdk support of `wait gpio` on rp2350b


jepler:
Hi! Over in CircuitPython I'm working on proper support for the high IO bank on the RP2350B.

As part of that, I think I've found a problem in pioasm and maybe `add_program_at_offset`.

I have the following (useless) pio program:
```
.pio_version 1
.program baker
    wait 0 gpio 18
    wait 0 gpio 34
    .word 0x2022
```

pioasm from 2.1.0 (& tip of develop, 2.1.0-6-g969f589) compiles it to the following:
```
static const uint16_t baker_program_instructions[] = {
            //     .wrap_target
    0x2012, //  0: wait   0 gpio, 18                 
    0x2022, //  1: wait   0 pin, 34                  
    0x2022, //  2: wait   0 pin, 2                   
            //     .wrap
};
```
Notice how the instruction encoding 0x2022 is used for both the `wait gpio` and `wait pin`, and the instruction written as `wait 0 gpio 34` has decoded to the impossible (?) `wait 0 pin, 34`.

I'm not sure what needs to happen here. I am pretty sure that at runtime you need a PIO peripheral with its gpio_offset set to 16, and that the instruction encodings loaded into `instr_mem` should actually be `0x2002` (gpio 16+2 = 18) and `0x2012` (gpio 16+18=34). but I don't know whether this should be the encoding in the `program_instructions[]` array, or if `add_program_at_offset` in the sdk should take care of altering the `wait gpio` instruction encodings similar to how it alters `jmp` encodings depending on the load address. 

This bears directly on how we implement this in CircuitPython, since we want to use `pio_add_program` and friends to load the instructions into the PIO peripheral, but also use our own python implementation of pioasm.

lurch:
Looking at section 11.4.3. of https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf shows that `Index` is only encoded with 5 bits (and thus has a maximum value of 32), which I guess means that the mapping of gpio34 -> gpio18 **has** to be done before encoding things into the `program_instructions[]` array? :thinking: And presumably at the moment trying to set an `Index` field of 34 is overflowing into the `Source` field? Which would explain why `wait 0 gpio 34` is encoding to the same thing as `wait 0 pin, 2` ( because `34 = (1 << 5) | (2 & 0x1f)` ) ?

[...]

PR #2265, opened 2025-02-14
Fix `WAIT GPIO n` for n > 32 and when using gpio_base = 16

kilograham:
- fix bug: pioasm was spilling bit 5 of arg2 into arg1
- make pio_add_program flip bit 5 of a WAIT GPIO instruction if gpio_base is 16

Note: pioasm will encode only the bits 4-0 of arg2 (the GPIO number), however it does indicate whether 0-15 means 0-15 or 32-47. This can only be fixed up when loading the program, and indeed once the GPIO base is known. pio_can_add_program and friends already detected the cases where this was impossible


kilograham:
fixes #2141

peterharperuk (approved):
Ok. After staring at this for awhile it looks good. We really should have an example of using this. Any suggestions?

PR #2270, opened 2025-02-15
fix up access type of CHAN_ABORT register in RP2040.svd and RP2350.svd

### LINKED ISSUE #1801: Contradictory SVD contents regarding CHAN_ABORT
With version 2.0.0, the `access` specification of `CHAN_ABORT` was changed to `write-only` in `RP2040.svd`:

```
            <register>
                <name>CHAN_ABORT</name>
                <addressOffset>0x00000444</addressOffset>
                <description>Abort an in-progress transfer sequence on one or more channels</description>
                <resetValue>0x00000000</resetValue>
                <fields>
                    <field>
                        <name>CHAN_ABORT</name>
                        <description>Each bit corresponds to a channel. Writing a 1 aborts whatever transfer sequence is in progress on that channel. The bit will remain high until any in-flight transfers have been flushed through the address and data FIFOs.

                            After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel.</description>
                        <bitRange>[15:0]</bitRange>
                        <access>write-only</access>
                    </field>
                </fields>
            </register>
```

I guess this was done because of RP2040-E13 and one should poll `BUSY` instead of `CHAN_ABORT`.

However, the `description` still says "After writing, this register must be polled until it returns all-zero." This can't be done on a write-only register.

kilograham:
fixes #1801

### DIFF
--- a/src/rp2040/hardware_regs/RP2040.svd
+++ b/src/rp2040/hardware_regs/RP2040.svd
@@ -32214,7 +32214,8 @@ SPDX-License-Identifier: BSD-3-Clause
 
                             After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel.</description>
                         <bitRange>[15:0]</bitRange>
-                        <access>write-only</access>
+                        <access>read-write</access>
+                        <modifiedWriteValues>clear</modifiedWriteValues>
                     </field>
                 </fields>
             </register>
--- a/src/rp2040/hardware_regs/RP2040.svd
+++ b/src/rp2040/hardware_regs/RP2040.svd
@@ -32215,7 +32215,7 @@ SPDX-License-Identifier: BSD-3-Clause
                             After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel.</description>
                         <bitRange>[15:0]</bitRange>
                         <access>read-write</access>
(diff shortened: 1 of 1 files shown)


peterharperuk (dismissed):

src/rp2040/hardware_regs/RP2040.svd
```diff
@@ -32214,7 +32214,8 @@ SPDX-License-Identifier: BSD-3-Clause
 
                             After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel.</description>
                         <bitRange>[15:0]</bitRange>
-                        <access>write-only</access>
+                        <access>read-write</access>
+                        <modifiedWriteValues>clear</modifiedWriteValues>
```
lurch, line 32218: https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf and https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf both document "DMA: CHAN_ABORT" as being Type `SC`. So out of curiosity, why does `RP2040.svd` use `clear` but `RP2350.svd` uses `oneToClear` ?

kilograham: good catch

lurch (approved):

PR #2296, opened 2025-02-18
Add new feature defines to pico_stdio_usb

mbrase:
This change adds the following to pico_stdio_usb:
- `PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK`
- `PICO_STDIO_USB_ENABLE_TINYUSB_INIT`
- `PICO_STDIO_USB_USE_DEFAULT_DESCRIPTORS`

These defines replace some ifdefs that were previously gated by the `LIB_TINYUSB_DEVICE` define and will allow finer grained control. All of these defines default to 1 if LIB_TINYUSB_DEVICE is not defined and 0 if LIB_TINYUSB_DEVICE is defined, which should preserve the old behavior.

### DIFF
--- a/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
+++ b/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
@@ -44,10 +44,30 @@
 // this variable is no longer set by default (one is claimed dynamically), but will be respected if specified
 #endif
 
+// PICO_CONFIG: PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK, Enable/disable the use of a background task to call tud_task(), type=bool, default=1 if the application is not using tinyUSB directly, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK
+#if !LIB_TINYUSB_HOST && !LIB_TINYUSB_DEVICE
+#define PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK 1
+#else
+#define PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK 0
+#endif
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_ENABLE_TINYUSB_INIT, Enable/disable calling tinyUSB tusb_init() during initialization, type=bool, default=1 if the application is not using tinyUSB directly, group=pico_stdio_usb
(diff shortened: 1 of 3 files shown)


src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -44,10 +44,30 @@
...
+#if !defined(LIB_TINYUSB_HOST) && !defined(LIB_TINYUSB_DEVICE)
+#define PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK 1
+#else
+#define PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK 0
+#endif
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_ENABLE_TINYUSB_INIT, Enable/disable calling tinyUSB tusb_init() during initialization, default=1 if the application is not using tinyUSB directly, group=pico_stdio_usb
+#ifndef PICO_STDIO_USB_ENABLE_TINYUSB_INIT
+#if !defined(LIB_TINYUSB_HOST) && !defined(LIB_TINYUSB_DEVICE)
```
armandomontanez, line 58: Copying from other PR:

These are the only instances in the Pico SDK that use `#if defined(LIB_*)`, I suspect they were intended to be `#if LIB_*`.

mbrase: Done.

src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -44,10 +44,30 @@
 // this variable is no longer set by default (one is claimed dynamically), but will be respected if specified
 #endif
 
+// PICO_CONFIG: PICO_STDIO_USB_ENABLE_IRQ_BACKGROUND_TASK, Enable/disable the use of a background task to call tud_task(), default=1 if the application is not using tinyUSB directly, group=pico_stdio_usb
```
lurch, line 47: It seems (to me) like "using tinyUSB directly" is more likely to be the common case, so I wonder if it might be worth rewording these to ", default=0 if the application is using tinyUSB directly, 1 otherwise" ?
Also, it's probably nice to add ", type=bool" too :slightly_smiling_face:

mbrase: Didn't realize I forgot the type=bool, thanks!

PR #2298, opened 2025-02-18
add cast to fix warning

kilograham:


lurch (approved):
Haven't tested, but I assume that this'll work :slightly_smiling_face:

peterharperuk (approved):
Fixes my problem

PR #2300, opened 2025-02-18
Add new stdio_usb_call_chars_available_callback() function

mbrase:
When the user links in tinyUSB directly, the pico_stdio_usb library disables some of its functionality, including its built-in background processing thread. The user can implement their own background thread in order to continue using the stdio functionality, except that there is no wey to trigger the registered chars_available_callback. This commit adds a new `stdio_usb_calls_chars_available_callback()` method to allow user's background threads to call the callback.

### DIFF
--- a/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
+++ b/src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
@@ -163,7 +163,7 @@ bool stdio_usb_deinit(void);
 bool stdio_usb_connected(void);
 
 #if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
-/*! \brief Explicitly runs the registered USB stdio chars_available_callback
+/*! \brief Explicitly calls the registered USB stdio chars_available_callback
  *  \ingroup pico_stdio_usb
  *
  * \ref This method is normally called by the internal USB stdio background thread when there is new USB CDC
@@ -171,7 +171,7 @@ bool stdio_usb_connected(void);
  * directly links tinyUSB), the user will need to implement their own background thread and call this
  * method directly.
  */
-void stdio_usb_run_chars_available_callback(void);
+void stdio_usb_call_chars_available_callback(void);
 #endif
 
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -161,6 +161,21 @@ bool stdio_usb_deinit(void);
...
+#if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
+/*! \brief Explicitly runs the registered USB stdio chars_available_callback
+ *  \ingroup pico_stdio_usb
+ *
+ * \ref This method is normally called by the internal USB stdio background thread when there is new USB CDC
+ * data available to read. However, if the internal background thread is disabled (e.g. when the user
+ * directly links tinyUSB), the user will need to implement their own background thread and call this
+ * method directly.
+ *
+ *  \return true if stdio is connected over CDC
```
lurch, line 174: Huh, says `\return true` but the function is declared `void` ?

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -171,6 +171,11 @@ void stdio_usb_set_chars_available_callback(void (*fn)(void*), void *param) {
     chars_available_callback = fn;
     chars_available_param = param;
 }
+void stdio_usb_run_chars_available_callback(void) {
```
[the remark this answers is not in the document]
mbrase: Fixed in latest upload

src/rp2_common/pico_stdio_usb/stdio_usb.c
```diff
@@ -66,7 +66,7 @@ static void low_priority_worker_irq(void) {
 #endif
         mutex_exit(&stdio_usb_mutex);
 #if PICO_STDIO_USB_SUPPORT_CHARS_AVAILABLE_CALLBACK
-        if (chars_avail && chars_available_callback) chars_available_callback(chars_available_param);
+        if (chars_avail && chars_available_callback) stdio_usb_run_chars_available_callback();
```
kilograham, line 69: I think i'd revert this line... no point in calling another method here (which also checks `chars_avail_callback` - id just call it directly to save introduction of the new function (which likely isn't otherwise called when this code is included

kilograham (approved):

PR #2302, opened 2025-02-19
Missed doc change in #850

kilograham:
found this dangling while cleaning up github branchs

### DIFF
--- a/src/rp2_common/hardware_gpio/include/hardware/gpio.h
+++ b/src/rp2_common/hardware_gpio/include/hardware/gpio.h
@@ -615,6 +615,8 @@ static inline void gpio_add_raw_irq_handler(uint gpio, irq_handler_t handler) {
  *
  * This method removes such a callback, and enables the "default" callback for the specified GPIOs.
  *
+ * \note You should always use the same gpio_mask as you used when you added the raw IRQ handler.
+ *
  * @param gpio_mask a bit mask of the GPIO numbers that will now be passed to the default callback for this core
  * @param handler the handler to remove from the list of GPIO IRQ handlers for this core
  */


src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -615,6 +615,8 @@ static inline void gpio_add_raw_irq_handler(uint gpio, irq_handler_t handler) {
  *
  * This method removes such a callback, and enables the "default" callback for the specified GPIOs.
  *
+ * \note You should always use the same gpio_mask as you used when you added the raw IRQ handler.
```
lurch, line 618: I wondered why this was the case, since it seems like you _should_ be able to do:
```c
gpio_add_raw_irq_handler_masked((1<<0) | (1<<1) | (1<< 2), my_handler);
gpio_remove_raw_irq_handler_masked((1<<1), my_handler);
```
and then expect to have `my_handler` still running for GPIOs 0 and 2. But looking at the code for `gpio_remove_raw_irq_handler_masked` I see that it just does a blunt:
```c
irq_remove_handler(IO_IRQ_BANK0, handler);
```
rather than only removing the handler when _all_ of its masked pins have been removed (IYSWIM).
Would supporting this more refined approach (which I guess would require storing a mapping from `handler`s to `mask`s) be too slow or use up too much memory?

And just out of curiosity, how does
```c
void gpio_remove_raw_irq_handler_masked(uint32_t gpio_mask, irq_handler_t handler) {
    assert(raw_irq_mask[get_core_num()] & gpio_mask); // should not remove handlers that are not added
    irq_remove_handler(IO_IRQ_BANK0, handler);
    raw_irq_mask[get_core_num()] &= ~gpio_mask;
}
```
re-enable the "default" callback for each of the pins in the mask? Similarly, I can't see how
```c
void gpio_add_raw_irq_handler_with_order_priority_masked(uint32_t gpio_mask, irq_handler_t handler, uint8_t order_priority) {
    hard_assert(!(raw_irq_mask[get_core_num()] & gpio_mask)); // should not add multiple handlers for the same event
    raw_irq_mask[get_core_num()] |= gpio_mask;
    irq_add_shared_handler(IO_IRQ_BANK0, handler, order_priority);
}

void gpio_add_raw_irq_handler_masked(uint32_t gpio_mask, irq_handler_t handler) {
    gpio_add_raw_irq_handler_with_order_priority_masked(gpio_mask, handler, GPIO_RAW_IRQ_HANDLER_DEFAULT_ORDER_PRIORITY);
}
```
disables the "default" callback for the masked pins? :thinking:

kilograham: 1. it does track them
2. the `raw_irq_mask` array determines this

PR #2303, opened 2025-02-19
Another orphaned branch - removing superfluous dependency

kilograham:


src/rp2_common/pico_lwip/CMakeLists.txt
```diff
@@ -149,6 +149,7 @@ if (EXISTS ${PICO_LWIP_PATH}/${LWIP_TEST_PATH})
             ${PICO_LWIP_PATH}/src/netif/ppp/polarssl/sha1.c
             )
 
+    pico_add_library(pico_lwip_apps)
```
peterharperuk: Where did pico_lwip_apps come from?

kilograham: think you are looking at the wrong diffs?

peterharperuk (approved):

PR #2306, opened 2025-02-19
Remove assignment of cyw43 last_size/header/backplane_window variables

### LINKED ISSUE #2305: `_cyw43_write_reg` assigns to `last_size`/`_header`/`_backplane_window` variables which are never used
The cyw43 SPI transport `_cyw43_write_reg()` function has this bit of code in it:
```c
    if (fn == BACKPLANE_FUNCTION) {
        // In case of f1 overflow
        self->last_size = 8;
        self->last_header[0] = buf[0];
        self->last_header[1] = buf[1];
        self->last_backplane_window = self->cur_backplane_window;
    }
```
(`self` is a cyw43 driver instance.)

These `last_xxx` variables are never used by `cyw43-driver` because `F1_OVERFLOW_CHANGE` is permanently disabled.  And that feature even has comments that it never actually worked.

It would be good to get rid of these variables in the cyw43 driver data structure, to save a little bit of RAM and clean it up.  So I propose removing their use from this sdk.

Alternatively the above code could be guarded by a `#if F1_OVERFLOW_CHANGE`, but then that config option would need to be made public, probably named `CYW43_F1_OVERFLOW_CHANGE`.

dpgeorge:
They variables are never used by the cyw43-driver code, because `F1_OVERFLOW_CHANGE` is disabled.

Fixes #2305 (see that issue for more detailed discussion).

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -451,13 +451,6 @@ static inline int _cyw43_write_reg(cyw43_int_t *self, uint32_t fn, uint32_t reg,
     uint32_t buf[2];
     buf[0] = make_cmd(true, true, fn, reg, size);
     buf[1] = val;
-    if (fn == BACKPLANE_FUNCTION) {
-        // In case of f1 overflow
-        self->last_size = 8;
-        self->last_header[0] = buf[0];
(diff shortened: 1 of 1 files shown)


dpgeorge:
I have tested this change on a Pico W and Pico 2 W (using MicroPython, running some network tests on those boards with this change).  It works.

peterharperuk (approved):

PR #2315, opened 2025-02-24
Add support for creating self-decrypting binaries

will-v-pi:
This adds extra functionality to the `pico_encrypt_binary` function to allow creating self-decrypting binaries, including specifying the OTP page to use for the AES key.

```cmake
pico_encrypt_binary(hello_encrypted
    ${CMAKE_CURRENT_LIST_DIR}/privateaes.bin
    ${CMAKE_CURRENT_LIST_DIR}/ivsalt.bin
    EMBED
    OTP_KEY_PAGE 29)
```

The main non-backwards-compatible change is the addition of a new IV salt bin file which is required and must be passed as the second argument. This will break any uses of `pico_encrypt_binary`, and has been added for security purposes, as we now salt the public IV with a salt stored in OTP.

The other non-backwards-compatible change it that if you previously called:
```cmake
pico_encrypt_binary(my_target my_aesfile.bin my_sigfile.pem)
```
you now need to call
```cmake
pico_encrypt_binary(my_target my_aesfile.bin my_iv_salt.bin SIGFILE my_sigfile.pem)
```
due to the new argument parsing. I think that this is fine, because the only time you'd pass a separated `SIGFILE` to `pico_encrypt_binary` is when you're using a different signing key for the binary vs the encrypted blob, which is not a common use case.

This PR requires use of the picotool encrypted-shares branch (https://github.com/raspberrypi/picotool/pull/207), so should be merged after that.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -460,7 +460,8 @@ endfunction()
 # This sets the target property PICOTOOL_EMBED_DECRYPTION to TRUE.
 #
 # Optionally, use MBEDTLS to to use the MbedTLS based decryption stage - this
-# is faster, but less secure.
+# is faster, but offers no security against power or timing sniffing attacks,
+# and takes up more code size.
 # This sets the target property PICOTOOL_USE_MBEDTLS_DECRYPTION to TRUE.
 #
 # Optionally, use OTP_KEY_PAGE to specify the OTP page storing the AES key.
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -42,6 +42,12 @@ define_property(TARGET
     FULL_DOCS "AES key for encrypting"
 )
 define_property(TARGET
+    PROPERTY PICOTOOL_EMBED_DECRYPTION
+    INHERITED
+    BRIEF_DOCS "Embed decryption stage into encrypted binary"
+    FULL_DOCS "Embed decryption stage into encrypted binary"
+)
+define_property(TARGET
     PROPERTY PICOTOOL_ENC_SIGFILE
     INHERITED
     BRIEF_DOCS "Private key for signing encrypted binaries"
@@ -424,7 +430,7 @@ endfunction()
 # \brief_nodesc\ Encrypt the taget binary
 #
 # Encrypt the target binary with the given AES key (should be a binary
-# file containing 32 bytes of a random key), and sign the encrypted binary.
+# file containing 128 bytes of a random key), and sign the encrypted binary.
 #
 # This sets the target properties PICOTOOL_AESFILE to AESFILE,
 # and PICOTOOL_ENC_SIGFILE to SIGFILE if present, else PICOTOOL_SIGFILE.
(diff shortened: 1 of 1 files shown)


tools/CMakeLists.txt
```diff
@@ -380,19 +404,58 @@ function(pico_embed_pt_in_binary TARGET PTFILE)
...
+# Salts the public IV with the provided IVFILE (should be a binary file
+# containing 16 bytes of a random IV), to give the IV used by the encryption.
+# This sets PICOTOOL_AESFILE to AESFILE, PICOTOOL_IVFILE to IVFILE, and
+# PICOTOOL_ENC_SIGFILE to SIGFILE if specified, else PICOTOOL_SIGFILE.
+#
+# Optionally, use EMBED to embed a decryption stage into the encrypted binary.
+# This sets PICOTOOL_EMBED_DECRYPTION to TRUE.
+#
+# Optionally, use MBEDTLS to to use the MbedTLS based decryption stage - this
+# is faster, but offers no security against power or timing sniffing attacks,
```
lurch, line 419: Should this caveat also be mentioned somewhere in the picotool PR related to this?

will-v-pi: It'll be documented more thoroughly in the C SDK book, but I'll add a quick note to the help text for the option

tools/CMakeLists.txt
```diff
@@ -380,19 +404,58 @@ function(pico_embed_pt_in_binary TARGET PTFILE)
...
 endfunction()
 
-# pico_encrypt_binary(TARGET AESFILE [SIGFILE])
+# pico_encrypt_binary(TARGET AESFILE IVFILE [SIGFILE <file>] [EMBED] [MBEDTLS] [OTP_KEY_PAGE <page>])
 # Encrypt the target binary with the given AES key (should be a binary
-# file containing 32 bytes of a random key), and sign the encrypted binary.
-# This sets PICOTOOL_AESFILE to AESFILE, and PICOTOOL_ENC_SIGFILE to SIGFILE
-# if present, else PICOTOOL_SIGFILE.
-function(pico_encrypt_binary TARGET AESFILE)
+# file containing 128 bytes of a random key), and sign the encrypted binary.
```
lurch, line 409: I guess this is now a 128-byte file (key share) OR a 32 byte file (key) OR a 64-character hex string (key) ?

tools/CMakeLists.txt
```diff
@@ -380,19 +404,58 @@ function(pico_embed_pt_in_binary TARGET PTFILE)
...
+# Salts the public IV with the provided IVFILE (should be a binary file
+# containing 16 bytes of a random IV), to give the IV used by the encryption.
```
lurch, line 411: And similarly I guess this is now a 16-byte file OR a 32-char hex string?

will-v-pi: Yeah, I think I'll say that the SDK only supports passing files - you really shouldn't be putting a secret key in plain text in your CMakeLists.txt file

tools/CMakeLists.txt
```diff
@@ -380,19 +404,59 @@ function(pico_embed_pt_in_binary TARGET PTFILE)
     )
 endfunction()
 
-# pico_encrypt_binary(TARGET AESFILE [SIGFILE])
+# pico_encrypt_binary(TARGET AESFILE IVFILE [SIGFILE <file>] [EMBED] [MBEDTLS] [OTP_KEY_PAGE <page>])
```
lurch, line 407: Is it worth preemptively adding your `\param\` stuff from #2422 here too, or is it too early for that?

will-v-pi: I'll wait till that's merged, then rebase this and fix the issues that presumably pop up

will-v-pi: Merged and rebased

tools/CMakeLists.txt
```diff
@@ -420,25 +444,70 @@ function(pico_embed_pt_in_binary TARGET PTFILE)
...
+# file containing 128 bytes of a random key share, or 32 bytes of a random key),
+# and sign the encrypted binary.
+#
+# Salts the public IV with the provided IVFILE (should be a binary file
+# containing 16 bytes of a random IV), to give the IV used by the encryption.
+#
+# This sets the target properties PICOTOOL_AESFILE to AESFILE, PICOTOOL_IVFILE to IVFILE, and
+# PICOTOOL_ENC_SIGFILE to SIGFILE if specified, else PICOTOOL_SIGFILE.
+#
+# Optionally, use EMBED to embed a decryption stage into the encrypted binary.
```
lurch, line 460: Another stupid question: when/why would you _not_ want to embed a decryption stage into the encrypted binary?

tools/CMakeLists.txt
```diff
@@ -420,25 +444,70 @@ function(pico_embed_pt_in_binary TARGET PTFILE)
...
+# This sets the target property PICOTOOL_EMBED_DECRYPTION to TRUE.
 #
-# This sets the target properties PICOTOOL_AESFILE to AESFILE,
-# and PICOTOOL_ENC_SIGFILE to SIGFILE if present, else PICOTOOL_SIGFILE.
+# Optionally, use MBEDTLS to to use the MbedTLS based decryption stage - this
+# is faster, but offers no security against power or timing sniffing attacks,
+# and takes up more code size.
+# This sets the target property PICOTOOL_USE_MBEDTLS_DECRYPTION to TRUE.
+#
+# Optionally, use OTP_KEY_PAGE to specify the OTP page storing the AES key.
```
lurch, line 468: Does this have some default or implicit value, if not explicitly set?

lurch: Is this same OTP page (or some fixed offset from it) also used for storing the IV salt, or _might_ it make sense to have an `OTP_SALT_PAGE` parameter?

kilograham (approved):

Issue #2334, opened 2025-03-06
RP2350: pad_checksum appears to set incorrect cpu


AnnsAnns:
It appears like the pad_checksum function accidentally sets the cpu as the `cortex-m0plus`, however the rp2350 uses the `cortex-m33`.

kilograham:
as per the PR, this actually may just want some removal of dead build code... certainly doesn't make any sense to do pad/checksum, but i'm not sure off the top of my head what is needed in terms of `pico_define_boot_stage2` and friends)

moving to 2.2.0 as this is a more involved change (not the `.cpu` wasn't that important as the entire asm files is `.byte`)

AnnsAnns:
Thank you for the update

will-v-pi:
I've added a new PR which should remove the dead code #3011

lurch:
Fixed on the `develop` branch.

PR #2335, opened 2025-03-06
rp2350/pad_checksum: fix set cpu

AnnsAnns:
See #2334 

This PR should hopefully fix the incorrect cpu for the rp2350 in the pad_checksum function.

### DIFF
--- a/src/rp2350/boot_stage2/pad_checksum
+++ b/src/rp2350/boot_stage2/pad_checksum
@@ -44,7 +44,7 @@ odata = idata + bytes(args.pad - len(idata))
 with open(args.ofile, "w") as ofile:
     ofile.write("// Padded and checksummed version of: {}\n\n".format(args.ifile))
     if args.arch == "arm":
-        ofile.write(".cpu cortex-m0plus\n")
+        ofile.write(".cpu cortex-m33\n")
         ofile.write(".thumb\n\n")
     ofile.write(".section .boot2, \"ax\"\n\n")
     ofile.write(".global __boot2_entry_point\n")


src/rp2350/boot_stage2/pad_checksum
```diff
@@ -44,7 +44,7 @@ odata = idata + bytes(args.pad - len(idata))
 with open(args.ofile, "w") as ofile:
     ofile.write("// Padded and checksummed version of: {}\n\n".format(args.ifile))
```
lurch, line 45: Perhaps this "and checksummed" file-comment is also misleading, as there's a code-comment a few lines above saying "No CRC..." ?
Also, looks like the `--seed` argument is _entirely_ unused by the RP2350 version of this script? :thinking: (Ahhh, but perhaps it's needed for CLI-compatibility with the [RP2040 version](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/boot_stage2/pad_checksum))

AnnsAnns:
After further investigation, it appears like this entire script is completely misleading, based on chapter 5.9.5 of the RP2350 documentation the boot2 checksum is not needed for the rp2350:

> Unlike RP2040, there is no requirement for flash binaries to have a checksummed "boot2" flash setup function at flash address 0. The RP2350 bootrom performs a simple best-effort XIP setup during flash scanning, and a flash-resident program can continue executing in this state, or can choose to reconfigure the QSPI interface at a later time for best performance.

I am not really sure how to proceed with this PR, wouldn't it make sense to completely get rid of that script based on the fact that this isn't something the pico2 actually needs?

AnnsAnns:
Closing this PR as the issue appears bigger than was expected in the scope of this PR, see issue comments

PR #2351, opened 2025-03-12
Add shortcuts to Pigweed formatters

armandomontanez:
Adds commands to easily run Pigweed's formatters:

* bazel run //tools:format_code
* bazel run //tools:check_code_formatting

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -16,7 +16,7 @@ bazel_dep(name = "rules_cc", version = "0.0.10")
 bazel_dep(name = "pigweed", version = "0.0.1", dev_dependency = True)
 git_override(
     module_name = "pigweed",
-    commit = "5ec4ad9051a66bbcd003b1961936624fe6ee952f",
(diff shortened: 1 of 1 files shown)


armandomontanez:
Needs a `.clang-format` config file at the root for `clang-format` to pick up. I'm biased, but my suggestion would be use Pigweed's as a starting point and then just tweak options to your liking.

lurch:
> Adds commands to easily run Pigweed's formatters:
> 
> * bazel run //tools:format_code

The comment in the code says `bazelisk run //tools:format_code` ?

armandomontanez:
The right answer is `bazelisk`, I just have `bazel` as an alias for `bazelisk` so I have a habit of typing that.

Also, this was put up as a WIP to discuss/demo live sometime in the future. While I find it handy for formatting the Bazel build files, it can format C/C++/Python files too. (I wasn't expecting anyone to pay any mind to it while it was marked as a draft)

PR #2352, opened 2025-03-14
Throw a warning when using the example signing/encryption keys

will-v-pi:
This throws a CMake warning when the example signing/encryption keys from pico-examples are used to sign/encrypt a binary, as the user should replace these with their own keys when actually developing secure applications

This is automatically disabled for pico-examples (as that can use the example keys), and can be manually disabled by setting PICO_ALLOW_EXAMPLE_KEYS


tools/CMakeLists.txt
```diff
@@ -449,6 +471,17 @@ function(picotool_postprocess_binary TARGET)
     set_target_properties(${TARGET} PROPERTIES
         PICOTOOL_PROCESSING_CONFIGURED true
     )
+
+    # Allow using the example keys for pico-examples
+    if ((NOT DEFINED PICO_ALLOW_EXAMPLE_KEYS) AND (DEFINED PICO_EXAMPLES_PATH))
```
lurch, line 476: I'm not sure if `(DEFINED PICO_EXAMPLES_PATH)` is the check that we want to use, as some people might set `-DPICO_EXAMPLES_PATH` unconditionally, as it's e.g. needed when you want to build the pico-sdk documentation. Would it be "better" to just `set(PICO_ALLOW_EXAMPLE_KEYS 1)` in the pico-examples `CMakeLists.txt` ?

tools/CMakeLists.txt
```diff
@@ -153,6 +153,28 @@ function(picotool_check_configurable TARGET)
...
 endfunction()
 
+# Compare 2 key files, used by picotool_check_default_keys
+function(picotool_compare_keys TARGET KEY DEFAULT TYPE)
+    if (KEY)
+        execute_process(COMMAND ${CMAKE_COMMAND} -E compare_files "${KEY}" "${PICO_SDK_PATH}/tools/example_keys/${DEFAULT}"
+            RESULT_VARIABLE compare_result
+        )
+        if(compare_result EQUAL 0)
+            message(WARNING "${TARGET} is using a default ${TYPE} key - this must be changed before production")
```
lurch, line 163: Would there be any value in also having a `PICO_FORBID_EXAMPLE_KEYS` setting, which changes this WARNING to an ERROR ? :shrug:

tools/CMakeLists.txt
```diff
@@ -153,6 +153,28 @@ function(picotool_check_configurable TARGET)
...
+        endif()
+    endif()
+endfunction()
+
+# Check if default signing/encryption keys are being used
```
lurch, line 168: Does there also need to be an example salt file?

tools/CMakeLists.txt
```diff
@@ -153,6 +153,28 @@ function(picotool_check_configurable TARGET)
...
+function(picotool_check_default_keys TARGET)
+    get_target_property(picotool_sigfile ${TARGET} PICOTOOL_SIGFILE)
+    picotool_compare_keys(${TARGET} ${picotool_sigfile} private.pem "signing")
+    get_target_property(picotool_aesfile ${TARGET} PICOTOOL_AESFILE)
+    picotool_compare_keys(${TARGET} ${picotool_aesfile} privateaes.bin "encryption")
+    get_target_property(picotool_enc_sigfile ${TARGET} PICOTOOL_ENC_SIGFILE)
+    picotool_compare_keys(${TARGET} ${picotool_enc_sigfile} private.pem "encrypted signing")
```
lurch, line 175: Maybe a stupid question, but would you "normally" use the same `.pem` file for both "signing" and "encrypted signing" ?

tools/CMakeLists.txt
will-v-pi: I don't think so, because the salt file could match the example one without issue, provided that you have a different AES key

tools/CMakeLists.txt
will-v-pi: Yes, normally you would use the same for both - I think the only use case for using different keys is if you want the public key for your encrypted binary to be kept secret, in which case you would use your extra secret key for signing the binary, and the less secret key for signing the encrypted binary. The examples all use the same key for both

tools/CMakeLists.txt
will-v-pi: Yes, that sounds sensible, if `PICO_EXAMPLES_PATH` is being set for other projects

tdewey-rpi (dismissed):
Good use of CMake functionality, and I approve of the concept entirely.

Make doing the right thing easy, and the wrong thing hard.

tools/CMakeLists.txt
```diff
@@ -449,6 +471,12 @@ function(picotool_postprocess_binary TARGET)
     set_target_properties(${TARGET} PROPERTIES
         PICOTOOL_PROCESSING_CONFIGURED true
     )
+
+    # PICO_CMAKE_CONFIG: PICO_ALLOW_EXAMPLE_KEYS, Don't throw warning when using default signing/encryption keys, type=bool, default=0, group=build
```
kilograham, line 475: nit, but you don't "throw" a warning; perhaps "raise"

kilograham (approved):

PR #2357, opened 2025-03-18
Fix PICO_PREVIOUS_PLATFORM error

will-v-pi:
Currently the error is not being thrown when switching platforms without deleting the build directory (eg https://forums.raspberrypi.com/viewtopic.php?t=385540)

This fixes that - `INTERNAL` should be the type of the variable, not placed at the end of the `set` invocation


kilograham:
thanks! I could swear i had implemented something - sure I tested it though (bit of a head scratcher)

PR #2379, opened 2025-03-28
Move unique_id (configurably) earlier in the static init process

### LINKED ISSUE #2373: unique_id is empty if called by C++ static initializers
I had been stumped by the fact that this code was constantly returning 0000000000000000:

```cxx
#include "pico/unique_id.h"
struct unique_board_id_hex_str
{
	char c_str[PICO_UNIQUE_BOARD_ID_SIZE_BYTES * 2 + 1];
	unique_board_id_hex_str()
	{
		pico_get_unique_board_id_string(c_str, sizeof(c_str));
	}
};
const unique_board_id_hex_str usbd_serial_str{};

// later, in main:  usbd_serial_str.c_str == "0000000000000000"
```

After much confusion, I realized that `pico_get_unique_board_id_string` doesn't initialize `retrieved_id` and `_retrieve_unique_id_on_boot` is instead set up as a constructor:

```c
/// unique_id.c
static pico_unique_board_id_t retrieved_id;
static void __attribute__((constructor)) _retrieve_unique_id_on_boot(void) { retrieved_id = /* snip */; }
void pico_get_unique_board_id_string(char *id_out, uint len) { /* snip*/ }
```

It appears that `pico_get_unique_board_id_string` either must check for an uninitialized value and initialize it itself, or that the constructor must be run earlier via numbering it:

Example:
```c
static void __attribute__((constructor(1000))) _retrieve_unique_id_on_boot(void) { retrieved_id = /* snip */; }
```

This will let users order the sequence of startup (example follows), and will before un-numbered constructors/static c++ initialization by default. I can create a PR for either approach, but I wasn't sure which option was preferred by the maintainers.

```cxx
[[gnu::init_priority(2000)]]
const unique_board_id_hex_str usbd_serial_str{};
```

byteit101:
Fixes #2373

Move unique_id (configurably) earlier in the static initialization process . This enables using the unique_id in C++ static initializers by default, and avoid the static initialization order fiasco.

I added the configuration in the off chance that someone would need to adjust it, but I think 1000 is probably a good-enough default. Values range from 101-65535, or "none", which is after 65535. Normal C++ static initializers are "none". Users can still manually specify to be earlier (101-999) without any configuration changes, via any of the following:

```cxx
// C  & C++
__attribute__((constructor(101))) some_function() {}
// C++ only
__attribute__((init_priority(101))) some_class instance;
[[gnu::init_priority(101)]] some_class instance;
```

The following will all be initialized after unique_id is initialized, by default

```cxx
// C  & C++
__attribute__((constructor(1001))) some_function() {}
__attribute__((constructor)) some_function() {}
// C++ only
__attribute__((init_priority(1001))) some_class instance;
[[gnu::init_priority(1001)]] some_class instance;
some_class instance;
```

### DIFF
--- a/src/rp2_common/pico_unique_id/include/pico/unique_id.h
+++ b/src/rp2_common/pico_unique_id/include/pico/unique_id.h
@@ -42,6 +42,28 @@ extern "C" {
 #define PICO_UNIQUE_BOARD_ID_SIZE_BYTES 8
 
 /**
+ * \brief Static initialization order
+ * \ingroup pico_unique_id
+ *
+ * This defines the init_priority of the pico_unique_id. By default, it is 1000. The valid range is
+ * from 101-65535. Set to -1 to set the priority to none, thus putting it after 65535. Changing
+ * This value will initialize the unique_id earlier or later in the static initialization order.
+ * This is most useful for C++ consumers of the pico-sdk.
+ *
+ * See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-constructor-function-attribute
+ * and https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html#index-init_005fpriority-variable-attribute
+ *
+ * This is an example C++ static initializer that will to run before, and then after pico_unique_id:
+ *
+ * [[gnu::init_priority(500)]] my_class before_instance;
+ * [[gnu::init_priority(2000)]] my_class after_instance;
+ *
+ */
+#ifndef PICO_UNIQUE_BOARD_ID_INIT_PRIORITY
+#define PICO_UNIQUE_BOARD_ID_INIT_PRIORITY 1000
+#endif
+
+/**
  * \brief Unique board identifier
  * \ingroup pico_unique_id
  *
--- a/src/rp2_common/pico_unique_id/include/pico/unique_id.h
+++ b/src/rp2_common/pico_unique_id/include/pico/unique_id.h
@@ -46,14 +46,14 @@ extern "C" {
  * \ingroup pico_unique_id
  *
  * This defines the init_priority of the pico_unique_id. By default, it is 1000. The valid range is
- * from 101-65535. Set to -1 to set the priority to none, thus putting it after 65535. Changing
- * This value will initialize the unique_id earlier or later in the static initialization order.
+ * from 101-65535. Set it to -1 to set the priority to none, thus putting it after 65535. Changing
+ * this value will initialize the unique_id earlier or later in the static initialization order.
  * This is most useful for C++ consumers of the pico-sdk.
  *
  * See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-constructor-function-attribute
  * and https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html#index-init_005fpriority-variable-attribute
  *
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_unique_id/include/pico/unique_id.h
```diff
@@ -41,6 +41,28 @@ extern "C" {
 
 #define PICO_UNIQUE_BOARD_ID_SIZE_BYTES 8
 
+/**
+ * \brief Static initialization order
+ * \ingroup pico_unique_id
+ *
+ * This defines the init_priority of the pico_unique_id. By default, it is 1000. The valid range is
+ * from 101-65535. Set to -1 to set the priority to none, thus putting it after 65535. Changing
+ * This value will initialize the unique_id earlier or later in the static initialization order.
```
lurch, line 50: "Changing This value" -> "Changing this value" ?

src/rp2_common/pico_unique_id/include/pico/unique_id.h
```diff
@@ -41,6 +41,28 @@ extern "C" {
 
 #define PICO_UNIQUE_BOARD_ID_SIZE_BYTES 8
 
+/**
+ * \brief Static initialization order
+ * \ingroup pico_unique_id
+ *
+ * This defines the init_priority of the pico_unique_id. By default, it is 1000. The valid range is
+ * from 101-65535. Set to -1 to set the priority to none, thus putting it after 65535. Changing
```
lurch, line 49: I'm afraid that static initializers aren't something I know much about, but would there ever be a reason that you'd _want_ to set this to -1 ?

src/rp2_common/pico_unique_id/include/pico/unique_id.h
```diff
@@ -41,6 +41,28 @@ extern "C" {
...
+ * This value will initialize the unique_id earlier or later in the static initialization order.
+ * This is most useful for C++ consumers of the pico-sdk.
+ *
+ * See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-constructor-function-attribute
+ * and https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html#index-init_005fpriority-variable-attribute
+ *
+ * This is an example C++ static initializer that will to run before, and then after pico_unique_id:
```
lurch, line 56: "that will to run before"
???

byteit101: Whoops, I kept re-writing that line will fix

src/rp2_common/pico_unique_id/include/pico/unique_id.h
byteit101: The current, existing behavior can be reverted to by setting to -1. I don't think there is a reason you couldn't use 65535 instead, but the default linker script order is that what I've called -1 is after 65535 (I could call it 65536, but I think that may be more confusing, oddly enough)

Linker script: https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_crt0/rp2040/memmap_default.ld#L79-L80

Note the `.ctor.*` are the ones with init_priority, and `.ctor` is the -1/default/old behavior

byteit101: If you would prefer, I can remove -1 and we then wait until someone hollers that we broke something? (I think that is unlikely, but not sure, as I haven't explicitly used init_priority before this year)

kilograham (approved):

PR #2393, opened 2025-04-02
Force inline critical_section_{enter,exit}

tannewt:
Otherwise it may end up in flash when we want it to follow the caller's placement (maybe in RAM.)

### DIFF
--- a/src/common/pico_sync/include/pico/critical_section.h
+++ b/src/common/pico_sync/include/pico/critical_section.h
@@ -7,6 +7,7 @@
 #ifndef _PICO_CRITICAL_SECTION_H
 #define _PICO_CRITICAL_SECTION_H
 
+#include "pico.h"
 #include "pico/lock_core.h"
 
 #ifdef __cplusplus
@@ -58,7 +59,7 @@ void critical_section_init_with_lock_num(critical_section_t *crit_sec, uint lock
  *
  * \param crit_sec Pointer to critical_section structure
  */
-static inline void critical_section_enter_blocking(critical_section_t *crit_sec) {
+__force_inline static void critical_section_enter_blocking(critical_section_t *crit_sec) {
     crit_sec->save = spin_lock_blocking(crit_sec->spin_lock);
 }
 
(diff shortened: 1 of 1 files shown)


kilograham (approved):

src/common/pico_sync/include/pico/critical_section.h
```diff
@@ -7,6 +7,7 @@
 #ifndef _PICO_CRITICAL_SECTION_H
 #define _PICO_CRITICAL_SECTION_H
 
+#include "pico/platform/compiler.h"
```
kilograham, line 10: not a fan of this - it should be included by `pico.h`

tannewt: np. Didn't know what the right include was.

lurch:
Bazel CI checks are failing with:
```
In file included from src/common/pico_sync/critical_section.c:7:
src/common/pico_sync/include/pico/critical_section.h:62:15: error: expected ';' before 'static'
   62 | __force_inline static void critical_section_enter_blocking(critical_section_t *crit_sec) {
      |               ^~~~~~~
      |               ;
src/common/pico_sync/include/pico/critical_section.h:71:15: error: expected ';' before 'static'
   71 | __force_inline static void critical_section_exit(critical_section_t *crit_sec) {
      |               ^~~~~~~
      |               ;
```

https://github.com/search?q=repo%3Araspberrypi%2Fpico-sdk%20__force_inline&type=code seem to show a mix of `__force_inline static ...` and `static __force_inline ...` so this doesn't _appear_ to be as simple as the keywords being listed in the wrong order?
ping @armandomontanez for suggestions.

armandomontanez:
`__force_inline` is defined in `src/rp2_common/pico_platform_compiler/include/pico/platform/compiler.h`, which notably is in `rp2_common` and doesn't appear to have a host port. Sounds like either CMake isn't building the `pico_sync_critical_section` library for host, or somehow the `rp2_common` header is slipping into the CMake host build?

I can reduce the scope of the Bazel host build to work around this if needed.

kilograham:
yeah; i'll fix that in the `host` headers

kilograham:
> yeah; i'll fix that in the `host` headers

don't seem to be able to push to your PR, so added this here #2396 

I'm fine with the `pico.h` include (it should always be included first), however it is already included by `pico/lock_core.h` so not clear why you need it?

tannewt:
Ok, I rebased it on develop and squashed the three commits. Hopefully it'll pass now

kilograham (approved):

PR #2395, opened 2025-04-03
Bazel build implementation, doesn't provide options for the compilation modes opt and debug

### LINKED ISSUE #2394: Bazel build implementation, doesn't provide options for the compilation modes opt and debug.
The Bazel toolchain implementation does not have default args for the compilation modes opt and debug
 i,e --compilation_mode=dbg and --compilation_mode=opt

Also the rule "@pico-sdk//bazel/toolchain:override_debug" provides defaults of -Og and -g3, but these are enabled by default in the toolchains, and are not changed with the above bazel cmd line --compilation_mode

From the bazel docs the defaults for opt should be -O2 and -DNDEBUG
https://bazel.build/docs/user-manual#compilation-mode

NolanBrad:
Fixes #2394 

The Bazel toolchain implementation does not have default args for the compilation modes opt and debug
i,e --compilation_mode=dbg and --compilation_mode=opt

Also the rule "@pico-sdk//bazel/toolchain:override_debug" provides defaults of -Og and -g3, but these are enabled by default in the toolchains, and are not changed with the above bazel cmd line --compilation_mode

From the bazel docs the defaults for opt should be -O2 and -DNDEBUG
https://bazel.build/docs/user-manual#compilation-mode

I have added 2 boolean flags 
//bazel/config:PICO_COMPILATION_DEBUG_OVERRIDE
//bazel/config:PICO_COMPILATION_OPT_OVERRIDE
These can be set to true ( default is False) which will then remove the defaults from the toolchain. the user can then provide their own args with the cmd line --copts="...."


lurch:
ping @armandomontanez for his thoughts / review

bazel/config/BUILD.bazel
```diff
@@ -294,3 +294,13 @@ label_flag(
     name = "PICO_FREERTOS_LIB",
     build_setting_default = "//bazel:empty_cc_lib",
 )
+
```
armandomontanez, line 297: Please add docstrings above these and add them to the exclusion list in `BAZEL_ONLY_ALLOWLIST` [here](https://github.com/raspberrypi/pico-sdk/blob/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c/tools/compare_build_systems.py#L111C1-L111C21).

bazel/toolchain/BUILD.bazel
```diff
@@ -134,16 +152,27 @@ configurable_toolchain_feature(
 
 # TODO: Make this shim unnecessary.
```
armandomontanez, line 153: These shims aren't necessary anymore, you can just directly list `opt_args` and `debug_args` in the `cc_feature` rules.

bazel/config/BUILD.bazel
NolanBrad: Ok, Thanks. These have been added.

bazel/toolchain/BUILD.bazel
NolanBrad: Ok, Thanks. You are right, they weren't needed. Removed

NolanBrad:
I have also added for completeness the same implementations for the fastbuild compilation mode, which is the default. 
But the bazel documents said -gmlt -Wl,-S (https://bazel.build/docs/user-manual#compilation-mode) which didnt make sense, nor work. So I have left the default as nothing. Which is also the same as the override.  
Please check this if this new fastbuild infrastructure should be added, or removed, as at the moment it does nothing.

Thanks
Brad.

bazel/config/BUILD.bazel
```diff
@@ -294,3 +294,21 @@ label_flag(
...
+# PICO_BAZEL_CONFIG: PICO_COMPILATION_OPT_OVERRIDE, Override the default opt compilation mode arguments to none, type=bool, default=0, group=build
```
lurch, line 298: Perhaps it would make sense if I knew more about bazel, but IMHO "Override the default xxxxx arguments to none" sounds kinda weird?

NolanBrad: Your right, OVERRIDE could be replaced with something like REMOVE ?
Ideally I wanted to override the defaults with user defined replacements. But it quickly got beyond my bazel knowledge. 
So I resorted to an override Boolean to remove the defaults, and then the user could optionally replace using 
build: --copt="-Os" for example.

I'm good with changing it to REMOVE?

NolanBrad: I've got a commit ready to go, if PICO_COMPILATION_xxx_REMOVE_DEFS is better?

armandomontanez: Here's a suggestion for a better naming/phrasing:
```
# PICO_BAZEL_CONFIG: PICO_COMPILATION_NO_OPT_ARGS, Makes the opt compilation mode override a no-op so users can inject custom optimization arguments via the --copt flag, type=bool, default=0, group=build
```

It's not possible today to have a `string_list_flag` end up in a block of `cc_args`, but I can add that to my TODO list.

NolanBrad: Ok, good with that suggestion, I have committed the changes. Thanks

bazel/config/BUILD.bazel
```diff
@@ -300,3 +300,19 @@ label_flag(
     name = "PICO_MBEDTLS_CONFIG",
     build_setting_default = "//bazel:empty_cc_lib",
 )
+# PICO_BAZEL_CONFIG: PICO_COMPILATION_NO_OPT_ARGS, Makes the opt compilation mode a no-op so custom optimization arguments can be injected via the --copt flag instead, type=bool, default=0, group=build
+bool_flag(
+    name = "PICO_COMPILATION_NO_OPT_ARGS",
+    build_setting_default = False,
+)
+# PICO_BAZEL_CONFIG: PICO_COMPILATION_NO_DEBUG_ARGS, Makes the debug compilation mode a no-op so custom optimization arguments can be injected via the --copt flag instead, type=bool, default=0, group=build
```
lurch, line 308: This might be a silly question, but is the "custom optimization arguments" here supposed to say "custom debug arguments" ?

armandomontanez: Bazel's compilation modes (fastbuild, debug, and opt) are basically equivalent conceptually to `CMAKE_BUILD_TYPE` in CMake. They're mutually exclusive choices, and they all bring along flags that tweak optimization levels or do other arbitrary things.

Since Bazel doesn't offer a `MinSizeRel` analogue, I suspect part of @NolanBrad's motivation is to commandeer one of the existing modes to inject `-Os` in their own project.

NolanBrad: @lurch, Sorry for the tardy spelling and comments, hopefully all fixed up now

kilograham (approved):
@armandomontanez has already approved

PR #2397, opened 2025-04-04
Rework use of pico_cmake_set in board headers to make it slightly less magic/confusing

kilograham:
Particularly, the fact that the lines are commented is confusing

- prefer "pico_cmake_set(var, value)" over "// pico_cmake_set var = value"
- prefer "pico_cmake_set_default(var, value)" over "// pico_cmake_set_default var = value"
- move these inside the header include guards as CLion complains

Note that the macros are defined in "pico.h" however that is not explicitly included by the board headers; this will probably confuse some VS code syntax highligting, so lets see how it looks - i'd prefer to avoid having to include a header just for this


src/boards/include/boards/0xcb_helios.h
```diff
@@ -73,7 +73,7 @@
 #endif
 
 // board has 16M onboard flash
-// pico_cmake_set_default PICO_FLASH_SIZE_BYTES = (16 * 1024 * 1024)
+pico_cmake_set_default(PICO_FLASH_SIZE_BYTES, 16 * 1024 * 1024)
```
lurch, line 76: Now that this "looks" like a proper macro, is there a risk that people might expect it to behave like a proper macro? E.g. if somebody did:
```c
// board has 16M onboard flash
#define DEFAULT_MEMORY_SIZE (16 * 1024 * 1024)
pico_cmake_set_default(PICO_FLASH_SIZE_BYTES, DEFAULT_MEMORY_SIZE)
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES DEFAULT_MEMORY_SIZE
#endif
```
would it work as expected?

lurch:
IMHO it would be great if this PR could also update https://github.com/raspberrypi/pico-sdk/blob/develop/tools/check_board_header.py cope with this new syntax. That's normally something that I'd do myself, but I'm about to go on holiday :sunglasses:

will-v-pi (commented):
This does seem to work fine with VS Code when using `compile_commands.json` - although it might make people think they can use `pico_cmake_set` outside of the board headers, so it might be good to add some sort of guard against that? Or change the definition to `pico_board_cmake_set` to make it clearer that it only works in board headers?

kilograham:
`pico_board_cmake_set` good idea - this also helps with something that had been bothering me that a commented out new style one looks a lot like an old style one!

will-v-pi:
I've updated the python script, added back the brackets for `PICO_FLASH_SIZE_BYTES` (as the python script checks for exact matches), and renamed to `pico_board_cmake_set` and `pico_board_cmake_set_default`

If you do want the brackets removed, then I can modify the python script to ignore the brackets instead

kilograham:
no brackets are good ;-) i was gonna put them back, but lost interest ;-)

cmake/generic_board.cmake
```diff
@@ -20,22 +20,22 @@ if (EXISTS ${PICO_BOARD_HEADER_FILE})
...
+        elseif (LINE MATCHES "^[ \t\]*pico_board_cmake_set_default[ \t\]*\\([ \t\]*([a-zA-Z_][a-zA-Z0-9_]*)[ \t\]*,[ \t\]*(.*)[ \t\]*\\)[ \t\]*")
             if (NOT DEFINED "${CMAKE_MATCH_1}")
                 set("${CMAKE_MATCH_1}" "${CMAKE_MATCH_2}")
             else()
                 list(APPEND PICO_BOARD_CMAKE_OVERRIDES ${CMAKE_MATCH_1})
             endif()
         # continue to support these for now
-        # // pico_cmake_set var = value
-        elseif (LINE MATCHES "^[ \t\]*//[ \t\]*pico_cmake_set[ \t\]*([a-zA-Z_][a-zA-Z0-9_]*)[ \t\]*=[ \t\]*(.*)")
+        # // pico_board_cmake_set var = value
```
kilograham, line 34: these should not be renamed

will-v-pi (approved):

PR #2400, opened 2025-04-06
Add pioasm c++ constexpr output target

audiobird:
Adds pioasm output that creates a c++ header file with a `constexpr` `std::array` containing the raw instructions.

Similar to the hex output option, only the raw instructions are generated. However, unlike the hex output option, multiple programs are supported in a single pio source file, with a `std::array` generated for each one.

One thing I'm unsure of is what namespace to put the generated arrays in. I went with "Pioasm" here but we could also go with "PIOInstructions" or something else entirely.

Input example: "ws2812.pio"
```
.program ws2812 
.side_set 1

.wrap_target
bitloop:
    out x, 1       side 0 [2];
    jmp !x do_zero side 1 [1];
do_one:
    jmp  bitloop   side 1 [4];
do_zero:
    nop            side 0 [4];
.wrap
```

`pioasm -o constexpr "ws2812.pio.hpp" "ws2812.pio"`

Output: "ws2812.pio.hpp"
```
#pragma once

#include <array>
#include <cstdint>

namespace Pioasm {

inline constexpr std::array<uint16_t, 4> ws2812 = {
            //     .wrap_target
    0x6221, //  0: out    x, 1            side 0 [2]
    0x1123, //  1: jmp    !x, 3           side 1 [1]
    0x1400, //  2: jmp    0               side 1 [4]
    0xa442, //  3: nop                    side 0 [4]
            //     .wrap
};

}
```

### DIFF
--- a/tools/pioasm/CMakeLists.txt
+++ b/tools/pioasm/CMakeLists.txt
@@ -27,6 +27,7 @@ add_executable(pioasm
 target_sources(pioasm PRIVATE c_sdk_output.cpp)
 target_sources(pioasm PRIVATE python_output.cpp)
 target_sources(pioasm PRIVATE hex_output.cpp)
+target_sources(pioasm PRIVATE constexpr_output.cpp)
 target_sources(pioasm PRIVATE json_output.cpp)
 target_sources(pioasm PRIVATE ada_output.cpp)
 target_sources(pioasm PRIVATE go_output.cpp)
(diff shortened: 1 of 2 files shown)


kilograham:
seems like you could add an output option for the namespace

audiobird:
Great, currently it looks like none of the output formats actually use the `std::vector<std::string> output_options` argument in the virtual output function, so I have no reference on what the preferred syntax would look like for the switch. 

How about: 
`pioasm -o constexpr -p namespace=some_namespace ...`

And would it be best to have a default namespace value or to require one?

kilograham:
yes it seems `ada` has one, but i think i prefer yours (name=value)

i think it should have a default value (though perhaps that is no namespace)

kilograham:
note that #2651 added support for constexpr - is this PR still useful?

audiobird:
I'm satisfied if #2651 is merged

PR #2403, opened 2025-04-08
Minor fixups to compile with TF-M

will-v-pi:
TF-M (Arm Trusted Firmware M) requires c99 compatibility, which throws errors at these lines. There is currently a patch for 2.0.0 in TF-M which fixes these errors, which TF-M applies when cloning the SDK. I have updated it to work with 2.1.0, 2.1.1, and the `develop` branch [here](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/37001), but it would be nice to get these fixups into the SDK so TF-M doesn't require this patch anymore.

### DIFF
--- a/src/rp2_common/hardware_flash/flash.c
+++ b/src/rp2_common/hardware_flash/flash.c
@@ -299,7 +299,11 @@ static void flash_devinfo_update_field(uint16_t wdata, uint16_t mask) {
     // Boot RAM does not support exclusives, but does support RWTYPE SET/CLR/XOR (with byte
     // strobes). Can't use hw_write_masked because it performs a 32-bit write.
     io_rw_16 *devinfo = flash_devinfo_ptr();
-    *(io_rw_32 *)hw_xor_alias_untyped(devinfo) = (*devinfo ^ wdata) & mask;
+#ifdef __STRICT_ANSI__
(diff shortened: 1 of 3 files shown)


src/rp2_common/hardware_flash/flash.c
```diff
@@ -299,7 +299,7 @@ static void flash_devinfo_update_field(uint16_t wdata, uint16_t mask) {
     // Boot RAM does not support exclusives, but does support RWTYPE SET/CLR/XOR (with byte
     // strobes). Can't use hw_write_masked because it performs a 32-bit write.
     io_rw_16 *devinfo = flash_devinfo_ptr();
-    *hw_xor_alias(devinfo) = (*devinfo ^ wdata) & mask;
+    *(io_rw_32 *)hw_xor_alias_untyped(devinfo) = (*devinfo ^ wdata) & mask;
```
kilograham, line 302: this is why typeof is good :-0

will-v-pi: Yeah - would you prefer these changes to be #ifdefed out by `__STRICT_ANSI__`, like the sem change is already?

kilograham: yes, probably (and in case it flew under the radar in previous comment, fix the cast on this one)

will-v-pi: Thanks - fixed

kilograham (approved):

PR #2405, opened 2025-04-08
Update mbedtls, lwip and cyw43-driver

peterharperuk:
Update lwip, mbedtls and cyw43-driver
Includes fixes to the build for mbedtls and some docs changes
It seemed less confusing to push all these changes in one PR?


peterharperuk:
I have tested this with pico-examples (needs this change https://github.com/raspberrypi/pico-examples/pull/631

kilograham:
what happens if the user hasn't initialized mbetls` sub-modules?

peterharperuk:
> what happens if the user hasn't initialized mbedtls` sub-modules?

Good question. The cmake workflow fails on github. But I think this is just caused by picotool trying to use lib/mbedtls which has a cmake file in lib/mbedtls/framework.

So pico-examples builds ok if the submodule is not there.

kilograham:
Cc @will-v-pi

will-v-pi:
Picotool currently includes mbedtls with `add_subdirectory`, but I can move it to a style like the SDK where it creates the library itself instead, which won't rely on the submodules

kilograham (approved):

PR #2410, opened 2025-04-10
Add CYW43_WL_GPIO_SMPS_PIN for pico_w and pico2_w

### LINKED ISSUE #2361: We should probably add CYW43_WL_GPIO_SMPS_PIN for Pico W
As PICO_SMPS_MODE_PIN is not set for Pico W (and Pico 2 W) we should add something to make it clear a modem pin is used for this.

peterharperuk:
Fixes #2361

### DIFF
--- a/src/boards/include/boards/pico2_w.h
+++ b/src/boards/include/boards/pico2_w.h
@@ -89,6 +89,12 @@
 #define CYW43_WL_GPIO_LED_PIN 0
 #endif
 
+// Drive high to force power supply into PWM mode (lower ripple on 3V3 at light loads)
+// As this is a CYW43 pin you can do this by calling cyw43_gpio_set
+#ifndef CYW43_WL_GPIO_SMPS_PIN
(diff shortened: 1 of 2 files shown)


kilograham:
where is WL_GPIO_SMPS_PIN used - i don't see it referenced or PICO_CONFIGed?

peterharperuk:
I'm going to re-push this to try and understand the github build failures. Please ignore.

peterharperuk:
> where is WL_GPIO_SMPS_PIN used - i don't see it referenced or PICO_CONFIGed?

It's not used. I recently tried setting it to see if it was having any effect on Wireless stability. I realised at that point that I never had a define for it. We have PICO_DEFAULT_SMPS_PIN for non-wifi devices. Do you want me to add PICO_CONFIG for all the defines in the board file?

kilograham:
no it is good; i misunderstood what it was about - might be worth adding it to an example

kilograham (approved):

PR #2412, opened 2025-04-11
Change BT flash storage default for rp2350 A2

### LINKED ISSUE #2322: Move btstack flash bank storage from the end of flash?
Saw this mentioned on the forum somewhere. By default the bt stack flash storage uses the last two sectors in flash. However picotool has started using the last page to workaround RP2350-E10.

We could.
- ignore the problem
- change the default to 3 sectors from the end
- move it somewhere else, e.g. end of the binary or inside the binary (I seem to remember circuit python do this?)

peterharperuk:
The workaround for errata RP2350-E10 overwrites the last block in flash. This will overwrite the BT flash storage causing a paired BT connection to fail. Move the default flash storage location to 3 sectors from the end of flash for RP2350 where A2 support is required.

This will require existing BT pairings to a Pico device to be removed and readded.

Fixes #2322


will-v-pi (approved):
LGTM

PR #2414, opened 2025-04-11
Update memmap_copy_to_ram.ld

### LINKED ISSUE #2413: Misleading comment about bootloader in rp2350/memmap_copy_to_ram.ld
`rp2_common/pico_crt0/rp2350/memmap_copy_to_ram.ld` has a comment

```
    /* Second stage bootloader is prepended to the image. It must be 256 bytes big
       and checksummed. It is usually built by the boot_stage2 target
       in the Raspberry Pi Pico SDK
    */
```

At least `rp2_common/pico_crt0/rp2350/memmap_no_flash.ld` explains that:

 "On Arm, the bootrom expects a VT at the start of the  image by default; on RISC-V, the default is to enter the image at its  lowest address, so an IMAGEDEF item is required to specify the nondefault entry point."

It is unclear to me how RISC-V code can force an entry point to be at the lowest address, at least the comment hints at using a IMAGEDEF.

This issue though is about the comment about the "Second stage bootloader" being "prepended to the image.", since it can not be that way in the ARM code, right?

Suggesting

```
       /* On Arm, the bootrom expects a VT at the start of the
       image by default; on RISC-V, the default is to enter the image at its
       lowest address, so an IMAGEDEF item is required to specify the
       nondefault entry point. */
```

d-c-d:
Fix #2413

### DIFF
--- a/src/rp2_common/pico_crt0/rp2350/memmap_copy_to_ram.ld
+++ b/src/rp2_common/pico_crt0/rp2350/memmap_copy_to_ram.ld
@@ -33,9 +33,10 @@ ENTRY(_entry_point)
 
 SECTIONS
 {
-    /* Second stage bootloader is prepended to the image. It must be 256 bytes big
-       and checksummed. It is usually built by the boot_stage2 target
-       in the Raspberry Pi Pico SDK
+    /* On Arm, the bootrom expects a VT at the start of the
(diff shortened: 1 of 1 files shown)


d-c-d:
oops, need to switch this to develop

PR #2416, opened 2025-04-11
Add --recursive flag to git submodule cmd

peterharperuk:
Update all github workflows because picotool is (properly) using mbedtls in pico-sdk and it now has its own submodules.


peterharperuk:
@kilograham I think this will fix the problems we are seeing - I hadn't updated all the workflows.

kilograham (approved):

PR #2422, opened 2025-04-14
Add script to extract CMake function descriptions

will-v-pi:
This adds a script to extract CMake functions and output them as a TSV file, similar to `extract_cmake_configs.py` for `PICO_CMAKE_CONFIG` variables

Also adds basic descriptions for all functions missing them, other than some which are set as allowed to be missing them in the script

### DIFF
--- a/src/common/pico_binary_info/CMakeLists.txt
+++ b/src/common/pico_binary_info/CMakeLists.txt
@@ -9,11 +9,15 @@ endif()
 
 target_link_libraries(pico_binary_info INTERFACE pico_binary_info_headers)
 
+# pico_set_program_name(TARGET name)
+# Set the program name for the target
 function(pico_set_program_name TARGET name)
     # PICO_BUILD_DEFINE: PICO_PROGRAM_NAME, value passed to pico_set_program_name, type=string, group=pico_binary_info
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_NAME="${name}")
 endfunction()
 
+# pico_set_program_description(TARGET description)
+# Set the program description for the target
 function(pico_set_program_description TARGET description)
     # since this is the command line, we will remove newlines
     string(REPLACE "\n" " " description ${description})
@@ -22,11 +26,15 @@ function(pico_set_program_description TARGET description)
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_DESCRIPTION="${description}")
 endfunction()
 
+# pico_set_program_url(TARGET url)
+# Set the program URL for the target
 function(pico_set_program_url TARGET url)
     # PICO_BUILD_DEFINE: PICO_PROGRAM_URL, value passed to pico_set_program_url, type=string, group=pico_binary_info
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_URL="${url}")
 endfunction()
 
+# pico_set_program_version(TARGET version)
+# Set the program version for the target
 function(pico_set_program_version TARGET version)
     # PICO_BUILD_DEFINE: PICO_PROGRAM_VERSION_STRING, value passed to pico_set_program_version, type=string, group=pico_binary_info
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_VERSION_STRING="${version}")
--- a/src/common/pico_binary_info/CMakeLists.txt
+++ b/src/common/pico_binary_info/CMakeLists.txt
@@ -10,14 +10,18 @@ endif()
 target_link_libraries(pico_binary_info INTERFACE pico_binary_info_headers)
 
 # pico_set_program_name(TARGET name)
-# Set the program name for the target
+# \brief\ Set the program name for the target
+#
+# \param\ name The program name to set
 function(pico_set_program_name TARGET name)
     # PICO_BUILD_DEFINE: PICO_PROGRAM_NAME, value passed to pico_set_program_name, type=string, group=pico_binary_info
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_NAME="${name}")
 endfunction()
 
 # pico_set_program_description(TARGET description)
-# Set the program description for the target
+# \brief\ Set the program description for the target
+#
+# \param\ description The program description to set
 function(pico_set_program_description TARGET description)
     # since this is the command line, we will remove newlines
     string(REPLACE "\n" " " description ${description})
@@ -27,14 +31,18 @@ function(pico_set_program_description TARGET description)
 endfunction()
 
 # pico_set_program_url(TARGET url)
-# Set the program URL for the target
+# \brief\ Set the program URL for the target
+#
+# \param\ url The program URL to set
 function(pico_set_program_url TARGET url)
     # PICO_BUILD_DEFINE: PICO_PROGRAM_URL, value passed to pico_set_program_url, type=string, group=pico_binary_info
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_URL="${url}")
 endfunction()
 
 # pico_set_program_version(TARGET version)
-# Set the program version for the target
+# \brief\ Set the program version for the target
+#
+# \param\ version The program version to set
 function(pico_set_program_version TARGET version)
     # PICO_BUILD_DEFINE: PICO_PROGRAM_VERSION_STRING, value passed to pico_set_program_version, type=string, group=pico_binary_info
     target_compile_definitions(${TARGET} PRIVATE -DPICO_PROGRAM_VERSION_STRING="${version}")
--- a/src/rp2040/boot_stage2/CMakeLists.txt
+++ b/src/rp2040/boot_stage2/CMakeLists.txt
@@ -34,8 +34,12 @@ pico_add_library(boot_stage2_headers)
 target_include_directories(boot_stage2_headers SYSTEM INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
 
 # pico_define_boot_stage2(NAME SOURCES)
-# Define a boot stage 2 target.
+# \brief\ Define a boot stage 2 target.
+#
 # By convention the first source file name without extension is used for the binary info name
+#
+# \param\ NAME The name of the boot stage 2 target
+# \param\ SOURCES The source files to link into the boot stage 2
 function(pico_define_boot_stage2 NAME SOURCES)
     add_executable(${NAME}
             ${SOURCES}
@@ -100,7 +104,11 @@ endmacro()
 pico_define_boot_stage2(bs2_default ${PICO_DEFAULT_BOOT_STAGE2_FILE})
 
 # pico_clone_default_boot_stage2(NAME)
+# \brief_nodesc\ Clone the default boot stage 2 target.
+#
 # Create a new boot stage 2 target using the default implementation for the current build (PICO_BOARD derived)
+#
+# \param\ NAME The name of the new boot stage 2 target
 function(pico_clone_default_boot_stage2 NAME)
     pico_define_boot_stage2(${NAME} ${PICO_DEFAULT_BOOT_STAGE2_FILE})
 endfunction()
--- a/src/rp2040/boot_stage2/CMakeLists.txt
+++ b/src/rp2040/boot_stage2/CMakeLists.txt
@@ -33,7 +33,7 @@ set(PICO_BOOT_STAGE2_DIR "${CMAKE_CURRENT_LIST_DIR}" CACHE INTERNAL "")
 pico_add_library(boot_stage2_headers)
 target_include_directories(boot_stage2_headers SYSTEM INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
 
-# pico_define_boot_stage2(TARGET NAME SOURCES)
+# pico_define_boot_stage2(NAME SOURCES)
 # Define a boot stage 2 target.
 # By convention the first source file name without extension is used for the binary info name
 function(pico_define_boot_stage2 NAME SOURCES)
@@ -99,7 +99,7 @@ endmacro()
 
 pico_define_boot_stage2(bs2_default ${PICO_DEFAULT_BOOT_STAGE2_FILE})
 
-# pico_clone_default_boot_stage2(TARGET NAME)
+# pico_clone_default_boot_stage2(NAME)
 # Create a new boot stage 2 target using the default implementation for the current build (PICO_BOARD derived)
 function(pico_clone_default_boot_stage2 NAME)
     pico_define_boot_stage2(${NAME} ${PICO_DEFAULT_BOOT_STAGE2_FILE})
--- a/src/rp2040/boot_stage2/CMakeLists.txt
+++ b/src/rp2040/boot_stage2/CMakeLists.txt
@@ -33,7 +33,9 @@ set(PICO_BOOT_STAGE2_DIR "${CMAKE_CURRENT_LIST_DIR}" CACHE INTERNAL "")
 pico_add_library(boot_stage2_headers)
 target_include_directories(boot_stage2_headers SYSTEM INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
 
-# by convention the first source file name without extension is used for the binary info name
+# pico_define_boot_stage2(TARGET NAME SOURCES)
+# Define a boot stage 2 target.
+# By convention the first source file name without extension is used for the binary info name
 function(pico_define_boot_stage2 NAME SOURCES)
     add_executable(${NAME}
             ${SOURCES}
@@ -97,6 +99,7 @@ endmacro()
 
 pico_define_boot_stage2(bs2_default ${PICO_DEFAULT_BOOT_STAGE2_FILE})
 
+# pico_clone_default_boot_stage2(TARGET NAME)
 # Create a new boot stage 2 target using the default implementation for the current build (PICO_BOARD derived)
 function(pico_clone_default_boot_stage2 NAME)
     pico_define_boot_stage2(${NAME} ${PICO_DEFAULT_BOOT_STAGE2_FILE})
--- a/src/rp2_common/pico_btstack/CMakeLists.txt
+++ b/src/rp2_common/pico_btstack/CMakeLists.txt
@@ -302,6 +302,7 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
 
     pico_promote_common_scope_vars()
 
+    # pico_btstack_make_gatt_header(TARGET_LIB TARGET_TYPE GATT_FILE)
     # Make a GATT header file from a BTstack GATT file
     # Pass the target library name library type and path to the GATT input file
     # To add additional directories to the gatt #import path, add them to the end of the argument list.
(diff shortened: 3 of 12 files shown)


src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -301,8 +301,9 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
 
     pico_promote_common_scope_vars()
 
-    # Make a GATT header file from a BTstack GATT file
-    # Pass the target library name library type and path to the GATT input file
+    # pico_btstack_make_gatt_header(TARGET_LIB TARGET_TYPE GATT_FILE)
+    # Make a GATT header file from a BTstack GATT file.
+    # Pass the target library name library type and path to the GATT input file.
```
lurch, line 306: Perhaps some commas would make this line more readable? :wink:

tools/CMakeLists.txt
```diff
@@ -233,6 +234,7 @@ endfunction()
 # Package a UF2 output to be written to the PACKADDR address. This can be
 # used with a no_flash binary to write the UF2 to flash when dragging &
 # dropping, and it will be copied to SRAM by the bootrom before execution.
+#
```
lurch, line 237: Apologies if this is a stupid question, but why have you added semicolons `;` to the descriptions in some cmake files, but empty `#` lines to descriptions in this cmake file?

lurch:
Looks like 79b3eb2 probably shouldn't be in this PR?

tools/CMakeLists.txt
will-v-pi: Semicolons are where line breaks should be kept in subsequent tooling, as opposed to line breaks in other descriptions which are just because of line length so should be removed in subsequent tooling.

Empty lines are where actual empty lines should appear in subsequent tooling.

lurch: Ahhh. So a cmake-description like:
```
# Hello hello; is anyone
# listening to this
# drivel
#
# Perhaps not?;
# Oh well!
```
would appear in the "rendered output" as:
```
Hello hello
is anyone listening to this drivel

Perhaps not?
Oh well!
```
?

IIRC there's other file-formats where we've needed to do something similar, and there we've used the [Asciidoc convention](https://docs.asciidoctor.org/asciidoc/latest/blocks/hard-line-breaks/) of a line ending with a space and plus character (` +`) to signify a forced-line-break. I.e. the previous example would be:
```
# Hello hello +
# is anyone
# listening to this
# drivel
# +
# Perhaps not? +
# Oh well!
```

will-v-pi: I considered the asciidoc convention, but thought it would be confusing for functions like `pico_minimize_runtime` to have a bunch of +es in their descriptions?

Also, it wouldn’t add new line breaks where you have a semicolon (so no break before “is” above), it would just preserve existing ones after the semicolon.

lurch: > I considered the asciidoc convention, but thought it would be confusing for functions like `pico_minimize_runtime` to have a bunch of +es in their descriptions?

:man_shrugging: This is one of those situations where there's no perfect solution; so your solution here is just as good as any other!

will-v-pi:
(Rebased onto #2423 for now, will rebase onto develop once that's merged)

src/rp2040/boot_stage2/CMakeLists.txt
```diff
@@ -33,7 +33,9 @@ set(PICO_BOOT_STAGE2_DIR "${CMAKE_CURRENT_LIST_DIR}" CACHE INTERNAL "")
 pico_add_library(boot_stage2_headers)
 target_include_directories(boot_stage2_headers SYSTEM INTERFACE ${CMAKE_CURRENT_LIST_DIR}/include)
 
-# by convention the first source file name without extension is used for the binary info name
+# pico_define_boot_stage2(TARGET NAME SOURCES)
```
lurch, line 36: Possibly a silly question, but why does this line say `pico_define_boot_stage2(TARGET NAME SOURCES)` when a few lines below it says `function(pico_define_boot_stage2 NAME SOURCES)`? Note that `TARGET` is listed in one but not the other.
(same thing is true for some other new descriptions in this PR too)

will-v-pi: That's a mistake - will fix it

will-v-pi: I think this is just a typo for the boot_stage2 descriptions (rp2040 & rp2350) - let me know if you spot any others

lurch: Ah - if i's just a typo, perhaps i's worth expanding your Python regexes so that you can check the description-signature matches the actual cmake-function-signature? :slightly_smiling_face:

will-v-pi: The issue it that it doesn't always match when doing argument parsing - it's probably possible to check the first argument or something, but I think I'll leave that for a separate PR

tools/CMakeLists.txt
```diff
@@ -361,7 +370,9 @@ function(pico_sign_binary TARGET)
 endfunction()
 
 # pico_hash_binary(TARGET)
-# Hash the target binary. This sets PICOTOOL_HASH_OUTPUT to true.
+# Hash the target binary.
+#
```
kilograham, line 374: just picked this at random, but this sets the target property PICOTOOL_HASH_OUTPUT

tools/extract_cmake_functions.py
```diff
@@ -56,6 +56,49 @@
 
 all_functions = {}
 
+# Supported commands:
+# \brief\ <brief description, which should be included in the main description>
+# \brief_nodesc\ <brief description, which should be excluded from the main description>
+# \param\ <parameter_name> <parameter description>
```
lurch, line 62: IMHO it might be nicer if these were `\brief`, `\param`, etc. so that they're share the same "syntax" as doxygen commands, rather than requiring them to also have a `\` at the end.
You could parse this by changing:
```python
        line = line.strip()
        if line.startswith('\\'):
            command = line.split('\\')[1]
```
to something like
```python
        line = line.strip()
        m = re.match('\\(\w+)', line)
        if m:
            command = m.group(1)
```
?

tools/extract_cmake_functions.py
```diff
@@ -75,10 +118,9 @@
                 for match in CMAKE_FUNCTION_RE.finditer(text):
                     name = match.group(4)
                     signature = match.group(1).strip()
-                    description = match.group(2)
-                    description = description.replace('#', '').strip()
                     if signature.startswith(name):
-                        all_functions[name] = (signature, description, group)
+                        description, brief, params = process_commands(match.group(2).replace('#', ''), name, group, signature)
+                        all_functions[name] = (group, signature, brief, description, params)
```
lurch, line 123: Probably makes sense to use a dictionary here rather than a tuple, which would simplify the file-writing loop.

src/common/pico_binary_info/CMakeLists.txt
```diff
@@ -10,14 +10,18 @@ endif()
 target_link_libraries(pico_binary_info INTERFACE pico_binary_info_headers)
 
 # pico_set_program_name(TARGET name)
-# Set the program name for the target
+# \brief\ Set the program name for the target
+#
+# \param\ name The program name to set
```
lurch, line 15: Why is there a `\param\ name` but not a `\param\ TARGET` ?

tools/extract_cmake_functions.py
will-v-pi: I was going for similar commands but different syntax to doxygen (hence the second slash), because the parsing is entirely different and will not handle the vast majority of doxygen stuff, so I wanted to prevent copy/pasting but still have similar commands

src/common/pico_binary_info/CMakeLists.txt
will-v-pi: I removed all the TARGET params as it means the same everywhere so I don't think it needs to be documented for every function, and loads of CMake functions have TARGET (eg `add_executable` which is needed for every executable) so it should be clear enough

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,137 @@
...
+        if filename == 'CMakeLists.txt' or file_ext == '.cmake':
+            file_path = os.path.join(dirpath, filename)
+
+            with open(file_path, encoding="ISO-8859-1") as fh:
+                text = fh.read()
+                for match in CMAKE_FUNCTION_RE.finditer(text):
+                    name = match.group(4)
+                    signature = match.group(1).strip()
+                    if signature.startswith(name):
+                        description, brief, params = process_commands(match.group(2).replace('#', ''), name, group, signature)
```
lurch, line 122: One of the descriptions has a genuine `#` inside it (`# To add additional directories to the gatt #import path, add them to the end of the argument list.`), so IMHO the `.replace('#', '')` here ought to be moved inside the line-by-line processing in `process_commands` and then you could do something like `line = re.sub('^\s+#', '', line)` to only remove the `#` from the start of lines.

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,137 @@
...
+# Supported commands:
+# \brief\ <brief description, which should be included in the main description>
+# \brief_nodesc\ <brief description, which should be excluded from the main description>
+# \param\ <parameter_name> <parameter description>
+#
+# Commands in the middle of a line are not supported
+def process_commands(description, name, group, signature):
+    brief = ''
+    params = []
+    ret = ''
```
lurch, line 68: I guess `ret` here is actually `desc`? (which might make the code easier to read)

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,137 @@
...
+                ret += brief + '\n'
+            elif command == 'brief_nodesc':
+                # Brief description which should not be included in the main description
+                brief = line.split('\\')[2].strip()
+            else:
+                logger.error("{}:{} has unknown command: {}".format(group, name, command))
+        elif '\\' in line:
+            logger.error("{}:{} has a line containing '\\': {}".format(group, name, line))
+        else:
+            ret += line + '\n'
```
lurch, line 88: Could I request that this becomes `ret += line + ' '` ? Whilst our current tooling works fine with actual-newlines inside the middle of a field in a TSV file, it feels kinda non-portable? And I don't think we actually need the real-newlines to be preserved in the descriptions?

lurch: If you _do_ want to preserve hard line-breaks inside TSV fields (i.e. those lines that you're currently ending with `;`, IMHO it's better to replace them with a literal `\n` (i.e. a backlash character followed by a `n`). Note that the `TsvDictReader` that you're using in the other PR already has special-case handling for literal `\n` within a field, and automatically converts it back into an actual line-break character (which I guess you'd then want to escape to ` +\n` for the AsciiDoc output).

tools/extract_cmake_functions.py
will-v-pi: Yeah, the script should throw an error if the command isn't known, which would happen if it didn't have the second backslash
```python
logger.error("{}:{} has unknown command: {}".format(group, name, command))
```

tools/extract_cmake_functions.py
will-v-pi: I think I'll go with `ret += line + '\\n'` then (and `ret += brief + '\\n'` above), as the double line breaks still need to be preserved, so I think it's easier to preserve all line breaks in the TSV and then the other tooling still knows where all the line breaks are

lurch: Cool :+1:

tools/extract_cmake_functions.py
will-v-pi: I think it's simpler to just not support `#` in the descriptions, so that would make `#`, `;` and `\` unsupported/special characters (I'll add a note in the script) - I'll remove that one, as it's not really needed

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,166 @@
...
+                    if name not in all_functions and name not in allowed_missing_functions:
+                        logger.warning("{} function has no description in {}".format(name, file_path))
+
+
+with open(outfile, 'w', newline='') as csvfile:
+    fieldnames = ('name', 'group', 'signature', 'brief', 'description', 'params')
+    writer = csv.DictWriter(csvfile, fieldnames=fieldnames, extrasaction='ignore', dialect='excel-tab')
+
+    writer.writeheader()
+    for name, row in sorted(all_functions.items(), key=sort_functions):
```
lurch, line 165: `name` isn't used here, so I guess you could even simplify this to `for row in sorted(all_functions.values(), key=sort_functions):` ?

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,166 @@
...
+                            'group': group,
+                            'signature': signature,
+                            'brief': brief,
+                            'description': description,
+                            'params': params
+                        }
+
+                for match in CMAKE_PICO_FUNCTIONS_RE.finditer(text):
+                    name = match.group(1)
+                    if name not in all_functions and name not in allowed_missing_functions:
```
lurch, line 156: Perhaps this should log a warning if the same `name` is seen twice?

src/rp2_common/pico_btstack/CMakeLists.txt
```diff
@@ -302,9 +302,15 @@ if (EXISTS ${PICO_BTSTACK_PATH}/${BTSTACK_TEST_PATH})
 
     pico_promote_common_scope_vars()
 
-    # Make a GATT header file from a BTstack GATT file
-    # Pass the target library name library type and path to the GATT input file
+    # pico_btstack_make_gatt_header(TARGET_LIB TARGET_TYPE GATT_FILE)
+    # \brief\ Make a GATT header file from a BTstack GATT file.
+    #
+    # Pass the target library name, library type, and path to the GATT input file.
     # To add additional directories to the gatt #import path, add them to the end of the argument list.
```
lurch, line 309: You said you'd remove the extra `#` here? :slightly_smiling_face:

tools/extract_cmake_functions.py
will-v-pi: The same name will be seen twice for boot stage 2 stuff, as the functions currently appear twice - so probably a warning if the same name is seen again with different descriptions

will-v-pi: Have added this, so it'll warn with the file path of the new one and use the new definition if the `name` appears again

lurch:
Looking at the latest TSV file that this generates, each row is indeed now a single line ( :100: ), but there seems to be multiple `\n\n` at the start and end of each `description`, which seems a bit superfluous?

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,170 @@
...
+    brief = ''
+    params = []
+    desc = ''
+    for line in description.split('\n'):
+        line = line.strip()
+        if line.startswith('\\'):
+            command = line.split('\\')[1]
+            if command == 'param':
+                # Parameter name and description
+                params.append(line.split('\\')[2].strip())
```
lurch, line 78: You've got `line.split('\\')[2].strip()` appearing multiple times, so perhaps it's better to do something like:
```python3
_, command, remainder = line.split('\\', 2)
remainder = remainder.strip()
```

lurch:
IMHO the extracted TSV file is looking much better now, thanks :+1: (If you've not already seen it, there's a really nice "Rainbow CSV" extension for VSCode which gives the different columns different colours, which is pretty handy!)

But the descriptions still have things like:
```
... included)\n\nDEFAULT_ALARM_POOL    - default alarm pool setup;\nPRINTF                - full printf ...
```
so perhaps it's worth doing `desc = re.sub(' +', ' ', desc)` too?

tools/CMakeLists.txt
```diff
@@ -269,10 +283,19 @@ function(pico_load_map_clear_sram TARGET)
...
+# rollback version, along with the rollback rows.
+#
+# These are appended as arguments to the target property PICOTOOL_EXTRA_PROCESS_ARGS if setting the
+# rollback version, or set as compile definitions if only setting the major/minor
+# versions.
+#
+# \param\ MAJOR The major version to set
+# \param\ MINOR The minor version to set
+# \param\ ROLLBACK The rollback version to set
+# \param\ ROWS The OTP rows to use for the rollback version
```
lurch, line 298: While looking at the documentation generated from this PR, I just noticed that you have `\param\ ROWS` here, but `[ROLLBACK_ROWS <rows...>]` in the signature line!
Here's a possible fix to `extract_cmake_functions.py` which catches that:
```diff
diff --git a/tools/extract_cmake_functions.py b/tools/extract_cmake_functions.py
index da87fc6..d57d367 100755
--- a/tools/extract_cmake_functions.py
+++ b/tools/extract_cmake_functions.py
@@ -94,9 +94,11 @@ def process_commands(description, name, group, signature):
     if any([';' in x for x in params]):
         logger.error("{}:{} has a parameter description containing ';'".format(group, name))
     # Check that all parameters are in the signature
+    signature_words = set(re.split('\W+', signature))
     for param in params:
-        if param.split(' ')[0] not in signature:
-            logger.error("{}:{} has a parameter {} which is not in the signature {}".format(group, name, param.split(' ')[0], signature))
+        param_name = param.split(' ', maxsplit=1)[0]
+        if param_name not in signature_words:
+            logger.error("{}:{} has a parameter {} which is not in the signature {}".format(group, name, param_name, signature))
     # Check that the brief description is not empty
     if not brief:
         logger.warning("{}:{} has no brief description".format(group, name))
```
:slightly_smiling_face:

lurch (dismissed):
Brilliant, thanks for taking all my feedback on-board :slightly_smiling_face:

lurch:
Could you rebase this onto the latest `develop` branch please?

lurch (dismissed):

tools/extract_cmake_functions.py
```diff
@@ -54,12 +54,38 @@
...
+    'pico_pio': ('Pico PIO', 'CMake functions to generate PIO headers'),
+    'other': ('Other', 'Other CMake functions'),
+}
+
+
 all_functions = {}
 
+for group, (brief, description) in group_names_descriptions.items():
+    all_functions['_desc_{group}'.format(group=group)] = {
+        'name': '_desc_{group}'.format(group=group),
```
lurch, line 76: Haha, sneaky :wink:

tools/CMakeLists.txt
```diff
@@ -268,11 +283,20 @@ function(pico_load_map_clear_sram TARGET)
     )
 endfunction()
 
-# pico_set_binary_version(<TARGET> [MAJOR <version>] [MINOR <version>] [ROLLBACK <version>] [ROLLBACK_ROWS <rows...>])
+# pico_set_binary_version(<TARGET> [MAJOR <version>] [MINOR <version>] [ROLLBACK <version>] [ROWS <rows...>])
```
lurch, line 286: I've just noticed that this one says `pico_set_binary_version(<TARGET> ...)` wheres all the other CMake signatures say e.g. `pico_set_otp_key_output_file(TARGET ...)` ?

lurch (approved):

tools/extract_cmake_functions.py
```diff
@@ -0,0 +1,207 @@
...
+    "pico_init_picotool",
+    "pico_add_platform_library",
+    "pico_get_runtime_output_directory",
+    "pico_set_printf_implementation",
+    "pico_expand_pico_platform",
+])
+
+# Group descriptions
+group_names_descriptions = {
+    'boot_stage2': ('Boot Stage 2', 'CMake functions to create stage 2 bootloaders'),
```
kilograham, line 59: I had thought to add these in the CMake files themselves, but this is certainly fine for now

PR #2423, opened 2025-04-15
Make doxygen buildable again

lurch:
Unfortunately in #2411 @peterharperuk used a doxygen `\par` keyword which isn't supported by our [`doxygentoasciidoc`](https://github.com/raspberrypi/doxygentoasciidoc) repo, and this broke the building of both https://www.raspberrypi.com/documentation/pico-sdk/ and https://datasheets.raspberrypi.com/pico/raspberry-pi-pico-c-sdk.pdf

@mudge It fails with:
```
....
  File ".../lib/doxygentoasciidoc/nodes.py", line 223, in nodefor
    return {
KeyError: 'title'
```

Removing these "erroneous" 3 lines from `pico_lwip/doc.h` is probably easier than fixing up the rest of our doxygen pipeline :wink: 
And I don't _think_ that this change has too much of an impact on the `pico_lwip_http` documentation?

(ping also @will-v-pi as he pointed out the problem to me)

### DIFF
--- a/src/rp2_common/pico_lwip/doc.h
+++ b/src/rp2_common/pico_lwip/doc.h
@@ -51,18 +51,12 @@
  *
  * This library enables you to make use of the LwIP HTTP client and server library
  *
- * \par LwIP HTTP server
- *
  * To make use of the LwIP HTTP server you need to provide the HTML that the server will return to the client.
  * This is done by compiling the content directly into the executable.
- *
- * \par makefsdata
  *
  * LwIP provides a c-library tool `makefsdata` to compile your HTML into a source file for inclusion into your program.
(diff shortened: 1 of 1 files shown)


peterharperuk (approved):
Odd, I remember testing this.

lurch:
> Odd, I remember testing this.

Yeah, the plain doxygen build itself works (i.e. `make docs` from within the CMake `build` directory), but it's the separate `doxygentoasciidoc` stage where it fails (see https://github.com/raspberrypi/documentation/blob/develop/CONTRIBUTING.md#build-the-pico-c-sdk-doxygen-documentation )

mudge:
See also https://github.com/raspberrypi/doxygentoasciidoc/pull/9 which would add support for `\par` in doxygentoasciidoc.

PR #2431, opened 2025-04-18
Use cyw43_delay_ms() in cyw43_spi_reset() instead of sleep_ms()

mbrase:
Since `cyw43_spi_reset()` may be executed from an async context, we should use `cyw43_delay_ms()` instead of `sleep_ms()`. This is particularly a problem when using the `async_context_threadsafe_background` backend, because `sleep_ms()` will assert in an ISR.

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c
@@ -359,9 +359,9 @@ void cyw43_spi_gpio_setup(void) {
 // Reset wifi chip
 void cyw43_spi_reset(void) {
     gpio_put(CYW43_PIN_WL_REG_ON, false); // off
-    sleep_ms(20);
+    cyw43_delay_ms(20);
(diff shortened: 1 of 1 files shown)


mbrase:
For additional context, this is the code that I'm running that crashes without this patch:
```cc
#include "pico/cyw43_arch.h"

void StartBlinkingLed() {
  static bool state = false;
  static async_at_time_worker_t worker = {
      .next = nullptr,
      .do_work =
          [](async_context_t* context, async_at_time_worker_t* worker) {
            bool& state = *static_cast<bool*>(worker->user_data);
            state = !state;
            cyw43_arch_gpio_put(CYW43_WL_GPIO_LED_PIN, state);
            async_context_add_at_time_worker_in_ms(context, worker, 1000);
          },
      .next_time = 0,
      .user_data = &state,
  };
  cyw43_arch_init();
  async_context_add_at_time_worker_in_ms(
      cyw43_arch_async_context(), &worker, 50);
}
```

peterharperuk (approved):
I think this should be ok. Ran some tests and everything worked.

PR #2435, opened 2025-04-20
Add board header for Waveshare RP2350-USB-A

### LINKED ISSUE #2434: Missing board header for Waveshare RP2350-USB-A
[This board](https://www.waveshare.com/wiki/RP2350-USB-A) combines an RP2350A with a USB-A port that can be used with a bit-banged implementation of the USB protocol such as [Pico-PIO-USB](https://github.com/sekigon-gonnoc/Pico-PIO-USB).

sonodima:
This PR introduces support for the [Waveshare RP2350-USB-A](https://www.waveshare.com/wiki/RP2350-USB-A).

This board combines an RP2350A with a USB-A port that can be used with a bit-banged implementation of the USB protocol such as [Pico-PIO-USB](https://github.com/sekigon-gonnoc/Pico-PIO-USB).

Fixes #2434


src/boards/include/boards/waveshare_rp2350_usb_a.h
```diff
@@ -0,0 +1,97 @@
...
+#ifndef PICO_DEFAULT_SPI_RX_PIN
+#define PICO_DEFAULT_SPI_RX_PIN 4
+#endif
+#ifndef PICO_DEFAULT_SPI_CSN_PIN
+#define PICO_DEFAULT_SPI_CSN_PIN 5
+#endif
+
+// --- PIO USB ---
+#ifndef PICO_DEFAULT_PIO_USB_DP_PIN
+#define PICO_DEFAULT_PIO_USB_DP_PIN 12
```
lurch, line 74: There's no SDK support for this, so IMHO prefixing it with `PICO_DEFAULT_` is a bit misleading. Perhaps `WAVESHARE_RP2350_USB_A_PIO_USB_DP_PIN` would be better?
Also, if you're defining the `PIO_USB_DP_PIN`, would it make sense to also define the `PIO_USB_DM_PIN` ??

kilograham: Is this define used by Pico-PIO-USB, if so im fine with it

kilograham: ok, i guess it doesn't ... i'd go with `PIO_USB_DP_PIN` and then maybe encourage Pico-PIO-USB to provide a confif that uses those along with 

```
#ifndef PIO_USB_DP_PIN
#define PIO_USB_DP_PIN PIO_USB_DP_PIN_DEFAULT
#endif

lurch: Doesn't look like it? https://github.com/search?q=repo%3Asekigon-gonnoc%2FPico-PIO-USB%20PICO_DEFAULT_PIO_USB_DP_PIN&type=code

sonodima: That macro is the one used in TinyUSB for the Pico PIO implementation.

I agree it may be a bit misleading. For further context, as far as I know, the only other RP2040 board with a USB-A port is the Adafruit RP2040 USB Host, and it doesn’t define anything specific for the USB-A port in its board header.

TinyUSB sets the default pins for that board [here](https://github.com/hathach/tinyusb/blob/master/hw/bsp/rp2040/boards/adafruit_feather_rp2040_usb_host/board.h).

It might be worth adding the custom board header to the TinyUSB repo instead?

ramseymcgrath: Any chance we could move that bit into a seperate PR and merge the rest?

kilograham: sure, go ahead

lurch:
> i'm confused as to the status of this now

@sonodima appears to have gone quiet, so I've just pushed an additional commit to this PR which implements my earlier suggestion.

sonodima:
> @sonodima appears to have gone quiet, so I've just pushed an additional commit to this PR which implements my earlier suggestion.

I'm sorry, I somehow missed the updates on this issue
I believe that yours is the right choice 👍🏼

PR #2436, opened 2025-04-22
async_context_freertos: Add support for configSUPPORT_STATIC_ALLOCATION

mbrase:
The implementation of async_context_freertos currently assumes that FreeRTOS has been configured with `configSUPPORT_DYNAMIC_ALLOCATION`, which causes it to allocate semaphores, timers and tasks from the heap. However, some projects may prefer `configSUPPORT_STATIC_ALLOCATION`, which requires memory to be allocated ahead of time. This change allows async_context_freertos to support either static or dynamic allocation.

The way this works is when `configSUPPORT_STATIC_ALLOCATION` is enabled, `async_context_freertos` struct will reserve extra space for the static objects (e.g. `StaticSemaphore_t`) and it will prefer to use the static creation functions (e.g. `xSemaphoreCreateBinaryStatic()`). For the task creation, the user will be responsible for allocating the stack memory and passing it in through the config. If a stack is not provided, then it will fallback to using dynamic allocation for the stack (e.g. `xTaskCreate()`).

### DIFF
--- a/src/rp2_common/pico_async_context/async_context_freertos.c
+++ b/src/rp2_common/pico_async_context/async_context_freertos.c
@@ -109,6 +109,23 @@ bool async_context_freertos_init(async_context_freertos_t *self, async_context_f
     self->core.type = &template;
     self->core.flags = ASYNC_CONTEXT_FLAG_CALLBACK_FROM_NON_IRQ;
     self->core.core_num = get_core_num();
+#if configSUPPORT_STATIC_ALLOCATION
+    self->lock_mutex = xSemaphoreCreateRecursiveMutexStatic(&self->lock_mutex_buf);
+    self->work_needed_sem = xSemaphoreCreateBinaryStatic(&self->work_needed_sem_buf);
+    self->timer_handle = xTimerCreateStatic( "async_context_timer",       // Just a text name, not used by the kernel.
+                                             portMAX_DELAY,
+                                             pdFALSE,        // The timers will auto-reload themselves when they expire.
+                                             self,
+                                             timer_handler,
+                                             &self->timer_buf);
+    self->task_handle = xTaskCreateStatic( async_context_task,
+                                           "async_context_task",
+                                           config->task_stack_size,
+                                           self,
+                                           config->task_priority,
+                                           config->task_stack,
+                                           &self->task_buf);
+#else
     self->lock_mutex = xSemaphoreCreateRecursiveMutex();
     self->work_needed_sem = xSemaphoreCreateBinary();
     self->timer_handle = xTimerCreate( "async_context_timer",       // Just a text name, not used by the kernel.
@@ -116,12 +133,18 @@ bool async_context_freertos_init(async_context_freertos_t *self, async_context_f
                                     pdFALSE,        // The timers will auto-reload themselves when they expire.
                                     self,
                                     timer_handler);
+#endif
 
     if (!self->lock_mutex ||
         !self->work_needed_sem ||
         !self->timer_handle ||
+#if configSUPPORT_STATIC_ALLOCATION
+        !self->task_handle
+#else
         pdPASS != xTaskCreate(async_context_task, "async_context_task", config->task_stack_size, self,
-                config->task_priority, &self->task_handle)) {
+                config->task_priority, &self->task_handle)
+#endif
+    ) {
         async_context_deinit(&self->core);
         return false;
     }
@@ -179,6 +202,9 @@ void async_context_freertos_lock_check(__unused async_context_t *self_base) {
 typedef struct sync_func_call{
     async_when_pending_worker_t worker;
     SemaphoreHandle_t sem;
+#if configSUPPORT_STATIC_ALLOCATION
+    StaticSemaphore_t sem_buf;
+#endif
     uint32_t (*func)(void *param);
     void *param;
     uint32_t rc;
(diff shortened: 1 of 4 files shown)


src/rp2_common/pico_async_context/async_context_freertos.c
```diff
@@ -109,19 +109,47 @@ bool async_context_freertos_init(async_context_freertos_t *self, async_context_f
...
                                     self,
                                     timer_handler);
+#endif
 
     if (!self->lock_mutex ||
         !self->work_needed_sem ||
         !self->timer_handle ||
-        pdPASS != xTaskCreate(async_context_task, "async_context_task", config->task_stack_size, self,
-                config->task_priority, &self->task_handle)) {
+        !(self->task_handle ||
```
kilograham, line 144: this seems a bit wordy.

Maybe just do 
```
     !self->timer_handle ||
#if configSUPPORT_STATIC_ALLOCATION
         !self->task_handler
#else
          pdPASS != xTaskCreate(async_context_task, "async_context_task", config->task_stack_size, self, ... )
#endif
```
              config->task_priority, &self->task_handle)

src/rp2_common/pico_async_context/include/pico/async_context_freertos.h
```diff
@@ -68,9 +76,21 @@ typedef struct async_context_freertos_config {
 struct async_context_freertos {
     async_context_t core;
     SemaphoreHandle_t lock_mutex;
+#if configSUPPORT_STATIC_ALLOCATION
```
kilograham, line 79: Just for style; i think i'd put all the `_bufs` in a single #ifdef after the handles (I donl think there is any particular code size reason for them being close together)

src/rp2_common/pico_async_context/async_context_freertos.c
mbrase: What I was trying to do here is make it so that if someone has both `configSUPPORT_STATIC_ALLOCATION` and `configSUPPORT_DYNAMIC_ALLOCATION` enabled, and they don't provide a `task_stack`, then it will fallback to using `xTaskCreate()`.  I'm open to suggestions on making this cleaner.

kilograham: Is that a valid configuration? (both defined)

mbrase: I'm not sure if that is valid, and to be fair I am only using `configSUPPORT_STATIC_ALLOCATION` in my project.  Perhaps we should keep this simple until a case arises where someone needs both.  I'll upload a new patch.

src/rp2_common/pico_async_context/include/pico/async_context_freertos.h
mbrase: Done

mbrase:
I pushed a new change that does not try to fall back to `xTaskCreate()` if no stack is provided.  I also modified the `cyw43_arch_init_default_async_context()` function to statically allocate a stack when `configSUPPORT_STATIC_ALLOCATION` is enabled.

src/rp2_common/pico_async_context/async_context_freertos.c
```diff
@@ -109,19 +109,42 @@ bool async_context_freertos_init(async_context_freertos_t *self, async_context_f
...
 
     if (!self->lock_mutex ||
         !self->work_needed_sem ||
         !self->timer_handle ||
-        pdPASS != xTaskCreate(async_context_task, "async_context_task", config->task_stack_size, self,
-                config->task_priority, &self->task_handle)) {
+#if configSUPPORT_STATIC_ALLOCATION
+        !self->task_handle
+#else
+        pdPASS == xTaskCreate(async_context_task, "async_context_task", config->task_stack_size, self,
```
kilograham, line 144: seems like this should be pdPass !=

kilograham, line 144: this needs fixing

kilograham (changes requested):

src/rp2_common/pico_async_context/include/pico/async_context_freertos.h
```diff
@@ -56,6 +56,14 @@ typedef struct async_context_freertos_config {
      * \brief Stack size for the async_context task
      */
     configSTACK_DEPTH_TYPE task_stack_size;
+     /**
+     * \brief Pointer to stack memory for the async_context task.
+     * If this is not provided, then a stack will be allocated from the
+     * freertos heap.
```
peterharperuk, line 62: is this comment still true?

src/rp2_common/pico_cyw43_arch/cyw43_arch_freertos.c
```diff
@@ -25,13 +25,20 @@
 
 static async_context_freertos_t cyw43_async_context_freertos;
 
+#if configSUPPORT_STATIC_ALLOCATION
+static StackType_t cyw43_async_context_freertos_task_stack[CYW43_TASK_STACK_SIZE];
```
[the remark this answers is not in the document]
peterharperuk: If a caller didn't use cyw43_arch_init_default_async_context, this static allocation would be wasted. It should maybe be up to the caller to allocate the stack or we need a way to remove it.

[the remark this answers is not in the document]
peterharperuk: I've pushed a change that adds CYW43_NO_DEFAULT_TASK_STACK just in case anyone is needs it.

peterharperuk:
> I'm curious, is there any reason that LwIP NO_SYS=0 mode can't use static allocation?

There's this assert in LwIP. I didn't look any deeper.

```
#if !configSUPPORT_DYNAMIC_ALLOCATION
# error "lwIP FreeRTOS port requires configSUPPORT_DYNAMIC_ALLOCATION"
#endif

```

peterharperuk (approved):

kilograham (approved):

PR #2441, opened 2025-04-25
Update hellbender_2350A_devboard.h for release version of the board.

### LINKED ISSUE #2440: hellbender_2350A_devboard.h is out of date
The version of the header in the sdk is for a pre-release version of our devboard. A PR with an update will follow.

CBJamo:
We've only used the old version of this board internally so there's no reason to maintain both versions.

Fixes #2440

### DIFF
--- a/src/boards/include/boards/hellbender_2350A_devboard.h
+++ b/src/boards/include/boards/hellbender_2350A_devboard.h
@@ -22,16 +22,17 @@ pico_board_cmake_set(PICO_PLATFORM, rp2350)
 #define HB_2350A_IMU_INT_PIN 1
 #define HB_2350A_USER_QWIIC_SDA_PIN 2
 #define HB_2350A_USER_QWIIC_SCL_PIN 3
-#define HB_2350A_RTC_INT_PIN 8
+#define HB_2350A_RTC_INT_PIN 22
 #define HB_2350A_SDCARD_CS_PIN 9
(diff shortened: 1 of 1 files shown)


lurch:
Can you rebase this onto the latest version of `develop` please? Note that `// pico_cmake_set PICO_PLATFORM=rp2350` has now become `pico_board_cmake_set(PICO_PLATFORM, rp2350)` https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/hellbender_2350A_devboard.h#L15 
(as well as `// pico_cmake_set_default name value` changing to `pico_board_cmake_set_default(name, value)`)

lurch:
...and if I make those changes locally, `tools/check_board_header.py` complains with:
`Exception: src/boards/include/boards/hellbender_2350A_devboard.h:81  PICO_DEFAULT_SPI_CSN_PIN is set to 28 which isn't a valid pin for CSN on SPI 0`
and Table 3 in https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf indeed shows that GPIO28 has SPI function SPI1 RX.

CBJamo:
Rebased, sorry about missing that.

I removed the default cs pin define, the bus is connected to both an expansion header and a micro sd card.

lurch (approved):
Thanks for the update :+1:

PR #2446, opened 2025-05-01
Crashes after calling btstack_cyw43_deinit

peterharperuk:
You can deinitialise cyw43 and btstack by calling btstack_cyw43_deinit but its pending and timeout workers are not removed which means they can keep running, which causes a crash.

Add a btstack_run_loop_async_context_deinit method and call this from btstack_cyw43_deinit.


kilograham (approved):

kilograham:
bit weird that the run loop itself doesnt have a deinit hook; oh well

Issue #2449, opened 2025-05-02
Typo for fractional part of `lposc_freq_khz`?


tommy-gilligan:
Currently:

https://github.com/raspberrypi/pico-sdk/blob/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c/src/rp2_common/hardware_powman/powman.c#L72

I think this line should be referencing `lposc_freq_hz` instead of `lposc_freq_khz`

Given an input value of `lposc_freq_hz = 32768`
`lposc_freq_khz` will be `32768 / 1000 = 32`
This is the integer part
Fractional part cannot be derived from this

~I am not sure if the rest of the maths makes sense either.~

edit: The maths is correct but it does seem the wrong variable is being referenced.

kilograham:
Please feel free to make a PR (for the khz->hz), i think the `65536 / 1000` is good as is, because otherwise you'd be multiplying by 655 not 655.36

tommy-gilligan:
Misread the size of variable.  Without parentheses is better.

lurch:
Fix merged into `develop`

PR #2450, opened 2025-05-02
Fix wrong value used for lposc_freq_khz_frac (#2449)

tommy-gilligan:
Looks like it was just a typo.  The code was attempting to derive the fractional part for lposc_freq_khz from an already truncated integer part.


kilograham (approved):

kilograham:
thx

will-v-pi:
This PR changes the `xosc_freq_...` stuff but not the `lposc_freq_...` stuff - should they both be changed?

kilograham:
> This PR changes the `xosc_freq_...` stuff but not the `lposc_freq_...` stuff - should they both be changed?

ah good catch!

tommy-gilligan:
just saw this >.< updating now

kilograham (approved):

PR #2452, opened 2025-05-02
Add PICO_CRT0_NEAR_CALLS to indicate that crt0 can call runtimine_init/main/exit etc via near calls

kilograham:


src/rp2_common/pico_crt0/crt0.S
```diff
@@ -15,6 +15,11 @@
 #include "boot/picobin.h"
 #include "pico/bootrom.h"
 
+// PICO_CONFIG: PICO_CRT0_NEAR_CALLS, Whether calls from crt0 into the binary are near (<16M away) - ignored for PICO_COPY_TO_RAM, default=9, type=bool, group=pico_crt0
```
will-v-pi, line 18:
```suggestion
// PICO_CONFIG: PICO_CRT0_NEAR_CALLS, Whether calls from crt0 into the binary are near (<16M away) - ignored for PICO_COPY_TO_RAM, default=0, type=bool, group=pico_crt0
```

will-v-pi (commented):
One small typo - rest looks good

will-v-pi (dismissed):

src/rp2_common/pico_runtime/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ endif()
 # FPGA_CHECK            - checks for FPGA which allows Raspberry Pi to run your binary on FPGA;
 # PANIC                 - default panic impl which brings in stdio;
 # AUTO_INIT_MUTEX       - auto init mutexes, without this you get no printf mutex either;
+# CRT0_FAR_CALLS        - use blx not bl far calls from crt0
```
lurch, line 71: Is the "far calls from crt0" here supposed to be "for calls from crt0" ?

kilograham: yes

PR #2453, opened 2025-05-02
Add PICO_CRT0_NO_RESET_SECTION

kilograham:
Needed if you want to replace the startup code without removing pico_crt0 as the reset section is unconditionally included by the linker scripts and changing that seems riskier


will-v-pi (dismissed):

lurch (approved):
Re-approving in William's absence

PR #2458, opened 2025-05-07
Add support for wiznet_w5100s_evb_pico2

scaprile:
This board has an RP2350 and a smaller 2MB flash
Building for this board requires manually setting flash size. There already is a board definition for its RP2040 sibling, why not adding one for this board ?
No related issues

### DIFF
--- a/src/boards/include/boards/wiznet_w5100s_evb_pico2.h
+++ b/src/boards/include/boards/wiznet_w5100s_evb_pico2.h
@@ -17,6 +17,9 @@ pico_board_cmake_set(PICO_PLATFORM, rp2350)
 // For board detection
 #define WIZNET_W5100S_EVB_PICO2
 
+// --- RP2350 VARIANT ---
+#define PICO_RP2350A 1
+
 // --- BOARD SPECIFIC ---
 #ifndef W5100S_EVB_PICO2_INTN_PIN
 #define W5100S_EVB_PICO2_INTN_PIN 21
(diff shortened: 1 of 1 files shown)


src/boards/include/boards/wiznet_w5100s_evb_pico2.h
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
+//       SHOULD ONLY CONSIST OF PREPROCESSOR DIRECTIVES
+// -----------------------------------------------------
+
+// pico_cmake_set PICO_PLATFORM=rp2350
```
lurch, line 12: The syntax of these cmake-specific lines have changed - see https://github.com/raspberrypi/pico-sdk/pull/2397 and https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/wiznet_w5100s_evb_pico.h

lurch:
This is missing a `#define PICO_RP2350A 1` line (see e.g. https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/pico2.h#L22-L23 ) which tells the SDK whether this board is fitted with an RP2350A or an RP2350B.

lurch (approved):
Thanks for the updates, LGTM now!

PR #2465, opened 2025-05-13
Remove warning

kilograham:


kilograham:
> This seems to be accidentally including the change from #2457 too? 🤔 (i.e. does this PR need to be rebased?)

oops; force pushed

lurch:
Removing myself from reviewers list, as I don't know enough about ARM assembly.

PR #2466, opened 2025-05-13
Pico runtime skip init all

kilograham:


src/rp2_common/pico_runtime_init/include/pico/runtime_init.h
```diff
@@ -80,7 +80,7 @@ void runtime_init_bootrom_reset(void);
 #endif
 
 // ---------------------------------------------------------------------------------------
-// Non-boot core eset of bootrom state, not needed if only using core 0 not used on RP2040
+// Non-boot core reset of bootrom state, not needed if only using core 0 not used on RP2040
```
lurch, line 83: Should there be a comma after "core 0" here, to avoid this being misread as "core 0 not used on RP2040" ?

src/rp2_common/pico_runtime/include/pico/runtime.h
```diff
@@ -49,13 +49,26 @@ void runtime_init(void);
 
 void runtime_run_initializers(void);
 void runtime_run_per_core_initializers(void);
+// PICO_CONFIG: PICO_RUNTIME_SKIP_INIT_ALL, Skip calling of `runtime_init_` functions during runtime init, type=bool, default=0, group=pico_runtime_init
```
lurch, line 52: So presumably this "overrides" e.g. `PICO_RUNTIME_SKIP_INIT_POST_CLOCK_RESETS` ?

lurch:
P.S. I just did a `git grep "#define PICO_RUNTIME_INIT_FUNC_HW"` and I see that the `host` and `rp2_common` versions of the macro have different signaures:
```
src/host/pico_runtime/include/pico/runtime.h:#define PICO_RUNTIME_INIT_FUNC_HW(func, priority_string1, priority_string2) PICO_RUNTIME_INIT_FUNC_FLAGS(func, PICO_RUNTIME_INIT_TYPE_HW, priority_string1, priority_string2)
src/rp2_common/pico_runtime/include/pico/runtime.h:#define PICO_RUNTIME_INIT_FUNC_HW(func, priority_string) PICO_RUNTIME_INIT_FUNC(func, priority_string)
```
Is that intentional?

peterharperuk:
Why is this needed?

kilograham:
> Why is this needed?

Good question; it is a bit of a hack, and i was on the fence. you pushed me off - i will remove them by updating the linker script in the encrypted bootloader

peterharperuk:
It wasn't a criticism. I was just wondering

PR #2467, opened 2025-05-14
Fix multicore_lockout features so that the victim core cannot become stuck in an infinite loop if the lockout attempt times out

### LINKED ISSUE #2454: When using flash_safe_execute, if the victim core does not respond within the timeout, it will be locked out forever
My Pico program uses two cores. Here is a sequence of events leading to the bug
- bootup: core 0: call `flash_safe_execute_core_init`
- core 0: start code on core 1
- ...normal operation for a while...
- core 0: enter long-running interrupt handler
- core 1: call `flash_safe_execute` with short timeout parameter
- core 1: call `multicore_lockout_start_block_until`
- core 1: send `LOCKOUT_MAGIC_START` and wait for an acknowledgment
- core 1: timeout reached
- core 1: `multicore_lockout_handler` returns `PICO_ERROR_TIMEOUT` is returned
- core 0: leave long-running interrupt handler and enter `multicore_lockout_handler`

**Observed behaviour**: 
- core 1 continues execution as usual
- core 0 stays within `multicore_lockout_handler` forever, in the following loop:
```
            while (multicore_fifo_pop_blocking_inline() != LOCKOUT_MAGIC_END) {
                tight_loop_contents(); // not tight but endless potentially
            }
```
- Interrupts are disabled, as core 0 is waiting for `multicore_lockout_end_block_until` to be called.
- Pico stops responding to USB and any other interrupts which are serviced on core 0.

**Expected behaviour**:
- core 0: `multicore_lockout_handler` exits quickly as the lockout request has been abandoned

**Analysis**: 
The timeout caused `flash_safe_execute` to abandon the attempt to lock out core 0, so `LOCKOUT_MAGIC_END` is never sent. Core 0 is not aware that the lockout request was abandoned and waits forever.

**Workarounds:**
Core 0 can be recovered by a second attempt to use `flash_safe_execute`. Alternatively, using a very long timeout will avoid the problem.

**Steps to reproduce**
The attached project flash-lockout-timeout-bug.tar.gz contains a small program which reproduces the problem.

The two cores run in loops, with cycle counts in global variables for easier debugging. One core runs `flash_safe_execute` periodically, while the other disables interrupts periodically. When a call to `flash_safe_execute` on core Y coincides with disabled interrupts on core X, `flash_safe_execute` returns `PICO_ERROR_TIMEOUT` (-2). Unfortunately this also causes core X to get stuck in `multicore_lockout_handler`. If core X = core 1, then USB continues to work, and the status can be seen on the serial line (the cycle count for core X stops, while the cycle count for core Y continues to increment). If core X = core 0, then USB stops working (no interrupts) but a hardware debugger can still be used to observe the cycle counts and see where core 0 has become stuck.

jwhitham:
Fixes #2454 

This bug can be triggered if the lockout request times out. If the victim CPU core has interrupts disabled for a long time, it will not respond quickly enough to the lockout request, which times out. However, the request is still pending in the FIFO, and when it is eventually handled by the victim CPU core, that core enters an infinite loop, waiting for a LOCKOUT_MAGIC_END message which never arrives.

I considered the solution of always sending a LOCKOUT_MAGIC_END message even on timeout, but this wouldn't work if the FIFO was already full. I could not use a blocking wait, as this would not respect the timeout, and as far as I can tell, there is no hardware support for clearing the FIFO without also resetting the CPU.

In this PR, the lockout state is controlled by a shared variable. The FIFO is used to begin a lockout and acknowledge it as before. But the end of the lockout is now signalled by updating the shared variable. This ensures that the end of the lockout request is recognised reliably by the victim CPU core, regardless of whether the end was caused by a timeout or whether the lockout completed normally. __wfe and __sev are used to signal updates to the shared variable in order to avoid polling.

The semantics of the `multicore_lockout_end_...` functions change: they will no longer wait for the lockout to end. This is described further in the updated doxygen and in the comments below.

### DIFF
--- a/src/rp2_common/pico_multicore/multicore.c
+++ b/src/rp2_common/pico_multicore/multicore.c
@@ -202,8 +202,10 @@ void multicore_launch_core1_raw(void (*entry)(void), uint32_t *sp, uint32_t vect
     irq_set_enabled(irq_num, enabled);
 }
 
+// A non-zero initialisation value is used in order to reduce the chance of
+// entering the lock handler on bootup due to a 0-word being present in the FIFO
+static volatile uint32_t lockout_request_id = 0x73a8831eu;
 static mutex_t lockout_mutex;
-static volatile uint32_t lockout_request_id = 0;
 
 // note this method is in RAM because lockout is used when writing to flash
 // it only makes inline calls
--- a/src/rp2_common/pico_multicore/multicore.c
+++ b/src/rp2_common/pico_multicore/multicore.c
@@ -202,10 +202,8 @@ void multicore_launch_core1_raw(void (*entry)(void), uint32_t *sp, uint32_t vect
     irq_set_enabled(irq_num, enabled);
 }
 
-#define LOCKOUT_MAGIC_START 0x73a8831eu
-
 static mutex_t lockout_mutex;
-static io_rw_32 lockout_request_id = LOCKOUT_MAGIC_START;
+static io_rw_32 lockout_request_id = 0;
 
 // note this method is in RAM because lockout is used when writing to flash
 // it only makes inline calls
--- a/src/rp2_common/pico_multicore/multicore.c
+++ b/src/rp2_common/pico_multicore/multicore.c
@@ -203,24 +203,26 @@ void multicore_launch_core1_raw(void (*entry)(void), uint32_t *sp, uint32_t vect
 }
 
 #define LOCKOUT_MAGIC_START 0x73a8831eu
-#define LOCKOUT_MAGIC_END (~LOCKOUT_MAGIC_START)
 
 static mutex_t lockout_mutex;
-static bool lockout_in_progress;
+static io_rw_32 lockout_request_id = LOCKOUT_MAGIC_START;
 
 // note this method is in RAM because lockout is used when writing to flash
 // it only makes inline calls
 static void __isr __not_in_flash_func(multicore_lockout_handler)(void) {
     multicore_fifo_clear_irq();
     while (multicore_fifo_rvalid()) {
-        if (sio_hw->fifo_rd == LOCKOUT_MAGIC_START) {
+        uint32_t request_id = sio_hw->fifo_rd;
+        if (request_id == lockout_request_id) {
+            // valid lockout request received
             uint32_t save = save_and_disable_interrupts();
-            multicore_fifo_push_blocking_inline(LOCKOUT_MAGIC_START);
-            while (multicore_fifo_pop_blocking_inline() != LOCKOUT_MAGIC_END) {
-                tight_loop_contents(); // not tight but endless potentially
+            multicore_fifo_push_blocking_inline(request_id);
+            // wait for the lockout to expire
+            while (request_id == lockout_request_id) {
+                // when lockout_request_id is updated, the other CPU core calls __sev
+                __wfe();
             }
             restore_interrupts_from_disabled(save);
-            multicore_fifo_push_blocking_inline(LOCKOUT_MAGIC_END);
         }
     }
 }
@@ -257,7 +259,7 @@ void multicore_lockout_victim_deinit(void) {
     }
 }
 
-static bool multicore_lockout_handshake(uint32_t magic, absolute_time_t until) {
+static bool multicore_lockout_handshake(uint32_t request_id, absolute_time_t until) {
     uint irq_num = SIO_FIFO_IRQ_NUM(get_core_num());
     bool enabled = irq_is_enabled(irq_num);
     if (enabled) irq_set_enabled(irq_num, false);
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_multicore/multicore.c
```diff
@@ -276,22 +278,36 @@ static bool multicore_lockout_handshake(uint32_t magic, absolute_time_t until) {
...
         }
     } while (!rc);
     if (enabled) irq_set_enabled(irq_num, true);
     return rc;
 }
 
+static uint32_t update_lockout_request_id() {
+    // generate new number and then update shared variable
+    uint32_t new_request_id = lockout_request_id + 1;
+    lockout_request_id = new_request_id;
```
lurch, line 292: AFAICT it looks like `lockout_request_id` only gets incremented and never gets decremented - does that mean that theoretically it might eventually overflow?

jwhitham: That's correct, but overflow (with wrap to 0) is ok. The numerical value of the lockout_request_id does not matter, all that is important is that the current lockout request has a different value to any others that may still be pending in the FIFO.

jwhitham:
> Does this PR mean that any of the Doxygen at https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_multicore/include/pico/multicore.h#L418 also needs updating to match, or is that all still "correct"?

You're right, this did need updating. And, I didn't realise this until now, but the semantics of `multicore_lockout_end_blocking` and `multicore_lockout_end_timeout_us` are different now, because these functions do not wait for the lockout to end. They just guarantee that it will end.

I thought about whether it was a good idea to push ahead with this behaviour change. I think it is a good idea, because it's simpler, and I don't see the value in knowing that the end of the lockout has been acknowledged by the victim core. The new design avoids the possibility for `multicore_lockout_end_timeout_us` that "a timeout here will leave the "lockout" functionality in a bad state" (as previously stated in doxygen). Timeouts always leave the lockout functionality in a usable state, both for start and end. I think the old blocking behaviour for the `multicore_lockout_end_...` functions could be preserved but at the cost of a more complex design, and I feel like it wouldn't be justified... however, I may be missing something!

src/rp2_common/pico_multicore/multicore.c
lurch: In that case, perhaps the `LOCKOUT_MAGIC_START` is a bit of a red herring, and it'd be better to remove it entirely and just initialise `lockout_request_id` with `0` ?

jwhitham: Agree, done.

src/rp2_common/pico_multicore/multicore.c
```diff
@@ -202,25 +202,25 @@ void multicore_launch_core1_raw(void (*entry)(void), uint32_t *sp, uint32_t vect
     irq_set_enabled(irq_num, enabled);
 }
 
-#define LOCKOUT_MAGIC_START 0x73a8831eu
-#define LOCKOUT_MAGIC_END (~LOCKOUT_MAGIC_START)
-
 static mutex_t lockout_mutex;
-static bool lockout_in_progress;
+static io_rw_32 lockout_request_id = 0;
```
kilograham, line 206: Small style point, we would probably use `volatile uint32_t` explicitly here vs `io_rw_32` which is really intended for memory mapped IO

src/rp2_common/pico_multicore/multicore.c
```diff
@@ -276,22 +276,36 @@ static bool multicore_lockout_handshake(uint32_t magic, absolute_time_t until) {
...
-        if (word == magic) {
+        if (word == request_id) {
             rc = true;
         }
     } while (!rc);
     if (enabled) irq_set_enabled(irq_num, true);
     return rc;
 }
 
+static uint32_t update_lockout_request_id() {
```
kilograham, line 287: can you add a `(void)` i'm pretty sure without this it makes at least one of the many GCC or Clang compiler versions unhappy

kilograham (changes requested):
Thanks, this looks great except a few nits

src/rp2_common/pico_multicore/multicore.c
jwhitham: Thanks, replaced.

src/rp2_common/pico_multicore/multicore.c
jwhitham: Thanks, fixed.

src/rp2_common/pico_multicore/multicore.c
jwhitham: After further testing I found that beginning at 0 is a bad idea because a 0-value word is received by CPU core 0 via the FIFO after reset. So, core 0 can become stuck in `multicore_lockout_victim_init` until core 1 requests a lockout. To avoid this the number should begin at a non-zero value. I restored the original "magic" number.

kilograham (approved):

PR #2475, opened 2025-05-19
Add printout of picotool location

will-v-pi:
This prints out the location of the picotool executable being used when configuring CMake. Note that for a picotool fetched by the SDK, this may not exist yet, as it is only build during the build step.

For example:
* `Using picotool from /usr/local/bin/picotool` for a system install
* `Using picotool from /home/name/pico-examples/build/_deps/picotool/picotool` for a picotool fetched by the SDK

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -139,6 +139,14 @@ function(pico_init_picotool)
         endif()
     endif()
 
+    if (TARGET picotool AND NOT DEFINED ENV{_PICOTOOL_FOUND_THIS_RUN})
+        # Internal environment variable to prevent printing multiple times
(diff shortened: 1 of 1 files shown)


tools/CMakeLists.txt
```diff
@@ -139,6 +139,14 @@ function(pico_init_picotool)
         endif()
     endif()
 
+    if (TARGET picotool AND NOT DEFINED ENV{_PICOTOOL_FOUND_THIS_RUN})
```
lurch, line 142: Should this also check `picotool_FOUND` ?

will-v-pi: I don't think it needs to - if the `picotool` target has been created, then it should work fine

kilograham (approved):

PR #2480, opened 2025-05-21
Remove accidental comment words in stdio.h

lurch:
Looks like these might have been a copy&paste typo?

### DIFF
--- a/src/rp2_common/pico_stdio/include/pico/stdio.h
+++ b/src/rp2_common/pico_stdio/include/pico/stdio.h
@@ -186,35 +186,35 @@ int stdio_get_until(char *buf, int len, absolute_time_t until);
  */
 int stdio_put_string(const char *s, int len, bool newline, bool cr_translation);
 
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_stdio/include/pico/stdio.h
```diff
@@ -186,35 +186,35 @@ int stdio_get_until(char *buf, int len, absolute_time_t until);
  */
 int stdio_put_string(const char *s, int len, bool newline, bool cr_translation);
 
-/*! \brief stdio_getchar Alias for \ref getchar that definitely does not go thru the implementation
```
robertrico, line 189: I think this one is right, the others need to be renamed not removed.

lurch: Yeah I wasn't sure - I'll wait to see what @kilograham says...
(but if your supposition is correct, we'll still need to change "Alias" to "alias")

kilograham (approved):

Issue #2479, opened 2025-05-21
CRT0.s is tailored for a cm0 not a cm33


tbandtg:
Working on fault handlers for my project(2350) noticed that busfault and usage fault are not available.

from the normal cortex m33.s file below is my patch 

```
--- C:\Work\old.txt
+++ C:\Work\new 3.txt
@@ -29,22 +29,22 @@
 .global __vectors, __VECTOR_TABLE
 __VECTOR_TABLE:
 __vectors:
-.word __StackTop
-.word _reset_handler
-.word isr_nmi
-.word isr_hardfault
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_svcall
-.word isr_invalid // Reserved, should never fire
-.word isr_invalid // Reserved, should never fire
-.word isr_pendsv
-.word isr_systick
+    .word   __StackTop          // Initial Main Stack Pointer
+    .word   _reset_handler      // Reset Handler
+    .word   isr_nmi             // NMI Handler
+    .word   isr_hardfault       // HardFault Handler
+    .word   isr_memmanage       // MemManage Fault Handler
+    .word   isr_busfault        // BusFault Handler
+    .word   isr_usagefault      // UsageFault Handler
+	.word isr_invalid // Reserved, should never fire
+	.word isr_invalid // Reserved, should never fire
+	.word isr_invalid // Reserved, should never fire
+	.word isr_invalid // Reserved, should never fire
+    .word   isr_svcall          // SVCall Handler
+    .word   isr_debugmon        // Debug Monitor Handler
+	.word isr_invalid // Reserved, should never fire
+    .word   isr_pendsv          // PendSV Handler
+    .word   isr_systick         // SysTick Handler
 #if PICO_NO_STORED_VECTOR_TABLE && !PICO_NO_FLASH // note in no flash binary, we only have the single RAM vector table anyway
 #if PICO_NO_RAM_VECTOR_TABLE
 #error Can't specify PICO_NO_STORED_VECTOR_TABLE and PICO_NO_RAM_VECTOR_TABLE
@@ -164,13 +164,15 @@
     bkpt #0
 .endm
 
-// these are separated out for clarity
[...]

PR #2482, opened 2025-05-21
add new M33 processor exception handlers

kilograham:


kilograham:
fixes #2479

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -52,15 +52,26 @@ __vectors:
...
 .word isr_invalid // Reserved, should never fire
 .word isr_invalid // Reserved, should never fire
 .word isr_invalid // Reserved, should never fire
 .word isr_invalid // Reserved, should never fire
+#else
+.word isr_memmanage
+.word isr_busfault
+.word isr_usagefault
+.word isr_securefault
+#endif
```
lurch, line 65: Minor nitpick, but it kinda feels like `isr_memmanage` is a "specific" action, and `isr_invalid` is a "default" action, so I wonder if it might make more sense to write this as:
```c
#if PICO_RP2350
.word isr_memmanage
.word isr_busfault
.word isr_usagefault
.word isr_securefault
#else
.word isr_invalid // Reserved, should never fire
.word isr_invalid // Reserved, should never fire
.word isr_invalid // Reserved, should never fire
.word isr_invalid // Reserved, should never fire
#endif
```

? :shrug:

kilograham: convention is based on what is the more likely path in the future... we assume RP2040 is the minority

Issue #2163, opened 2025-01-06
pioasm: allow optional target for .wrap


magy00:
It could sometimes be convenient (and possibly allow more descriptive source code) if `.wrap` would optionally accept the same target expressions as `jmp` (with `.wrap_target` as the default):

    .wrap (<target>)

E.g.

```
loop:
    mov osr, rxfifo[y]
    ...
    jmp y--, loop
    .wrap loop

```
would be equivalent to
```
.wrap_target
loop:
    mov osr, rxfifo[y]
    ...
    jmp y--, loop
    .wrap

```

magy00:
Most of above proposed changes also summarized in #2484 (with some contrived tests in `amethyst.pio`).

PR #2484, opened 2025-05-22
Add target for .wrap, dot symbol, .word as instruction (#2163)

magy00:
Diff as suggested in #2163. Requires rebuild of flex/bison targets and `pioasm` docs updates like e.g.:

page 58 from RP-009085-KB-2-raspberry-pi-pico-c-sdk

### DIFF
--- a/tools/pioasm/lexer.ll
+++ b/tools/pioasm/lexer.ll
@@ -220,6 +220,8 @@ output_fmt    [^%\n]+
 
 {comment}                           { }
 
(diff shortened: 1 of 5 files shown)


magy00:
Note that the "." (dot) symbol was added at the end of the lexer (and parser) list where previously an _invalid character_ syntax error would have been printed (and so should be fully backwards compatible like all other changes).

`.word` behaves similar as in MicroPython (all values are simply ORed).

PR #2490, opened 2025-05-29
Use cyw43 macros for bluetooth logging

### LINKED ISSUE #2253: Tidy up cybt_printf
It looks like you can't disable cybt_printf in all source files. Maybe just reuse the cyw43 driver logging if that's more sensible.

peterharperuk:
CYW43_PRINTF is used for errors
CYW43_DEBUG is for messages in debug builds only
CYW43_VDEBUG is disabled by default

Fixes #2253


kilograham:
not yet convinced as CYW43_PRINTF seems to be on by default (?) which means we're now getting a bunch **more** logging

peterharperuk:
I've replaced calls to cybt_printf with calls to CYW43_PRINTF. So as far as I can see this should NOT produce more logging. I don't see extra logging when I test bluetooth.

kilograham:
I'm confused (from cyw43_config.h)

```
#ifndef CYW43_PRINTF
#include <stdio.h>
#define CYW43_PRINTF(...) printf(__VA_ARGS__)
#endif

#ifndef CYW43_VDEBUG
#define CYW43_VDEBUG(...) (void)0
#define CYW43_VERBOSE_DEBUG 0
#endif

#ifndef CYW43_DEBUG
#ifdef NDEBUG
#define CYW43_DEBUG(...) (void)0
#else
#define CYW43_DEBUG(...) CYW43_PRINTF(__VA_ARGS__)
#endif
#endif

#ifndef CYW43_INFO
#define CYW43_INFO(...) CYW43_PRINTF(__VA_ARGS__)
#endif

#ifndef CYW43_WARN
#define CYW43_WARN(...) CYW43_PRINTF("[CYW43] " __VA_ARGS__)
#endif
```

Which is why i thought CYW43_PRINTF **is** supposed to print, and you should use the other macros to route to it.

I don't see anything in pico-sdk at least that disables CYW43_PRINTF by default

kilograham:
Note also that the cyw43_driver only uses it directly in a `dump_` debugging method

peterharperuk:
You're saying that my change enables more logging. I'm just telling you that it behaves the same.

peterharperuk:
ok, you're right cybt_printf was off in release.

peterharperuk:
Second attempt...

src/rp2_common/pico_cyw43_driver/cybt_shared_bus/cybt_logging.h
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright (c) 2025 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef CYBT_LOGGING_H
+#define CYBT_LOGGING_H
+
+// Error messages only enabled in debug by default
+#ifndef NDEBUG
```
kilograham, line 11: should be guarded by `#ifndef CYBT_ERROR_ENABLED` or it isn't a default!

kilograham: same for the others

Issue #1471, opened 2023-08-10
Possible lockup-issue with `i2c_write_blocking` versus `i2c_write_timeout_us`


ZodiusInfuser:
Hi, I am working on a Pimoroni RP2040 product that involves regular communication with some onboard I2C devices at 400KHz.

During my testing I was encountering frequent lock-ups after a call to `i2c_write_blocking` had completed. I checked this by setting a pin high before the call and to low afterwards. The pin would go low, suggesting the function had completed, then the lockup would occur before reaching my next line of code.

This was very strange to me, as I had successfully got the product working with a custom build of CircuitPython, which I had written C level drivers for. Checking how they implement I2C, I saw that `i2c_write_timeout_us` was used, with a 1 second timeout.

Switching to this call in my code the lock-ups went away, without any timeouts actually getting triggered. Checking the pico-sdk, I see that both functions call `i2c_write_blocking_internal`, with the latter having a callback that results in the below if statement getting executed.

https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/hardware_i2c/i2c.c#L166-L172

Suspecting that the longer loop time the timeout call was having a positive effect, I added a `sleep_us(10)` as an else condition, and sure enough the lock-ups from `i2c_write_blocking_internal` go away. Switching to a debug build also made the issue go away, thus ruling out SWD debugging.

Also, sending CTRL-C via the USB terminal would break the board out of the lock-up, though devices like PIO would not resume.

I am completely baffled by this! 

For now I can switch my code to using the timeout function, and carry on. I just thought I would raise it in case there is an actual issue that needs resolving.

P.S: It may be inconsequential, but I have observed that my "debug" pin goes low before I2C comms complete, for the `i2c_write_blocking` code path:
![image](https://github.com/raspberrypi/pico-sdk/assets/7437269/62c711e8-e8be-4015-bbdb-26f5b2d061c2)

Whereas my "debug" pin goes low after the I2C comms complete, for the `i2c_write_timeout_us` code path:
![image](https://github.com/raspberrypi/pico-sdk/assets/7437269/9ce6fb94-b152-4be5-87eb-399a5da37301)

ZodiusInfuser:
[...]

PR #2493, opened 2025-05-30
i2c abort change

peterharperuk:
Elsewhere we just read to clear clr_tx_abrt if the abort reason is set.

Speculative fix for #1471


peterharperuk:
Going to close this - it didn't seem to make any difference.

Issue #188, opened 2021-02-25
Add pico_status_led library


kilograham:
The 1.0.2 release removed the requirement for PICO_DEFAULT_LED_PIN since not all boards have a LED
Equally PICO_DEFAULT_WS2812_PIN was added.

Right now user code has to check for the existence of the pins, set them up etc. We should add a friendly library which is a no-op if none are defined, which also handles output on common pin types (e.g. WS2812).

lurch:
> Equally PICO_DEFAULT_WS2812_PIN was added.

Was it? :confused:

hathach:
great news, it is worth mentioning that some board also has dedicate pin for neopixel power enable as well.

kilograham:
> > Equally PICO_DEFAULT_WS2812_PIN was added.
> 
> Was it? confused
it isn't committed yet

PR #2501, opened 2025-06-03
add pico_status_led

peterharperuk:
Add a status led library to make adapting to a "normal" led and the cyw43 led a bit easier.
Bonus support for ws2812 leds.

### DIFF
--- a/src/rp2_common/pico_status_led/include/pico/status_led.h
+++ b/src/rp2_common/pico_status_led/include/pico/status_led.h
@@ -72,10 +72,10 @@ extern "C" {
 #define PICO_COLORED_STATUS_LED_COLOR_FROM_WRGB(w, r, g, b) (((w) << 24) | ((r) << 16) | ((g) << 8) | (b))
 #endif
 
-// PICO_CONFIG: PICO_DEFUALT_COLORED_STATUS_LED_ON_COLOR, the default pixel color value of the colored status LED when it is on, type=int, group=pico_status_led
-#ifndef PICO_DEFUALT_COLORED_STATUS_LED_ON_COLOR
+// PICO_CONFIG: PICO_DEFAULT_COLORED_STATUS_LED_ON_COLOR, the default pixel color value of the colored status LED when it is on, type=int, group=pico_status_led
+#ifndef PICO_DEFAULT_COLORED_STATUS_LED_ON_COLOR
 #if PICO_COLORED_STATUS_LED_USES_WRGB
-#define PICO_DEFUALT_COLORED_STATUS_LED_ON_COLOR PICO_COLORED_STATUS_LED_COLOR_FROM_WRGB(0xaa, 0, 0, 0)
+#define PICO_DEFAULT_COLORED_STATUS_LED_ON_COLOR PICO_COLORED_STATUS_LED_COLOR_FROM_WRGB(0xaa, 0, 0, 0)
 #else
 #define PICO_DEFAULT_COLORED_STATUS_LED_ON_COLOR PICO_COLORED_STATUS_LED_COLOR_FROM_RGB(0xaa, 0xaa, 0xaa)
 #endif
@@ -118,7 +118,7 @@ bool status_led_init_with_context(struct async_context *context);
  * \return true if the colored status LED API is available and expected to produce visible results
  * \sa PICO_COLORED_STATUS_LED_AVAILABLE
  */
-static inline bool colored_status_led_supported() {
+static inline bool colored_status_led_supported(void) {
     return PICO_COLORED_STATUS_LED_AVAILABLE;
 }
 
@@ -127,7 +127,7 @@ static inline bool colored_status_led_supported() {
  * \return true if the olored status LED is being used for the non-colored `status_led_` API
  * \sa PICO_STATUS_LED_VIA_COLORED_STATUS_LED
  */
-static inline bool status_led_via_colored_status_led() {
+static inline bool status_led_via_colored_status_led(void) {
     return PICO_STATUS_LED_VIA_COLORED_STATUS_LED;
 }
 
@@ -138,7 +138,7 @@ static inline bool status_led_via_colored_status_led() {
  * \sa PICO_STATUS_LED_AVAILABLE
  * \sa PICO_STATUS_LED_VIA_COLORED_STATUS_LED
  */
-static inline bool status_led_supported() {
+static inline bool status_led_supported(void) {
     if (status_led_via_colored_status_led()) {
         return colored_status_led_supported();
     }
@@ -162,7 +162,7 @@ bool colored_status_led_set_state(bool led_on);
  *
  * \return true if the colored status LED is on, or false if the colored status LED is off
  */
-bool colored_status_led_get_state();
+bool colored_status_led_get_state(void);
     
 /*! \brief Ensure the colored status LED is on, with the specified color
  *  \ingroup pico_status_led
--- a/src/rp2_common/pico_status_led/status_led.c
+++ b/src/rp2_common/pico_status_led/status_led.c
@@ -6,7 +6,7 @@
 
 #include "pico/status_led.h"
 
-#if PICO_STATUS_LED_AVAILABLE && defined(CYW43_WL_GPIO_LED_PIN)
+#if PICO_STATUS_LED_AVAILABLE && defined(CYW43_WL_GPIO_LED_PIN) && !defined(PICO_DEFAULT_LED_PIN)
 #define STATUS_LED_USING_WL_GPIO 1
 #else
 #define STATUS_LED_USING_WL_GPIO 0
@@ -18,7 +18,7 @@
 #define STATUS_LED_USING_GPIO 0
 #endif
 
-#if PICO_COLORED_STATUS_LED_HW_AVAILABLE && defined(PICO_DEFAULT_WS2812_PIN)
+#if PICO_COLORED_STATUS_LED_AVAILABLE && defined(PICO_DEFAULT_WS2812_PIN)
 #define COLORED_STATUS_LED_USING_WS2812_PIO 1
 #else
 #define COLORED_STATUS_LED_USING_WS2812_PIO 0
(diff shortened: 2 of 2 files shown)


src/rp2_common/pico_status_led/include/pico/status_led.h
```diff
@@ -0,0 +1,121 @@
...
+#if defined PICO_STATUS_LED_WS2812_PIN
+
+/*! \brief Set the ws2812 status led on or off
+ *  \ingroup pico_status_led
+ *
+ * \param led_on True to turn the led on. Pass False to turn the led off
+ */
+void pico_status_led_ws2812_set(bool led_on);
+
+/*! \brief Set the ws2812 status led pixel value when on or off
```
lurch, line 58: "status led pixel value" -> "status led colour" ?
(Also, should WS2812 be capitalised here and elsewhere?)

src/rp2_common/pico_status_led/include/pico/status_led.h
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * \return True if the led is on, or False if the led is off
+ */
+static inline bool pico_status_led_get(void) {
+#if defined PICO_STATUS_LED_PIN
+    return gpio_get(PICO_STATUS_LED_PIN);
+#elif defined PICO_STATUS_LED_WS2812_PIN
+    return pico_status_led_ws2812_get();
+#elif defined CYW43_WL_GPIO_LED_PIN
+    // Ask the wifi "driver" to set the GPIO on or off
```
lurch, line 103: I guess this comment needs tweaking, given that this is the `_get` function :wink:

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -0,0 +1,84 @@
...
+#elif defined(CYW43_WL_GPIO_LED_PIN)
+    rc = cyw43_arch_init();
+#endif
+#if PICO_STATUS_LED_WS2812_PIN
+    ws2812_pixel[0] = PICO_STATUS_LED_WS2812_DEFAULT_PIXEL_OFF;
+    ws2812_pixel[1] = PICO_STATUS_LED_WS2812_DEFAULT_PIXEL;
+    if (!pio_claim_free_sm_and_add_program_for_gpio_range(&ws2812_program, &pio, &sm, &offset, PICO_STATUS_LED_WS2812_PIN, 1, true)) {
+        rc = PICO_ERROR_INSUFFICIENT_RESOURCES;
+    }
+    ws2812_program_init(pio, sm, offset, PICO_STATUS_LED_WS2812_PIN, PICO_STATUS_LED_WS2812_FREQ, false);
```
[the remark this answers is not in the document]
peterharperuk: I'd have to support setting the white pixel. So I don't want to support it if it's not needed.

kilograham:
I think the color setting should be orthogonal to the On/Off and the color method should not mention WS2812 in their name - they should probably (and all the methods really) have a bool which says whether they do anything (so you can still use the library with no LEDs at all)

I'm not quite sure about the color APIs right now - somewhat leaning towards uint32_t so you can read, and allow r,g,b or w,r,g,b methods - maybe make it a #define what WRGB LEDs do when you set r,g,b (either 0 for WRGB or auto set)

lurch:
> as most boards seem to have both a simple led and a ws2812 led.

There are definitely several boards with _only_ a WS2812 and no normal LED though.

lurch:
> I'll get rid of the on off thing and just have a method that sets the RGB of a colour status led.

IMHO being able to simply turn on/off an RGB LED would mean that a board with an RGB LED (using this status_led library) could be used identically (after a recompile of course) to a board with a normal LED? (e.g. the canonical "status-LED-blink" example)
Perhaps we've just got a different mental-model of what it is that this status_led library is actually for?

lurch:
> But most boards are configured with both. Who's to say that you shouldn't be able to control both individually?

I'm not saying that you _shouldn't_ be able to control both LEDs individually, I guess I was just kinda expecting a `status_led` library to control only one of them, with the other LED then free for user-purposes. :bulb: 

So, entirely IMVHO:
* board has a regular LED (i.e. `PICO_DEFAULT_LED_PIN` or `CYW43_WL_GPIO_LED_PIN`) but no RGB LED : `status_led` on/off turns the regular LED on/off
* board has an RGB LED (i.e. `PICO_DEFAULT_WS2812_PIN`) but no regular LED : `status_led` on/off turns the RGB led on/off, with optional control of the "on" colour (defaults to "white")
* board has both a regular LED and an RGB LED (i.e. `PICO_DEFAULT_LED_PIN` and  `PICO_DEFAULT_WS2812_PIN`) : `status_led` on/off turns the regular LED on/off, with the RGB LED then left free for any other purposes
* board has both a regular LED and an RGB LED **and** the user sets the `PICO_STATUS_LED_USE_WS2812_PIN` define (or equivalent) : `status_led` on/off turns the RGB led on/off, with the regular LED then left free for any other purpose
* board has lots of regular LEDs and lots of RGB LEDs : user is free to set whatever `PICO_CONFIG` defines they want, in order to customise which LED the `status_led` on/off functions will control

That way any example-code that's using the `status_led` on/off functions will "just work" on any type of board? :shrug:

src/rp2_common/pico_status_led/include/pico/status_led.h
```diff
@@ -0,0 +1,124 @@
...
+ * \note: If your hardware does not support a status led (\see PICO_DEFAULT_LED_PIN), this function does nothing and returns false.
+ *
+ * \param led_on True to turn the led on. Pass False to turn the led off
+ * \param True if the status led could be set, otherwise false
+ */
+static inline bool pico_status_led_set(bool led_on) {
+#if defined PICO_DEFAULT_LED_PIN
+    gpio_put(PICO_DEFAULT_LED_PIN, led_on);
+#elif defined CYW43_WL_GPIO_LED_PIN
+    cyw43_arch_gpio_put(CYW43_WL_GPIO_LED_PIN, led_on);
```
peterharperuk, line 55: Not sure I understand how this links as it's in pico_cyw43_arch. we need to include the right header file in here?

kilograham:
> > But most boards are configured with both. Who's to say that you shouldn't be able to control both individually?
> 
> I'm not saying that you _shouldn't_ be able to control both LEDs individually, I guess I was just kinda expecting a `status_led` library to control only one of them, with the other LED then free for user-purposes. 💡
> 
> So, entirely IMVHO:
> 
>     * board has a regular LED (i.e. `PICO_DEFAULT_LED_PIN` or `CYW43_WL_GPIO_LED_PIN`) but no RGB LED : `status_led` on/off turns the regular LED on/off
> 
>     * board has an RGB LED (i.e. `PICO_DEFAULT_WS2812_PIN`) but no regular LED : `status_led` on/off turns the RGB led on/off, with optional control of the "on" colour (defaults to "white")
> 
>     * board has both a regular LED and an RGB LED (i.e. `PICO_DEFAULT_LED_PIN` and  `PICO_DEFAULT_WS2812_PIN`) : `status_led` on/off turns the regular LED on/off, with the RGB LED then left free for any other purposes
> 
>     * board has both a regular LED and an RGB LED **and** the user sets the `PICO_STATUS_LED_USE_WS2812_PIN` define (or equivalent) : `status_led` on/off turns the RGB led on/off, with the regular LED then left free for any other purpose
> 
>     * board has lots of regular LEDs and lots of RGB LEDs : user is free to set whatever `PICO_CONFIG` defines they want, in order to customise which LED the `status_led` on/off functions will control
> 
> 
> That way any example-code that's using the `status_led` on/off functions will "just work" on any type of board? 🤷

This is somewhat similar to what i was expecting.

1. The ability to blink a LED on any old board. If it has a WS812 or similar, the ability to set the color too (the API can exist but return false)
2. Possibly in the future, some helpers, like "blink error pattern"
3. The ability to target these APIs at a particular LED pin or WS2812 pin, or WIFI LED is probably nice too

Sorry, i haven't given a lot of though yet to how i would like this to look

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -0,0 +1,113 @@
...
+    return true;
+}
+
+void pico_status_led_deinit(struct async_context *context) {
+    // Note: We cannot deinit cyw43 in case it has other users
+#if defined(PICO_DEFAULT_LED_PIN)
+    gpio_deinit(PICO_DEFAULT_LED_PIN);
+#elif defined(CYW43_WL_GPIO_LED_PIN)
+    assert((context && !status_led_context.core.type) || (!context && status_led_context.core.type));
+    if (context) {
```
peterharperuk, line 101: Should be !context

kilograham:
looking nice - one comment in code

peterharperuk:
I added some more config so you can make it use WS2812 by default

src/rp2_common/pico_status_led/include/pico/status_led.h
```diff
@@ -0,0 +1,165 @@
...
+#define PICO_STATUS_LED_WS2812_OFF PICO_STATUS_LED_RGB(0, 0, 0)
+#endif
+#endif
+
+/*! \brief Initialise the status leds
+ *  \ingroup pico_status_led
+ *
+ *  Initialise the status leds and the resources they need before use.
+ *
+ * \param context The hardware for some devices (e.g. Pico W) requires an async context.
```
kilograham, line 69: this is a bit confusing and vague, however fine if there is a pico-example

src/rp2_common/pico_status_led/include/pico/status_led.h
```diff
@@ -0,0 +1,165 @@
...
+ * \note: If your hardware does not support a colored status led (\see PICO_DEFAULT_WS2812_PIN), this function always returns 0x0.
+ *
+ * \return The color of the colored status led in 0xWWRRGGBB format.
+ */
+uint32_t pico_status_led_color_get(void);
+
+/*! \brief Generate an RGB colour value for /ref pico_status_led_color_set
+ *  \ingroup pico_status_led
+ */
+#define PICO_STATUS_LED_RGB(R, G, B) (((R) << 16) | ((G) << 8) | (B))
```
kilograham, line 99: i'd #ifndef this

kilograham:
looking great; a few comments in the code, but also

I still think
c
the status LED may or may not have a settable color
the status LED may or may not be on/off

i.e. turning on/off a WS2812 LED should set it to the current

peterharperuk:
Another attempt pushed. Note that I don't have a WS2812 atm, so will test more on Monday.

peterharperuk:
@kilograham Can you tell me if this is closer to what you want before I waste more time on it?

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -0,0 +1,155 @@
...
+
+bool set_ws2812(uint32_t value) {
+    if (offset > -1) {
+#if PICO_STATUS_LED_WS2812_WRGB
+        // Convert to 0xWWGGRRBB
+        pio_sm_put_blocking(pio, sm, WHITE(ws2812_value) << 24 | GREEN(ws2812_value) << 16 | RED(ws2812_value) << 8 | BLUE(ws2812_value));
+#else
+        // Convert to 0xGGRRBB00
+        pio_sm_put_blocking(pio, sm, GREEN(ws2812_value) << 24 | RED(ws2812_value) << 16 | BLUE(ws2812_value) << 8);
+#endif //P ICO_STATUS_LED_WS2812_WRGB
```
peterharperuk, line 43: typo

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -0,0 +1,155 @@
...
+        return true;
+    }
+}
```
peterharperuk, line 46: should return false?

kilograham:
fixes #188

Issue #2506, opened 2025-06-06
powman_power_state returned by powman_get_power_state inverted


eightycc:
The RP2350 POWMAN `STATE` register bit-fields `REQ` and `CURRENT` represent the requested and current power state bits. According to the datasheet, the hardware defines these as 0 for `powered-up` and 1 for `powered-down.`

The SDK defines a type `powman_power_state` to hold a power state, along with helper functions `powman_power_state_with_domain_on()`,  `powman_power_state_with_domain_off()`, and `powman_power_state_is_domain_on()` to manipulate and query its domain power state bits.

Macro `POWMAN_POWER_STATE_NONE` is defined as 0, indicating all power domains off.

It's clear from this type, helpers, and the macro, that the SDK wants to handle domain ON/OFF states intuitively as 1 for `ON` and 0 for `OFF`.

The implementation of `powman_get_power_state()` returns the raw `CURRENT` field of the `STATE` register as `powman_power_state` type. To be consistent with SDK usage, it should logically complement the field.

The implementation of `powman_set_power_state()` correctly complements its `state` argument before loading it into the `REQ` field of the `STATE` register.

kilograham:
fix merged into develop

Issue #2511, opened 2025-06-08
powman_configure_wakeup_state() broken


eightycc:
Power Manager function `powman_configure_wakeup_state()` should do the following:

- Validate that the requested sleep state and the wake-up state adhere to the rules outlined in section 6.2.3 "Power State Transitions" of the RP2350 Datasheet.
- Configure the power sequencer `SEQ_CFG` register bits `HW_PWRUP_SRAM0` and `HW_PWRUP_SRAM1` to power the SRAM power domains when waking up as specified by the `wakeup_state` argument.

The current implementation has these issues:

- It will reject valid power state transitions such as P0.0 -> P1.7 -> P0.0, where the current state is P0.0, the sleep state is P1.7, and the wake-up state is P0.0.
- It does not take the current state into consideration when validating the sleep state.
- It does not validate that the sleep state is one of the P1.x states, or that the wake-up state is one of the P0.x states.
- It takes only the sleep state into consideration when setting `SEQ_CFG` register bits `HW_PWRUP_SRAM0` and `HW_PWRUP_SRAM1`. Since these bits are defined: 0 for power-up and 1 for no change, both the sleep state and the wake-up state must be considered to set them correctly.

kilograham:
fix merged into develop

PR #2516, opened 2025-06-11
Fix powman powman_configure_wakeup_state and powman_get_power_state

eightycc:
This PR resolves issues #2506 and #2511:

To conform with SDK usage, `powman_get_power_state` now returns the complemented contents of field `POWMAN_STATE_CURRENT_BITS` of the `POWMAN` `STATE` register (#2506).

Corrects operation of `powman_configure_wakeup_state` (#2511):

- Validates that sleep state is one of the `P1.x` states, and that the wake-up state is one of the `P0.x` states.
- Correctly validates power manager state transitions while taking into account the current state, the sleep state, and the wake-up state. Validation conforms to section 6.2.3 of the RP2350 datasheet.
- Sets the `SEQ_CFG` register bits `HW_PWRUP_SRAM0` and `HW_PWRUP_SRAM1` while taking into account both the sleep state and the wake-up state.

### DIFF
--- a/src/rp2_common/hardware_powman/powman.c
+++ b/src/rp2_common/hardware_powman/powman.c
@@ -181,19 +181,30 @@ int powman_set_power_state(powman_power_state state) {
 }
 
 bool powman_configure_wakeup_state(powman_power_state sleep_state, powman_power_state wakeup_state) {
-    // When powman wakes up it can keep the state of the sram0 and sram1 banks. Note, it can't
-    // explicitly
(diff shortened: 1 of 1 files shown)


peterharperuk (approved):
Thanks for fixing our code! I've spent some time staring at it and done some testing with my powman example and I think it makes sense. It fixes the weird situation where you couldn't specify SRAM on in the wakeup state - which never made sense to me.

PR #2524, opened 2025-06-14
Add board definition for Adafruit Feather RP2040 Adalogger

### LINKED ISSUE #2522: Support for Adafruit Feather RP2040 "Adalogger" board?
https://www.adafruit.com/product/5980

Recently picked up this board, and then realised that the documentation only mentions using it with CircuitPython/MicroPython or earlephilhower RP2040 Arduino core. 

Not sure if this is trivial/just a matter of adding a header, or if it would take a lot more work to properly support this board and use it with Pico SDK.

I am aware that using some of the peripherals on the board like the microSD would be more involved and probably up to the user to implement, but I guess the pin definitions etc. would be nice at least so that basic functionality is possible with Pico SDK without needing any tweaks or using other abstraction layers.

b4shful:
Fixes #2522 

Based off of the existing Adafruit Feather RP2040 header (`adafruit_feather_rp2040.h`)

The only changes made from the regular Feather RP2040 board header are:

- Change NeoPixel/WS2812 pin `PICO_DEFAULT_WS2812_PIN` from 16 to 17

- Add defs for SD card pins (see SD section in header).

### Note on SD card pin defs:

I followed the conventions used for other boards with SD card pins defined, although none of the other boards I looked at had the SD "Card Detect" pin mapped so I named its pin `PICO_SD_CARD_DETECT_PIN`.

I think this should be all that is needed to provide *minimal* support for Pico SDK on this board variant, but please do let me know if I'm missing something.

### DIFF
--- a/src/boards/include/boards/adafruit_feather_rp2040_adalogger.h
+++ b/src/boards/include/boards/adafruit_feather_rp2040_adalogger.h
@@ -59,19 +59,19 @@ pico_board_cmake_set(PICO_PLATFORM, rp2040)
 
 //------------- SPI -------------//
 #ifndef PICO_DEFAULT_SPI
-#define PICO_DEFAULT_SPI 0
+#define PICO_DEFAULT_SPI 1
 #endif
 
 #ifndef PICO_DEFAULT_SPI_TX_PIN
-#define PICO_DEFAULT_SPI_TX_PIN 19
+#define PICO_DEFAULT_SPI_TX_PIN 15
 #endif
 
 #ifndef PICO_DEFAULT_SPI_RX_PIN
-#define PICO_DEFAULT_SPI_RX_PIN 20
+#define PICO_DEFAULT_SPI_RX_PIN 8
 #endif
 
 #ifndef PICO_DEFAULT_SPI_SCK_PIN
-#define PICO_DEFAULT_SPI_SCK_PIN 18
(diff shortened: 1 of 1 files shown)


b4shful:
I realised that `PICO_DEFAULT_SPI` is currently set to 0. However this is the SPI peripheral Adafruit specifies is used for the SD card. 

SPI1 is unused by default and corresponds to the MOSI/MISO/SCK silkscreen labels on the Feather RP2040 Adalogger, so perhaps I should have defined `PICO_DEFAULT_SPI` as 1 instead?

b4shful:
Sorry, got a bit confused there.

The default SPI instance and pins from the original Feather RP2040 header file are what the Adalogger board uses for the SD card. So I will need to edit the `PICO_DEFAULT_SPI` defs to instead use SPI1 and the correct pins for this as silkscreened on the board.

lurch:
> Committed the required edits for this.

Based on that, I'll mark this as ready for review.

> As a note, the macros `PICO_DEFAULT_SPI_TX_PIN` and `PICO_DEFAULT_SPI_RX_PIN` seem confusingly named as TX and RX could be the complete opposite depending on the role the SPI peripheral is configured in.

Yup, see section 4.4.3.15 of https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf

lurch (approved):

PR #2530, opened 2025-06-17
Make Python `re` strings in `tools/*.py` be raw to avoid escape sequence warnings

### LINKED ISSUE #2529: tools/check_board_header.py et al give SyntaxWarnings for invalid escape sequences
tools/check_board_header.py gives syntax warnings for invalid escape sequences when used with Python 3.12 (at least). The `re` strings should be raw strings.

example:
```
/home/halbert/repos/adafruit/pico-sdk/tools/check_board_header.py:223: SyntaxWarning: invalid escape sequence '\s'
  if re.match("^\s*// For board detection", line):
/home/halbert/repos/adafruit/pico-sdk/tools/check_board_header.py:228: SyntaxWarning: invalid escape sequence '\s'
  m = re.match("^\s*// This header may be included by other board headers as \"(.+?)\"", line)
```

I'll submit a PR for this shortly

dhalbert:
Fixes #2529.

All `re` strings in `tools/*.py` are now raw strings, to prevent warnings like:
```
/home/halbert/repos/adafruit/pico-sdk/tools/check_board_header.py:223: SyntaxWarning: invalid escape sequence '\s'
  if re.match("^\s*// For board detection", line):
```
Also (fixed by accident due to my editor settings), some trailing whitespace was removed in the files I edited.

### DIFF
--- a/tools/check_board_header.py
+++ b/tools/check_board_header.py
@@ -130,7 +130,7 @@ def read_defines_from(header_file, defines_dict):
             if m:
                 validity_stack.pop()
                 continue
-            
+
             if validity_stack[-1]:
                 # look for "#include "foo.h"
                 m = re.match(r"""^\s*#include\s+"(.+?)"\s*$""", line)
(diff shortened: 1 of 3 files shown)


will-v-pi:
These are also fixed in #2519 and #2525, which is a larger refactor so I'd prefer to just add these changes into those PRs

lurch:
I'll convert this to draft, and we can come back to it (if necessary) after those other PRs are merged.

dhalbert:
I will just close, since it was addressed elsewhere.

PR #2532, opened 2025-06-18
Add POWMAN_PASSWORD_BITS to pico_bootsel_via_double_reset.

Gadgetoid:
Fix pico_bootsel_via_double_reset for RP2350.

Was borrowing this code for a USB MSC user boot toggle. Found it didn't work, delved into the bootloader and found:

https://github.com/raspberrypi/pico-bootrom-rp2350/blob/fd6104450fa8f55c11c0c9b54dbc69a27537130f/src/main/arm/varm_boot_path.c#L774

### DIFF
--- a/src/rp2_common/pico_bootsel_via_double_reset/pico_bootsel_via_double_reset.c
+++ b/src/rp2_common/pico_bootsel_via_double_reset/pico_bootsel_via_double_reset.c
@@ -97,11 +97,11 @@ static inline bool double_tap_flag_is_set(void) {
 }
 
 static inline void set_double_tap_flag(void) {
-    hw_set_bits(&powman_hw->chip_reset, POWMAN_CHIP_RESET_DOUBLE_TAP_BITS);
+    hw_set_bits(&powman_hw->chip_reset, POWMAN_CHIP_RESET_DOUBLE_TAP_BITS | POWMAN_PASSWORD_BITS);
 }
 
 static inline void clear_double_tap_flag(void) {
-    hw_clear_bits(&powman_hw->chip_reset, POWMAN_CHIP_RESET_DOUBLE_TAP_BITS);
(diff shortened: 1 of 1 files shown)


kilograham (approved):
I'm assuming this is correct (can't see any docs which say which regs need the password ;-) )

Gadgetoid:
Section 6.4 from the good book for posterity:

All registers with address offsets up to and including 0x000000ac are password protected. Therefore, the following writeable registers are unprotected and have 32-bit write access:

• POWMAN_SCRATCH0 → POWMAN_SCRATCH7
• POWMAN_SCRATCH0 → POWMAN_SCRATCH7
• POWMAN_BOOT0 → POWMAN_BOOT3
• POWMAN_BOOT0 → POWMAN_BOOT3
• POWMAN_INTR
• POWMAN_INTE
• POWMAN_INTF

Doesn’t help that this section lands on a page boundary and it looks like a formatting error leading into *another* list of registers right below...

lurch:
Maybe it'd be better to use [powman_set_bits](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_powman/include/hardware/powman.h#L110) ? :thinking:

Gadgetoid:
> Maybe it'd be better to use [powman_set_bits](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_powman/include/hardware/powman.h#L110) ? 🤔

Agreed! And updated.

lurch:
> yeah lets revert that last change, as it obviously breaks RP2040, and i don't know if it is worth having to do a platform test in the CMake and bazel

Sorry @Gadgetoid for my duff suggestion :man_facepalming:

Gadgetoid:
> Sorry @Gadgetoid for my duff suggestion 🤦‍♂️

It was a *good* and reasonable suggestions, but SDK says no 😆

Reverted and push'd!

PR #2535, opened 2025-06-19
Provide an easy way to disable cyw43 logging

### LINKED ISSUE #2523: Feature Request: Disable CYW43 debug output
I'm working on firmware for a device that can do host comms both over usb and over wifi as a tcp server.

The CYW43_WARN, _INFO, etc. periodically put stuff out on the usb (via their printf calls) that interferes with the usb communication.

There should be some defines that can disable that capability.

peterharperuk:
All cyw43 logging uses CYW43_PRINTF by default. To disable logging you have to define this to do nothing. There's no simple way to achieve this. Make it easier by adding PICO_CYW43_LOGGING_ENABLED which can be set to zero to disable CYW43_PRINTF.

Fixes #2523

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
+++ b/src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h
@@ -198,6 +198,15 @@ void cyw43_post_poll_hook(void);
 #define cyw43_free free
 #endif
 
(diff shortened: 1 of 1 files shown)


lurch (approved):

peterharperuk:
> CYW43_DEBUG_ENABLED etc to mirror the new CYBT_DEBUG_ENABLED

Maybe it would be better to get that added to cyw43-driver if you think it's important?

kilograham (approved):

PR #2539, opened 2025-06-23
Use platform/generator agnostic CMake commands

will-v-pi:
Change the README to use platform/generator agnostic commands to create and build the CMake project

This means the generic commands work with Ninja or Makefiles, and avoids the need to `cd` into build directories

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
       ```sh
       $ cmake --build build --target hello_world
       ```
-> The directory-name supplied to the `--build` flag needs to match the directory-name that was passed to the `-B` flag in the earlier cmake command.
+      > The directory-name supplied to the `--build` flag needs to match the directory-name that was passed to the `-B` flag in the earlier cmake command.
 
 1. You now have `hello_world.elf` to load via a debugger, or `hello_world.uf2` that can be installed and run on your Raspberry Pi Pico-series device via drag and drop.
 
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
       ```
       $ cmake -S . -B build
       ```
-      > The cmake -S flag indicates the source directory, and -B tells cmake where to put the output files.
+      > The cmake -S flag indicates the source directory, and the -B flag tells cmake the name of the output-directory to create. This doesn't have to be named "build", you can call it whatever you want.
    
    When building for a board other than the Raspberry Pi Pico, you should pass `-DPICO_BOARD=board_name` to the `cmake` command above, e.g. `cmake -S . -B build -DPICO_BOARD=pico2` or `cmake -S . -B build -DPICO_BOARD=pico_w` to configure the SDK and build options accordingly for that particular board.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,9 +177,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
 1. Setup a CMake build directory.
       For example, if not using an IDE:
       ```
-      $ mkdir build
-      $ cd build
-      $ cmake ..
+      $ cmake -S . -B build
```
lurch, line 180: IMHO we should either explain what these short command-line switches mean...

README.md
```diff
@@ -193,7 +191,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
 
 1. Make your target from the build directory you created.
       ```sh
-      $ make hello_world
+      $ cmake --build build --target hello_world
```
lurch, line 194: ...or we should be consistent and just use the long command-line switches? (which are "more self-documenting" than the short options)

README.md
will-v-pi: Something like adding
> `cmake -S <source_dir> -B <build_dir>`

below the command?

lurch: Yeah, or just something like
> The cmake `-S` flag indicates the source directory, and `-B` tells cmake where to put the output files.

? :shrug:

README.md
```diff
@@ -193,7 +192,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
...
```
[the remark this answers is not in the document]
will-v-pi: I was wondering if it'd be cleaner to use `./build` or `build/` or something, to make it clearer it's a directory?

README.md
```diff
@@ -177,12 +177,11 @@ instructions for other platforms, and just in general, we recommend you see [Ras
 1. Setup a CMake build directory.
       For example, if not using an IDE:
       ```
-      $ mkdir build
-      $ cd build
-      $ cmake ..
-      ```   
+      $ cmake -S . -B build
+      ```
+      > The cmake -S flag indicates the source directory, and -B tells cmake where to put the output files.
```
lurch, line 182:
```suggestion
      > The cmake -S flag indicates the source directory, and the -B flag tells cmake the name of the output-directory to create. This doesn't have to be named "build", you can call it whatever you want.
```

README.md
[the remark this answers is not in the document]
lurch: Maybe a bit wordy, but I've suggested one possible way that we could explain this more clearly.

README.md
```diff
@@ -193,7 +192,7 @@ instructions for other platforms, and just in general, we recommend you see [Ras
 
 1. Make your target from the build directory you created.
       ```sh
-      $ make hello_world
+      $ cmake --build build --target hello_world
       ```
```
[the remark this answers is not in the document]
will-v-pi: Probably as an inline quote, like the other similar comment above
```suggestion
> The directory-name supplied to the `--build` flag needs to match the directory-name that was passed to the `-B` flag in the earlier cmake command.

```

README.md
```diff
@@ -193,8 +192,9 @@ instructions for other platforms, and just in general, we recommend you see [Ras
...
+> The directory-name supplied to the `--build` flag needs to match the directory-name that was passed to the `-B` flag in the earlier cmake command.
```
will-v-pi, line 197:
```suggestion
      > The directory-name supplied to the `--build` flag needs to match the directory-name that was passed to the `-B` flag in the earlier cmake command.
```

PR #2545, opened 2025-06-30
Add FLASH option to pico_package_uf2_output

will-v-pi:
Allows passing FLASH instead of 0x10000000 (eg `pico_package_uf2_output(hello_serial FLASH)`) to package a UF2 to flash

Also adds a `pico_ensure_load_map(TARGET)` function, to ensure a load_map is added to the binary (eg for universal no_flash binaries)

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -95,6 +95,12 @@ define_property(TARGET
     BRIEF_DOCS "Extra arguments to pass to uf2 conversion"
     FULL_DOCS "Extra arguments to pass to uf2 conversion"
 )
+define_property(TARGET
+    PROPERTY PICOTOOL_LOAD_MAP
+    INHERITED
+    BRIEF_DOCS "Ensure a load map is added"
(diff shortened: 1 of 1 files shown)


will-v-pi:
> Would it also make sense to add an "SRAM" option?

Not really - packaging is where you put an SRAM binay in Flash, so it can be stored in non volatile memory, and the bootrom copies it to SRAM at runtime. It doesn't make sense to do the opposite and package a Flash binary to SRAM. It might make sense to package an SRAM binary from one bit of SRAM to another, but you'd be specifying a custom address for that anyway.

kilograham:
I'm a bit confused; you _always_ want to package it into flash, no? so perhaps allowing you to omit the argument is better?

will-v-pi:
> I'm a bit confused; you _always_ want to package it into flash, no? so perhaps allowing you to omit the argument is better?

I've switched it to this, as currently yes, you do always want to package it into flash

kilograham (approved):

Issue #2338, opened 2025-03-09
irq clear prev not working


pauldevine:
On the rp2350b I'm synchronizing two state machines on different PIO instances so that they can address different pin ranges. 0-31 with one state machine and 32-47 on the other. I'm able to successfully set an interrupt and have it triggered on the second machine across the PIO boundary. But I'm not able to clear it with `irq clear`. The compiler won't allow me to reference the previous pio block with:

```
.wrap_target
    wait 0 irq prev 0      ; Wait for IRQ0
    irq clear prev 0        ; Clear IRQ0
.wrap
```

that results in a syntax error `15-18: syntax error, unexpected prev, expecting ( or identifier or integer.` I get the same syntax error if I try the documented `irq clear 0 prev`. Since the word prev seems to work for the other statements when it precedes the interrupt number I assumed that syntax should be valid here.

I can get the pio assembly to compile if I use `irq prev clear 0` however that doesn't clear the interrupt. 

If instead I clear the interrupt immediately with `wait 1 irq prev 0` that works as expected. I was hoping to have a few pio commands executed before clearing the interrupt as I'm trying to get a bus to the right state before continuing the other state machine.

lurch:
Similar issue to https://github.com/raspberrypi/pico-feedback/issues/440 ?

pauldevine:
That earlier comment is pointing out a documentation error. I believe I'm experiencing a functional error.

KewlKris:
I also had trouble with IRQs between state machines in different blocks (even when synchronizing with CTRL.NEXTPREV_CLKDIV_RESTART).

I eventually used `wait JMPPIN + n` (to wait for a signal) and `jmp PIN` (to branch on a signal) to synchronize and communicate between state machines. I wish I could get things working with an internal solution rather than using external pins, though.

kilograham:
Can you please attach a small example showing the problem; does it work as expected without the `next`?

pauldevine:
I can put together a small reproducer over the weekend.

pauldevine:
[irq_test.zip](https://github.com/user-attachments/files/21206933/irq_test.zip)
Attached is a small reproducer. If you run the existing setup as shown you'll get output like:
Starting IRQ test
IRQ 0 state: 1
IRQ 0 state: 1
IRQ 0 state: 1
IRQ 0 state: 1
IRQ 0 state: 1
IRQ 0 state: 1
IRQ 0 state: 1
IRQ 0 state: 1
[...]

PR #2554, opened 2025-07-09
add pioasm --version, and print version number in generated files

### LINKED ISSUE #2307: pioasm: add command line option --version and print SDK/git revision with --help
`pioasm` should have an option `--version` (and uppercase `-V`) and/or include SDK version or git revision with `--help`, and possibly also in the generated file ([from forum](https://forums.raspberrypi.com/viewtopic.php?p=2296588#p2296556)).

kilograham:
fixes #2307


lurch:
Is the change to `amehyst.pio` here intentional, or is that part of #2338 ?
This LGTM, but I won't approve because CI is failing :wink: 

@will-v-pi I wonder if it's worth having a CI job that builds a sample PIO file with each of the different output-formats that `pioasm` supports, and then comparing those newly-created outputs to "known good" output files?

Also, after this PR is merged I guess we'll need to update the pioasm-generated headers in the pico-examples repo.

tools/pioasm/python_output.cpp
```diff
@@ -61,7 +61,9 @@ struct python_output : public output_format {
         FILE *out = open_single_output(destination);
         if (!out) return 1;
 
-        header(out, "This file is autogenerated by pioasm; do not edit!");
+        std::stringstream header_string;
+        header_string << "This file is autogenerated by pioasm version " << PIOASM_VERSION_STRING << "; do not edit!";
+        header(out, header_string.str());
```
lurch, line 66: This code-snippet is identical in each of the different output-format generators - perhaps it's worth abstracting out a new common function, and then this could be simplified to:
```c
        std::stringstream header_string = standard_header_comment();
        header(out, header_string.str());
```
(and I guess it would also mean that each of the different output-format generators don't need to separately include `version.h`)

lurch: (Tangential unimportant observation: The ADA generator's `header()` function has an `indent` parameter, which none of the `header()` functions for the other language generators have! :slightly_smiling_face: )

kilograham: I saw both of these, and decided i didn't care that much ;-)

PR #2566, opened 2025-07-16
Add init_submodules.sh

kilograham:


lurch:
Hmmmm...
* [`.github/workflows/cmake.yml`](https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/cmake.yml) does `git submodule update --init --recursive`
* [`.github/workflows/macOS.yml`](https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/macOS.yml) does `git submodule update --init`
* [`.github/workflows/multi-gcc.yml`](https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/multi-gcc.yml) does `git submodule update --init`
* [`.github/workflows/windows.yml`](https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/windows.yml) does `git submodule update --init`

So perhaps we ought to modify `init_submodules.sh` to do `git submodule update --init`, and then update the above CI scripts to use `init_submodules.sh`, for consistency? :thinking:

will-v-pi (commented):
The script isn't executable, and also might be better placed under the `scripts` directory?

PR #2568, opened 2025-07-16
minor tidying of hardware_rcp header; add an `rcp_is_true()` that is …

kilograham:


will-v-pi (approved):
LGTM - checked that the rcp checks still work for my Risc-V stuff

PR #2577, opened 2025-07-21
Improve rom_pick_ab_update_partition docs

will-v-pi:
Should be merged with https://github.com/raspberrypi/pico-examples/pull/677

### DIFF
--- a/src/rp2_common/pico_bootrom/include/pico/bootrom.h
+++ b/src/rp2_common/pico_bootrom/include/pico/bootrom.h
@@ -750,7 +750,7 @@ static inline int rom_load_partition_table(uint8_t *workarea_base, uint32_t work
  * NOTE: This method does not look at owner partitions, only the A partition passed and it's corresponding B partition.
  * 
  * NOTE: You should not call this method directly when performing a Flash Update Boot before calling `explicit_buy`, as it may prevent
- * any version downgrade from occuring - instead see \ref rom_pick_ab_update_partition() which wraps this function.
+ * any version downgrade from occuring - instead see \ref rom_pick_ab_partition_during_update() which wraps this function.
  * 
  * \param workarea_base base address of work area
  * \param workarea_size size of work area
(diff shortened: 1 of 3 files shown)


src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1101,6 +1101,9 @@ int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32
  * a main image TBYB boot. It requires the same minimum workarea size as `rom_pick_ab_partition`.
  * \see rom_pick_ab_partition()
  * 
+ * This should be used instead of `rom_pick_ab_partition` when performing a Flash Update Boot before calling `explicit_buy`, and can still be used without
```
kilograham, line 1104: Commenting on a wider scope:

* any in progress update or TBYB boot" perhaps "any in progress Flash Update boot or TBYB boot" (note boot repeated), though a TBYB is a flash update boot, so maybe TBYB operation IDK
* I'd put this function right after rom_pick_ab_partition in the header
* we should use \ref to refer to other functions not `func` (cc @lurch)
* arguably "normally" is "always" so really you could lose the word
* bit of a weird '-' in the middle of the For example sentence
* erase 2 partitions -> erase two partitions
* It also checks" It is a big dangly here (perhaps this function)

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1101,6 +1101,9 @@ int rom_add_flash_runtime_partition(uint32_t start_offset, uint32_t size, uint32
...
+ * issue when a Flash Update Boot is not in progress.
+ *
```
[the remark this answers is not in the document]
kilograham: I don't mind renaming it, not least as I cant imagine that many people have used it - i had considered "during_update" however your comment mentioned that it works fine not during update. That said, i think for want of something better, yes, what you said

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -766,6 +767,32 @@ static inline int rom_pick_ab_partition(uint8_t *workarea_base, uint32_t workare
     return rc;
 }
 
+/*! \brief Pick A/B partition without disturbing any in progress Flash Update boot or TBYB boot
+ * \ingroup pico_bootrom
+ *
+ * This will call \ref rom_pick_ab_partition() using the `flash_update_boot_window_base` from the current boot, while performing extra checks to prevent disrupting
```
kilograham, line 773: If we aren't copying the text from rom_pick_ab_partiton() I think we should focus on the fact that this performs the same function as rom_pick_ab_partiont(), but does a few other things. I guess this is implicit, but using "calls" just makes me think the documentation is just a precis of the function code

src/rp2_common/pico_bootrom/include/pico/bootrom.h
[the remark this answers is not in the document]
will-v-pi: Have renamed it to `rom_pick_ab_update_partition_during_update` - corresponding pico-examples PR https://github.com/raspberrypi/pico-examples/pull/677

src/rp2_common/pico_bootrom/include/pico/bootrom.h
will-v-pi: I've replaced "call" with "perform the same function as" to make that clear

PR #2582, opened 2025-07-22
add some doxygen to hard_assert

### LINKED ISSUE #2314: Document hard_assert()
Several SDK examples (like [`hello.c`](https://github.com/raspberrypi/pico-examples/blob/master/pio/hello_pio/hello.c#L38)) call `hard_assert()` but it is apparently not documented (in the PDF version of the C/C++ SDK or in the datasheets).

It's not so clear how `"pico/assert.h"` actually gets included and which parts of it are public and when to use it instead of `assert()` from `<assert.h>` (or `<cassert>`).

kilograham:
fixes #2314


will-v-pi (dismissed):
LGTM, and useful because I didn't realise this was what it was for

will-v-pi (approved):

PR #2587, opened 2025-07-23
add a task completion semaphore to guarantee async_context task is finished before async_context cleanup finishes

### LINKED ISSUE #2487: Possible race condition between async_context_task() and async_context_freertos_deinit()
Tested with pico_cyw43_arch_lwip_sys_freertos and commit [ee68c78](https://github.com/raspberrypi/pico-sdk/commit/ee68c78d0afae2b69c03ae1a72bf5cc267a2d94c).

If the thread executing _async_context_freertos_deinit()_ has higher priority than _async_context_task()_ then the _async_context_task()_ does not end immediately and _async_context_freertos_deinit()_ clears the `async_context_freertos_t` struct before _async_context_task()_ exit. This results in _async_context_task()_ using already freed freertos objects.

The course of events:

1. When _async_context_freertos_deinit()_ is called, it adds a synch work that sets `task_should_exit = true` requesting ending of  the _async_context_task()_. Then _async_context_freertos_deinit()_ pends on semaphore which will signal completing of the work.
2. Context is switched to _async_context_task()_ which will execute queued work and set `task_should_exit = true`. Then the semaphore is signaled as the work is completed.
3. Since higher priority task, that is _async_context_freertos_deinit()_, is now unblocked, the scheduler wakes it up.
4. _async_context_freertos_deinit()_ then destroys all the allocated freertos object  and clears the `async_context_freertos_t` struct (_memset()_ at the end of the function). This effectively sets `task_should_exit` to `false` again.
5. At later time when scheduler wakes up the  _async_context_task()_, it finds `task_should_exit` `false` and continues to run trying to take semaphore (by calling _async_context_freertos_acquire_lock_blocking()_) which was already freed by the _async_context_freertos_deinit()_  and is `NULL` now. This will cause the following _assert()_ error: `assertion "pxMutex" failed: file "FreeRTOS-Kernel/queue.c", line 826`.

Steps to reproduce:

1. Call the _cyw43_arch_deinit()_ from a task that has priority equal or higher to `tskIDLE_PRIORITY + 4`.

Possible solutions:

1. Setting default value of `CYW43_TASK_PRIORITY` to `(configMAX_PRIORITIES - 1)` instead of `(tskIDLE_PRIORITY + 4)`.
2. Implementing synchronization between _async_context_freertos_deinit()_ and _async_context_task()_. For example, _async_context_task()_ could signal an event after going out of do-while loop.

kilograham:
fixes #2487

### DIFF
--- a/src/rp2_common/pico_async_context/async_context_freertos.c
+++ b/src/rp2_common/pico_async_context/async_context_freertos.c
@@ -73,6 +73,7 @@ static void async_context_task(__unused void *vself) {
         async_context_freertos_release_lock(&self->core);
         __sev(); // it is possible regular code is waiting on a WFE on the other core
     } while (!self->task_should_exit);
+    xSemaphoreGive(self->task_complete_sem);
     vTaskDelete(NULL);
 }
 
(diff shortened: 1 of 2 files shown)


peterharperuk (approved):
Looks good. And seems to fixe the issue

PR #2588, opened 2025-07-23
Update docs, and add methods for new RP2350 DMA features

### LINKED ISSUE #2059: Add DMA APIs for count modes
There is no ability to configure the Count Modes field of the DMA`s CHx_TRANS_COUNT register for RP2350 - its processing is missing in the SDK API version 2.0.0.

From RP2350 datasheet:
image
image

File "~\\.pico-sdk\sdk\2.0.0\src\rp2_common\hardware_dma\include\hardware\dma.h"
Actual code:
```
static inline void dma_channel_set_trans_count(uint channel, uint32_t trans_count, bool trigger) {
    if (!trigger) {
        dma_channel_hw_addr(channel)->transfer_count = trans_count;
    } else {
        dma_channel_hw_addr(channel)->al1_transfer_count_trig = trans_count;
    }
}
```
Expected code example:
```
static inline void dma_channel_set_trans_count(uint channel, uint32_t mode, uint32_t trans_count, bool trigger) {
    assert(mode == DMA_CH0_TRANS_COUNT_MODE_VALUE_NORMAL || mode == DMA_CH0_TRANS_COUNT_MODE_VALUE_TRIGGER_SELF || mode == DMA_CH0_TRANS_COUNT_MODE_VALUE_ENDLESS);
    assert(trans_count < (1 << DMA_CH0_TRANS_COUNT_MODE_LSB));
    if (!trigger) {
        dma_channel_hw_addr(channel)->transfer_count = trans_count;
    } else {
        dma_channel_hw_addr(channel)->al1_transfer_count_trig = trans_count;
    }
    dma_channel_hw_addr(channel)->transfer_count |= mode << DMA_CH0_TRANS_COUNT_MODE_LSB;
}
```

### LINKED ISSUE #2444: Expose DMA reverse/inc_by_two functionality for RP2350

kilograham:
* self-trigering/endless transfers
* reversed/double-increment DMA

fixes #2059 
fixes #2444

### DIFF
--- a/src/rp2_common/hardware_dma/include/hardware/dma.h
+++ b/src/rp2_common/hardware_dma/include/hardware/dma.h
@@ -146,30 +146,116 @@ enum dma_channel_transfer_size {
     DMA_SIZE_32 = 2    ///< Word transfer (32 bits)
 };
 
+/*! \brief Enumeration of types of updates that can be made to the DMA read or write address after each transfer
+ *  \ingroup hardware_dma
+ */
+enum dma_address_update_type {
+    DMA_ADDRESS_UPDATE_NONE = 0,                ///< The address remains the same after each transfer
+    DMA_ADDRESS_UPDATE_INCREMENT = 1,           ///< The address is incremented by the transfer size after each transfer
+#if !PICO_RP2040
+    DMA_ADDRESS_UPDATE_INCREMENT_BY_TWO = 2,    ///< (RP2350 only) The address is incremented by twice the transfer size after each transfer
+    DMA_ADDRESS_UPDATE_DECREMENT = 3,           ///< (RP2350 only) The address is decremented by the transfer size after each transfer
+#endif
+};
+
 typedef struct {
     uint32_t ctrl;
 } dma_channel_config;
 
-/*! \brief  Set DMA channel read increment in a channel configuration object
+#ifndef DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS
+#if PICO_RP2040
+#define DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS(u) (((u)&1) << DMA_CH0_CTRL_TRIG_INCR_READ_LSB)
+#else
+#define DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS(u) \
+    ((((u)&1) << DMA_CH0_CTRL_TRIG_INCR_READ_LSB) | \
+    (((u)&2) << (DMA_CH0_CTRL_TRIG_INCR_READ_REV_LSB - 1)))
+#endif
+#endif
+
+#ifndef DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_WRITE_BITS
+#if PICO_RP2040
+#define DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_WRITE_BITS(u) (((u)&1) << DMA_CH0_CTRL_TRIG_INCR_WRITE_LSB)
+#else
+#define DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_WRITE_BITS(u) \
+((((u)&1) << DMA_CH0_CTRL_TRIG_INCR_WRITE_LSB) | \
+(((u)&2) << (DMA_CH0_CTRL_TRIG_INCR_WRITE_REV_LSB - 1)))
+#endif
+#endif
+
+#ifndef DMA_CH_CTRL_ALL_ADDRESS_UPDATE_READ_BITS
+#if PICO_RP2040
+#define DMA_CH_CTRL_ALL_ADDRESS_UPDATE_READ_BITS DMA_CH0_CTRL_TRIG_INCR_READ_BITS
+#else
+#define DMA_CH_CTRL_ALL_ADDRESS_UPDATE_READ_BITS (DMA_CH0_CTRL_TRIG_INCR_READ_BITS | DMA_CH0_CTRL_TRIG_INCR_READ_REV_BITS)
+#endif
+#endif
+
+#ifndef DMA_CH_CTRL_ALL_ADDRESS_UPDATE_WRITE_BITS
+#if PICO_RP2040
+#define DMA_CH_CTRL_ALL_ADDRESS_UPDATE_WRITE_BITS DMA_CH0_CTRL_TRIG_INCR_WRITE_BITS
+#else
+#define DMA_CH_CTRL_ALL_ADDRESS_UPDATE_WRITE_BITS (DMA_CH0_CTRL_TRIG_INCR_WRITE_BITS | DMA_CH0_CTRL_TRIG_INCR_WRITE_REV_BITS)
+#endif
+#endif
+
+/*! \brief  Set DMA channel read address update type in a channel configuration object
  *  \ingroup channel_config
  *
  * \param c Pointer to channel configuration object
- * \param incr True to enable read address increments, if false, each read will be from the same address
- *             Usually disabled for peripheral to memory transfers
+ * \param update_type The type of adjustment to make to the read address after each transfer.
+ *             Usually set to DMA_ADDRESS_UPDATE_NONE for peripheral to memory transfers
+ * \sa channel_config_set_read_increment
  */
+static inline void channel_config_set_read_address_update_type(dma_channel_config *c, enum dma_address_update_type update_type) {
+    c->ctrl = (c->ctrl & ~DMA_CH_CTRL_ALL_ADDRESS_UPDATE_READ_BITS) |
+        DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS(update_type);
+}
+
+/*! \brief  Set DMA channel write address update type in a channel configuration object
+ *  \ingroup channel_config
+ *
+ * \param c Pointer to channel configuration object
+ * \param update_type The type of adjustment to make to the read address after each transfer.
+ *             Usually set to DMA_ADDRESS_UPDATE_NONE for peripheral to memory transfers
+ * \sa channel_config_set_write_increment
+ */
+static inline void channel_config_set_write_address_update_type(dma_channel_config *c, enum dma_address_update_type update_type) {
+    c->ctrl = (c->ctrl & ~DMA_CH_CTRL_ALL_ADDRESS_UPDATE_WRITE_BITS) |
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -146,30 +146,116 @@ enum dma_channel_transfer_size {
...
+        DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS(update_type);
+}
+
+/*! \brief  Set DMA channel write address update type in a channel configuration object
  *  \ingroup channel_config
  *
  * \param c Pointer to channel configuration object
- * \param incr True to enable read address increments, if false, each read will be from the same address
- *             Usually disabled for peripheral to memory transfers
+ * \param update_type The type of adjustment to make to the read address after each transfer.
```
lurch, line 218: Given that this function is named `channel_config_set_write_address_update_type`, was this supposed to say:
```suggestion
 * \param update_type The type of adjustment to make to the write address after each transfer.
```
?

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +505,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
+ * \endif
+ * \return the encoded_transfer_count
+ * \sa dma_channel_set_transfer_count
+ * \sa dma_channel_configure
+ * \sa dma_channel_transfer_from_buffer_now
+ * \sa dma_channel_transfer_to_buffer_now
+ */
+static inline uint32_t dma_encode_transfer_count(uint transfer_count) {
+#if !PICO_RP2040
+    invalid_params_if(HARDWARE_DMA, transfer_count & DMA_CH0_TRANS_COUNT_MODE_BITS);
```
lurch, line 527: Would this flag a `transfer_count` of 0 s an invalid_param? The doxygen comment above implies that 0 is in the valid range? Also, I _think_ that perhaps this should be using `<` rather than `&` (with the comparison-value obviously modified as appropriate), because otherwise a value above 2^28 that has some of the lower bits also set, would still count as "valid"? :thinking:

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +505,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
+ * \if rp2350_specific
+ * On RP2350 the valid range is 0 -> 2^28 - 1
+ * \endif
+ * \return the encoded_transfer_count
+ * \sa dma_channel_set_transfer_count
+ * \sa dma_channel_configure
+ * \sa dma_channel_transfer_from_buffer_now
+ * \sa dma_channel_transfer_to_buffer_now
+ */
+static inline uint32_t dma_encode_transfer_count(uint transfer_count) {
```
lurch, line 525: If the valid range for RP2040 is 0 -> 2^32 - 1, should this be
```suggestion
static inline uint32_t dma_encode_transfer_count(uint32_t transfer_count) {
```
?

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +505,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
+#else
+    return dma_encode_transfer_count(transfer_count) | (DMA_CH0_TRANS_COUNT_MODE_VALUE_TRIGGER_SELF << DMA_CH0_TRANS_COUNT_MODE_LSB);
+#endif
+}
+
+/*! \brief  Return an endless transfer as an "encoded_transfer_length" value suitable for the referenced methods
+ *  \ingroup hardware_dma
+ *
+ * \if rp2040_specific
+ * On RP2040 endless DMA transfers are not supported, so this method may not be used
```
lurch, line 563:
```suggestion
 * On RP2040 endless DMA transfers are not supported, so this method must not be used
```

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +505,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
  * \param channel DMA channel
- * \param trans_count The number of transfers (not NOT bytes, see channel_config_set_transfer_data_size)
+ * \param encoded_transfer_count The encoded transfer count
+ *
+ * \if rp2040_specific
+ * On RP2040 this is just the number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size) from 0 -> 2^32 - 1.
+ * \endif
+ *
+ * \if rp2350_specific
+ * On RP2350 the low 28 bits are used to encode a number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size) and non-zero values of the top 4 bits are used to specify other options.
```
lurch, line 592: It appears in multiple places, but I wonder if "non-zero values of the top 4 bits are used to" could be shortened to "the top 4 bits are used to", since you'd never indicate anything with a zero-value?

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -146,30 +146,116 @@ enum dma_channel_transfer_size {
...
 } dma_channel_config;
 
-/*! \brief  Set DMA channel read increment in a channel configuration object
+#ifndef DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS
+#if PICO_RP2040
+#define DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS(u) (((u)&1) << DMA_CH0_CTRL_TRIG_INCR_READ_LSB)
+#else
+#define DMA_ADDRESS_UPDATE_TYPE_TO_DMA_CH_CTRL_READ_BITS(u) \
+    ((((u)&1) << DMA_CH0_CTRL_TRIG_INCR_READ_LSB) | \
+    (((u)&2) << (DMA_CH0_CTRL_TRIG_INCR_READ_REV_LSB - 1)))
```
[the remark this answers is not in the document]
Wren6991: Yeah, I deliberately put the bits next to each other so you could do that. Graham's code is better because it doesn't rely on that assumption. We usually expect these to be used with constants so there is no additional cost

src/rp2_common/hardware_dma/include/hardware/dma.h
Wren6991: The test is correct. `invalid_params_if` is a negative assert, so it flags inputs which have any of the top 4 bits set. MODE_BITS is `0xf << 28`

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +505,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
+ * \if rp2350_specific
+ * On RP2350 the low 28 bits are used to encode a number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size) and non-zero values of the top 4 bits are used to specify other options.
+ * \endif
+ *
+ * The best practice is always to use either \ref dma_encode_transfer_count, \ref dma_encode_transfer_count_with_self_trigger, or \ref dma_encode_endless_transfer_count to generate a value
+ * to pass for this argument
  * \param trigger True to start the transfer immediately
  */
-static inline void dma_channel_set_trans_count(uint channel, uint32_t trans_count, bool trigger) {
+static inline void dma_channel_set_transfer_count(uint channel, uint32_t encoded_transfer_count, bool trigger) {
```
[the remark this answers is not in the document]
kilograham: good suggestion, but i think i'll leave it

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +505,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
+    return transfer_count;
+#endif
+}
+
+/*! \brief  Encode the specified transfer length, along with a flag to indicate the DMA transfer should be self-triggering, into an "encoded_transfer_length" value suitable for the referenced methods
+ *  \ingroup hardware_dma
+ * \param transfer_count the number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size)
+ *
+ * \if rp2040_specific
+ * On RP2040 self-triggering DMA is not supported, so this method may not be used
```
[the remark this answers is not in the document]
lurch: Okay, perhaps "should" then? IMHO "should" feels more active (i.e. something the user does) whereas "may" feels more passive (i.e. something the SDK does) ? :man_shrugging:

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -140,36 +140,128 @@ bool dma_channel_is_claimed(uint channel);
...
+/*! \brief  Set DMA channel read increment in a channel configuration object
+*  \ingroup channel_config
+*
+* \note this method is equivalent to
+* \code
+* channel_config_set_read_address_update_type(c, incr ? DMA_ADDRESS_UPDATE_INCREMENT : DMA_ADDRESS_UPDATE_NONE)
+* \endcode
+*
+* \param c Pointer to channel configuration object
+* \param incr True to enable read address increments, whereby the read address increments by the transfer size with each transfer, false to perform each read from the same address.
```
lurch, line 242: Small nitpick, but this is probably more readable as
```suggestion
* \param incr True to enable read address increments, whereby the read address increments by the transfer size with each transfer; false to perform each read from the same address.
```
or
```suggestion
* \param incr True to enable read address increments, whereby the read address increments by the transfer size with each transfer. False to perform each read from the same address.
```

src/rp2_common/hardware_dma/include/hardware/dma.h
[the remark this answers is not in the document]
kilograham: i dont think i care enough to spend more time

src/rp2_common/hardware_dma/include/hardware/dma.h
```diff
@@ -419,37 +511,137 @@ static inline void dma_channel_set_write_addr(uint channel, volatile void *write
...
+    return transfer_count;
+#endif
+}
+
+/*! \brief  Encode the specified transfer length, along with a flag to indicate the DMA transfer should be self-triggering, into an "encoded_transfer_length" value suitable for the referenced methods
+ *  \ingroup hardware_dma
+ * \param transfer_count the number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size)
+ *
+ * \if rp2040_specific
+ * On RP2040 self-triggering DMA is not supported, so this method may not be used
```
lurch, line 545:
```suggestion
 * On RP2040 self-triggering DMA is not supported, so this method should not be used
```

peterharperuk (approved):

Issue #2595, opened 2025-07-26
Enable hardware_irq and hardware_claim library also in host common and not only in rp2_common


el-bert0:
The "host" folder already contains "hardware_irq" source code.
The "common" folder already contains "hardware_claim" source code.

For testing purposes, it could be useful to enable hardware_irq and hardware_claim build by default also in PICO_PLATFORM=host.

kilograham:
merged into develop

PR #2596, opened 2025-07-26
Enable hardware_irq library also in host and not only in rp2_common (#2595)

el-bert0:
The "host" folder already contains "hardware_irq" source code.
The "common" folder already contains "hardware_claim" source code.

For testing purposes, it could be useful to enable hardware_irq and hardware_claim build by default also in PICO_PLATFORM=host.


.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/georgerobotics/cyw43-driver.git
 [submodule "lib/lwip"]
 	path = lib/lwip
-	url = https://github.com/lwip-tcpip/lwip.git
+	url = git@github.com:el-bert0/lwip.git
```
lurch, line 9: I'm assuming this change wasn't supposed to be in this PR?

el-bert0: Correct, it shouldn't have been a Pull Request.

el-bert0: I removed it.

kilograham (approved):

kilograham:
fixes #2595

Issue #2576, opened 2025-07-20
Define pico_rand library also in host and not only in rp2_common


el-bert0:
It could be useful to define the **pico_rand** library also in host library (PICO_PLATFORM = HOST) and not only in rp2_common.

This could help people to simulate some tasks while prototyping in host platform.

PR #2598, opened 2025-07-27
Enable pico_rand library also in host and not only in rp2_common (#2576)

el-bert0:
When PICO_PLATFORM=host use only time as random value (define PICO_RAND_ENTROPY_SRC_TIME and PICO_RAND_SEED_ENTROPY_SRC_TIME).


el-bert0:
Need to integrate also Bazel build.

kilograham (approved):
I was considering we should just move it to common, however I think this is fine for now (it does something useful) and it is possible we might want to add calls to proper h/w rand() funcs here in the future anyway

PR #2629, opened 2025-08-14
Add library for reset interface

### LINKED ISSUE #2713: Bug in !PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED implementation?
Just out of curiosity, I went looking for where this was used, and found it [here](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_stdio_usb/reset_interface.c#L128).
```c
#if !PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED
            if (request->wValue & 0x100) {
                gpio = request->wValue >> 9u;
            }
            active_low = request->wValue & 0x200;
#endif
```

To my non-expert eyes, it looks like the flag for `active_low` actually overlaps with the value being read out for `gpio` ?

will-v-pi:
This separates the stdio_usb reset interface out into a separate library, which can be used by applications which use TinyUSB directly.

No change in behaviour when using the typical stdio_usb (eg in the hello_usb example).

Keeps the `PICO_STDIO_USB_` prefix on the config defines to maintain compatibility, and also maintains compatibility with picotool.

Instructions to use this are added to the header file.

See implementation in [debugprobe](https://github.com/raspberrypi/debugprobe/pull/178) and the [dev_multi_cdc example](https://github.com/raspberrypi/pico-examples/pull/694)


src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface.h
```diff
@@ -25,4 +41,128 @@
 // regular flash boot
 #define RESET_REQUEST_FLASH 0x02
 
+#if LIB_PICO_USB_RESET_INTERFACE
```
kilograham: i would be tempted to move this into a separate header under pico_usb_reset_interface_headers, that can be included if desired

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface.h
```diff
@@ -25,4 +41,128 @@
...
+  'c', 0,     'd', 0,     '-', 0,     '4', 0,     'c', 0,     'b', 0, \
+  '7', 0,     '-', 0,     '9', 0,     '8', 0,     'b', 0,     '8', 0, \
+  '-', 0,     '9', 0,     '1', 0,     '3', 0,     'a', 0,     '8', 0, \
+  'f', 0,     'c', 0,     'a', 0,     '7', 0,     'b', 0,     'f', 0, \
+  '6', 0,     '}', 0,       0, 0
+
+#include "stdint.h"
+#include "device/usbd_pvt.h"
+
+void resetd_init(void);
```
kilograham: not sure `resetd` is a sufficient/good prefix for the public API

kilograham: note also, on this and the below, this is peculiar to TinyUSB - as we think about CherryUSB etc, we (i haven't looked in detail) should consider what if any of this is exposed (and whether we should expose them again in a separate header - e.g. one for TinyUSB one for CherryUSB)

will-v-pi: I will rename them to `pico_usb_reset_interface_...`

Note that these functions are not supposed to be exposed as a public API - the only thing that should be used publicly is the `_resetd_driver` struct, I will try to refactor it so that's the case

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface.h
will-v-pi: Have made a separate `usb_reset_interface_device.h` file which should be included instead of `usb_reset_interface.h`

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface.h
will-v-pi: I made a separate TinyUSB header, and made it so this should be compatible with the CherryUSB PR (as that only uses the `usb_reset_interface.h` header)

I couldn't refactor the functions out of the header, because I think the `pico_usb_reset_interface_driver` has to be defined in the header as `static const` in order for the debugprobe inclusion to compile, and therefore the functions must be in the header too

kilograham: > I will rename them to `pico_usb_reset_interface_...`
> 
> Note that these functions are not supposed to be exposed as a public API - the only thing that should be used publicly is the `_resetd_driver` struct, I will try to refactor it so that's the case

note `pico_`/`hardware_` is not included in function names; e.g. `dma_foo`

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface_config.h
```diff
@@ -0,0 +1,75 @@
...
+// PICO_CONFIG: PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_MS_OS_20_DESCRIPTOR, If vendor reset interface is included add support for Microsoft OS 2.0 Descriptor, type=bool, default=1 if application is not using TinyUSB directly, 0 otherwise, group=pico_usb_reset_interface
+#ifndef PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_MS_OS_20_DESCRIPTOR
+#if !LIB_TINYUSB_HOST && !LIB_TINYUSB_DEVICE
+#define PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_MS_OS_20_DESCRIPTOR 1
+#else
+#define PICO_STDIO_USB_RESET_INTERFACE_SUPPORT_MS_OS_20_DESCRIPTOR 0
+#endif
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_RESET_INTERFACE_MS_OS_20_DESCRIPTOR_ITF, If vendor reset interface is included the USB interface number for the reset interface, type=int, default=2 if application is not using TinyUSB directly, undefined otherwise, group=pico_usb_reset_interface
```
lurch:
```suggestion
// PICO_CONFIG: PICO_STDIO_USB_RESET_INTERFACE_MS_OS_20_DESCRIPTOR_ITF, If vendor reset interface is included this specifies the USB interface number for the reset interface, type=int, default=2 if application is not using TinyUSB directly, undefined otherwise, group=pico_usb_reset_interface
```
Perhaps this is slightly more readable?

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface_config.h
```diff
@@ -0,0 +1,75 @@
...
+#define PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE 1
+#else
+#define PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE 0
+#endif
+#endif
+
+// PICO_CONFIG: PICO_STDIO_USB_RESET_INCLUDE_APP_DRIVER_CB, Set to 0 if your application defines usbd_app_driver_get_cb, type=bool, default=1 when using pico_usb_reset_interface, 0 otherwise, group=pico_usb_reset_interface
+#ifndef PICO_STDIO_USB_RESET_INCLUDE_APP_DRIVER_CB
+#define PICO_STDIO_USB_RESET_INCLUDE_APP_DRIVER_CB LIB_PICO_STDIO_USB
+#endif
```
lurch:
```suggestion
// PICO_CONFIG: PICO_STDIO_USB_RESET_INCLUDE_DEFAULT_APP_DRIVER_CB, Set to 0 if your application defines its own usbd_app_driver_get_cb function, type=bool, default=1 when using pico_usb_reset_interface, 0 otherwise, group=pico_usb_reset_interface
#ifndef PICO_STDIO_USB_RESET_INCLUDE_DEFAULT_APP_DRIVER_CB
#define PICO_STDIO_USB_RESET_INCLUDE_DEFAULT_APP_DRIVER_CB LIB_PICO_STDIO_USB
#endif
```
Maybe? :man_shrugging:

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface_device.h
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (c) 2025 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _PICO_USB_RESET_INTERFACE_DEVICE_H
+#define _PICO_USB_RESET_INTERFACE_DEVICE_H
+
+/** \file usb_reset_interface_device.h
+ *  \defgroup pico_usb_reset_interface pico_usb_reset_interface
```
lurch, line 11: Does this new `pico_usb_reset_interface` group need adding to `docs/index.h` ?

popcornmix:
Any reason this has stalled? I'd quite like to use this, so my project can handle reset to reconfigure in a new mode.

mtfurlan:
I've been using this PR in a project for a while now, which isn't great, but it has been working.

lurch:
Looks like this PR has conflicts which need resolving? @will-v-pi

will-v-pi:
> Looks like this PR has conflicts which need resolving? @will-v-pi

I've rebased it onto develop to resolve those - this PR is awaiting review by @kilograham

kilograham:
oops - sorry i seem to have ignored this!

1. we should probably remove the STDIO_ from the define names, and add backwards-compatibility in the least ugly place possible
2. we have hit a slight snafu obv with the fact that pico/usb_reset_interface (and particularly the header) existed before.. i'm kinda of loath to have a pico/foo where foo.h is not the API, so i wonder if we should rename this to something else

will-v-pi:
> oops - sorry i seem to have ignored this!
> 
> 1. we should probably remove the STDIO_ from the define names, and add backwards-compatibility in the least ugly place possible
> 2. we have hit a slight snafu obv with the fact that pico/usb_reset_interface (and particularly the header) existed before.. i'm kinda of loath to have a pico/foo where foo.h is not the API, so i wonder if we should rename this to something else

I can rename the new library to `pico_usb_reset_interface_device`, as the header name you actually include to use it is already `usb_reset_interface_device.h`. That will make it the longest library name, but it's only slightly longer than `pico_bootsel_via_double_reset` so it's probably fine. I can also move the new headers out of `common` and into that library.

will-v-pi:
Actually, maybe just `pico_usb_reset` is fine for the library name, given it also includes the resetting via baud rate

src/common/pico_usb_reset_interface_headers/include/pico/usb_reset_interface_config.h
```diff
@@ -0,0 +1,132 @@
...
+#ifdef PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE // backwards compatibility with SDK <= 2.2.0
+#define PICO_ENABLE_USB_RESET_VIA_VENDOR_INTERFACE PICO_STDIO_USB_ENABLE_RESET_VIA_VENDOR_INTERFACE
+#elif !LIB_TINYUSB_HOST && !LIB_TINYUSB_DEVICE
+#define PICO_ENABLE_USB_RESET_VIA_VENDOR_INTERFACE 1
+#else
+#define PICO_ENABLE_USB_RESET_VIA_VENDOR_INTERFACE 0
+#endif
+#endif
+
+// PICO_CONFIG: PICO_USB_RESET_INCLUDE_DEFAULT_APP_DRIVER_CB, Set to 0 if your application defines its own usbd_app_driver_get_cb function, type=bool, default=1 when using pico_usb_reset_interface, 0 otherwise, group=pico_usb_reset_interface
```
lurch: "default=1 when using pico_usb_reset_interface, 0 otherwise"
:confused: 

Search&replace typo?

will-v-pi:
```suggestion
// PICO_CONFIG: PICO_USB_RESET_INCLUDE_DEFAULT_APP_DRIVER_CB, Set to 0 if your application defines its own usbd_app_driver_get_cb function, type=bool, default=1 when using pico_stdio_usb, 0 otherwise, group=pico_usb_reset_interface
```

src/rp2_common/pico_usb_reset/usb_reset.c
```diff
@@ -0,0 +1,160 @@
...
+#if !PICO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED
+            // wValue layout:
+            //   bits 0-1  : forwarded as the bootrom disable_interface_mask
+            //   bit 7     : 1 if the activity-LED GPIO is active-low
+            //   bit 8     : 1 if an activity-LED GPIO is being specified
+            //   bits 9-15 : activity-LED GPIO number (0-127)
+            if (request->wValue & 0x100) {
+                gpio = request->wValue >> 9u;
+                active_low = request->wValue & 0x80;
+            }
```
lurch, line 114: It _might_ be worth swapping around two of these lines, so that the code better matches the comment?
```suggestion
            // wValue layout:
            //   bits 0-1  : forwarded as the bootrom disable_interface_mask
            //   bit 7     : 1 if the activity-LED GPIO is active-low
            //   bit 8     : 1 if an activity-LED GPIO is being specified
            //   bits 9-15 : activity-LED GPIO number (0-127)
            if (request->wValue & 0x100) {
                active_low = request->wValue & 0x80;
                gpio = request->wValue >> 9u;
            }
```

kilograham:
ok, so there was maybe a bit of a miscommunication on the renaming, but i have looked at it again anyway.

I am cool with leaving it called `pico_usb_reset_interface` since thats really what it is - the original headers can remain.

even though they are long, we should probably keep PICO_USB_RESET_INTERFACE_ in the names to be consistent (i could handle contraction to IFC i guess)

It's kinda ugly, but the real headers really belong in rp2_common, so i'd actually put them with the code in rp2_common, but add them to the pico_usb_reset_interface_headers, and #include them from the original one guarded by a `#if LIB_PICO_USB_RESET_INTERFACE`

will-v-pi:
> ok, so there was maybe a bit of a miscommunication on the renaming, but i have looked at it again anyway.
> 
> I am cool with leaving it called `pico_usb_reset_interface` since thats really what it is - the original headers can remain.
> 
> even though they are long, we should probably keep PICO_USB_RESET_INTERFACE_ in the names to be consistent (i could handle contraction to IFC i guess)
> 
> It's kinda ugly, but the real headers really belong in rp2_common, so i'd actually put them with the code in rp2_common, but add them to the pico_usb_reset_interface_headers, and #include them from the original one guarded by a `#if LIB_PICO_USB_RESET_INTERFACE`

Sorry, I was occupied with other stuff - I think the rename to `pico_usb_reset` makes all of this cleaner so I will do that, as it allows for a separate pico_usb_reset_headers library so everything can be in rp2_common without any hacks in the existing header

src/rp2_common/pico_usb_reset/include/pico/usb_reset_config.h
```diff
@@ -0,0 +1,126 @@
...
+
+#ifndef _PICO_USB_RESET_CONFIG_H
+#define _PICO_USB_RESET_CONFIG_H
+
+// PICO_CONFIG: PICO_USB_RESET_BOOTSEL_ACTIVITY_LED, Optionally define a pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE), type=int, min=0, max=47 on RP2350B, 29 otherwise, group=pico_usb_reset
+#ifdef PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED // backwards compatibility with SDK <= 2.2.0
+#define PICO_USB_RESET_BOOTSEL_ACTIVITY_LED PICO_STDIO_USB_RESET_BOOTSEL_ACTIVITY_LED
+#endif // default is undefined
+
+// PICO_CONFIG: PICO_USB_RESET_BOOTSEL_ACTIVITY_LED_ACTIVE_LOW, Whether pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE) is active low, type=bool, default=0, group=pico_usb_reset
```
lurch: https://github.com/raspberrypi/pico-sdk/pull/2971/files mentions that the activity led active_low setting is ignored on RP2040. Is the same true here? (And if so, should it be mentioned?)

src/rp2_common/pico_usb_reset/include/pico/usb_reset_config.h
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright (c) 2025 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _PICO_USB_RESET_CONFIG_H
+#define _PICO_USB_RESET_CONFIG_H
+
+// PICO_CONFIG: PICO_USB_RESET_BOOTSEL_ACTIVITY_LED, Optionally define a pin to use as bootloader activity LED when BOOTSEL mode is entered via USB (either VIA_BAUD_RATE or VIA_VENDOR_INTERFACE), type=int, min=0, max=47 on RP2350B, 29 otherwise, group=pico_usb_reset
```
lurch: For readability purposes, I wonder if it might make sense to move all these `PICO_USB_RESET_BOOTSEL_` settings below `PICO_USB_RESET_SUPPORT_RESET_TO_BOOTSEL` ? (Since if the latter is set to 0, I guess none of the former settings are applicable?)

src/rp2_common/pico_usb_reset/include/pico/usb_reset_tusb.h
```diff
@@ -0,0 +1,64 @@
...
+static usbd_class_driver_t const usb_reset_interface_driver =
+{
+#if CFG_TUSB_DEBUG >= 2
+    .name = "RESET",
+#endif
+    .init             = usb_reset_interface_init,
+    .reset            = usb_reset_interface_reset,
+    .open             = usb_reset_interface_open,
+    .control_xfer_cb  = usb_reset_interface_control_xfer_cb,
+    .xfer_cb          = usb_reset_interface_xfer_cb,
```
lurch, line 60: To match the renaming done in a recent commit, would it also make sense to rename these functions to `usb_reset_init` etc.? Or does the `usb_reset_interface_` naming still make sense here? :shrug:

will-v-pi: I think it still makes sense here, as these functions are specific to the USB interface that handles the reset

src/rp2_common/pico_usb_reset/usb_reset.c
will-v-pi: I've swapped them, and replaced `0x100` etc with `(1u << 8)` to make that clearer as well

src/rp2_common/pico_usb_reset/include/pico/usb_reset_tusb.h
```diff
@@ -0,0 +1,64 @@
...
+  \
+  /* MS OS 2.0 Compatible ID descriptor: length, type, compatible ID, sub compatible ID */ \
+  U16_TO_U8S_LE(0x0014), U16_TO_U8S_LE(MS_OS_20_FEATURE_COMPATBLE_ID), 'W', 'I', 'N', 'U', 'S', 'B', 0x00, 0x00, \
+  0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, /* sub-compatible */ \
+  \
+  /* MS OS 2.0 Registry property descriptor: length, type */ \
+  U16_TO_U8S_LE(0x0080), U16_TO_U8S_LE(MS_OS_20_FEATURE_REG_PROPERTY), \
+  U16_TO_U8S_LE(0x0001), U16_TO_U8S_LE(0x0028), /* wPropertyDataType, wPropertyNameLength and PropertyName "DeviceInterfaceGUID" in UTF-16 */ \
+  'D', 0x00, 'e', 0x00, 'v', 0x00, 'i', 0x00, 'c', 0x00, 'e', 0x00, 'I', 0x00, 'n', 0x00, 't', 0x00, 'e', 0x00, \
+  'r', 0x00, 'f', 0x00, 'a', 0x00, 'c', 0x00, 'e', 0x00, 'G', 0x00, 'U', 0x00, 'I', 0x00, 'D', 0x00, 0x00, 0x00, \
```
lurch, line 31: Not directly-related to this PR (so feel free to ignore this suggestion), but _maybe_ this could be reformatted to:
```suggestion
  'D', 0,     'e', 0,     'v', 0,     'i', 0,     'c', 0,     'e', 0, \
  'I', 0,     'n', 0,     't', 0,     'e', 0,     'r', 0,     'f', 0, \
  'a', 0,     'c', 0,     'e', 0,     'G', 0,     'U', 0,     'I', 0, \
  'D', 0,     0, 0, \
```
as this would then match the formatting of the `{bc7398c1-...` UTF-16 string below?

kilograham (approved):

PR #2635, opened 2025-08-22
Fix race in gpio_set_irq_enabled_with_callback

will-v-pi:
When disabling, if you get an interrupt after `gpio_set_irq_callback` before `gpio_set_irq_enabled` then you get an `unhandled_user_irq`

Fix is to change where you `gpio_set_irq_callback` depending on whether you're enabling or disabling

### DIFF
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -196,9 +196,11 @@ void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled) {
 }
 
 void gpio_set_irq_enabled_with_callback(uint gpio, uint32_t events, bool enabled, gpio_irq_callback_t callback) {
-    // first set callback, then enable the interrupt
-    gpio_set_irq_callback(callback);
+    // when enabling, first set callback, then enable the interrupt
+    // when disabling, first disable the interrupt, then clear callback
+    if (enabled) gpio_set_irq_callback(callback);
     gpio_set_irq_enabled(gpio, events, enabled);
+    if (!enabled) gpio_set_irq_callback(callback);
     if (enabled) irq_set_enabled(IO_IRQ_BANK0, true);
 }
 


src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -196,9 +196,11 @@ void gpio_set_irq_enabled(uint gpio, uint32_t events, bool enabled) {
...
 
 void gpio_set_irq_enabled_with_callback(uint gpio, uint32_t events, bool enabled, gpio_irq_callback_t callback) {
-    // first set callback, then enable the interrupt
-    gpio_set_irq_callback(callback);
+    // when enabling, first set callback, then enable the interrupt
+    // when disabling, first disable the interrupt, then clear callback
+    if (enabled) gpio_set_irq_callback(callback);
     gpio_set_irq_enabled(gpio, events, enabled);
+    if (!enabled) gpio_set_irq_callback(callback);
     if (enabled) irq_set_enabled(IO_IRQ_BANK0, true);
```
lurch, line 204: Is this function likely to get used in performance-critical code? If it is, perhaps it's worth reducing the branches? :thinking:  (or is this one of those cases where the compiler is able to do this optimisation automatically?)
```suggestion
    // when enabling, first set callback, then enable the interrupt
    // when disabling, first disable the interrupt, then clear callback
    if (enabled) {
        gpio_set_irq_callback(callback);
        gpio_set_irq_enabled(gpio, events, enabled);
        irq_set_enabled(IO_IRQ_BANK0, true);
    } else {
        gpio_set_irq_enabled(gpio, events, enabled);
        gpio_set_irq_callback(callback);
    }
```

will-v-pi: I'll defer to @kilograham on this - I'm happy with either

kilograham: it is fine; the compiler will sort it out

kilograham (approved):

PR #2637, opened 2025-08-26
remove spurious cmake code, which seems to break some versions of cmake

kilograham:


peterharperuk (approved):
No idea why that's there!

Issue #2495, opened 2025-06-01
tud_task() inhibits wfe on rp2350, turns low power loops into busy loops


rlcamp:
The common pattern of

    while (!condition) {
        tud_task();
        __wfe();
    }

which should put the processor to sleep between interrupts, becomes a busy loop by default on RP2350 due to the use of pico-sdk spinlocks within `tud_task()`, which behave like `sev()` due to reasons discussed in https://github.com/raspberrypi/pico-sdk/issues/1812. Mitigations include disabling EXTEXCALL (which modifies the behaviour of ALL atomics and other exclusive load/stores between cores), or putting

    #undef CFG_TUSB_OS
    #define CFG_TUSB_OS OPT_OS_NONE

in tusb_config.h, disabling the pico-specific code paths in the TinyUSB OS abstraction layer. The latter is less of a big hammer but still has consequences wr interacting with TinyUSB from both cores.

kilograham:
I think you need to do

```
while (!condition) {
    tud_task();
    if (!tud_task_event_ready()) {
        __wfe();
    }
}
```

As such it can't really be moved into `tud_task()`

I would argue that relying on processor events should never be idiomatic

Note, TinyUSB however could add a `tud_task_and_block()` of course in that case you have no way to break it out of the  blocking for your condition, you'd have to rely on (as you already did) that presumably you cause an event when setting your condition (and you'd have to assume that TinyUSB is also waiting on the event)

kilograham:
ah, sorry ignore me..

kilograham:
Ok, trying again:

```
    while (!condition) {
        tud_task();
        // do these at the end of tud_task() on RP2350 to consume the event (I'd suggest putting them in there)
        __sev();
        __wfe();
        // require tud_task_event_ready() to avoid getting stuck if another IRQ happened prior to the above `__wfe()`
        if (!tud_task_event_ready()) {
            __wfe();
        }
    }
```

Note we can't really do `__sev(); __wfe()` in the spin lock code, as eating events is certainly a non-desirable side effect in many cases.

However I'd suggest that someone calling `tud_task()` should not reasonably expect any guarantees related to processor events. For this reason I'd suggest that calling `tud_task()` followed by `__wfe()` is a poor idiom in general, that has just happened to work up until now.

arguably there should be a `tud_block_until_event()` or similar

[...]

PR #2645, opened 2025-09-04
Fix hardware spinlock IDs on RP2350

will-v-pi:
Use spinlock IDs that are unaffected by errata E2

This allows use of regular spinlocks rather than software spinlocks, which should fix issues like https://github.com/raspberrypi/pico-sdk/issues/2495

### DIFF
--- a/src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
+++ b/src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
@@ -10,7 +10,7 @@
 #include "pico.h"
 #include "hardware/sync.h"
 
-// PICO_CONFIG: PICO_USE_SW_SPIN_LOCKS, Use software implementation for spin locks, type=bool, default=1 on RP2350 due to errata, group=hardware_sync
+// PICO_CONFIG: PICO_USE_SW_SPIN_LOCKS, Use software implementation for spin locks, type=bool, default=1 on RP2350 due to errata E2, group=hardware_sync
 #ifndef PICO_USE_SW_SPIN_LOCKS
 #if PICO_RP2350
 #define PICO_USE_SW_SPIN_LOCKS 1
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
```diff
@@ -19,53 +19,93 @@
...
+#endif
 
 // PICO_CONFIG: PICO_SPINLOCK_ID_TIMER, Spinlock ID for Timer protection, min=0, max=31, default=10, group=hardware_sync
 #ifndef PICO_SPINLOCK_ID_TIMER
+#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS
+#define PICO_SPINLOCK_ID_TIMER 6
+#else
 #define PICO_SPINLOCK_ID_TIMER 10
 #endif
+#endif
```
lurch, line 36: RP2350-E2 says that 10 is a "safe" spinlock register, so is there any reason to re-define spinlock 10 as `PICO_SPINLOCK_ID_RAND` rather than leaving it defined as `PICO_SPINLOCK_ID_TIMER` ?

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
```diff
@@ -19,53 +19,93 @@
 
 // PICO_CONFIG: PICO_SPINLOCK_ID_IRQ, Spinlock ID for IRQ protection, min=0, max=31, default=9, group=hardware_sync
 #ifndef PICO_SPINLOCK_ID_IRQ
+#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS
+#define PICO_SPINLOCK_ID_IRQ 5
+#else
 #define PICO_SPINLOCK_ID_IRQ 9
```
lurch, line 25: @kilograham In the "original" version of this code, do you remember why the "defined" spinlocks start at 9? I.e. is it actually okay for the "modified" version of this code to use spinlocks 5, 6 and 7?

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
will-v-pi: I thought it better to keep them in the same order

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
kilograham: i was getting to comment on this; the main reason is that they are used by pico-extras stuff, so yeah, we want to be careful about what we assign into that range (and pick something that isn't likely to be used at runtime) - i just havent spent any brain cycles picking which, but @will-v-pi you are free to ;-)

will-v-pi: In pico-extras, scan video uses 2-5, and audio uses 6&7 - so that would mean starting SDK ones from 10. Note that the pico-extras spin locks will be affected by E2 too, so if we wanted to assign all those safe ones too, the SDK ones would need to start from 19

Alternatively, I could put errors in the ifdefs in pico-extras which these clash with, to say that if on RP2350 and not using software spin locks, you need to define your own IDs for the clashes? I think I’d prefer this option, as it gives the SDK all the spin locks, and the people it affects can pick their own spin locks to use (either SDK bits they don’t use, the OS ones, or from the ranges at the end)

will-v-pi: I've opened https://github.com/raspberrypi/pico-extras/pull/99 in pico-extras to throw errors when not using software spin locks, with the error message explaining that you need to define the IDs

kilograham (approved):

PR #2651, opened 2025-09-06
Use 'constexpr' when including pio header files from a C++ compiler

### LINKED ISSUE #2650: Cannot perform static_assert in C++ on header files generated by pioasm
Because the ```pio_program``` structure is not declared ```constexpr```, it's not possible to perform compile-time checks on the contents.

In my project, I have a couple of PIO programs that I would like to ensure both fit in a single PIO instance (i.e. in the 32-word memory) and discover at compile-time if changes are made that would prevent this, rather than with a panic at run-time.

Obviously, ```constexpr``` is a C++ keyword so can't be used if the header file is included from a plain C file, but the keyword to use can be determined via checking whether ```__cplusplus``` is defined.

rhulme:
In particular this allows compile-time checks to be performed on the contents of pio_program structs via 'static_assert'.

This could be used (for example) to confirm that multiple PIO programs fit in a single PIO module and abort compilation if not.

Fixes #2650

### DIFF
--- a/tools/pioasm/c_sdk_output.cpp
+++ b/tools/pioasm/c_sdk_output.cpp
@@ -79,6 +79,13 @@ struct c_sdk_output : public output_format {
         fprintf(out, "#endif\n");
         fprintf(out, "\n");
 
+        fprintf(out, "#ifdef __cplusplus\n");
+        fprintf(out, "#define PIO_CONST constexpr\n");
+        fprintf(out, "#else\n");
+        fprintf(out, "#define PIO_CONST const\n");
+        fprintf(out, "#endif\n");
+        fprintf(out, "\n");
+
         output_symbols(out, "", source.global_symbols);
 
         for (const auto &program : source.programs) {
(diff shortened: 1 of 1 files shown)


tools/pioasm/c_sdk_output.cpp
```diff
@@ -79,6 +79,13 @@ struct c_sdk_output : public output_format {
         fprintf(out, "#endif\n");
         fprintf(out, "\n");
 
+        fprintf(out, "#ifdef __cplusplus\n");
```
lurch: https://en.cppreference.com/w/cpp/language/constexpr.html says that `constexpr` was only added in C++11, so should this line perhaps be
```suggestion
        fprintf(out, "#if defined(__cplusplus) && defined(__cpp_constexpr)\n");
```
so that it doesn't break older C++ compilers? (Or perhaps simply `#ifdef __cpp_constexpr` is sufficient?)

rhulme: Yes, fair point.  I've update the check.

kilograham (dismissed):

kilograham (approved):

PR #2652, opened 2025-09-06
src/boards/include/boards/*.h: remove executable bit

josch:
find src -type f -perm /a+x -name '*.h' -print0 | xargs --null chmod -x


lurch:
Yeah, the executable-bit sometimes gets added to files when people use `git` on Windows.

lurch (approved):

PR #2656, opened 2025-09-10
Add mention of PICO_PLATFORM=rp2350-arm-ns to README

lurch:
...as discussed with @will-v-pi


kilograham:
Yeah, this we should keep around, although u can’t strictly build much with -ns yet

lurch:
Feel free to defer this to a later milestone, if you think that'd be appropriate :slightly_smiling_face:

Issue #2662, opened 2025-09-12
Unused parameter warning for PICO_RP2040 build


martinbudden:
For `PICO_RP2040` builds `dma_encode_transfer_count_with_self_trigger` creates an unused parameter warning when built with `-Wunused-parameter` compiler flag. See below.

```c
static inline uint32_t dma_encode_transfer_count_with_self_trigger(uint transfer_count) {
#if PICO_RP2040
    panic_unsupported();
#else
    return dma_encode_transfer_count(transfer_count) | (DMA_CH0_TRANS_COUNT_MODE_VALUE_TRIGGER_SELF << DMA_CH0_TRANS_COUNT_MODE_LSB);
#endif
}
```

martinbudden:
@peterharperuk Hi!

peterharperuk:
It;'s a small world @martinbudden !

lurch:
Fix merged into `develop`

PR #2663, opened 2025-09-12
Fix unused parameter warning.

martinbudden:
Fixes issue #2662.

Adds `(void)transfer_count;` for parameter `transfer_count` to avoid warning.

### DIFF
--- a/src/rp2_common/hardware_dma/include/hardware/dma.h
+++ b/src/rp2_common/hardware_dma/include/hardware/dma.h
@@ -556,6 +556,7 @@ static inline uint32_t dma_encode_transfer_count(uint transfer_count) {
  */
 static inline uint32_t dma_encode_transfer_count_with_self_trigger(uint transfer_count) {
 #if PICO_RP2040
+    (void)transfer_count;
     panic_unsupported();
 #else
     return dma_encode_transfer_count(transfer_count) | (DMA_CH0_TRANS_COUNT_MODE_VALUE_TRIGGER_SELF << DMA_CH0_TRANS_COUNT_MODE_LSB);


lurch:
Would it be "better" to use the `__unused` attribute, e.g. https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_dma/include/hardware/dma.h#L59 ?

martinbudden:
I'm not familiar with the `__unused` attribute. Do you mean?

```c
static inline uint32_t dma_encode_transfer_count_with_self_trigger(uint transfer_count) {
#if PICO_RP2040
    __unused transfer_count;
    panic_unsupported();
#else
    return dma_encode_transfer_count(transfer_count) | (DMA_CH0_TRANS_COUNT_MODE_VALUE_TRIGGER_SELF << DMA_CH0_TRANS_COUNT_MODE_LSB);
#endif
}
```

Since `transfer_count` is only unused in the `PICO_RP2040` build.

Or do you mean

```c
static inline uint32_t dma_encode_transfer_count_with_self_trigger(__unused uint transfer_count) {
```

Either way, I'll happily update the PR.

lurch:
I'm not super-familiar either, I just remember seeing it used a lot of times (e.g. see the line that I linked to).
Perhaps @will-v-pi or @kilograham will be able to confirm the preferred style in `pico-sdk`. (I suspect they're just different ways of doing the same thing?)

martinbudden:
I've also seen `(void)` used in the code. As you say, it's just a question of confirming the prefered style.

kilograham (approved):

kilograham:
yeah; i think `(void)` is probably preferred when there is an #ifdef path where it is still used (though the existing uses aren't necessarily that way)

kilograham (approved):

PR #2671, opened 2025-09-16
Add PICO_CONFIG entries for the metadata block configuration defines

will-v-pi:
Includes rejigging it a bit, as the checking script requires #defines for all configs

### DIFF
--- a/src/rp2_common/pico_crt0/embedded_start_block.inc.S
+++ b/src/rp2_common/pico_crt0/embedded_start_block.inc.S
@@ -1,10 +1,12 @@
-#if !defined(PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM)
+// PICO_CONFIG: PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM, Include an image type item in the metadata block, default=0 on RP2040, 1 otherwise, type=bool, group=pico_crt0
+#ifndef PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM
 // todo decide whether we want this item for RP2040 by default, probably not (there are a zilloon binaries out there without it)
 #if !PICO_RP2040
 #define PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM 1
 #endif
 #endif
 
+// PICO_CONFIG: PICO_CRT0_INCLUDE_PICOBIN_VECTOR_TABLE_ITEM, Include a vector table item in the metadata block, default=1 for no_flash binaries, 0 otherwise or on RP2040, type=bool, group=pico_crt0
 #ifndef PICO_CRT0_INCLUDE_PICOBIN_VECTOR_TABLE_ITEM
 // If no_flash bin, then include a vector table item
 #if PICO_NO_FLASH && !PICO_RP2040
@@ -12,6 +14,7 @@
 #endif
 #endif
 
+// PICO_CONFIG: PICO_CRT0_INCLUDE_PICOBIN_ENTRY_POINT_ITEM, Include an entry point item in the metadata block, default=1 for Risc-V, 0 otherwise, type=bool, group=pico_crt0
 #ifndef PICO_CRT0_INCLUDE_PICOBIN_ENTRY_POINT_ITEM
 // On RISC-V the default entry point from bootrom is the start of the binary, but
 // we have our vtable at the start, so we must include an entry point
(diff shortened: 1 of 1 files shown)


andygpz11 (approved):
LGTM

src/rp2_common/pico_crt0/embedded_start_block.inc.S
```diff
@@ -20,55 +23,72 @@
 #endif
 #endif
 
+// PICO_CONFIG: PICO_CRT0_INCLUDE_PICOBIN_BLOCK, Include a metadata block at the start of the image, default=PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM, type=bool, group=pico_crt0
 #ifndef PICO_CRT0_INCLUDE_PICOBIN_BLOCK
 #define PICO_CRT0_INCLUDE_PICOBIN_BLOCK PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM
```
lurch, line 28: Would there ever be a situation where you want to have `PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM` set to 0, but `PICO_CRT0_INCLUDE_PICOBIN_BLOCK` set to 1?
(I'm wondering if the default value ought to be something like
```c
#define PICO_CRT0_INCLUDE_PICOBIN_BLOCK (PICO_CRT0_INCLUDE_PICOBIN_IMAGE_TYPE_ITEM || PICO_CRT0_INCLUDE_PICOBIN_VECTOR_TABLE_ITEM || PICO_CRT0_INCLUDE_PICOBIN_ENTRY_POINT_ITEM || etc. )
```

?)

src/rp2_common/pico_crt0/embedded_start_block.inc.S
```diff
@@ -20,55 +23,72 @@
...
+#endif
+
 #if PICO_CRT0_IMAGE_TYPE_TBYB
 #define CRT0_TBYB_FLAG PICOBIN_IMAGE_TYPE_EXE_TBYB_BITS
 #else
 #define CRT0_TBYB_FLAG 0
 #endif
 
+// PICO_CONFIG: PICO_CRT0_VERSION_MAJOR, The major version for the metadata block (ignored if 0), default=0, type=int, group=pico_crt0
+// PICO_CONFIG: PICO_CRT0_VERSION_MINOR, The minor version for the metadata block (ignored if 0), default=0, type=int, group=pico_crt0
```
lurch, line 48:
```suggestion
// PICO_CONFIG: PICO_CRT0_VERSION_MAJOR, The major version for the metadata block (ignored if 0), default=0 if PICO_CRT0_VERSION_MINOR is defined, type=int, group=pico_crt0
// PICO_CONFIG: PICO_CRT0_VERSION_MINOR, The minor version for the metadata block (ignored if 0), default=0 if PICO_CRT0_VERSION_MAJOR is defined, type=int, group=pico_crt0
```
?

src/rp2_common/pico_crt0/embedded_start_block.inc.S
will-v-pi: Yes (eg see RP2040 universal examples) - but you wouldn’t want any of those other items as those are only valid for image_def blocks, instead you would just want a block with a single ignored item (which is what get produced when you set those values), so I think the default is fine

src/rp2_common/pico_crt0/embedded_start_block.inc.S
will-v-pi: I think that might create confusion - those bits are just to ensure that if you only define one of them, the other is defined so you get a valid version item. The default is that neither of them are defined, so you don’t get a version item

src/rp2_common/pico_crt0/embedded_start_block.inc.S
```diff
@@ -20,55 +23,72 @@
...
        PICOBIN_IMAGE_TYPE_EXE_CHIP_AS_BITS(RP2350) | \
        CRT0_TBYB_FLAG
 #elif PICO_RP2040
-.hword PICOBIN_IMAGE_TYPE_IMAGE_TYPE_AS_BITS(EXE) | \
+#define PICO_CRT0_IMAGE_TYPE_ITEM_VALUE \
+       PICOBIN_IMAGE_TYPE_IMAGE_TYPE_AS_BITS(EXE) | \
        PICOBIN_IMAGE_TYPE_EXE_SECURITY_AS_BITS(NS) | \
        PICOBIN_IMAGE_TYPE_EXE_CPU_AS_BITS(ARM) | \
        PICOBIN_IMAGE_TYPE_EXE_CHIP_AS_BITS(RP2040) | \
        CRT0_TBYB_FLAG
```
lurch, line 69: Does it even make sense to include `CRT0_TBYB_FLAG` for `PICO_RP2040` ?

will-v-pi: You might have user code that selects based on it (eg an RP2350 booting some RP2040s from its partitions), so it’s best to leave it in

src/rp2_common/pico_crt0/embedded_start_block.inc.S
lurch: Okay :slightly_smiling_face:

kilograham (approved):

PR #2709, opened 2025-10-29
[Bazel] Add bzl_library targets

RyanDraves:
Add some `bzl_library` targets for `.bzl` files. This enables [stardoc to do its thing](https://blog.aspect.build/stardocs-on-bcr) and is helpful in its own right for validating rule & macro dependencies.

Notes:
- Open question: should this PR add bzl_library targets for all `.bzl` files? I added the ones I personally depend on, but adding it for more files requires bumping the `rules_cc` version; might be poking the bear?
- I haven't filed an issue yet, will do as desired

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,6 +4,7 @@ module(
 )
 
 bazel_dep(name = "platforms", version = "0.0.9")
+bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "bazel_skylib", version = "1.6.1")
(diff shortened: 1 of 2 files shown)


kilograham:
cc @armandomontanez

armandomontanez:
Bumping rules_cc should be fine.

I should probably take a pass at some of the .bzl libraries and re-assess their intended visibility. For example, `configurable_feature`  was an implementation detail to make it easier to write/maintain the toolchains. I don't think it should be considered part of the public interface.

Btw, thanks for kicking this off, stardoc in the BCR is very cool.

Issue #2736, opened 2025-11-29
Host GPIO improvements


rhulme:
The host platform GPIO code has a few issues:
* Only one function has been declared 'weak' so any project wishing to implement a more concrete GPIO simulation has to disable that part of the SDK (by setting SKIP_HARDWARE_GPIO to 1 before importing the SDK) and provide it's own implementation of the full interface.
* The function parameters are not marked as 'unused' so generate lots of warnings if a higher warning level has been set.
* Many functions from rp2_common are missing

kilograham:
fix merged into develop

PR #2737, opened 2025-11-29
Host GPIO improvements

rhulme:
As stated in #2736, the host GPIO implementation is currently a little too basic and is missing quite a few functions that have been introduced in recent SDK versions.

This PR has several commits to make the implementation a bit more usable for running tests on the host.

### DIFF
--- a/src/host/hardware_gpio/include/hardware/gpio.h
+++ b/src/host/hardware_gpio/include/hardware/gpio.h
@@ -12,9 +12,11 @@ extern "C" {
 #endif
 
 #include "pico.h"
+#include "hardware/irq.h"
 
-enum gpio_function {
+typedef enum gpio_function {
     GPIO_FUNC_XIP = 0,
+    GPIO_FUNC_XIP_CS1 = 0,
     GPIO_FUNC_SPI = 1,
     GPIO_FUNC_UART = 2,
     GPIO_FUNC_I2C = 3,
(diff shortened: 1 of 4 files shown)


src/host/hardware_gpio/include/hardware/gpio.h
```diff
@@ -12,9 +12,11 @@ extern "C" {
 #endif
 
 #include "pico.h"
+#include "hardware/irq.h"
 
-enum gpio_function {
+typedef enum gpio_function {
```
lurch, line 17: Looks like this is a "superset" of RP2040's and RP2350's `gpio_function` enums? That makes sense, but I'm not sure if there's any point in having `GPIO_FUNC_XIP` and `GPIO_FUNC_XIP_CS1` both set to `0`?

src/host/hardware_gpio/include/hardware/gpio.h
```diff
@@ -24,7 +26,32 @@ enum gpio_function {
...
+    GPIO_FUNC_UART_AUX = 13,
+    GPIO_FUNC_NULL = 0x1f,
+}gpio_function_t;
+
+enum gpio_dir {
+    GPIO_OUT = 1u, ///< set GPIO to output
+    GPIO_IN = 0u,  ///< set GPIO to input
+};
+
+enum gpio_irq_level {
```
lurch, line 41: I wonder if it _might_ make sense to move some of these enums to `src/common/` to prevent having to maintain separate copies in both `src/rp2_common/` and `src/host/`? (But perhaps that'd be more effort than it's worth? :shrug: )

src/host/hardware_gpio/include/hardware/gpio.h
```diff
@@ -94,61 +158,91 @@ void gpio_init_mask(uint gpio_mask);
...
+// 0 = in
+uint gpio_get_dir(uint gpio);
+
+#if PICO_SECURE
+void gpio_assign_to_ns(uint gpio, bool ns);
+#endif
+
 // debugging
 #ifndef PICO_DEBUG_PIN_BASE
 #define PICO_DEBUG_PIN_BASE 19u
```
lurch: For completeness, maybe you also want `PICO_DEBUG_PIN_COUNT` ?

src/host/hardware_gpio/include/hardware/gpio.h
rhulme: Yes, it's a superset.  I set `GPIO_FUNC_XIP_CS1` to the same value as `GPIO_FUNC_XIP` as each enum only exists on one platform (rp2040/rp2350b) so I don't think there should be a problem but it's not something I have strong feelings on either way.

src/host/hardware_gpio/include/hardware/gpio.h
rhulme: I agree but I didn't want to make too many assumptions.  I think there are a lot of enums and #defines that could theoretically be moved to `src/common/` to aid additional implementations under `src/host/` (e.g. `pio_instructions.h`).
Happy to make any changes though.

lurch: I'll leave this up to the pico-sdk maintainers :slightly_smiling_face:

src/host/hardware_gpio/include/hardware/gpio.h
rhulme: Yes, good point.  I think all the 'debug pin' lines can be taken from the rp2_common header.

src/host/hardware_gpio/include/hardware/gpio.h
kilograham: I think we can move "commoning" up of defines to a separate PR

kilograham (approved):

PR #2739, opened 2025-11-30
Fix incorrect Bazel dependency in host/hardware_irq

### LINKED ISSUE #2738: Bazel dependency on hardware_claim in host/hardware_irq is incorrect
The Bazel file for host/hardware_irq has a dependency on host/hardware_claim, which doesn't exist.

It should be **common**/hardware_claim.

rhulme:
Bazel configuration adds a dependency on hardware_claim in 'host' but it doesn't exist there.  It's in 'common'.

Fixes #2738

### DIFF
--- a/src/common/pico_base_headers/BUILD.bazel
+++ b/src/common/pico_base_headers/BUILD.bazel
@@ -92,6 +92,7 @@ cc_library(
     includes = ["include"],
     visibility = [
         "//src/common:__subpackages__",
+        "//src/host/hardware_irq:__pkg__",
         "//src/host/hardware_sync:__pkg__",
         "//src/host/hardware_timer:__pkg__",
         "//src/host/pico_platform:__pkg__",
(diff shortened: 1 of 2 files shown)


armandomontanez (approved):
Thanks! Because of #2596, you might be able to remove `tags = ["manual"]` from this too

rhulme:
> Thanks! Because of #2596, you might be able to remove `tags = ["manual"]` from this too

I haven't got my head around how exactly Bazel does things yet.  I just tried running the command from the Bazel build checks (which seem to be permanently failing anyway).

I noticed `host/hardware_irq` still has issues because `PICO_NUM_VTABLE_IRQS` is not defined.  That's not a Bazel issue directly, more a debug/release build issue.  It seems that Bazel creates a debug build by default, so the assertion in `check_irq_param` becomes 'live' and the compiler complains about the undefined value.

lurch:
> I noticed `host/hardware_irq` still has issues because `PICO_NUM_VTABLE_IRQS` is not defined.

More PRs are always welcome! :smile:

rhulme:
> > I noticed `host/hardware_irq` still has issues because `PICO_NUM_VTABLE_IRQS` is not defined.
> 
> More PRs are always welcome! 😄

I was trying to decide what the "right" way to fix it would be :wink:

kilograham (approved):

PR #2751, opened 2025-12-04
Add pico_unique_id library to host platform

### LINKED ISSUE #2750: pico_unique_id missing from host platform
The host platform is missing the `pico_unique_id` library.

rhulme:
Fixes #2750

### DIFF
--- a/src/host/pico_unique_id/unique_id.c
+++ b/src/host/pico_unique_id/unique_id.c
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2021 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "pico/unique_id.h"
+
+PICO_WEAK_FUNCTION_DEF(pico_get_unique_board_id)
+void PICO_WEAK_FUNCTION_IMPL_NAME(pico_get_unique_board_id)(pico_unique_board_id_t *id_out) {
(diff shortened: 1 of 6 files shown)


src/host/pico_unique_id/unique_id.c
```diff
@@ -0,0 +1,38 @@
...
+    for (int i = 0; i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES; i++) {
+        retrieved_id.id[i] = 0xa0 + i;
+    }
+}
+
+PICO_WEAK_FUNCTION_DEF(pico_get_unique_board_id)
+void PICO_WEAK_FUNCTION_IMPL_NAME(pico_get_unique_board_id)(pico_unique_board_id_t *id_out) {
+    *id_out = retrieved_id;
+}
+
```
lurch, line 15: _Perhaps_ it's worth adding a comment (either here or in the header file) that explicitly says what `pico_get_unique_board_id_string` will return for "host" builds?

rhulme: Yep.  I've added comments and removed the static part.  If someone wants to get a different ID, they just need to implement their own version of `pico_get_unique_board_id`, and `pico_get_unique_board_id_string` will automatically follow suit.

lurch: Thanks. But I was suggesting clarifying if (the default implementation of) `pico_get_unique_board_id_string` returns `"A0A1A2A3A4A5A6A7"` or `"A7A6A5A4A3A2A1A0"` :wink:

kilograham (approved):

PR #2759, opened 2025-12-12
Flash functions in PICO_NO_FLASH binaries

### LINKED ISSUE #2796: flash_get_unique_id panic with not supported in no flash builds
We assume that PICO_NO_FLASH binaries means the device does not have a flash chip. But it's possible that you want a ram binary AND want to mess about with flash, e.g. to get the unique flash id.

Suggest adding a config variable that says - I have a flash chip even though the code is a  ram binary. Or maybe the no_flash binary type is misnamed?

peterharperuk:
We assume that PICO_NO_FLASH binaries means the device does not have a flash chip. But it's possible that you want a ram binary AND want to mess about with flash, e.g. something that uses the unique flash id.

Add PICO_INCLUDE_FLASH_FUNCTIONS  which forces some flash functions to be included.

### DIFF
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -52,6 +52,11 @@
 
 // PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes, type=int, default=Usually provided via board header, group=hardware_flash
 
+// PICO_CONFIG: PICO_INCLUDE_FLASH_ID_FUNCTIONS, Force inclusion of flash unique id functionality in PICO_NO_FLASH binaries, type=bool, default=0, group=hardware_flash
+#ifndef PICO_INCLUDE_FLASH_ID_FUNCTIONS
+#define PICO_INCLUDE_FLASH_ID_FUNCTIONS 0
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -52,9 +52,9 @@
 
 // PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes, type=int, default=Usually provided via board header, group=hardware_flash
 
-// PICO_CONFIG: PICO_INCLUDE_FLASH_ID_FUNCTIONS, Force inclusion of flash unique id functionality in PICO_NO_FLASH binaries, type=bool, default=0, group=hardware_flash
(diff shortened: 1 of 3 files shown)


peterharperuk:
Needed for this https://github.com/raspberrypi/pico-examples/pull/653

will-v-pi:
Probably would be nicer if `PICO_INCLUDE_FLASH_UTILS ` defaulted to `!PICO_NO_FLASH`, rather than needing to check both in all the `#if`s?
```
#ifndef PICO_INCLUDE_FLASH_UTILS 
#define PICO_INCLUDE_FLASH_UTILS  !PICO_NO_FLASH
#endif
```
It would also need a `PICO_CONFIG` comment, so it appears in the docs

peterharperuk:
> defaulted to !PICO_NO_FLASH

Then you would be able to remove these functions by setting PICO_INCLUDE_FLASH_UTILS=0 in flash builds. Which will cause odd behaviour?

will-v-pi:
I've ended up adding the first bit to allow `flash_do_cmd` in no_flash binaries in my PSRAM PR #2919 - maybe the other half of this would be better expressed as `PICO_INCLUDE_FLASH_ID_FUNCTIONS` instead?

kilograham:
can we revisit this now with @will-v-pi referenced changes

peterharperuk:
I've updated this.

will-v-pi (dismissed):

kilograham (changes requested):
i will come back to this - need to think of a better name i think

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -52,6 +52,11 @@
 
 // PICO_CONFIG: PICO_FLASH_SIZE_BYTES, size of primary flash in bytes, type=int, default=Usually provided via board header, group=hardware_flash
 
+// PICO_CONFIG: PICO_INCLUDE_FLASH_ID_FUNCTIONS, Force inclusion of flash unique id functionality in PICO_NO_FLASH binaries, type=bool, default=0, group=hardware_flash
```
kilograham:
```suggestion
// PICO_CONFIG: PICO_ALWAYS_INCLUDE_FLASH_ID_FUNCTIONS, Force inclusion of flash unique id functionality even in PICO_NO_FLASH binaries, type=bool, default=0, group=hardware_flash
```

kilograham (changes requested):

src/rp2_common/hardware_flash/include/hardware/flash.h
peterharperuk: done

will-v-pi (approved):

kilograham (approved):

PR #2760, opened 2025-12-15
Fix uninit r2 in divmod_u32u32_unsafe and divmod_s32s32_unsafe

### LINKED ISSUE #2773: Uninitialized r2 in divmod_x32x32_unsafe without PICO_DIVIDER_DISABLE_INTERRUPTS
When the ``divmod_s32s32_unsafe`` and ``divmod_u32u32_unsafe`` is called without enabled ``PICO_DIVIDER_DISABLE_INTERRUPTS``, ``ldr r2, =SIO_BASE`` will be bypassed. Later access to ``SIO_DIV`` will crash.

andylinpersonal:
When the ``divmod_s32s32_unsafe`` and ``divmod_u32u32_unsafe``
is called without enabled ``PICO_DIVIDER_DISABLE_INTERRUPTS``,
``ldr r2, =SIO_BASE`` will be bypassed. Later access to ``SIO_DIV``
will crash.

Fixes #2773

### DIFF
--- a/src/rp2_common/pico_divider/divider_hardware.S
+++ b/src/rp2_common/pico_divider/divider_hardware.S
@@ -111,6 +111,7 @@ regular_func divmod_s32s32
     lsrs r3, #SIO_DIV_CSR_DIRTY_SHIFT_FOR_CARRY
     bcs divmod_s32s32_savestate
 regular_func divmod_s32s32_unsafe
+    ldr r2, =SIO_BASE
 #else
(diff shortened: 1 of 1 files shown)


kilograham:
good spot; i think i'd probably want to extract the unsafe variants into their own code/section (I doubt they are that ocmmonly used) to avoid the extra 2 cycles in the common divide code path

andylinpersonal:
Maybe it's useful when used inside RTOS. In those case we can delegate the saving stuff to context-switch code.

kilograham:
superseded by #2976

PR #2772, opened 2025-12-27
Fix RP2040 crash with PICO_*_IN_RAM and NaN propagation

### LINKED ISSUE #2771: Enabling both PICO_*_IN_RAM and PICO_*_PROPAGATE_NANS will break __check_nan_* calls on RP2040
When we enable ``PICO_<FLOAT|DOUBLE>_IN_RAM`` and ``PICO_<FLOAT|DOUBLE>_PROPAGATE_NANS`` simutaneously, the ``__check_nan_<f|d><1|2>`` will break on RP2040 port.

andylinpersonal:
``__check_nan<f|d><1|2>`` is hard-coded into ``.text`` instead of being wrapped with ``<double|float>_section`` like other ``__aeabi*`` functions. When those functions are moved to SRAM, this leaves ``__check_nan*`` in flash, causing GCC to emit a veneer call that clobbers ``ip`` and corrupts the saved ``lr`` in ``wrapper_func_<d|f><1|2>``.

Fixes #2771

### DIFF
--- a/src/rp2_common/pico_double/double_aeabi_rp2040.S
+++ b/src/rp2_common/pico_double/double_aeabi_rp2040.S
@@ -44,11 +44,9 @@ PICO_RUNTIME_INIT_FUNC_RUNTIME(__aeabi_double_init, PICO_RUNTIME_INIT_AEABI_DOUB
 #endif
 .endm
 
-.section .text
-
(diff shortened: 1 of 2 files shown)


kilograham:
nice catch!

kilograham (approved):

PR #2775, opened 2025-12-29
Update the localtime_r declaration for LLVM libc

petrhosek:
This is to match upstream declaration and avoid conflicting declaration error. This declaration will be removed after the next Clang update since LLVM libc now implements localtime_r but we keep the declaration for now to allow soft transition for SDK users.


petrhosek:

kilograham (approved):

PR #2779, opened 2025-12-30
Allow 'host' platform to simulate either RP2040 or RP2350

### LINKED ISSUE #2778: Host platform has no support for RP2350-based projects
The `host` platform only really partially simulates an RP2040 environment so far, with no way to specify a particular board either, so attempting to build a host version of an existing code that uses an RP2350 or relies on values defined in a board file is either difficult or impossible.

Expanding the `host` platform to something like `host-rp2040` and `host-rp2350`, and allowing matching board files to be specified would make it relatively straightforward to define the base environment and pull in the chip-specific platform definitions.

rhulme:
The 'host' platform only really partially simulates an RP2040 environment, with no way to specify a particular board either, so attempting to build a host version of a existing code that uses an RP2350 or relies on values defined in a board file is either difficult or impossible.

Instead of setting PICO_PLATFORM to just 'host', it can now be set to 'host-rp2040' or 'host-rp2350' and PICO_BOARD can be set to any board configuration that is compatible with the given chip type. Platform definitions for the given chip will be fetched, meaning defines such as 'NUM_IRQS' will be set correctly.

Fixes #2778

Note that some changes are needed to the `picotool` repository when building for a genuine `rp2350` target as it does not initialise the pico-sdk completely and so will fail to find `platform_defs.h`.


rhulme:
No changes were made to the Bazel environment as this appears to be more complicated (and a brand new area for me) and I'd rather not invest the time without knowing if this approach is going to be acceptable.

PR #2780, opened 2025-12-31
Omit semicolons after _Pragma directives

petrhosek:
These are not statements and shouldn't be terminated by semicolons; in newer versions of Clang separating these by semicolons results in a build error.

We can also further simplify `remove_volatile_cast` by using the comma operator and avoiding the reliance on non-standard GNU extensions.


petrhosek:

kilograham (approved):

PR #2781, opened 2025-12-31
Use alternate macro mode to evaluate expressions

petrhosek:
Without the alternate macro mode, when building with Clang, the arguments to macros weren't evaluated.

### DIFF
--- a/src/rp2_common/pico_crt0/crt0_riscv.S
+++ b/src/rp2_common/pico_crt0/crt0_riscv.S
@@ -171,11 +171,13 @@ __soft_vector_table:
 .word isr_irq\n
 .endm
 
+.altmacro
 .set IRQN, 0
 .rept PICO_NUM_VTABLE_IRQS
-vtable_irq_n IRQN
+vtable_irq_n %IRQN
 .set IRQN, IRQN + 1
 .endr
+.noaltmacro
 
 // all default trap handlers do nothing, and we can check for them being set to our
 // default values by seeing if they point to somewhere between __defaults_isrs_start and __default_isrs_end
@@ -222,11 +224,13 @@ decl_isr_bkpt isr_riscv_machine_soft_irq
 .endif
 .endm
 
+.altmacro
 .set IRQN, 0
 .rept PICO_NUM_VTABLE_IRQS
-decl_isr_n IRQN
+decl_isr_n %IRQN
 .set IRQN, IRQN + 1
 .endr
+.noaltmacro
     // fall through
 
 // All unhandled USER IRQs fall through to here. Note there is no way to get


petrhosek:

kilograham:
yeah, i guess we haven't tested on RISC-V clang - so i should probably test the public one of those - can you give me a link to yours also

kilograham:
see also #2228

Wren6991:
@lenary is this an expected difference between GAS and the LLVM assembler?

Wren6991 (approved):
Looks sensible and minimal.

Wren6991:
Actually, the `.altmacro` seems to be required by GNU assembler too. Extracting just the relevant parts for repro:

```
#define PICO_NUM_VTABLE_IRQS 54
#define NUM_IRQS 54


.p2align 4
.global __soft_vector_table
__soft_vector_table:
.macro vtable_irq_n n
.word isr_irq\n
.endm

.set IRQN, 0
.rept PICO_NUM_VTABLE_IRQS
vtable_irq_n IRQN
.set IRQN, IRQN + 1
.endr

.macro decl_isr name
.weak \name
\name:
.endm

.macro decl_isr_n n
.if \n < PICO_NUM_VTABLE_IRQS
    decl_isr isr_irq\n
.elseif \n < NUM_IRQS
    // We declare a strong label, so user can't override, since there is no vtable entry
    .type isr_irq\n,%function
    .thumb_func
    isr_irq\n:
.endif
.endm

.set IRQN, 0
.rept PICO_NUM_VTABLE_IRQS
decl_isr_n IRQN
.set IRQN, IRQN + 1
.endr
    // fall through

// All unhandled USER IRQs fall through to here. Note there is no way to get
// the "current exception" on RISC-V (as there is no such thing -- the
// hardware does not model the exception lifecycle like on Arm) so instead we
// just pass the IRQ number into the IRQ handler in a0.
.global __unhandled_user_irq
__unhandled_user_irq:
.global unhandled_user_irq_num_in_a0
unhandled_user_irq_num_in_a0:
    // The dispatch code will have left IRQ << 2 in a0 from its meinext read,
    // so just need to shift it back down
    srli a0, a0, 2
    ebreak
```
Disassembling with relocations shows:

```
0000000e <__soft_vector_table>:
	...
			e: R_RISCV_32	isr_irqIRQN
			12: R_RISCV_32	isr_irqIRQN
			16: R_RISCV_32	isr_irqIRQN
			1a: R_RISCV_32	isr_irqIRQN
			1e: R_RISCV_32	isr_irqIRQN
			22: R_RISCV_32	isr_irqIRQN
			26: R_RISCV_32	isr_irqIRQN
			2a: R_RISCV_32	isr_irqIRQN
```

The difference between GNU and LLVM seems to be that the GNU assembler is permitting multiple weak definitions of the bogus symbol `isr_irqIRQN` within one translation unit. @kilograham this is definitely our bug.

PR #2785, opened 2026-01-05
Enable hard-float ABI support for pico_float and pico_double for m33

### LINKED ISSUE #1993: RP2350 FPU compiler flags
See https://github.com/earlephilhower/arduino-pico/issues/2535

According to [this](https://github.com/earlephilhower/arduino-pico/blob/edba2faa3424737b1b24d5583a330d1c54f0e7df/tools/makeboards.py#L311) it seems like it is using softmp mode for floating point operation for the RP2350.

SoftFP (soft floating point): Use hardware floating-point operations, but function calls pass parameters via general-purpose registers (compatible with non-FPU systems).

HardFP (hard floating point): Pass floating-point values in floating-point registers during function calls (which is faster but requires hardware FPU support).

Is there any reason for this and would hardfp make floating point performance better (I am using the FPU primarely for Audio DSP).

### LINKED ISSUE #2783: Enable hard-float ABI support for pico_float and pico_double for m33
`pico_float` and `pico_double` currently only support `softfp` ABI.

Adding hard-float ABI support would increase flexibility when targeting cortex-m33 and improve interoperability with existing toolchains and libraries.

andylinpersonal:
Enable support for `-mfloat-abi=hard` in `pico_float:vfp` and `pico_double:pico-dcp`.
Exclude `pico_float:dcp`, since this variant is intended for non-FPU configurations.

Toolchain:
* Tested with the [official](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads) `arm-gnu-toolchain-14.3.rel1`

Implementation:
* `__aeabi_*` runtime helper functions (`__aeabi_x2y`) have more direct path than custom functions (`xtoy`).
* Complex math functions are called through wrapper.

Test:
* `__aeabi_*` functions should be explicitly marked as `__attribute__((pcs("aapcs")))` when the hard abi is enabled.

Resolves #2783, resolves #1993 (maybe duplicated)

BTW, it seems like `src/rp2_common/pico_double/double_fma_dcp.S` is not covered by any test case yet.


kilograham:
subsumed by #2822

PR #2794, opened 2026-01-12
Handle the missing DCP state in pico_float_pico_dcp

### LINKED ISSUE #2793: Some functions in `pico_float_pico_dcp` do not check the coprocessor state
In the dcp variant of the floating-point library, some functions skipped the DCP wrapper.

Specifically:
https://github.com/raspberrypi/pico-sdk/blob/9c6713a8ed39d7719e353ec35c4124ae4ac05755/src/rp2_common/pico_float/float_aeabi_dcp.S#L187-L188
and
https://github.com/raspberrypi/pico-sdk/blob/9c6713a8ed39d7719e353ec35c4124ae4ac05755/src/rp2_common/pico_float/float_aeabi_dcp.S#L159

andylinpersonal:
Fixes #2793

### DIFF
--- a/src/rp2_common/pico_float/float_aeabi_dcp.S
+++ b/src/rp2_common/pico_float/float_aeabi_dcp.S
@@ -156,7 +156,7 @@ saving_func wrapper __aeabi_ui2f uint2float
   saving_func_return
 
 float_section float2fix_z
-regular_func float2fix_z
+saving_func regular float2fix_z
   ubfx r2, r0, #23, #8
   cbz r2, 2f // input is zero or denormal
   cmp r2, #0xff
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_float/float_aeabi_dcp.S
```diff
@@ -199,10 +198,10 @@ regular_func float2ufix_z
   b float2uint_z_entry
 2:
   movs r0, #0
-  bx lr
+  saving_func_return
```
lurch, line 201: I know absolutely nothing about what this code is doing, but should `float2fix` also be updated to use `saving_func_return` instead of `bx lr` ?

andylinpersonal: Just keep consistent with elsewhere. Indeed `saving_func_return` is also `bx lr` 😌.
https://github.com/raspberrypi/pico-sdk/blob/9c6713a8ed39d7719e353ec35c4124ae4ac05755/src/rp2_common/pico_float/float_aeabi_dcp.S#L59-L61

kilograham:
On looking at this, i think i'd rather just remove the `_entry` functions, leave the functions as `regular_` and just branch to the saving entry point

kilograham:
my suggestion now subsumed by #2849

PR #2821, opened 2026-02-08
Fix one output interleavement issue with pico_printf_compiler

kilograham:
Force `fflush()` of `stdout` after `printf` when using `pico_set_printf_implementation(compiler)` and `PICO_STDIO_SHORT_CIRCUIT_CLIB_FUNCS` as a semi-workaround for the c library buffering `stdout` during `printf`

Note: the correct solution would be to flush `stdout` in the short-circuited functions, but this rather destroys the point of making them fast.

This takes care of the common case. For complete/best support use `PICO_STDIO_SHORT_CIRCUIT_CLIB_FUNCS=0` when using `pico_printf_compiler`


will-v-pi (approved):
LGTM

PR #2822, opened 2026-02-08
Add support for hard float on M33 via CMake variable PICO_HARD_FLOAT_ABI=1

kilograham:
Note there is currently no hard float library in the `llvm_libc`


dpgeorge:
Really looking forward to this!  Unless there are serious performance regressions (I expect there will not be) we will enable it in MicroPython.  That's because our ABI for dynamically loadable native code only supports `-mfloat-abi=soft` or `-mfloat-abi=hard`, we don't support `softfp` (we could, but it's extra work and an extra combination of targets to support across all the tools).

test/pico_float_test/CMakeLists.txt
```diff
@@ -5,110 +5,133 @@ endif()
...
-    pico_add_extra_outputs(pico_double_test)
-    #pico_set_float_implementation(pico_double_test compiler)
-    #pico_set_double_implementation(pico_double_test compiler)
+    add_executable(custom_float_funcs_test_${FLOAT_TYPE} custom_float_funcs_test.c)
+    pico_set_float_implementation(custom_float_funcs_test_${FLOAT_TYPE} ${FLOAT_TYPE})
+    target_link_libraries(custom_float_funcs_test_${FLOAT_TYPE} PRIVATE pico_stdlib)
+    pico_add_extra_outputs(custom_float_funcs_test_${FLOAT_TYPE})
+    pico_set_printf_implementation(custom_float_funcs_test_${FLOAT_TYPE} compiler)
+    if (PICO_C_COMPILER_IS_CLANG)
+        # llvm by default with treat our out of range float->int conversions as undefined behavior
```
lurch:
```suggestion
        # llvm by default will treat our out of range float->int conversions as undefined behavior
```

test/pico_float_test/CMakeLists.txt
```diff
@@ -5,110 +5,133 @@ endif()
...
+        endif()
+
+        add_executable(custom_double_funcs_test_${DOUBLE_TYPE} custom_double_funcs_test.c)
+        pico_set_double_implementation(custom_double_funcs_test_${DOUBLE_TYPE} ${DOUBLE_TYPE})
+        target_link_libraries(custom_double_funcs_test_${DOUBLE_TYPE} PRIVATE pico_stdlib)
+        pico_add_extra_outputs(custom_double_funcs_test_${DOUBLE_TYPE})
+        pico_set_printf_implementation(custom_double_funcs_test_${DOUBLE_TYPE} compiler)
+
+        if (PICO_C_COMPILER_IS_CLANG)
+            # llvm by default with treat our out of range float->int conversions as undefined behavior
```
lurch:
```suggestion
            # llvm by default will treat our out of range float->int conversions as undefined behavior
```

src/rp2_common/pico_double/double_aeabi_dcp.S
Wren6991, line 41: There are some dummy `r0`s in other push/pops in this diff (e.g. in `regular_func double2float`)  to maintain stack alignment. The save hook here pushes `lr` only. Worth fixing? If not, worth removing `r0` elsewhere?

Wren6991: For correctness I think it's ok as nothing observes the alignment inside of the DCP funcs, and IRQs realign the stack. Just seemed inconsistent.

src/rp2_common/pico_double/double_conv_m33.S
```diff
@@ -94,13 +108,27 @@ regular_func fix642double
  bic r0,r0,#1 @ force to even
  b 5b
 
+#if defined(__ARM_PCS_VFP)
+regular_func uint642double
+ movs r2,#0
+regular_func ufix642double
+ push {lr}
+ bl ufix642double_entry
+ vmov d0,r0,r1
```
Wren6991, line 117: This seems weird; `vmov` into `d0` is injected after `__aeabi_ul2d`, which is already a double-returning function that should have its result in `d0`

src/rp2_common/pico_double/double_fma_dcp.S
```diff
@@ -604,6 +620,10 @@ regular_func mla
  dcp_dadd_m r0,r1,r0,r1,r2,r3
 // todo optimize this based on final decision on saving_func_entry
  pop {r4,r5,r14}
+#if defined(__ARM_PCS_VFP)
+ vmov d0,r0,r1
+ add sp,sp,#8
```
Wren6991: Mismatched push/pop here: if engaged flag is set you have `vpush {d2}; push {lr}; add sp, sp, #8; pop {pc}`, so return address is corrupted.

Wren6991: Fix is probably to use the `saving_func` macro from the main impl file, so it's impossible to add stack ops above the `lr` save.

src/rp2_common/pico_float/CMakeLists.txt
```diff
@@ -121,30 +124,33 @@
...
-                #NO_AEABI_ARITHMETIC
-                #NO_AEABI_CMP
-                #NO_AEABI_CONV_32
-                #NO_AEABI_CONV_64
-                #NO_AEABI_CONV_DOUBLE
-                #NO_SQRTF
-                #NO_SCI
-                #NO_SCI_EXTRA
-        )
+        # dcp library not supported with hard float
```
Wren6991, line 129: ~Is this true? E.g. `double_eabi_dcp.S` was updated to support VFP PCS~

(ignore; this is float not double)

src/rp2_common/pico_float/float_common_m33.S
```diff
@@ -23,14 +23,29 @@ float_section WRAPPER_FUNC_NAME(\func)
 
 float_wrapper_section conv_tof
 
+#if defined(__ARM_PCS_VFP)
+regular_func int642float
+ movs r2,#0       @ fall through
+regular_func fix642float
+ push {lr}
+ bl fix642float_softfp
+ vmov s0,r0
```
Wren6991, line 32: `vmov s0,r0` has been injected after a function synonymous with `__aeabi_l2f`, which is a float-returning function so should already have its result in `s0`.

Am I missing something in the Arm docs? Do AEABI funcs not follow VFP PCS?

Wren6991 (changes requested):
Fairly certain the mismatched push/pop is a bug. Otherwise looks ok, so LGTM once that is re-checked

src/rp2_common/pico_double/double_aeabi_dcp.S
kilograham: i am fixing to remove the other r0 pushes

src/rp2_common/pico_double/double_conv_m33.S
kilograham: the code is confusing here because of the `#ifdef`, but the ufix2double_entry in this case is not a proper API as it returns in r0, r1

src/rp2_common/pico_double/double_fma_dcp.S
kilograham: i have reworked

src/rp2_common/pico_float/float_common_m33.S
kilograham: this took me a while to figure out - i wish i'd taken in your question at the bottom - indeed they do not use the VFP PCS

lurch:
CI appears to be failing with:
```
CMake Error at src/CMakeLists.txt:72 (target_link_options):
  Cannot specify link options for target "pico_double_pico_dcp" which is not
  built by this project.
Call Stack (most recent call first):
  src/rp2_common/pico_double/CMakeLists.txt:60 (pico_wrap_function)
  src/rp2_common/pico_double/CMakeLists.txt:124 (wrap_double_functions)
```

Wren6991 (dismissed):
CMake change in fc579bc looks weird, and may be related to CI failure. Otherwise looks like everything I mentioned was addressed so LGTM

lurch:
Lots of CI failures in the Bazel builds now, e.g.
```
[1,024 / 1,094] Compiling test/hardware_sync_spin_lock_test/hardware_sync_spin_lock_test.c; 0s processwrapper-sandbox, disk-cache ... (3 actions, 2 running)
INFO: From Compiling test/pico_float_test/pico_float_test.c:
In file included from external/+_repo_rules+arm_gcc_linux-x86_64/bin/../lib/gcc/arm-none-eabi/13.2.1/../../../../arm-none-eabi/include/stdio.h:35,
                 from test/pico_float_test/pico_float_test.c:15:
test/pico_float_test/pico_float_test.c: In function 'main':
test/pico_float_test/pico_float_test.c:348:85: warning: implicit declaration of function '__real_fmaf'; did you mean '__real_fmodf'? [-Wimplicit-function-declaration]
  348 | #define check_close3(func,p0,p1,p2) ({ typeof(p0) r = func(p0,p1,p2), r2 = __CONCAT(__real_, func)(p0,p1,p2); if (isnanf(p0) || isnanf(p1) || isnanf(p2)) assert_nan(r); else assert_close(r, r2); r; })
      |                                                                                     ^~~~~~~
test/pico_float_test/pico_float_test.c:506:17: note: in expansion of macro 'check_close3'
  506 |                 check_close3(fmaf, a, b, c);
      |                 ^~~~~~~~~~~~
```

src/rp2_common/pico_float/float_common_m33.S
andylinpersonal: Relevant ABI specs for reference: [parameters to and results are passed in integer core registers](https://github.com/ARM-software/abi-aa/blob/0a230c6ac17fc889a57e43cb97d54fc44b708a61/rtabi32/rtabi32.rst?plain=1#L919)

PR #2838, opened 2026-02-20
Reset alarm pending flag when cancelling alarm.

### LINKED ISSUE #2836: Using threadsafe_background, removing an "at time" worker in a pending worker can stop future "at time" workers running
I've see this with BTSTACK not advancing its state machine.

In process_under_lock if we find that there are no timers scheduled we cancel the existing alarm but crucially don't reset the alarm_pending flag,
If another "at time" worker is requested the following check in process_under_lock succeeds, we think we have an alarm due so don't set another,

```
if (self->alarm_pending && absolute_time_diff_us(self->last_set_alarm_time, next_time) > 0) break;
```

The fix is quite simple.

```
        if (is_at_the_end_of_time(next_time)) {
            // cancel the alarm early (we will have been called soon after an alarm wakeup), so that
            // we don't risk alarm_id collision.
            if (self->alarm_id > 0) {
                alarm_pool_cancel_alarm(self->alarm_pool, self->alarm_id);
                self->alarm_id = 0;
                self->alarm_pending = false; // <-- FIX
            }
            break;
        }
```

peterharperuk:
Fixes #2836


test/pico_time_test/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ if (NOT PICO_TIME_NO_ALARM_SUPPORT)
     target_compile_definitions(pico_time_test PRIVATE
             PICO_TIME_DEFAULT_ALARM_POOL_MAX_TIMERS=250
     )
-    target_link_libraries(pico_time_test PRIVATE pico_test)
+    target_link_libraries(pico_time_test PRIVATE pico_test pico_async_context_threadsafe_background)
```
kilograham: it seems were shoe-horning a bunch of unrelated stuff into pico_time_test - perhaps should be an async_context_test; in either case, I'm pretty sure this test compiles for host (and if not host wtih SDL) so use of this library (or the new test if you split it) should be guarded in the CMake withn if (TARGET pico_async_context_threadsafe_background) and with an appropriate #if LIB_PICO_ in the code

kilograham (changes requested):

test/pico_time_test/CMakeLists.txt
peterharperuk: yes, you're right, this shouldn't be in here. I will move.

kilograham (approved):

kilograham:
not sure why this was sitting with "changes requested" when you had made the changes!

PR #2840, opened 2026-02-24
Fixes #2839 Picolibc compatibility issues

### LINKED ISSUE #2839: pico-sdk doesn't work when compiling with picolibc
There are several compatibility issues when trying to build with picolibc version 1.8.11:

 * Using GCC adds --specs=nosys.specs, a newlib-specific file
 * pico-sdk assumes the C library will define __printflike, picolibc defines __picolibc_format instead.
 * GCC's -Wredundant-decls complains about the symbol aliasing used to define stdout and stderr.
 * Picolibc will add another linker script unless the application uses the -T format instead of -Wl,--script.
 * Picolibc uses different constructor symbols (`__bothinit_array_start` and `__bothinit_array_end`).
 * Picolibc needs thread-local variables to be per-core instead of global.
 * When using picolibc's printf, we want to inherit the project printf feature selection

keith-packard:
As a part of the Debian migration from newlib to picolibc for the arm-none-eabi GCC toolchain, a number of picolibc portability issues were discovered in pico-sdk. Here's a set of fixes that allows use of pico-sdk with picolibc 1.8.11.

 * Only use --specs=nosys.specs with newlib.
 * Provide a definition of `__printflike` in terms of `__picolibc_format`
 * Ignore -Wredundant-decls messages when building picolibc_interface.c. This makes sure the 'kitchen-sink' example builds with its long list of warnings.
 * Switch from `-Wl,--script=` to `-T` when building a linker command line; this avoids referencing picolibc's default linker script. The alternative here would be to add `-T/dev/null` when using picolibc; other projects have done that to allow continued use of `LINKER:--script` for compatibility with toolchains which don't support `-T` (I'm not sure which ones those might be).
 * Add definitions for `__bothinit_array_start` and `__bothinit_array_end` to all of the linker scripts to allow picolibc's `__libc_init_array` function to correctly call constructors.
 * Add picolibc-specific printf format selection arguments to the linker command line when using pico_printf_compiler. This allows applications to use pico-sdk style options instead of needing to use target_linker_options to add the picolibc-specific versions.

A separate PR which stacks an implementation of per-core thread local variables is available as #2846


src/rp2040/boot_stage2/CMakeLists.txt
```diff
@@ -49,7 +49,9 @@ function(pico_define_boot_stage2 NAME SOURCES)
     if (PICO_C_COMPILER_IS_CLANG)
         target_link_options(${NAME} PRIVATE "-nostdlib")
     elseif (PICO_C_COMPILER_IS_GNU)
-        target_link_options(${NAME} PRIVATE "--specs=nosys.specs")
+        if (PICO_CLIB STREQUAL "newlib")
+	   target_link_options(${NAME} PRIVATE "--specs=nosys.specs")
+	endif()
```
lurch: Indentation looks a bit weird here, which might be because you've mixed tabs and spaces. Could you stick to using just spaces please? (and also in some of the other files you've edited here)

src/rp2_common/pico_clib_interface/picolibc_interface.c
```diff
@@ -129,26 +136,45 @@ void runtime_init(void) {
     __libc_init_array();
 }
 
-#if !PICO_RUNTIME_NO_INIT_PER_CORE_TLS_SETUP
```
lurch, line 132: The complete removal of these `#if` directives seems a bit odd?

keith-packard: I wasn't sure how these symbols could ever be defined so that I could test to make sure the code operated when set.

src/rp2040/boot_stage2/CMakeLists.txt
keith-packard: I tried (and appear to have failed). Working on a combination of projects, some of which require tabs and some (sensibly) don't allow them means my editor configuration is complicated...

keith-packard: I think I fixed all of the whitespace problems (and have set my editor configuration to avoid tabs in future)

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -191,4 +191,23 @@ if (NOT TARGET pico_standard_link)
...
         endif()
     endif()
+
+    # Convert pico-sdk printf variant selection into picolibc version
+    # when using pico_printf_compiler
+
+    if (PICO_CLIB STREQUAL "picolibc")
+        if (RUNTIME_INCLUDE_PRINTF_MINIMAL)
+          target_link_options(pico_standard_link INTERFACE
+            $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=m>)
```
lurch: One more small nitpick: most of our CMake files use a 4-space indent, but you seem to be using a 2-space indent here?

src/rp2040/boot_stage2/CMakeLists.txt
lurch: Thank you!

src/rp2_common/pico_clib_interface/picolibc_interface.c
lurch: I dunno if it helps answer your question, but see sections 5.5.19. and 6.1. of https://pip.raspberrypi.com/documents/RP-009085-KB
If that doesn't help (I'm afraid the code here is beyond my level of knowledge), then you'll have to wait until somebody else has time to look at this.

(The reason I found the removal of these directives surprising, is because I'm sure they must have been added for a good reason :wink: )

src/rp2_common/pico_standard_link/CMakeLists.txt
keith-packard: Yeah, that's what Zephyr uses... All fixed.

src/rp2_common/pico_clib_interface/picolibc_interface.c
keith-packard: Pretty sure they were added because there wasn't any per-core TLS support in pico-sdk before I copied bits over from picolibc's implementation (which is only available when you build picolibc from source; I haven't figured out a good way to expose it without that).

lurch: Ahh, a `git grep` reveals that these are indeed only used by the picolibc's bindings:
```console
$ git grep PER_CORE_TLS_SETUP
src/rp2_common/pico_clib_interface/picolibc_interface.c:#if !PICO_RUNTIME_NO_INIT_PER_CORE_TLS_SETUP
src/rp2_common/pico_clib_interface/picolibc_interface.c:#if !PICO_RUNTIME_SKIP_INIT_PER_CORE_TLS_SETUP
src/rp2_common/pico_clib_interface/picolibc_interface.c:PICO_RUNTIME_INIT_FUNC_PER_CORE(runtime_init_pre_core_tls_setup, PICO_RUNTIME_INIT_PER_CORE_TLS_SETUP);
src/rp2_common/pico_runtime_init/include/pico/runtime_init.h:// PICO_RUNTIME_SKIP_INIT_PER_CORE_TLS_SETUP defaults to 0
src/rp2_common/pico_runtime_init/include/pico/runtime_init.h:#ifndef PICO_RUNTIME_INIT_PER_CORE_TLS_SETUP
src/rp2_common/pico_runtime_init/include/pico/runtime_init.h:#define PICO_RUNTIME_INIT_PER_CORE_TLS_SETUP                 "10060"
```

I was mistakenly assuming that this was some kind of "standard feature" that all C-lib bindings were expected to support :facepalm:

src/rp2_common/pico_crt0/rp2040/memmap_default.ld
```diff
@@ -129,6 +133,30 @@ SECTIONS
...
+        *(.tdata .tdata.* .gnu.linkonce.td.*)
+        PROVIDE( __tdata_end = . );
+    } >FLASH
+
+    .tbss (NOLOAD) : {
+        *(.tbss .tbss.* .gnu.linkonce.tb.*)
+        *(.tcommon)
+        PROVIDE( __tls_end = . );
+        PROVIDE( __tbss_end = . );
+    } >FLASH
```
will-v-pi: What is the reason behind moving these sections to FLASH rather than keeping them in RAM?

kilograham:
where can we get/build a GCC compiler that uses picolibc to test with?

kilograham:
@keith-packard can we split the tls stuff into a separate PR - it is going to clash with some other PRs.

keith-packard:
> where can we get/build a GCC compiler that uses picolibc to test with?

You can either use the current debian unstable toolchain for arm-none-eabi for which picolibc is the only supported C library, or you can use the debian stable toolchain and add the picolibc package. For the latter, you must build and link with --specs=picolibc.specs. In either case, `sudo apt install picolibc-arm-none-eabi` should do the trick on any debian or ubuntu system.

For other systems, you can use the Zephyr SDK toolchain which is available for Linux, Mac OS X and Windows.

You can also use the ARM GNU Toolchain version 15.2.rel1 and add the pre-built picolibc add-ons available here: https://github.com/picolibc/picolibc/releases/download/1.8.11/picolibc-arm-none-eabi-1.8.11-15.2.rel1.zip However, this toolchain is flawed in disabling TLS usage entirely, so I recommend against it. The current debian toolchain uses the same sources with slightly different configuration to enable TLS and use picolibc by default.

keith-packard:
> @keith-packard can we split the tls stuff into a separate PR - it is going to clash with some other PRs.

Can do. I've posted a separate PR which stacks the TLS changes atop the rest of these so they don't get lost. #2846

src/rp2_common/pico_crt0/rp2040/memmap_default.ld
keith-packard: .tdata needs to be in FLASH so that it's initial contents are available to use in initializing TLS areas, .tbss needs to be sequential in addressing with .tdata so that the thread local offsets computed by the linker are correct. .tbss should not actually allocate any space in FLASH; the addresses generated are used only to compute TLS offsets during relocation. RAM for each of the per-core TLS areas is allocated below.

The previous kludge avoided the need to even initialize the TLS area by wedging it between .data and .bss and using the existing initializers of those to also initialize the TLS area. That relied upon careful layout of the TLS area to account for padding and alignment constraints, which are not entirely correct in the existing scripts.

The new way introduced here is far more traditional, with the exception of the static allocation of two TLS blocks in RAM.

edo9300:
for what it's worth, i could successfully use this branch to compile using the picolibc based toolchain https://wonderful.asie.pl/, (only things i had to do were explicitly set the libc to use in the cmake and toolchain path)

kilograham:
closed in favor of #2898 as there was a big merge conflict with recent linker script changes

Note, i did not port 

```

    if (PICO_CLIB STREQUAL "picolibc")
        if (RUNTIME_INCLUDE_PRINTF_MINIMAL)
            target_link_options(pico_standard_link INTERFACE
              $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=m>)
        elseif (RUNTIME_INCLUDE_PRINTF_FLOAT)
            # this is the default
        elseif (RUNTIME_INCLUDE_PRINTF_LONG_LONG)
            target_link_options(pico_standard_link INTERFACE
              $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=l>)
        else()
            target_link_options(pico_standard_link INTERFACE
              $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=i>)
        endif()
    endif()
```

because `RUNTIME_` CMake vars are only used in `pico_minimize_runtime` which isn't called by the regular build... you could make a PR to add these lines there, but i'm not sure what your intention was

keith-packard:
> Note, i did not port
> 
> ```
> 
>     if (PICO_CLIB STREQUAL "picolibc")
>         if (RUNTIME_INCLUDE_PRINTF_MINIMAL)
>             target_link_options(pico_standard_link INTERFACE
>               $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=m>)
>         elseif (RUNTIME_INCLUDE_PRINTF_FLOAT)
>             # this is the default
>         elseif (RUNTIME_INCLUDE_PRINTF_LONG_LONG)
>             target_link_options(pico_standard_link INTERFACE
>               $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=l>)
>         else()
>             target_link_options(pico_standard_link INTERFACE
>               $<$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_PRINTF_IMPL>,pico_printf_compiler>:--printf=i>)
>         endif()
>     endif()
> ```
> 
> because `RUNTIME_` CMake vars are only used in `pico_minimize_runtime` which isn't called by the regular build... you could make a PR to add these lines there, but i'm not sure what your intention was

These lines select a suitable version of printf from the library to provide only the necessary code, minimizing the size of the resulting application. Given that these properties exist, presumably you want this optimization to be enabled somewhere?

PR #2845, opened 2026-03-02
Fix description rp2350 svd

hpcraink:
The description of RISC-V's MTIME register is a duplicate of the High register MTIMEH.
This seems to be the low half of the machine-mode time.
The fallout is that comments in several autogenerated header files is affected, too -- hence submitted in a separate commit.


lurch:
The files you've edited here are all auto-generated from internal-metadata using internal-tooling; so if we merged this PR, your changes would just get overwritten the next time we updated these files. Instead, we need to update our internal-metadata so that our "upstream" files include your "downstream" changes.

kilograham:
converting to draft as I can't convert to an issue

kilograham:
@matiasilva can you make sure this is fixed upstream

kilograham:
subsumed by #2905

PR #2846, opened 2026-03-02
Provide per-core thread local variables when using picolibc

keith-packard:
This PR contains a single patch atop #2840 to allocate per-core thread local storage which is necessary for safe thread local variable usage (like errno).

Change the linker scripts to statically allocate thread local storage space for each core and to initialize that using picolibc's _init_tls instead wedging it between .data and .bss.
    
Implement pico-specific __aeabi_read_tp and_set_tls functions which use the core ID register at 0xd0000000 to index an array of TLS base pointers.
    
Perform per-core TLS region setup, calling _init_tls and _set_tls using the areas allocated by the linker script.


src/rp2_common/pico_crt0/rp2040/memmap_blocked_ram.ld
```diff
@@ -181,37 +205,34 @@ SECTIONS
...
+       PROVIDE(__tls0_base = .);
+       . = . + __tls_size_align;
+    } > RAM
+
+    .tls1 (NOLOAD) : {
+        . = ALIGN(__tls_align);
+        PROVIDE(__tls1_base = .);
+        . = . + __tls_size_align;
+    } > RAM
+
```
will-v-pi, line 235: Would it be possible to put these sections between `data` and `bss`, so that `bss` and `heap` are contiguous? Otherwise, this may break compatibility with some other changes that are being worked on.

keith-packard: Sure, but that might break code which assumes that data and bss are contiguous?

kilograham:
thank you; subsumed by #2910

PR #2863, opened 2026-03-11
Document Clock Divider = 1 in pio_get_default_sm_config  in pio.h

### LINKED ISSUE #2862: pio_get_default_sm_config documentation does not specify clkdiv
The documentation for pio_get_default_sm_config shows the default values for everything except it does not mention that default clkdiv defaults to 1.

peternlewis:
Add Clock Divider = 1 to documentation of pio_get_default_sm_config

Fixes #2862

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -741,6 +741,7 @@ static inline void sm_config_set_mov_status(pio_sm_config *c, enum pio_mov_statu
  *
  * Setting | Default
  * --------|--------
+ * Clock Divider | 1
  * Out Pins | 32 starting at 0
  * Set Pins | 0 starting at 0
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -741,6 +741,7 @@ static inline void sm_config_set_mov_status(pio_sm_config *c, enum pio_mov_statu
  *
  * Setting | Default
  * --------|--------
+ * Clock Divider | 1
```
lurch, line 744: The Clock Divider is fractional, so perhaps "1.0" would be clearer?

kilograham: I think "1" is fine - 1.0 is nominally a double not a float!

kilograham (approved):

PR #2866, opened 2026-03-13
Move CPP header guard in status_led

### LINKED ISSUE #2642: pico/status_led.h might need extern "C"
If I include pico/status_led.h from a cpp file, I end up with undefined symbols at link time. If I wrap the include in extern "C", it works.

The undefined symbols are cyw43_gpio_set and similar; I can see that there are no extern Cs by the time the includes get to cyw43_gpio_set in cyw43.h. I can bring up status_blink.cpp.obj file in an editor and it looks like it wants a mangled name (_Z14cyw43_gpio_setP8_cyw43_ti) but I'm guessing there.

To reproduce:
This is for pico2-w.
Create a new project from example status_blink.
Build it and copy the uf2 to the pico2-w and it works.
Rename status_blink.c to status_blink.cpp and change CMakeLists.txt accordingly.
Clean (delete build dir) and rebuild, and the link fails.
Wrap the #include "pico/status_led.h" in extern "C" and the build works, and the uf2 works.

The link error:
```
[build] C:/Users/johnf/.pico-sdk/toolchain/14_2_Rel1/bin/../lib/gcc/arm-none-eabi/14.2.1/../../../../arm-none-eabi/bin/ld.exe: CMakeFiles/status_blink.dir/status_blink.cpp.obj: in function `status_led_get_state':
[build] C:/Users/johnf/.pico-sdk/sdk/2.2.0/src/rp2_common/pico_status_led/include/pico/status_led.h:234:(.text.status_led_get_state+0x10): undefined reference to `cyw43_gpio_get(_cyw43_t*, int, bool*)'

```
I don't know where the extern "C" should go, or if it might cause other problems by putting it around the include of cyw43.h in status_led.h.

peterharperuk:
Need to handle missing header guard in cyw43.h

Fixes #2642

### DIFF
--- a/src/rp2_common/pico_status_led/include/pico/status_led.h
+++ b/src/rp2_common/pico_status_led/include/pico/status_led.h
@@ -19,15 +19,16 @@
 
 #include "hardware/gpio.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #if defined(CYW43_WL_GPIO_LED_PIN)
 #include "cyw43.h"
 #endif
 
 struct async_context;
 
-#ifdef __cplusplus
-extern "C" {
-#endif
 
 // PICO_CONFIG: PICO_STATUS_LED_AVAILABLE, Indicate whether a single-color status LED is available, type=bool, default=1 if PICO_DEFAULT_LED_PIN or CYW43_WL_GPIO_LED_PIN is defined; may be set by the user to 0 to not use either even if they are available, group=pico_status_led
 #ifndef PICO_STATUS_LED_AVAILABLE


src/rp2_common/pico_status_led/include/pico/status_led.h
```diff
@@ -17,6 +17,10 @@
 #ifndef _PICO_STATUS_LED_H
 #define _PICO_STATUS_LED_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
kilograham, line 25: as a rule/style we put this after other includes ... cyw43.h being a special case which really should be fixed there, so please move to after the hardware/gpio.h

kilograham (changes requested):

kilograham (approved):

Issue #2861, opened 2026-03-11
Custom panic handler is called with improper stack alignment on ARM


arthomnix:
When a custom panic handler is set using `PICO_PANIC_FUNCTION`, the stack pointer is not aligned to 8 bytes when the handler is called (as it should be according to the ARM calling convention). 

The definition of `panic` used when `PICO_PANIC_FUNCTION` is defined contains a single `push {lr}` instruction, which causes the stack pointer to become misaligned when the custom handler is called:

https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/pico_platform_panic/panic.c#L45-L47

This can result in strange, difficult-to-debug behaviour in the panic handler, as the C compiler assumes that functions will be called according to the calling convention with proper stack alignment. For example, this program

```c
#include <stdio.h>
#include "pico/stdlib.h"

void handle_panic(void) {
    printf("%f\n", 1.234567f);
    while (1);
}

int main(void) {
    stdio_init_all();
    sleep_ms(5000);
    panic("test");
}
```

(with `add_compile_definitions(PICO_PANIC_FUNCTION=handle_panic)` in CMakeLists.txt) prints `2.681562e+154` on the Pico W with the current SDK.

kilograham:
merged into develop (a while back)

PR #2867, opened 2026-03-13
Fix PICO_PANIC_FUNCTION= implementations

kilograham:
- stack should remain properly aligned
- risc-v was actually completely broken for user defined handlers because GCC was auto-pushing vaargs for naked function even with no stack frame
- new implementation supports full call stacks in gdb for Arm or RISC-V when hitting breakpoint in panic function
- new config defines make it explicit that gdb full stack trace comes at the cost of corrupting stack based vaargs - 4th is corrupted on Arm, later on RISC-V (7 maybe?)


kilograham:
fixes #2861

src/rp2_common/pico_platform_panic/include/pico/platform/panic.h
```diff
@@ -11,8 +11,24 @@
 extern "C" {
 #endif
 
-#ifndef __ASSEMBLER__
+// PICO_CONFIG: PICO_PANIC_FUNCTION, Name of a function to use in place of the stock panic function or empty string to simply breakpoint on panic, group=pico_runtime
+
+// PICO_CONFIG: PICO_PANIC_FUNCTION_WITH_CALL_STACK, Ensure the call stack is available when using a custom panic function via PICO_PANIC_FUNCTION=. When set to 1 it conflicts with PICO_PANIC_FUNCTION_WITH_ALL_VAARGS=1 as a stack fram is pushed onto the stack corrupting later vaargs. Note this defaults to 1 as most custom panic functions don't actually use the arguments and the call stack seems more useful in general, default=1, group=pico_runtime
```
lurch:
```suggestion
// PICO_CONFIG: PICO_PANIC_FUNCTION_WITH_CALL_STACK, Ensure the call stack is available when using a custom panic function via PICO_PANIC_FUNCTION=. When set to 1 it conflicts with PICO_PANIC_FUNCTION_WITH_ALL_VAARGS=1 as a stack frame is pushed onto the stack corrupting later vaargs. Note this defaults to 1 as most custom panic functions don't actually use the arguments and the call stack seems more useful in general, default=1, group=pico_runtime
```

src/rp2_common/pico_platform_panic/include/pico/platform/panic.h
```diff
@@ -11,8 +11,24 @@
...
+#ifndef PICO_PANIC_FUNCTION_WITH_CALL_STACK
+#define PICO_PANIC_FUNCTION_WITH_CALL_STACK 1
+#endif
+
+// PICO_CONFIG: PICO_PANIC_FUNCTION_WITH_ALL_VAARGS, Ensures that all vaargs are faithfully passed to the custom panic function via PICO_PANIC_FUNCTION=. When set to 1 it conflicts with PICO_PANIC_FUNCTION_WITH_CALL_STACK=1 because both the later vaargs and the stack frame would be expected at the top of the stack. Note this defaults to 0 as most custom panic functions don't actually use the arguments and the call stack seems more useful in general, default=0, group=pico_runtime
+#ifndef PICO_PANIC_FUNCTION_WITH_ALL_VAARGS
+#define PICO_PANIC_FUNCTION_WITH_ALL_VAARGS 0
+#endif
+
+// PICO_CONFIG: PICO_PANIC_FUNCTION_DOES_NOT_RETURN, Indicate that the user supplied PICO_PANIC_FUNCTION= does not return so the caller does not need to inject a breakpoint. When this is 1 both full call stacks are available and all vaargs are correctly passed to the user function, default=-0 group=pico_runtime
```
lurch: As written, this implies that there are two call stacks? Was it supposed to be something like
```suggestion
// PICO_CONFIG: PICO_PANIC_FUNCTION_DOES_NOT_RETURN, Indicate that the user supplied PICO_PANIC_FUNCTION= does not return so the caller does not need to inject a breakpoint. When this is 1 the full call stack is available and all vaargs are correctly passed to the user function, default=-0 group=pico_runtime
```
?

src/rp2_common/pico_platform_panic/include/pico/platform/panic.h
```diff
@@ -11,8 +11,24 @@
...
+#ifndef PICO_PANIC_FUNCTION_WITH_ALL_VAARGS
+#define PICO_PANIC_FUNCTION_WITH_ALL_VAARGS 0
```
lurch: Oops?
```suggestion
#ifndef PICO_PANIC_FUNCTION_DOES_NOT_RETURN
#define PICO_PANIC_FUNCTION_DOES_NOT_RETURN 0
```

src/rp2_common/pico_platform_panic/include/pico/platform/panic.h
lurch: Also, the `default=-0 group=pico_runtime` needs to be `default=0, group=pico_runtime`

lurch:
`tools/extract_configs.py` is complaining because it doesn't allow `=` to be used inside of a description, so I'm afraid you'll need to tweak the descriptions slightly.

src/rp2_common/pico_platform_panic/include/pico/platform/panic.h
```diff
@@ -1,44 +1,60 @@
...
-#endif
\ No newline at end of file
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _PICO_PLATFORM_PANIC_H
+#define _PICO_PLATFORM_PANIC_H
```
lurch, line 8: I've got **no** idea why GitHub is highlighting the _entirety_ of this file as "different" - if I check out the `develop` and `panic_function` branches locally, then the first difference is actually on line 14 :confused: 
Probably just a bug in the GitHub web interface?

kilograham: yeah it seems to be being dumb

PR #2875, opened 2026-03-20
Add justifications for missing `PICO_BAZEL_CONFIG` and `PICO_CMAKE_CONFIG` entries

### LINKED ISSUE #2868: Add any missing PICO_CONFIG to bazel?
Various new build flags have been added in other PRs but are not yet in bazel

will-v-pi:
Add justifications for why `PICO_BAZEL_CONFIG` and `PICO_CMAKE_CONFIG` entries differ - mostly they are configs that only make sense in one or the other, but some added to TODOs to implement in Bazel in the future

This means compare_build_systems.py doesn't raise any more warnings

Fixes #2868


tools/compare_build_systems.py
```diff
@@ -81,22 +91,44 @@
...
+    "PICO_TINYUSB_PATH",
+    "PICO_BTSTACK_PATH",
+    "PICO_CYW43_DRIVER_PATH",
+    "PICO_MBEDTLS_PATH",
+    "PICO_LWIP_PATH",
+    # These are legacy CMake options that should not be used.
+    "PICO_NO_FLASH",
+    "PICO_COPY_TO_RAM",
+    # These aren't supported as build flags in CMake either, but they appear
+    # as comments which this script can't tell.
```
lurch, line 130: Alternatively, it looks like this simple tweak allows this script to ignore "commented out" `PICO_CMAKE_CONFIG` options?
```patch
--- a/tools/compare_build_systems.py
+++ b/tools/compare_build_systems.py
@@ -187,6 +187,9 @@ def FindKnownOptions(option_pattern_matcher, file_paths):
     for p in file_paths:
         with open(p, "r") as f:
             for line in f:
+                if re.match("\s*#\s*#", line):
+                    continue
+
                 match = re.search(pattern, line)
                 if not match:
                     continue
```

tools/compare_build_systems.py
```diff
@@ -81,22 +91,44 @@
...
+    # Bazel doesn't provide a way of overriding the defaults for this.
+    # TODO: Provide a way to customise these toolchain flags.
+    "PICO_HARD_FLOAT_ABI",
+    "PICO_NO_CMSE",
     # Bazel supports this differently.
     # TODO: Provide a helper rule for explicitly generating a UF2 so users don't
     # have to write out a bespoke run_binary.
     "PICO_NO_UF2",
+    "PICO_NO_COPRO_DIS",
+    "PICO_ALLOW_EXAMPLE_KEYS",
```
lurch, line 113: Should there be an explicit TODO: to add a similar option to Bazel?

tools/compare_build_systems.py
```diff
@@ -81,22 +91,44 @@
...
     # Bazel uses PICO_CONFIG_HEADER.
     "PICO_BOARD_CMAKE_DIRS",
     "PICO_BOARD_HEADER_FILE",
     "PICO_BOARD_HEADER_DIRS",
+    # Bazel handles default boards differently.
+    "PICO_DEFAULT_BOARD_rp2040",
+    "PICO_DEFAULT_BOARD_rp2350",
+    "PICO_DEFAULT_BOARD_host",
+    # Bazel doesn't provide a way of overriding the defaults for this.
+    # TODO: Provide a way to customise these toolchain flags.
```
armandomontanez, line 105: These should be pretty easy. I'll take a look.

will-v-pi: These have both been added now - hard float in another PR, and cmse here

tools/compare_build_systems.py
will-v-pi: Thanks, added

tools/compare_build_systems.py
will-v-pi: I think it's fine - these all come under the same category as `PICO_NO_UF2`, so fixing that will make fixing these possible too

kilograham (approved):

PR #2879, opened 2026-03-25
Update Clang toolchain used by Bazel

### LINKED ISSUE #2859: Update clang toolchain revision in Bazel builds
Clang toolchain revision used are outdated and more than a year old. Update to a more recent revision available.

Prabhuk:
The Clang toolchain is updated to a newer revision  9431920bfaee50f4b1552fc01ea081a5c18a5a3b from March 20, 2026.
    
Fixes #2859

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -69,34 +69,34 @@ http_archive(
 http_archive(
     name = "clang_linux-x86_64",
     build_file = "//bazel/toolchain:clang.BUILD",
-    sha256 = "82302f8f0d9cb1062e60756147403c1525e965e1d7b777fab8076c74c7a5a19b",
+    sha256 = "7a4f040d87461825a507eaeaa4d51e3dcc685a238df5c4daa1cd3a6ad7c853cd",
     type = "zip",
-    url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/linux-amd64/+/git_revision:910be4ff90d7d07bd4518ea03b85c0974672bf9c",
+    url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/linux-amd64/+/git_revision:9431920bfaee50f4b1552fc01ea081a5c18a5a3b",
 )
 
 http_archive(
     name = "clang_win-x86_64",
     build_file = "//bazel/toolchain:clang.BUILD",
-    sha256 = "2e9b8ac889838754e5305b6fd73c7bba7a6ec7364f1ce8ac60268b6d3bc61e6c",
+    sha256 = "e333c66cbdc1a981aef0be34d7b582db4636e72ae89ebd9809da4e91fa2c9fe0",
     type = "zip",
     # Windows doesn't like `:` in the produced filename, so replace it with `%3A`.
-    url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/windows-amd64/+/git_revision:910be4ff90d7d07bd4518ea03b85c0974672bf9c".replace("git_revision:", "git_revision%3A"),
+    url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/windows-amd64/+/git_revision:9431920bfaee50f4b1552fc01ea081a5c18a5a3b".replace("git_revision:", "git_revision%3A"),
 )
 
 http_archive(
     name = "clang_mac-x86_64",
     build_file = "//bazel/toolchain:clang.BUILD",
-    sha256 = "d3f2ef6f391ef66141092cfdf07facd18d2587a25616e1251e6e6b13b05ab3df",
+    sha256 = "860920075ab45e12a36d3da0f871a28ddca71ae95b02bc1fac8f8a23a6c831c7",
     type = "zip",
-    url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/mac-amd64/+/git_revision:910be4ff90d7d07bd4518ea03b85c0974672bf9c",
+    url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/mac-amd64/+/git_revision:9431920bfaee50f4b1552fc01ea081a5c18a5a3b",
 )
 
 http_archive(
     name = "clang_mac-aarch64",
     build_file = "//bazel/toolchain:clang.BUILD",
-    sha256 = "61109b8464e9213ef8b9bfe55ce56298b94d4c66eaea308cf2b6556b0b85429e",
+    sha256 = "717d41447ff1a947dd376feddb9798155b575f7e642a5cd4582d0f3dad9067d4",
     type = "zip",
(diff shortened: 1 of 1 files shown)


Prabhuk:
CC: @petrhosek @kilograham

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
```diff
@@ -8,7 +10,8 @@ if (PICO_HARD_FLOAT_ABI)
     set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=hard")
     # todo - doesn't seem to be a hard_fp variant for google atm?
     # these are all the directories under LLVM embedded toolchain for ARM (newlib or picolibc)
-    set(PICO_CLANG_RUNTIMES armv8m.main_hard_fp armv8m.main_hard_fp_unaligned armv8m.main_hard_fp_unaligned_size)
+    set(PICO_CLANG_RUNTIMES armv8m.main_hard_fp armv8m.main_hard_fp_unaligned armv8m.main_hard_fp_unaligned_size armv8m.main-unknown-none-eabi)
+    message("ENABLING HARD FLOAT RUNTIME")
```
lurch: Was this just a "testing print", which now ought to be removed?

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
```diff
@@ -1,3 +1,5 @@
+set(PICO_HARD_FLOAT_ABI 1 CACHE BOOL "Use hard floating point ABI")
```
will-v-pi: This changes the default `PICO_HARD_FLOAT_ABI` for all Clang users, which I don't think is necessary for this PR - all the other Clang toolchains still support both hard and soft float ABIs, it's just the new Google toolchain that only supports hard float.

A better fix might be to remove `armv8m.main-unknown-none-eabi` from the softfp `PICO_CLANG_RUNTIMES` list, which should then throw the `Could not find an llvm runtime` error if `PICO_HARD_FLOAT_ABI=0` (assuming support for the older soft float toolchain is not required).

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
Prabhuk: yes. I'll drop the message.

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
Prabhuk: Done. PTAL

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
Prabhuk: Done. PTAL

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
```diff
@@ -8,11 +8,11 @@ if (PICO_HARD_FLOAT_ABI)
     set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=hard")
     # todo - doesn't seem to be a hard_fp variant for google atm?
     # these are all the directories under LLVM embedded toolchain for ARM (newlib or picolibc)
-    set(PICO_CLANG_RUNTIMES armv8m.main_hard_fp armv8m.main_hard_fp_unaligned armv8m.main_hard_fp_unaligned_size)
+    set(PICO_CLANG_RUNTIMES armv8m.main_hard_fp armv8m.main_hard_fp_unaligned armv8m.main_hard_fp_unaligned_size armv8m.main-unknown-none-eabi)
 else()
     set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=softfp")
     # these are all the directories under LLVM embedded toolchain for ARM (newlib or picolibc) and under llvm_libc
-    set(PICO_CLANG_RUNTIMES armv8m.main_soft_nofp armv8m.main_soft_nofp_unaligned armv8m.main_soft_nofp_unaligned_size armv8m.main-unknown-none-eabi)
+    set(PICO_CLANG_RUNTIMES armv8m.main_soft_nofp armv8m.main_soft_nofp_unaligned armv8m.main_soft_nofp_unaligned_size)
```
kilograham:
```suggestion
    set(PICO_CLANG_RUNTIMES armv8m.main_soft_nofp armv8m.main_soft_nofp_unaligned armv8m.main_soft_nofp_unaligned_size  armv8m.main-unknown-none-eabi)
```

kilograham: I'm going to revert this line, so you can build both hard-fp or soft-fp with google llvm (which means any compiler version is suppoed as long as you pass the right hard/soft float)

This is the best option i can see given the issues with google-llvm not indicating in any way that it has changed ABI

cmake/preload/toolchains/pico_arm_cortex_m33_clang.cmake
```diff
@@ -8,11 +8,11 @@ if (PICO_HARD_FLOAT_ABI)
     set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=hard")
     # todo - doesn't seem to be a hard_fp variant for google atm?
     # these are all the directories under LLVM embedded toolchain for ARM (newlib or picolibc)
-    set(PICO_CLANG_RUNTIMES armv8m.main_hard_fp armv8m.main_hard_fp_unaligned armv8m.main_hard_fp_unaligned_size)
+    set(PICO_CLANG_RUNTIMES armv8m.main_hard_fp armv8m.main_hard_fp_unaligned armv8m.main_hard_fp_unaligned_size armv8m.main-unknown-none-eabi)
 else()
     set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=softfp")
     # these are all the directories under LLVM embedded toolchain for ARM (newlib or picolibc) and under llvm_libc
-    set(PICO_CLANG_RUNTIMES armv8m.main_soft_nofp armv8m.main_soft_nofp_unaligned armv8m.main_soft_nofp_unaligned_size armv8m.main-unknown-none-eabi)
+    set(PICO_CLANG_RUNTIMES armv8m.main_soft_nofp armv8m.main_soft_nofp_unaligned armv8m.main_soft_nofp_unaligned_size  armv8m.main-unknown-none-eabi)
```
kilograham:
```suggestion
    set(PICO_CLANG_RUNTIMES armv8m.main_soft_nofp armv8m.main_soft_nofp_unaligned armv8m.main_soft_nofp_unaligned_size armv8m.main-unknown-none-eabi)
```

Prabhuk:
As discussed offline, we are bumping the toolchain revision to a new enough revision but prior to toolchain switching the float ABI for cortex-m33 runtime libraries to hard float.

kilograham (approved):

PR #2880, opened 2026-03-27
Add generated trees of the new linker scripts

will-v-pi:
Add some scripts to generate trees to show what the new linker scripts include, and CI to check they are up to date


kilograham (commented):
Id point out this is the default include tree (in the comment) and that the user can override

kilograham (approved):

PR #2883, opened 2026-03-29
fix: Add missing block_cipher.c to pico_mbedtls

### LINKED ISSUE #2882: pico_mbedtls doesn't compile block_cipher.c
I tried using AES-256-GCM using pico_mbedtls. However, linking always failed with this error message:
```
/usr/x86_64-pc-linux-gnu/arm-none-eabi/binutils-bin/2.46.0/ld: CMakeFiles/PROJECT/opt/raspberrypi/pico-sdk/lib/mbedtls/library/gcm.c.o: in function `gcm_mask':
/opt/raspberrypi/pico-sdk/lib/mbedtls/library/gcm.c:543:(.text.gcm_mask+0x16): undefined reference to `mbedtls_block_cipher_encrypt'
/usr/x86_64-pc-linux-gnu/arm-none-eabi/binutils-bin/2.46.0/ld: (mbedtls_block_cipher_encrypt): Unknown destination type (ARM/Thumb) in CMakeFiles/PROJECT/opt/raspberrypi/pico-sdk/lib/mbedtls/library/gcm.c.o
```

My mbedtls_config file looks like this:
```
#ifndef _MBEDTLS_CONFIG_H
#define _MBEDTLS_CONFIG_H

#if LIB_PICO_SHA256
// Enable hardware acceleration
#define MBEDTLS_SHA256_ALT
#define MBEDTLS_MD_CAN_SHA256
#else
#define MBEDTLS_SHA256_C
#endif

// KDF
#define MBEDTLS_MD_C
#define MBEDTLS_HKDF_C

// AES GCM
#define MBEDTLS_AES_C
#define MBEDTLS_GCM_C

#endif
```

I added `#define MBEDTLS_MD_CAN_SHA256` since otherwise, some header check told me that preconditions were not met.

Adding this line to my own CMakeLists.txt file fixed it:
```
add_executable(${CMAKE_PROJECT_NAME}
    ${PICO_SDK_PATH}/lib/mbedtls/library/block_cipher.c
   .... other sources
```

Thus, I assume the file is not referenced in pico_mbedtls by accident

kripton:
Fixes #2882

It looks like the reference to `block_cipher.c` was missed in pico_mbedtls.

I did NOT recompile any examples with this changes but the file exists on disk and will do nothing if the relevant `#define` is not set in `mbedtls_config.h`

### DIFF
--- a/src/rp2_common/pico_mbedtls/CMakeLists.txt
+++ b/src/rp2_common/pico_mbedtls/CMakeLists.txt
@@ -37,7 +37,6 @@ if (EXISTS ${PICO_MBEDTLS_PATH}/${MBEDTLS_TEST_PATH})
             asn1write.c
             base64.c
             bignum.c
-            block_cipher.c
             camellia.c
             ccm.c
             chacha20.c
@@ -122,6 +121,14 @@ if (EXISTS ${PICO_MBEDTLS_PATH}/${MBEDTLS_TEST_PATH})
                 sha3.c
             )
         endif()
+
+        # block_cipher.c was only added with v3.6.0. Support versions without it as well
+        if (EXISTS ${PICO_MBEDTLS_PATH}/library/block_cipher.c)
+            list(APPEND src_crypto
+                block_cipher.c
+            )
+        endif()
+
         list(TRANSFORM src_crypto PREPEND ${PICO_MBEDTLS_PATH}/library/)
         set(src_crypto ${src_crypto} PARENT_SCOPE)
     endfunction()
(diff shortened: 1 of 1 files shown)


kripton:
OK, I need to rework this. Merge conflict arose because I did the changes on main but switched the base branch of this PR to `develop` as instructed

kripton:
Ok, commit updated against latest `develop`. I also checked https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_mbedtls/mbedtls.BUILD#L8 but it shouldn't be an issue here since a glob pattern is used.

peterharperuk (changes requested):
I think support for this was only added in 3.6.0? In which case we should make an effort to still build with older versions of mbedtls.

kripton:
> I think support for this was only added in 3.6.0? In which case we should make an effort to still build with older versions of mbedtls.

Agreed, that makes sense. I changed the PR accordingly and tested my project (without the workaround described in #2882) compiles fine. However, I have not tested using an older version of mbedtls (<3.6.0)

kripton:
Polite ping @peterharperuk. I think I made the relevant changes and this PR awaits a re-review from you. Thanks :)

peterharperuk (approved):

PR #2885, opened 2026-04-01
enable float/double benchmark on risc-v

kilograham:


test/pico_float_test/float_benchmark.c
```diff
@@ -902,7 +919,7 @@ timer_func_def(fcopysign)(volatile float a, volatile float b) {
     uint32_t t0 = *systick_ptr;
     volatile float x = copysignf(a, b);
     uint32_t t1 = *systick_ptr;
-    return cycle_diff(t0, t1) - FLOAT_INPUT_COST - FLOAT_OUTPUT_COST;
+    return cycle_diff(t0, t1) - FLOAT_INPUT_COST * 2 - FLOAT_OUTPUT_COST;
```
lurch, line 922: Naive question: Should the other benchmarking functions that take two `float` inputs (e.g. `ffmod` and `fdrem`) also be using `FLOAT_INPUT_COST * 2` ?

kilograham: yes - are there some i didn't look in detail

kilograham: have fixed some others

PR #2887, opened 2026-04-05
add missing splitmix when combined random state from cncurrent updates

kilograham:


peterharperuk (approved):
Makes sense

lurch:
Ping also @andygpz11 as I think he wrote this library originally?

andygpz11:
Thanks @lurch

andygpz11 (approved):
LGTM

PR #2891, opened 2026-04-05
Refactor toolchain CMake files to avoid duplication

kilograham:


cmake/preload/toolchains/util/pico_gcc_common.cmake
```diff
@@ -0,0 +1,57 @@
+include(${CMAKE_CURRENT_LIST_DIR}/find_compiler.cmake)
+
+# include our Platform/PICO.cmake
+set(CMAKE_SYSTEM_NAME PICO)
+
+# PICO_CMAKE_CONFIG: PICO_GCC_TRIPLE, List of GCC_TRIPLES -- usually only one -- to try when searching for a compiler. This may be specified the user environment, type=int, default=PICO_DEFAULT_GCC_TRIPLE which is set based on PICO_COMPILER, group=pico_base, docref=cmake-toolchain-config
```
lurch:
```suggestion
# PICO_CMAKE_CONFIG: PICO_GCC_TRIPLE, List of GCC_TRIPLES -- usually only one -- to try when searching for a compiler. This may be specified as an environment variable, type=string, default=PICO_DEFAULT_GCC_TRIPLE which is set based on PICO_COMPILER, group=pico_base, docref=cmake-toolchain-config
```

Also, is this a space-separated list. comma-separated, or something else?

kilograham: it is a CMake list

lurch: The reason I asked was because I was wondering how you'd pass a list if you're using the environment-variable :wink:  Or does using the env-var limit you to just a single triple?

cmake/preload/toolchains/util/pico_arm_gcc_common.cmake
```diff
@@ -1,57 +1,11 @@
...
-        message(FATAL_ERROR "PICO_DEFAULT_GCC_TRIPLE is not defined")
-    endif()
+# PICO_CMAKE_CONFIG: PICO_HARD_FLOAT_ABI, use hard floating point ABI, type=bool, default=0, group=build, docref=cmake-toolchain-config
+
+# note to future self - putting the -mfloat-abi before the other -mcpu/-march/-mfpu cause GCC to
+# wrongly emit .fpu softvfp regardless of any other options! go figure!!!
+if (PICO_HARD_FLOAT_ABI)
+    set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=hard")
+else()
+    set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} -mfloat-abi=softfp")
```
will-v-pi, line 8: Just double checking that adding `-mfloat-abi=softfp` isn't going to change the behaviour on RP2040, where `-mfloat-abi` wasn't previously specified?

Also, the m23 file still sets `-mfloat-abi=softfp`, so that probably needs removing now it is set here

will-v-pi (commented):
Is it worth doing something similar for clang (probably just moving everything from `pico_arm_clang_common` to `pico_clang_common`), to make adding Risc-V clang easier in the future?

Also, is the `pico_arm_clang_arm.cmake` file used by anything, or should that be removed in this cleanup?

cmake/preload/toolchains/util/pico_gcc_common.cmake
kilograham: i think is is semi-colons

cmake/preload/toolchains/util/pico_arm_gcc_common.cmake
kilograham: ah. yeah it builds fine on RP2040

kilograham:
> Is it worth doing something similar for clang (probably just moving everything from `pico_arm_clang_common` to `pico_clang_common`), to make adding Risc-V clang easier in the future?

Lets make that a separate PR

> 
> Also, is the `pico_arm_clang_arm.cmake` file used by anything, or should that be removed in this cleanup?

Removed in separate PR  #2892 so we are explicit about it

PR #2897, opened 2026-04-13
Make all the tests use "PASSED" to indicate success vs some other string

kilograham:


will-v-pi (approved):
I approve of this, mostly because it means `kitchen_sink` now exits, so if you use `PICO_ENTER_USB_BOOT_ON_EXIT=1` it actually exits to bootsel mode, rather than hanging on the breakpoint in the default exception handler

lurch:
Looks plausible :+1: 
I've never actually run any of these tests, but it looks like everything inside the `test/` directory either does `printf("PASSED\n");` or `puts("PASSED");` or `ostr("PASSED\n");` or `PICOTEST_END_TEST();`. (I guess in an ideal world everything would `#include "pico/test.h"` and use `PICOTEST_END_TEST()`, but that's obviously a much bigger refactoring!)

* Does https://github.com/raspberrypi/pico-sdk/blob/develop/test/pico_float_test/m33.c#L251 also need updating, or is that a test which doesn't run on-device?
* Is there some combination of defines that would result in https://github.com/raspberrypi/pico-sdk/blob/develop/test/panic_function_test/panic_function_test.c#L66 not printing "PASSED" ?

PR #2898, opened 2026-04-13
Fix basic picolibc support

kilograham:


src/rp2_common/pico_standard_link/script_include/section_copy_to_ram_data.incl
```diff
@@ -35,6 +35,7 @@ SECTIONS
 
         . = ALIGN(4);
         /* preinit data */
+        PROVIDE_HIDDEN (__bothinit_array_start = .);
```
will-v-pi, line 39: `__bothinit_array_start` and `end` should be added to the comment at the top for these files

src/rp2_common/pico_standard_link/CMakeLists.txt
will-v-pi, line 172: Does this also need to be changed to `-T`, or is it only the main linker script that needs that?

will-v-pi (commented):
Did you want to add https://github.com/raspberrypi/pico-sdk/pull/2846 to this PR too (probably with some new section files)? With my comment to put the new stuff between `.data` and `.bss`, so that the low power PR can insert `.persistent_data` between `.bss` and `.heap`.

src/rp2_common/pico_standard_link/script_include/section_copy_to_ram_data.incl
kilograham: ok

src/rp2_common/pico_standard_link/CMakeLists.txt
kilograham: no, i think picolibc.specs just looks for any -T arg to suppress its own linker (this fix is a bit of a hack, we can revisit - if need be - it works on all clang and gcc we support)

will-v-pi (approved):

edo9300:
This PR didn't bring in all the changes from the one it superseeded, making picolibc builds still fail, more specifically, it didn't add the checks in the cmakelists files of the boot_stage2 projects https://github.com/raspberrypi/pico-sdk/pull/2840/files#diff-cce72f3991862657cdf9eaf928e7bdf89d7ff38a75656ffa438c07bef64d47a4 and https://github.com/raspberrypi/pico-sdk/pull/2840/files#diff-bfec06eb22ca43444b572d23036d14ab8ef4bc57a811a3125420e9b96ae76139

kilograham:
oops; thx #2917

PR #2900, opened 2026-04-14
Replace `new_git_repository` with `git_repository`

### LINKED ISSUE #2899: Replace deprecated bazel rule new_git_repository
Since Bazel 7.0 (and maybe even earlier), the repository rule `new_git_repository` has just been an alias of the `git_repository` rule.  See https://github.com/bazelbuild/bazel/blob/d798ebde6c6394203a87b5f1a6b62ecfc3880991/tools/build_defs/repo/git.bzl#L218

Recently, a change was made to make use of `new_git_repository` a breaking failure.  See https://github.com/bazelbuild/bazel/commit/71160cb619b5bfeb89c91fee53bb044438ee9b3f

People should migrate away from using `new_git_repository` and instead use `git_repository`.

willstranton:
Since Bazel 7, new_git_repository has just been an alias of git_repository. Recently, in
https://github.com/bazelbuild/bazel/commit/71160cb619b5bfeb89c91fee53bb044438ee9b3f the use of new_git_repository will fail. Users should migrate to using git_repository instead. This should be a noop change - no change in behavior

Fixes #2899

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -99,10 +99,10 @@ http_archive(
     url = "https://chrome-infra-packages.appspot.com/dl/fuchsia/third_party/clang/mac-arm64/+/git_revision:910be4ff90d7d07bd4518ea03b85c0974672bf9c",
 )
 
-new_git_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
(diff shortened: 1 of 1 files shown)


armandomontanez (approved):
Thanks, not sure how I ended up using `new_git_repository` instead of `git_repository` here.

kilograham (approved):

Issue #2643, opened 2025-09-02
gpio_init_mask does not initialize GPIOs 32-47 on RP2350B


romain145:
https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/hardware_gpio/gpio.c#L283

The gpio_mask is a uint, which is 32 bits. I suspect GPIOs above 31 are initialized incorrectly on RP2350B.

lurch:
Hmmm, maybe there ought to be a `gpio_init_mask64` function, similar to the various other `*_masked64` functions in that file?
(Also, perhaps it makes sense to change `void gpio_init_mask(uint gpio_mask)` to a more explicit `void gpio_init_mask(uint32_t gpio_mask)`, in order to be consistent with all the other functions that take a `gpio_mask` parameter?)

kilograham:
merged into develop

PR #2907, opened 2026-04-23
Add `gpio_init_mask64` function

FLyrfors-MI:
Add `gpio_init_mask64` function mirroring the other `mask64`/`masked64` GPIO functions used by RP2350B with more than 32 GPIO's.

See #2643

### DIFF
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -275,7 +275,7 @@ void gpio_init(uint gpio) {
     gpio_set_dir(gpio, GPIO_IN);
     gpio_put(gpio, 0);
     gpio_set_function(gpio, GPIO_FUNC_SIO);
-}
+}7
 
 void gpio_deinit(uint gpio) {
     gpio_set_function(gpio, GPIO_FUNC_NULL);
@@ -292,11 +292,12 @@ void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
 }
 
 void gpio_init_mask(uint32_t gpio_mask) {
-    gpio_set_dir_masked(gpio_mask, GPIO_IN);
+    gpio_set_dir_in_masked(gpio_mask);
     gpio_clr_mask(gpio_mask);
     gpio_set_function_masked(gpio_mask, GPIO_FUNC_SIO);
 }
 
+// these functions are provided (in gpio.h) as inlined calls to the 32 bit versions if we have <= 32 GPIOs
 #if NUM_BANK0_GPIOS > 32
 void gpio_set_function_masked64(uint64_t gpio_mask, gpio_function_t fn) {
     for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
@@ -308,9 +309,8 @@ void gpio_set_function_masked64(uint64_t gpio_mask, gpio_function_t fn) {
     }
 }
 
-// these functions collapse to the non 64 bit versions as inline funcs if we have < 32 GPIOs
 void gpio_init_mask64(uint64_t gpio_mask) {
-    gpio_set_dir_masked64(gpio_mask, GPIO_IN);
+    gpio_set_dir_in_masked64(gpio_mask);
     gpio_clr_mask64(gpio_mask);
     gpio_set_function_masked64(gpio_mask, GPIO_FUNC_SIO);
 }
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -282,25 +282,39 @@ void gpio_deinit(uint gpio) {
 }
 
 void gpio_init_mask(uint32_t gpio_mask) {
-    for(uint32_t i=0;i<NUM_BANK0_GPIOS;i++) {
+    for(uint i=0;i<NUM_BANK0_GPIOS;i++) {
         if (gpio_mask & 1) {
             gpio_init(i);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
 }
 
 void gpio_init_mask64(uint64_t gpio_mask) {
-    for(uint64_t i=0;i<NUM_BANK0_GPIOS;i++) {
-        if (gpio_mask & 1) {
+#if NUM_BANK0_GPIOS <= 32
+    // code is much smaller in this case, especially for Cortex-M0+ which is convenient since RP2040 has < 32 pins
+    uint32_t gpio_mask32 = (uint32_t)gpio_mask;
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
+        if (gpio_mask32 & 1u) {
+            gpio_init(i);
+        }
+        gpio_mask32 >>= 1;
+        if (!gpio_mask32) break;
+    }
+#else
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
+        if (gpio_mask & 1u) {
             gpio_init(i);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
+#endif
 }
 
 void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
-    for (uint i = 0; i < MIN(NUM_BANK0_GPIOS, 32u); i++) {
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
         if (gpio_mask & 1u) {
             gpio_set_function(i, fn);
         }
@@ -309,10 +323,23 @@ void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
 }
 
 void gpio_set_function_masked64(uint64_t gpio_mask, gpio_function_t fn) {
-    for (uint i = 0; i < MIN(NUM_BANK0_GPIOS, 64u); i++) {
+#if NUM_BANK0_GPIOS <= 32
+    // code is much smaller in this case, especially for Cortex-M0+ which is convenient since RP2040 has < 32 pins
+    uint32_t gpio_mask32 = (uint32_t)gpio_mask;
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
+        if (gpio_mask32 & 1u) {
+            gpio_set_function(i, fn);
+        }
+        gpio_mask32 >>= 1;
+        if (!gpio_mask32) break;
+    }
+#else
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
         if (gpio_mask & 1u) {
             gpio_set_function(i, fn);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
+#endif
 }
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -319,6 +319,7 @@ void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
             gpio_set_function(i, fn);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
 }
 
--- a/src/rp2_common/hardware_gpio/gpio.c
+++ b/src/rp2_common/hardware_gpio/gpio.c
@@ -281,16 +281,6 @@ void gpio_deinit(uint gpio) {
     gpio_set_function(gpio, GPIO_FUNC_NULL);
 }
 
-void gpio_init_mask(uint32_t gpio_mask) {
-    for(uint i=0;i<NUM_BANK0_GPIOS;i++) {
-        if (gpio_mask & 1) {
-            gpio_init(i);
-        }
-        gpio_mask >>= 1;
-        if (!gpio_mask) break;
-    }
-}
-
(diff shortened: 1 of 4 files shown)


src/rp2_common/hardware_gpio/include/hardware/gpio.h
```diff
@@ -41,23 +41,23 @@ extern "C" {
...
- * pins are used to execute code from an external flash device, leaving the User bank (GPIO0 to GPIO29) for the programmer to use. 
+ * RP2040 has 30 user GPIO pins in bank 0, and 6 QSPI pins in the QSPI bank 1 (QSPI_SS, QSPI_SCLK and QSPI_SD0 to QSPI_SD3). The QSPI
+ * pins are used to execute code from an external flash device, leaving the User bank (GPIO0 to GPIO29) for the programmer to use.
  * \endif
- * 
+ *
  * \if rp2350_specific
- * The number of GPIO pins available depends on the package. There are 30 user GPIOs in bank 0 in the QFN-60 package (RP2350A), or 48 user GPIOs 
+ * The number of GPIO pins available depends on the package. There are 30 user GPIOs in bank 0 in the QFN-60 package (RP2350A), or 48 user GPIOs
  * in the QFN-80 package. Bank 1 contains the 6 QSPI pins and the USB DP/DM pins.
```
lurch:
```suggestion
 * The number of GPIO pins available depends on the package. There are 30 user GPIOs in bank 0 in the QFN-60 package (RP2350A), or 48 user GPIOs
 * in the QFN-80 package (RP2350B). Bank 1 contains the 6 QSPI pins and the USB DP/DM pins.
```

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,8 +281,17 @@ void gpio_deinit(uint gpio) {
     gpio_set_function(gpio, GPIO_FUNC_NULL);
 }
 
-void gpio_init_mask(uint gpio_mask) {
-    for(uint i=0;i<NUM_BANK0_GPIOS;i++) {
+void gpio_init_mask(uint32_t gpio_mask) {
+    for(uint32_t i=0;i<NUM_BANK0_GPIOS;i++) {
```
lurch: `gpio_set_function_masked` does
```c
for (uint i = 0; i < MIN(NUM_BANK0_GPIOS, 32u); i++) {
```
Probably makes sense to do the same here?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,8 +281,17 @@ void gpio_deinit(uint gpio) {
...
+        if (gpio_mask & 1) {
+            gpio_init(i);
+        }
+        gpio_mask >>= 1;
+    }
+}
+
+void gpio_init_mask64(uint64_t gpio_mask) {
+    for(uint64_t i=0;i<NUM_BANK0_GPIOS;i++) {
```
lurch: Similarly, `gpio_set_function_masked64` does
```c
for (uint i = 0; i < MIN(NUM_BANK0_GPIOS, 64u); i++) {
```

src/rp2_common/hardware_gpio/gpio.c
kilograham: i reimplemented

src/rp2_common/hardware_gpio/gpio.c
kilograham: i reimplemented

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,29 +281,66 @@ void gpio_deinit(uint gpio) {
...
 void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
-    for (uint i = 0; i < MIN(NUM_BANK0_GPIOS, 32u); i++) {
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
         if (gpio_mask & 1u) {
             gpio_set_function(i, fn);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
 }
```
lurch, line 292: It's only a small optimisation, but is it worth doing:
```c
void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
    if (gpio_mask) {
        for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
            if (gpio_mask & 1u) {
                gpio_set_function(i, fn);
            }
            gpio_mask >>= 1;
            if (!gpio_mask) break;
        }
    }
}
```
(or is it better to assume that this function will never be called with an empty `gpio_mask`?)

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,29 +281,66 @@ void gpio_deinit(uint gpio) {
...
+    // code is much smaller in this case, especially for Cortex-M0+ which is convenient since RP2040 has < 32 pins
+    uint32_t gpio_mask32 = (uint32_t)gpio_mask;
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
+        if (gpio_mask32 & 1u) {
+            gpio_init(i);
+        }
+        gpio_mask32 >>= 1;
+        if (!gpio_mask32) break;
+    }
+#else
```
will-v-pi: Is it worth just doing this, rather than duplicating the code?
```suggestion
#if NUM_BANK0_GPIOS <= 32
    gpio_init_mask((uint32_t)gpio_mask);
#else
```

Or even just a define?
```c
#define gpio_init_mask64(gpio_mask) gpio_init_mask((uint32_t)gpio_mask)
```

src/rp2_common/hardware_gpio/gpio.c
kilograham: it is an unlikely parameter and the code still works, so we don't want to increase code size

src/rp2_common/hardware_gpio/gpio.c
kilograham: yeah probably cleaner to #ifdef it in the header - though some people may be sad that the function appears/disappears

will-v-pi: Maybe #ifdef it as an inline function in the header, so code using mask64 still compiles, and you can still do type checking? I agree having the mask64 functions disappear entirely would be bad, for anyone writing more universal code.

kilograham: yeah that's what i'm suggesting - some people may get confused/bitten by the fact that sometimes the outline function exists and sometimes it doesn't too, maybe I'll allow it with a #undef in the gpio.c

kilograham: actually i changed my mind; #defining one function to another makes the doxygen confusing, so using static inline instead

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,29 +281,45 @@ void gpio_deinit(uint gpio) {
...
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
         if (gpio_mask & 1u) {
             gpio_set_function(i, fn);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
+    }
+}
+
+#if NUM_BANK0_GPIOS >= 32
```
lurch: The condition in the header file is `#if NUM_BANK0_GPIOS <= 32` so this line ought to be
```suggestion
#if NUM_BANK0_GPIOS > 32
```
?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,29 +281,45 @@ void gpio_deinit(uint gpio) {
...
-void gpio_init_mask(uint gpio_mask) {
+void gpio_init_mask(uint32_t gpio_mask) {
     for(uint i=0;i<NUM_BANK0_GPIOS;i++) {
         if (gpio_mask & 1) {
             gpio_init(i);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
 }
```
lurch: Hmmm...
The implementation of `gpio_init_mask` (and also `gpio_init_mask64`) simply calls `gpio_init` in a loop. But `gpio_init` does:
```c
void gpio_init(uint gpio) {
    gpio_set_dir(gpio, GPIO_IN);
    gpio_put(gpio, 0);
    gpio_set_function(gpio, GPIO_FUNC_SIO);
}
```
and `gpio_set_dir` calls `gpio_set_dir_in_masked`, and `gpio_put` calls `gpio_clr_mask`.

So maybe it'd be more efficient if `gpio_init_mask` was changed to:
```c
void gpio_init_mask(uint32_t gpio_mask) {
    gpio_set_dir_in_masked(gpio_mask);
    gpio_clr_mask(gpio_mask);
    gpio_set_function_masked(gpio_mask, GPIO_FUNC_SIO);
}
```
? :thinking:

will-v-pi: Yeah, I agree with this - it's presumably a large speedup, given the `set_dir_in` and `clr_mask` can be single coprocessor instructions

kilograham: nice good catch - presumably applies to 64 bit variant too

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,29 +281,38 @@ void gpio_deinit(uint gpio) {
...
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
         if (gpio_mask & 1u) {
             gpio_set_function(i, fn);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
 }
+
+// these functions collapse to the non 64 bit versions as inline funcs if we have < 32 GPIOs
```
will-v-pi: This comment probably belongs next to `#if NUM_BANK0_GPIOS > 32` rather than here?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -281,29 +281,38 @@ void gpio_deinit(uint gpio) {
...
         if (gpio_mask & 1u) {
             gpio_set_function(i, fn);
         }
         gpio_mask >>= 1;
+        if (!gpio_mask) break;
     }
 }
 
+void gpio_init_mask(uint32_t gpio_mask) {
+    gpio_set_dir_masked(gpio_mask, GPIO_IN);
```
lurch: IMHO this should be
```suggestion
    gpio_set_dir_in_masked(gpio_mask);
```
as per my original suggestion. Note that the second parameter to `gpio_set_dir_masked` isn't `GPIO_IN`/`GPIO_OUT`, but a per-gpio direction value. See https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_gpio_1gaa64fdceff54dda03f4427f04c2f9ced2

kilograham: true dat

src/rp2_common/hardware_gpio/gpio.c
kilograham: si

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -275,35 +275,44 @@ void gpio_init(uint gpio) {
     gpio_set_dir(gpio, GPIO_IN);
     gpio_put(gpio, 0);
     gpio_set_function(gpio, GPIO_FUNC_SIO);
-}
+}7
```
lurch: Oops?

src/rp2_common/hardware_gpio/gpio.c
```diff
@@ -275,35 +275,44 @@ void gpio_init(uint gpio) {
...
-        if (gpio_mask & 1) {
-            gpio_init(i);
-        }
-        gpio_mask >>= 1;
-    }
-}
-
 void gpio_set_function_masked(uint32_t gpio_mask, gpio_function_t fn) {
-    for (uint i = 0; i < MIN(NUM_BANK0_GPIOS, 32u); i++) {
+    for (uint i = 0; i < NUM_BANK0_GPIOS; i++) {
```
will-v-pi, line 285: Is it worth using `__builtin_ctz` to skip over the 0s at the start of the mask, as it already skips 0s at the end with the `if (!gpio_mask)?

Could even skip over 0s in the middle, but that would probably give something that's slower for setting lots of pins, and would probably increase code size, so that's probably not worth it

will-v-pi (dismissed):
One possible improvement, but that can easily go in a separate PR if it's worth it, so approving

src/rp2_common/hardware_gpio/gpio.c
will-v-pi:
```suggestion
}
```

will-v-pi (approved):

src/rp2_common/hardware_gpio/gpio.c
lurch: Thinking on this some more, I guess you could just iterate between the first-set-bit and last-set-bit by using both `__builtin_ctz` and `__builtin_ltz`, and then you wouldn't need the `if (!gpio_mask)` check inside the loop-body?
And then would
```c
        if (gpio_mask & (1u << i)) {
            gpio_set_function(i, fn);
        }
```
be any more efficient than
```c
        if (gpio_mask & 1u) {
            gpio_set_function(i, fn);
        }
        gpio_mask >>= 1;
```
?

PR #2909, opened 2026-04-24
Add PICO_TEST_FILE_GENERATOR and PICO_TEST_ target properties

will-v-pi:
Add `PICO_TEST_FILE_GENERATOR` option, along with `PICO_TEST_` target properties to aid in generation of lists of tests for running in CI.

This allows pointing to a script file with `-DPICO_TEST_FILE_GENERATOR=/path/to/my_test_generator.cmake`, and that script file can generate JSON (or whatever format you want) listing whatever the CI runner needs to know.

The `PICO_TEST_` target properties help that generator understand what certain tests do, such as what string they print on success (`PICO_TEST_SUCCESS_STRING`) or how long they are expected to take (`PICO_TEST_TIMEOUT`). The script file can also access the usual target properties such as `PICO_TARGET_STDIO_USB` or `PICO_TARGET_BINARY_TYPE` if it needs that information, so those don't need to be added as separate `PICO_TEST_` properties.

This should not affect normal SDK builds, as the file is only included when it is specified as an environment variable or on the command line. I haven't added a `PICO_CMAKE_CONFIG` for it as I don't think it should be documented in the SDK PDF, as this is primarily for internal use.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -34,6 +34,7 @@ if (PICO_TEST_FILE_GENERATOR AND EXISTS ${PICO_TEST_FILE_GENERATOR})
     #   Can be used to force failure even when the test prints the success string
     # - PICO_TEST_BUDDY_FILE: File to run on buddy device for the test (default is NONE)
     #   For example, low power dormant tests on RP2040 require rtc_clksrc running on the buddy device
+    #   This can either be a path relative to the test target, or an absolute path
     # - PICO_TEST_SKIP_IN_CI: Whether to skip the test in ci (default is FALSE)
     #   If TRUE, the test is not included in the ci test suite
     include(${PICO_TEST_FILE_GENERATOR})
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -16,3 +16,25 @@ if (PICO_ON_DEVICE)
     add_subdirectory(pico_sha256_test)
     add_subdirectory(pico_async_context_test)
 endif()
+
+# PICO_TEST_FILE_GENERATOR: Path to the a CMake script to generate test files for ci
+if (DEFINED ENV{PICO_TEST_FILE_GENERATOR} AND NOT PICO_TEST_FILE_GENERATOR)
+    set(PICO_TEST_FILE_GENERATOR $ENV{PICO_TEST_FILE_GENERATOR})
+    message("Using PICO_TEST_FILE_GENERATOR from environment ('${PICO_TEST_FILE_GENERATOR}')")
+endif()
+
+if (PICO_TEST_FILE_GENERATOR AND EXISTS ${PICO_TEST_FILE_GENERATOR})
+    # Include script file to generate test files for ci
+    # This can use the following extra target properties from the test targets:
+    # - PICO_TEST_TIMEOUT: Timeout in seconds for the test (default is 5)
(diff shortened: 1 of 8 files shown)


test/pico_float_test/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ if (PICO_RISCV)
     target_link_libraries(pico_float_test PRIVATE pico_float pico_stdlib)
     target_include_directories(pico_float_test PRIVATE ${CMAKE_CURRENT_LIST_DIR})
     pico_add_extra_outputs(pico_float_test)
+    set_target_properties(pico_float_test PROPERTIES PICO_TEST_SUCCESS_STRING "Well done, you can relax now")
```
kilograham: seems like we should just change the string to PASSED

test/pico_float_test/CMakeLists.txt
```diff
@@ -138,7 +139,7 @@ if (PICO_RP2350 AND NOT PICO_RISCV)
     add_executable(m33
             m33.c
     )
-
+    set_target_properties(m33 PROPERTIES PICO_TEST_SUCCESS_STRING "Pass")
```
kilograham: also change to PASSED (I guess i missed some)

test/CMakeLists.txt
```diff
@@ -16,3 +16,26 @@ if (PICO_ON_DEVICE)
...
+    set(PICO_TEST_FILE_GENERATOR $ENV{PICO_TEST_FILE_GENERATOR})
+    message("Using PICO_TEST_FILE_GENERATOR from environment ('${PICO_TEST_FILE_GENERATOR}')")
+endif()
+
+if (PICO_TEST_FILE_GENERATOR AND EXISTS ${PICO_TEST_FILE_GENERATOR})
+    # Include script file to generate test files for ci
+    # This can use the following extra target properties from the test targets:
+    # - PICO_TEST_TIMEOUT: Timeout in seconds for the test (default is 5)
+    #   Maximum time the test is allowed to run for
+    # - PICO_TEST_SUCCESS_STRING: Success string for the test (default is "PASSED")
```
lurch, line 31: Where are the default values mentioned here actually set?

will-v-pi: They are set by the external test file generator, documented here for anyone else writing test file generators

test/pico_float_test/CMakeLists.txt
will-v-pi: Have updated this string and the `m33.c` string too

test/CMakeLists.txt
lurch: ...in which case these aren't really "defaults"? (as I guess the external test file generator could just set them to whatever it wants?) :confused:

test/pico_float_test/CMakeLists.txt
lurch: Heh :smiley:  https://github.com/raspberrypi/pico-sdk/pull/2897#issuecomment-4305356543

test/CMakeLists.txt
will-v-pi: You can, but then your external test system may not work if you don't use the correct defaults (eg if you're expecting everything to print "Success" then your external test system will not work). These are mostly for internal use anyway.

kilograham (approved):

PR #2910, opened 2026-04-25
Add pico_thread_local library providing thread local variable support

kilograham:
* Works on GCC/llvm, newlib/picolibc, Arm/RISC-V
* Provides picolibc style functions for FreeRTOS (and perhaps others to use)
* Attempts to minimize runtime overhead when not used:
   - note linker scripts are not competent enough to allow us to elide code based on the emptiness of sections
   - on Arm we have a signal of whether __aeabi_read_tp or __emutls_get_address are called (we dont use this to change the link, we just do lazy initializtion from these methods which removes most oof the code if there are no thread local variables (since the methods aren't ever called and we use gc-sections)
   - on RISC-V theres not much we can do since TP reg is used
   - support for resetting thread locals on core 1 re-init is optional
   - thread local support can be dialed back if not needed (see below)
* Three types of TLS support are provided:
   - per_thread (default) - proper per core (and potentially RTOS task) values for __thread variables
   - global - __thread is supported but only a single shared value is used
   - none - programs using __thread may fail to link and runtime results are undefined
* Note that tdata/tbss source data sections can be used if present
* Note that emutls source data sections can be used if present
* Code will still compile with older linker scripts missing the tls sections
* In either case two linear RAM sections .tls0 and .tls1 are used, and _init_tls is used to copy the relevant source data from there
* Note .tdata/.tbss sit between .data/.bss so that they are copied/cleared correctly by crt0. This was the case before, but I've recombined the separate linker script fragments to make this more concrete
* The CMake build has code to detect emutls vs __aeabi_read_tp on Arm which allows it to pass defines indicating support is only needed for one type. Of course, other build systems or pico_clib_interfaces are free to define these variables too, but they are not required for correct behavior (they may just save space)
* Note that the code was basde somewhat on earlier separate PRs. Despite being somewhat related to the c library being used, it is orthogonal in the pico-sdk code since it really is a compiler decision. If a particular pico_clib_interface wants to influence the behavior, it can do so via #defines

PER_THREAD tls mode
   - picolibc style tls funcs (_tls_size(), _init_tls()) work based on either tdata/tbss or emutls source data for the latter the emutls_objects are walked to figure out a linear size
   - two sections .tls0 and .tls1 are added to RAM to hold tdata and tbss for the two cores
   - .tls1 is re-initialized (or rather re-initialization is marked but deferred via per core runtime_init by default so values are reset on core 1 reset). There is a define to turn this off to save a tiny amount of space if not needed
   - note that .tbss actually currently wasted space in RAM in this scenario which is annoying but I think common to most runtimes
   - note similarly .tdata need not be in RAM here either as it is not directly
   - note w.r.t. to the last two points I don't want to rock the boar in terms of other code that may use these sections (e.g. other RTOS)
   - on RISC-V TP is initialed per core to .tls0 and .tls1

GLOBAL tls mode
   - the __aeabi_read_tp() value returned always points at .tdata/.tbss
   - RISC-V TP reg is initialized to .tdata./.tbss on core init
   - emutls falls back to the c library
   - the .tls0 and .tls1 sections are elided from the output via some sorcery
   - _tls_size() returns 0 to prevent anyone else allocating space

NONE tls mode
   - the .tls0 and .tls1 sections are elided from the output via some sorcery
   - no other code is included, so you are left with whatever c library support there is, but without crt having initialized it correctly. in other words, it ain't gonna work right!


kilograham:
cc @alastairpatrick @keith-packard

src/rp2_common/pico_clib_interface/newlib_interface.c
keith-packard, line 1: Looks like a few whitespace fixes unrelated to this patch. Dunno if that's OK in this project or not.

kilograham: thx - if they improve they can go in ;-)

src/rp2_common/pico_thread_local/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+            if (NOT PICO_RISCV)
+                message(WARNING "Could not identify TLS mechanism")
+            endif()
+            set(TLS_TYPE "unknown" CACHE INTERNAL "TLS Type")
+        endif()
+    else()
+        message(STATUS "Compiler does not support __thread keyword.")
+        set(TLS_TYPE "unsupported" CACHE INTERNAL "TLS Type")
+    endif()
+    set(${OUTPUT_VAR} ${TLS_TYPE} PARENT_SCOPE)
```
will-v-pi, line 79: Maybe delete the files before returning too, to clean up

```suggestion
    set(${OUTPUT_VAR} ${TLS_TYPE} PARENT_SCOPE)
    file(REMOVE ${CMAKE_CURRENT_BINARY_DIR}/tls_check.c ${CMAKE_CURRENT_BINARY_DIR}/tls_check.a)
```

src/rp2_common/pico_thread_local/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+            set(TLS_TYPE "arm-aeabi" CACHE INTERNAL "TLS Type")
+        else()
+            if (NOT PICO_RISCV)
+                message(WARNING "Could not identify TLS mechanism")
+            endif()
+            set(TLS_TYPE "unknown" CACHE INTERNAL "TLS Type")
+        endif()
+    else()
+        message(STATUS "Compiler does not support __thread keyword.")
+        set(TLS_TYPE "unsupported" CACHE INTERNAL "TLS Type")
```
will-v-pi: I don't think `TLS_TYPE` needs to be a `CACHE` variable, it can just be a normal one

```suggestion
            set(TLS_TYPE "emutls")
        elseif(NM_OUT MATCHES "__aeabi_read_tp")
            message(STATUS "Detected native TLS (__aeabi_read_tp)")
            set(TLS_TYPE "arm-aeabi")
        else()
            if (NOT PICO_RISCV)
                message(WARNING "Could not identify TLS mechanism")
            endif()
            set(TLS_TYPE "unknown")
        endif()
    else()
        message(STATUS "Compiler does not support __thread keyword.")
        set(TLS_TYPE "unsupported")
```

src/rp2_common/pico_thread_local/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+
+    # Create a temporary file for the source
+    file(WRITE "${CMAKE_BINARY_DIR}/tls_check.c" "${TLS_CHECK_SOURCE}")
+
+    set(TLS_CHECK_LIBRARY "${CMAKE_BINARY_DIR}/tls_check.a")
+    set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+    # Attempt to compile the source file
+    try_compile(CAN_COMPILE_TLS
+            "${CMAKE_BINARY_DIR}"                     # Directory for temporary build files
+            "${CMAKE_BINARY_DIR}/tls_check.c"        # Source file to compile
```
will-v-pi: It might be better to write this in `CMAKE_CURRENT_BINARY_DIR` to avoid polluting the root of the build directory


```suggestion
    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/tls_check.c" "${TLS_CHECK_SOURCE}")

    set(TLS_CHECK_LIBRARY "${CMAKE_CURRENT_BINARY_DIR}/tls_check.a")
    set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
    # Attempt to compile the source file
    try_compile(CAN_COMPILE_TLS
            "${CMAKE_CURRENT_BINARY_DIR}"                     # Directory for temporary build files
            "${CMAKE_CURRENT_BINARY_DIR}/tls_check.c"        # Source file to compile
```

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,137 @@
...
+// PICO_CONFIG: PICO_THREAD_LOCAL_SUPPORT_THREAD_POINTER, Thread local support should work with compilers that use a per thread pointer and provide .tdata and .tbss, type=bool, default=1, advanced=true, group=pico_thread_local
+
+// PICO_CONFIG: PICO_THREAD_LOCAL_EMUTLS_CONFIRMED, Passed from build if the compiler is known to use Emutls and so other types of support are not needed, type=bool, default=0, advanced=true, group=pico_thread_local
+// PICO_CONFIG: PICO_THREAD_LOCAL_THREAD_POINTER_ARM_CONFIRMED, Passed from build if the compiler is known to use a thread pointer accessed via __areabi_read_tp and so other types of support are not needed, type=bool, default=0, advanced=true, group=pico_thread_local
+// PICO_CONFIG: PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED, Passed from build if the compiler is known to use a thread pointer stored in the TP register and so other types of support are not needed, type=bool, default=0, advanced=true, group=pico_thread_local
+#ifndef PICO_THREAD_LOCAL_EMUTLS_CONFIRMED
+#define PICO_THREAD_LOCAL_EMUTLS_CONFIRMED 0
+#endif
+#ifndef PICO_THREAD_LOCAL_THREAD_POINTER_ARM_CONFIRMED
+#define PICO_THREAD_LOCAL_THREAD_POINTER_ARM_CONFIRMED 0
```
will-v-pi: It looks like there are 2 names used for this - `PICO_THREAD_LOCAL_THREAD_POINTER_ARM_CONFIRMED` here which has the `PICO_CONFIG` but doesn't affect the build, or `PICO_THREAD_LOCAL_THREAD_POINTER_ARM_AEABI_CONFIRMED` below which doesn't have a `PICO_CONFIG` but does affect the build. I assume these are supposed to be the same name?

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * The pico_thread_local library comes in three main flavors:
+ *
+ * 1. `pico_thread_local_per_thread` (default) - Full support for thread local variables. Each thread/core gets its own copy of the variable initialized when it starts executing. The library tries to minimize the overhead if there are no thread locals are used, however particularly on RISC-V there is always a small overhead.
+ * 2. `pico_thread_local_global` - Thread local variables are allowed in code, but each thread/core shares the same value (i.e. they aren't really thread local). There is very minimial overhead for this option.
+ * 3. `pico_thread_local_none` - No support for thread locals is provided. Code using them may not compile or link, and if it does the values won't be shared and may not even be initialized correctly. This mode is however guaranteed to have basically no overhead when thread locals are known not to be used.
+ *
+ * The TLS provided by the `pico_thread_local` library may be set from CMake via `pico_set_tls_implementation(<TARGET> per_thread|global|none)` and the default for all targets may be set via CMake variable (e.g. `set(PICO_DEFAULT_TLS_IMPL=pico_thread_local_none`).
```
will-v-pi: CMake `set` doesn't use `=`

```suggestion
 * The TLS provided by the `pico_thread_local` library may be set from CMake via `pico_set_tls_implementation(<TARGET> per_thread|global|none)` and the default for all targets may be set via CMake variable (e.g. `set(PICO_DEFAULT_TLS_IMPL pico_thread_local_none`).
```

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,137 @@
...
+
+#if PICO_THREAD_LOCAL_EMUTLS_CONFIRMED + PICO_THREAD_LOCAL_THREAD_POINTER_ARM_CONFIRMED + PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED > 1
+#error expected at most one of PICO_THREAD_LOCAL_EMUTLS_CONFIRMED, PICO_THREAD_LOCAL_THREAD_POINTER_ARM_CONFIRMED and PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED to be set
+#endif
+
+#if PICO_THREAD_LOCAL_EMUTLS_CONFIRMED
+#define PICO_THREAD_LOCAL_SUPPORT_THREAD_POINTER 0
+#define PICO_THREAD_LOCAL_SUPPORT_EMUTLS 1
+#elif PICO_THREAD_LOCAL_THREAD_POINTER_ARM_AEABI_CONFIRMED || PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED || defined(PICOLIBC_TLS) || __riscv
+// we assume here and it tls.c that if PICOLIBC_TLS is set, or we're on RISC-V
```
will-v-pi:
```suggestion
// we assume here and in thread_local.c that if PICOLIBC_TLS is set, or we're on RISC-V
```

src/rp2_common/pico_thread_local/include/picotls.h
kilograham: i will review - actually i think i'm gonna back out the part of @lurch config checker which requires initializers for defaults of 0 too because these defines just clutter things

bazel/config/BUILD.bazel
```diff
@@ -151,6 +151,17 @@ string_flag(
     ],
 )
 
+# PICO_BAZEL_CONFIG: PICO_DEFAULT_THREAD_LOCAL_IMPL, The default implementation for pico_thread_local to link; per_thread provides per thread locals global provide shared global values, type=string, default=per_thread, group=build
```
lurch:
```suggestion
# PICO_BAZEL_CONFIG: PICO_DEFAULT_THREAD_LOCAL_IMPL, The default implementation for pico_thread_local to link; per_thread provides per thread locals while global provides shared global values, type=string, default=per_thread, group=build
```

src/rp2_common/pico_standard_link/script_include/section_tls.incl
```diff
@@ -0,0 +1,63 @@
...
+    PROVIDE( __tls_size = __tls_end - __tls_start );
+    PROVIDE( __tls_align = MAX(ALIGNOF(.tdata), ALIGNOF(.tbss)) );
+    PROVIDE( __tls_size_align = ALIGN( __tls_size, __tls_align) );
+    PROVIDE( __arm32_tls_tcb_offset = MAX(8, __tls_align) );
+    PROVIDE( __tdata_size = SIZEOF(.tdata) );
+    PROVIDE( __tdata_source = LOADADDR(.tdata) );
+    PROVIDE( __tbss_size = SIZEOF(.tbss) );
+    PROVIDE( __tbss_offset = ADDR(.tbss) - __tls_start );
+
+    /* this marker is present if the code knows we don't need __tls0 and __tls1... we device
```
lurch:
```suggestion
    /* this marker is present if the code knows we don't need __tls0 and __tls1... we decide
```

src/rp2_common/pico_thread_local/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+
+    # 1. Try to compile a snippet using TLS
+    set(TLS_CHECK_SOURCE "
+            __thread int test_var = 42;
+            int tls_check() {
+                return test_var;
+            }
+        ")
+
+    # Create a temporary file for the source
```
lurch, line 45: _Perhaps_ it's worth doing
```cmake
set(TLS_CHECK_SOURCEFILE "${CMAKE_CURRENT_BINARY_DIR}/tls_check.c")
```
and then e.g. line 79 might "simplify" to:
```cmake
file(REMOVE ${TLS_CHECK_SOURCEFILE} ${TLS_CHECK_LIBRARY})
```

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,122 @@
+/*
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright (c) 2019 Keith Packard
+ * Copyright (c) 2026 Raspberry Pi (Trading) Ltd.
+ */
+#ifndef _PICO_THREAD_LOCAL_PICOTLS_H
+#define _PICO_THREAD_LOCAL_PICOTLS_H
+
+// Note the name of this file is slightly confusing - this is actually overriding <picotls.h> from picolibc (if present)!
```
lurch, line 10: ...and presumably nothing to do with mbedtls? :rofl:

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,122 @@
...
+ * This library provides transparent runtime support for thread locals for the different types of thread local implementations used by the SDK supported compilers and C libraries.
+ *
+ * This means that by default, each core will get its own value for the thread local variable, and the core 1 value will be re-initialized if the core is restarted.
+ *
+ * Additionally, this library provides the "picolibc" style methods `_tls_size()`, `_init_tls()` and `_set_tls()` even when not using picolibc, which makes
+ * it easy to enable thread locals per task on at least FreeRTOS (just set `#define configUSE_PICOLIBC_TLS 1` in your `FreeRTOSConfig.h`).
+ *
+ * The pico_thread_local library comes in three main flavors:
+ *
+ * 1. `pico_thread_local_per_thread` (default) - Full support for thread local variables. Each thread/core gets its own copy of the variable initialized when it starts executing. The library tries to minimize the overhead if there are no thread locals are used, however particularly on RISC-V there is always a small overhead.
```
lurch:
```suggestion
 * 1. `pico_thread_local_per_thread` (default) - Full support for thread local variables. Each thread/core gets its own copy of the variable initialized when it starts executing. The library tries to minimize the overhead if there are no thread locals used, however particularly on RISC-V there is always a small overhead.
```

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,122 @@
...
+ * 2. `pico_thread_local_global` - Thread local variables are allowed in code, but each thread/core shares the same value (i.e. they aren't really thread local). There is very minimial overhead for this option.
+ * 3. `pico_thread_local_none` - No support for thread locals is provided. Code using them may not compile or link, and if it does the values won't be shared and may not even be initialized correctly. This mode is however guaranteed to have basically no overhead when thread locals are known not to be used.
+ *
+ * The support provided by the `pico_thread_local` library may be set from CMake via `pico_set_thread_local_implementation(<TARGET> per_thread|global|none)` and the default for all targets may be set via CMake variable (e.g. `set(PICO_DEFAULT_THREAD_LOCAL_IMPL pico_thread_local_none`).
```
lurch:
```suggestion
 * The support provided by the `pico_thread_local` library may be set from CMake via `pico_set_thread_local_implementation(<TARGET> per_thread|global|none)` and the default for all targets may be set via CMake variable (e.g. `set(PICO_DEFAULT_THREAD_LOCAL_IMPL pico_thread_local_none)`).
```

src/rp2_common/pico_thread_local/include/picotls.h
```diff
@@ -0,0 +1,122 @@
...
+
+// Rest of file is contingent on not being in PICO_THREAD_LOCAL_MODE_NONE...
+#if !PICO_THREAD_LOCAL_MODE_NONE
+#if !PICO_THREAD_LOCAL_SUPPORT_EMUTLS && !PICO_THREAD_LOCAL_SUPPORT_THREAD_POINTER
+// PICO_THREAD_LOCAL_EMUTLS_CONFIRMED, PICO_THREAD_LOCAL_THREAD_POINTER_ARM_AEABI_CONFIRMED PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED are
+// optional defines provided by the build that allow us to improve code size slightly if only one is set
+
+#if PICO_THREAD_LOCAL_EMUTLS_CONFIRMED + PICO_THREAD_LOCAL_THREAD_POINTER_ARM_AEABI_CONFIRMED + PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED > 1
+#error expected at most one of PICO_THREAD_LOCAL_EMUTLS_CONFIRMED, PICO_THREAD_LOCAL_THREAD_POINTER_ARM_AEABI_CONFIRMED and PICO_THREAD_LOCAL_THREAD_POINTER_RISCV_REG_CONFIRMED to be set
+#endif
```
lurch, line 58: Cunning. Maybe it's worth doing something similar to check that only one of `PICO_THREAD_LOCAL_MODE_PER_THREAD`, `PICO_THREAD_LOCAL_MODE_GLOBAL` or `PICO_THREAD_LOCAL_MODE_NONE` is set, rather than the the multiple nested `#if`s that are being used currently?

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+#endif
+#if PICO_THREAD_LOCAL_MODE_NONE
+#error PICO_THREAD_LOCAL_MODE_PER_THREAD and PICO_THREAD_LOCAL_MODE_NONE are both specified
+#endif
+// ------------------------------------------------------------
+// Proper TLS support per thread (PICO_THREAD_LOCAL_MODE_PER_THREAD = 1)
+// -------------------------------------------------------------
+
+#if !PICO_THREAD_LOCAL_THREAD_POINTER_VIA_RISCV_REG
+// we don't have TP register so need to track one tls region per code
```
lurch:
```suggestion
// we don't have TP register so need to track one tls region per core
```

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+// From emutls.c:
+// 'For every TLS variable xyz, there is one __emutls_control variable named __emutls_v.xyz. If xyz has
+// non-zero initial value, __emutls_v.xyz's "value" will point to __emutls_t.xyz, which has the initial value.'
+//
+// The linker script groups all the __emutls_v.xyz variables into a single array and provides symbols
+// __emutls_array_start and __emutls_array_end, which can be used to iterate over the array. This allows
+// the storage for each core's thread local variables to be pre-allocated and pre-initialized, which leaves
+// minimal work for __wrap___emutls_get_address.
+//
+// This array is available to other TLS implementations too, such a TLS implementation for an RTOS.
```
lurch:
```suggestion
// This array is available to other TLS implementations too, such as a TLS implementation for an RTOS.
```
?

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+// This is called lazily (either as a result of __emutls_get_address
+// or someone (perhaps RTOS) calling _runtime_tls_size first) to figure out
+// the size of the TLS area per thread
+static void _emutls_one_time_init(void) {
+    // Use spinlock so we don't add dependency on mutex code - we don't call anything
+    spin_lock_t *lock = spin_lock_instance(PICO_SPINLOCK_ID_HARDWARE_CLAIM);
+    uint32_t save = spin_lock_blocking(lock);
+    if (!emutls_one_time_init_done) {
+        // Three passes:
+        // 1) Calculate the offset of each thread local variable and the total storage to be allocated for each thread.
```
lurch: This seems to missing code-comments for passes 2 and 3? :raised_eyebrow:

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+            _emutls_size = (_emutls_size + tls_obj->align - 1) & ~(tls_obj->align - 1);
+            tls_obj->offset = _emutls_size + TLS_ADJUST;
+            _emutls_size += tls_obj->size;
+        }
+        emutls_one_time_init_done = true;
+    }
+    spin_unlock(lock, save);
+}
+
+// When we support EMUTLS we have _tls_size() redirect here (from our replacement <picolibc.h>)
```
lurch: Was this supposed to say `replacement <picotls.h>` ? :shrug:

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+}
+
+static void *_emutls_per_core_init(void) {
+    uint core_num = get_core_num();
+    assert(!_get_tls_adjusted_for_core(core_num));
+    // do this first, as _emutls_align may be set as a side effect
+    size_t size = _tls_size();
+    assert(size);
+    if (size) {
+        // aligned_alloc is not available in all libraries wew support, and it isn't thread safe anyway,
```
lurch:
```suggestion
        // aligned_alloc is not available in all libraries we support, and it isn't thread safe anyway,
```

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+        // so we'll just do the padded malloc
+        // tls = aligned_alloc(size, _emutls_align);
+        void *tls = malloc(size + _emutls_align - 1);
+        // note we never free the memory, so bumping the pointer is fine
+        tls = (void *)((((uintptr_t)tls) + (_emutls_align - 1)) & ~(_emutls_align - 1));
```
lurch, line 240: Perhaps?
```suggestion
        uint32_t align_minus_1 = _emutls_align - 1;
        void *tls = malloc(size + align_minus_1);
        // note we never free the memory, so bumping the pointer is fine
        tls = (void *)((((uintptr_t)tls) + align_minus_1) & ~align_minus_1);
```
Or is that the kind of optimization that the compiler will do automatically?

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+        tls_adjusted = _emutls_per_core_init();
+    }
+    assert(tls_adjusted);
+    return tls_adjusted + ((tls_object_t *)obj)->offset;
+}
+#endif
+
+#if PICO_THREAD_LOCAL_SUPPORT_THREAD_POINTER && PICO_THREAD_LOCAL_THREAD_POINTER_VIA_RISCV_REG
+// on RISC-V we must set up the pointer each time, note we don't actually respect PICO_THREAD_LOCAL_CORE1_REINITIALIZE
+// on RISC-V as it is an optimization flag not a bejavioral flag (i.e. it is indended to be set to 0
```
lurch:
```suggestion
// on RISC-V as it is an optimization flag not a behavioral flag (i.e. it is intended to be set to 0
```

src/rp2_common/pico_thread_local/thread_local.c
```diff
@@ -0,0 +1,338 @@
...
+#endif
+#endif
+#endif
+
+#if PICO_THREAD_LOCAL_PROVIDE_INIT_TLS
+void _init_tls(void *tls) {
+    // we expect an impl in this case
+    _INIT_TLS_IMPL(tls);
+}
+#endif
```
lurch, line 317: Should this `#error` if `_INIT_TLS_IMPL` isn't defined?
(and obviously same question for `_SET_TLS_IMPL` below)

test/pico_thread_local_test/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+if (TARGET pico_multicore)
+    # Note the global versions don't actually need multicore, but it didn't seem worth adding #ifdef to them all
+    add_executable(pico_thread_local_test pico_thread_local_test.c)
+    target_link_libraries(pico_thread_local_test PRIVATE pico_stdlib pico_multicore)
+    pico_add_extra_outputs(pico_thread_local_test)
+
+    add_executable(pico_thread_local_test_global pico_thread_local_test.c)
+    target_link_libraries(pico_thread_local_test_global PRIVATE pico_stdlib pico_multicore)
+    pico_add_extra_outputs(pico_thread_local_test_global)
+    pico_set_thread_local_implementation(pico_thread_local_test_global global)
```
lurch, line 10: Should this also be testing the "none" implementation?
EDIT: Doh, forget I said that :wink:  "Code using them may not compile or link"

test/pico_thread_local_test/pico_thread_local_test.c
```diff
@@ -0,0 +1,97 @@
+/**
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
```
lurch: 2026?

test/pico_thread_local_test/pico_thread_local_test.c
```diff
@@ -0,0 +1,97 @@
...
+#include <string.h>
+#include <hardware/sync.h>
+#include "pico/stdlib.h"
+#include "pico/multicore.h"
+#include "pico/sem.h"
+
+#include <picotls.h>
+#define COUNT 10000
+
+volatile __thread int counter = 7;
```
lurch: Maybe it's worth doing
```c
#define COUNTER_START 7
```
and then all the `7`s in this file could be replaced with `COUNTER_START`?

test/pico_thread_local_test/pico_thread_local_test.c
```diff
@@ -0,0 +1,97 @@
...
+    //printf("AH have %d\n", _tls_size());
+    sem_init(&sem, 0, 1);
+    multicore_launch_core1(core1_entry);
+    int core0_count = do_count(1);
+    sem_acquire_blocking(&sem);
+    printf("Core 0: %d (expected %d)\n", core0_count, COUNT + 7);
+    printf("Core 1: %d (expected %d)\n", core1_count, COUNT*2 + 7);
+    if (core0_count != COUNT + 7 || core1_count != COUNT*2 + 7) {
+        return false;
+        return -1;
```
lurch: Huh?

tools/extract_configs.py
```diff
@@ -303,8 +303,10 @@ def ValidateAttrs(config_name, config_attrs, file_path, linenum, applicable):
                     first_define_value = get_first_dict_key(defines_obj)
                     first_define_file_path, first_define_linenum = defines_obj[first_define_value]
                     errors.append(Exception('Found {} at {}:{} with a default of {}, but #define says {} (at {}:{})'.format(config_name, file_path, linenum, config_default, first_define_value, first_define_file_path, first_define_linenum)))
-            else:
+            elif config_default != "0":
                 errors.append(Exception('Found {} at {}:{} with a default of {}, but no matching #define found'.format(config_name, file_path, linenum, config_default)))
+            else:
+                logger.info('Found {} at {}:{} with a default of {}, but no matching #define found'.format(config_name, file_path, linenum, config_default))
```
lurch: I think I'd be happier with this "a missing default #define is allowed if the default is 0" if it was restricted to only `bool` types? i.e.
```suggestion
            elif config_obj['attrs']['type'] == "bool" and config_default == "0":
                # a bool with a missing #define defaults to 0 (false) anyway
                logger.info('Found {} (bool) at {}:{} with a default of {}, but no matching #define found'.format(config_name, file_path, linenum, config_default))
            else:
                errors.append(Exception('Found {} at {}:{} with a default of {}, but no matching #define found'.format(config_name, file_path, linenum, config_default)))
```
(I've also swapped things around so that the Exception is in the final `else:` clause, which is how it was structured before)

kilograham: yes, i had already though to restrict it to bools

src/rp2_common/pico_thread_local/include/picotls.h
kilograham: correct - hence tls -> thread_local_storage where we have control

src/rp2_common/pico_thread_local/CMakeLists.txt
kilograham: possibly, but i dont care enough ;-)

src/rp2_common/pico_thread_local/thread_local.c
kilograham: yes

src/rp2_common/pico_thread_local/thread_local.c
kilograham: yes the compiler is quite capable

src/rp2_common/pico_thread_local/thread_local.c
kilograham: it will error anyway, i dont think there is a need for a nicer error

src/rp2_common/pico_thread_local/thread_local.c
kilograham: oops - comment is wrong

src/rp2_common/pico_thread_local/CMakeLists.txt
lurch: Haha, fair enough. In that case, do you need to ensure that line 79 is properly quoting the paths?

test/pico_thread_local_test/pico_thread_local_test.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+#if PICO_THREAD_LOCAL_MODE_PER_THREAD
+    //printf("AH have %d\n", _tls_size());
+    sem_init(&sem, 0, 1);
+    multicore_launch_core1(core1_entry);
+    int core0_count = do_count(1);
+    sem_acquire_blocking(&sem);
+    printf("Core 0: %d (expected %d)\n", core0_count, COUNT + COUNTER_INIT_VALUE);
+    printf("Core 1: %d (expected %d)\n", core1_count, COUNT*2 + COUNTER_INIT_VALUE);
+    if (core0_count != COUNT + 7 || core1_count != COUNT*2 + COUNTER_INIT_VALUE) {
```
lurch:
```suggestion
    if (core0_count != COUNT + COUNTER_INIT_VALUE || core1_count != COUNT*2 + COUNTER_INIT_VALUE) {
```

src/rp2_common/pico_thread_local/CMakeLists.txt
kilograham: ok, on reflection, i do feel like it!

will-v-pi:
This has broken the `picotool` test build of pico-examples using Clang (https://github.com/raspberrypi/picotool/actions/runs/25926790767/job/76210082546?pr=314) with the error:
```
src/rp2_common/pico_thread_local/thread_local.c:84:20: fatal error: unused function '_tls_init_from_emutls_or_tdata' [-Wunused-function]
```
when building the `tinyusb_dev_audio_4_channel_mic` example. This is using:
```
LLVM 19.1.1 from https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/releases/download/release-19.1.1/LLVM-ET-Arm-19.1.1-Linux-x86_64.tar.xz
```

PR #2913, opened 2026-04-28
make kitchen_sink include all hardware struct headers

kilograham:


lurch:
Hmmm, is it worth modifying https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/cmake.yml to explicitly build for both RP2040 and RP2350? (as I believe GitHub CI currently won't be testing the RP2350-specific headers?)

PR #2915, opened 2026-04-29
add __SOFTFP__ via build for GCC 6/7 when using softfp as those versions don't and test/cmsis code rely on it

kilograham:


kilograham:
note not done in `pico_arm_gcc_common.cmake` as CMake doesn't know the compiler version until later

will-v-pi (approved):
This fixes the GCC 6 & 7 build for normal Picos, but the GCC 6 & 7 build is still broken for Pico W https://github.com/raspberrypi/pico-sdk/actions/runs/25208235535 - it's failing in the cyw43-driver submodule with:
```
pico-sdk/lib/cyw43-driver/firmware/cyw43_btfw_43439.h:588:25: error: initializer element is not constant
 const size_t btfw_len = cyw43_btfw_43439_len;
```

As that's in a submodule, I guess that'll require a separate change, so approving this but noting that Multi GCC is still not passing

lurch:
> This fixes the GCC 6 & 7 build for normal Picos, but the GCC 6 & 7 build is still broken for Pico W https://github.com/raspberrypi/pico-sdk/actions/runs/25208235535 - it's failing in the cyw43-driver submodule with:
> 
> ```
> pico-sdk/lib/cyw43-driver/firmware/cyw43_btfw_43439.h:588:25: error: initializer element is not constant
>  const size_t btfw_len = cyw43_btfw_43439_len;
> ```
> 
> As that's in a submodule, I guess that'll require a separate change, so approving this but noting that Multi GCC is still not passing

That's being tracked by https://github.com/georgerobotics/cyw43-driver/issues/139

Issue #2372, opened 2025-03-25
PSRAM: SDK support for bringing up PSRAM


Gadgetoid:
Between us (Pimoroni), SparkFun and Adafruit we've been pushing for PSRAM support in MicroPython (or CircuitPython in Adafruit's case but potato pota..to), since the extra RAM benefits are huge.

See: https://github.com/micropython/micropython/pull/15620

However as part of this there's some bringup that (as mentioned also by MicroPython's maintainers) feels more like Pico SDK's responsibility.

This includes detecting and getting the size of a PSRAM chip:

https://github.com/micropython/micropython/blob/293d17e4463142588ce4fd45f82b0bd628735676/ports/rp2/rp2_psram.c#L9-L78

And setting up PSRAM for general use (it then gets used to store additional heap for MicroPython):

https://github.com/micropython/micropython/blob/293d17e4463142588ce4fd45f82b0bd628735676/ports/rp2/rp2_psram.c#L80-L167

This is all rather spicy code that could benefit from an official canonical solution for bringup and the additional test coverage and exercising that SDK inclusion would imply.

Tighter integration into the SDK could also do things like assert when attempting to write to flash from a PSRAM resident address (bad times) or even attempt to handle cases like this transparently.

All three of us (to the best of my knowledge) have shipping products using this (or similar) PSRAM code in either MicroPython or CircuitPython and seem to have ironed out any quirks.

Are there any plans for this? If not, could it be considered? Thank you!

kilograham:
Seems reasonable, putting in 2.2.0 for now

sfe-SparkFro:
Also worth adding, the Arduino Pico port also has its own PSRAM implementation: https://github.com/earlephilhower/arduino-pico/blob/ecf2b2e39f04d343e13be9151b67232e6e3643a8/cores/rp2040/psram.cpp with a test example here: https://github.com/earlephilhower/arduino-pico/blob/ecf2b2e39f04d343e13be9151b67232e6e3643a8/libraries/rp2350/examples/PSRAMTest/PSRAMTest.ino

Totally agree on having an implementation in the Pico SDK that everyone else can leverage. We at SparkFun have put PSRAM on basically every RP2350 board we're selling.

byzantic:
Can I *strongly* advocate for this addition please? I have several boards that use PSRAM, and I'm currently blocked from using them with the SDK because of this issue. 

[...]

PR #2919, opened 2026-05-01
Add PSRAM support

will-v-pi:
This adds a `hardware_psram` library to the SDK to add PSRAM support, as requested in https://github.com/raspberrypi/pico-sdk/issues/2372

Adds PSRAM defines to board headers for boards that have PSRAM - I may not have captured them all

Has three ways it will detect PSRAM when you link `hardware_psram`, in order of priority:
1. If flash_devinfo is setup (eg configured in OTP), it will initialise PSRAM with the flash_devinfo CS1 size and GPIO
2. If `PICO_AUTO_DETECT_PSRAM` is set it will attempt to detect PSRAM size and GPIO on CS1
3. If `PICO_PSRAM_SIZE_BYTES` is set (eg configured in board header, or with `pico_override_psram_size`) it will initialise PSRAM with that size and GPIO

Variables can be placed in PSRAM using `__psram` or `__psram_uninitialised` macros, and you can also write directly to the memory addresses.

If there are variables placed in PSRAM, it will setup ATRANS to prevent any access to PSRAM addresses greater than the size available (from flash_devinfo, auto-detection, or `PICO_PSRAM_SIZE_BYTES`), which will cause bus faults. The `psram_is_available` function should be used before accessing variables in PSRAM when auto-detection is on, to prevent these bus faults.

This support uses flash_devinfo to enable the bootrom support for PSRAM, then adds a callback with `flash_set_qmi_cs1_setup_function` which runs as part of `flash_rp2350_restore_qmi_cs1` to enable quad mode and setup faster timings. The RP2350-E14 workaround is included.

The CS pin auto-detection iterates through all possible CS pins, skipping any that are defined as defaults in board header files with `remove_defaults_from_cs_gpios` - this is an ugly function, but without it you get some garbled characters on the UART as pin 0 is a possible CS pin. Maybe some of the defaults could be removed from this, if they are unlikely to be affected by this auto-detection?

This has `psram_or_malloc` and `psram_or_free` macros, which statically allocate in PSRAM and use that if available, but use `malloc` if not available - not sure how useful these are, could be removed if we don't want them. They were to demostrate `psram_check_address` which checks if an address is in available PSRAM, so you can avoid busfaults when using variables in PSRAM.

In passing this also:
- Removes some Bazel TODOs - CMake gates RP2350 only stuff in src/cmake/rp2_common.cmake
- Expands RP2350 boot2 linker script by 4 bytes, as there is no checksum


src/rp2040/pico_platform/BUILD.bazel
```diff
@@ -99,6 +99,7 @@ cc_library(
         "//src/rp2040/pico_platform/script_include:rp2040_linker_scripts",
         "//src/rp2_common/pico_standard_link/script_include:rp2_linker_scripts",
         "//src/rp2_common/pico_standard_link:default_flash_region",
+        "//src/rp2_common/pico_standard_link:default_psram_region",
```
lurch: Does it make sense to include this into an RP2040-specific file?

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -142,12 +142,36 @@ void flash_get_unique_id(uint8_t *id_out);
  *  \param txbuf Pointer to a byte buffer which will be transmitted to the flash
  *  \param rxbuf Pointer to a byte buffer where data received from the flash will be written. txbuf and rxbuf may be the same buffer.
  *  \param count Length in bytes of txbuf and of rxbuf
+ *  \param cs Chip select index: generally 0 is Flash, 1 is PSRAM.
```
lurch: Should we document that the `cs` parameter is ignored  on RP2040?

lurch: Is it worth adding one of the `invalid_params_if` checks on RP2350 to check that `cs < 2` ?

lurch: Perhaps the earlier "Low-level function to execute a serial command on a flash device attached to the QSPI interface." wording needs to be updated to say something like "flash or PSRAM device" ? :shrug:

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -142,12 +142,36 @@ void flash_get_unique_id(uint8_t *id_out);
...
 void flash_flush_cache(void);
 
 #if !PICO_RP2040
+
+typedef void (*qmi_setup_function_t)(void);
+/*! \brief Set the function to be called to setup the QMI CS1 configuration
+ *  \ingroup hardware_flash
+ *
+ * \param function The function to be called to setup the QMI CS1 configuration
+ * \return True if the function was set, false if not (eg tried to set a function in flash)
```
lurch:
```suggestion
 * \return true if the function was set, false if not (e.g. tried to set a function in flash)
```

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+#ifndef PICO_AUTO_DETECT_PSRAM
+#define PICO_AUTO_DETECT_PSRAM 0
+#endif
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_SIZE, automatically detect psram size, type=bool, default=PICO_AUTO_DETECT_PSRAM, group=hardware_psram
+#ifndef PICO_AUTO_DETECT_PSRAM_SIZE
+#define PICO_AUTO_DETECT_PSRAM_SIZE PICO_AUTO_DETECT_PSRAM
+#endif
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS, automatically detect psram chip select pin, default=PICO_AUTO_DETECT_PSRAM, group=hardware_psram
```
lurch: `type=bool` ?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+#error "PICO_AUTO_DETECT_PSRAM_SIZE must be set to use PICO_AUTO_DETECT_PSRAM_CS"
+#endif
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS, skip default GPIOs when auto-detecting psram chip select pin, type=bool, default=PICO_AUTO_DETECT_PSRAM_CS, group=hardware_psram
+#ifndef PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS
+#define PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS PICO_AUTO_DETECT_PSRAM_CS
+#endif
+
+#if PICO_AUTO_DETECT_PSRAM_CS
+#if PICO_RP2350
```
lurch: Isn't the entirety of `hardware_psram` unusable if we're not running on RP2350?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+#endif
+
+// PICO_CONFIG: PICO_DEFAULT_PSRAM_ID, Default ID of psram used for auto-detection, type=int, default=0x5D, group=hardware_psram
+#ifndef PICO_DEFAULT_PSRAM_ID
+#define PICO_DEFAULT_PSRAM_ID 0x5D
+#endif
+
+// PICO_CONFIG: PICO_DEFAULT_PSRAM_MAX_FREQ, Default max frequency of psram, type=int, default=133000000, group=hardware_psram
+#ifndef PICO_DEFAULT_PSRAM_MAX_FREQ
+#define PICO_DEFAULT_PSRAM_MAX_FREQ 133000000
```
lurch: I guess this might be more readable as `133 * MHZ`? :wink:  Or would that introduce an unwanted header dependency?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+#endif
+
+// PICO_CONFIG: PICO_DEFAULT_PSRAM_MAX_SELECT, Default max select time in ns of psram, type=int, default=8000, group=hardware_psram
+#ifndef PICO_DEFAULT_PSRAM_MAX_SELECT
+#define PICO_DEFAULT_PSRAM_MAX_SELECT 8000
+#endif
+
+// PICO_CONFIG: PICO_DEFAULT_PSRAM_MIN_DESELECT, Default min deselect time in ns of psram, type=int, default=18, group=hardware_psram
```
lurch: IMHO "time in ns of psram" here should be "time of psram in ns" ?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+#ifndef PICO_DEFAULT_PSRAM_MIN_DESELECT
+#define PICO_DEFAULT_PSRAM_MIN_DESELECT 18
+#endif
+
+
+// PICO_CONFIG: PICO_RUNTIME_SKIP_INIT_PSRAM, Skip calling of `runtime_init_psram` function during runtime init, type=bool, default=1 if PICO_PSRAM_SIZE_BYTES is not set, group=pico_runtime_init
+// PICO_CONFIG: PICO_RUNTIME_NO_INIT_PSRAM, Do not include SDK implementation of `runtime_init_psram` function, type=bool, default=1 if PICO_PSRAM_SIZE_BYTES is not set, group=pico_runtime_init
```
lurch: The defaults described here don't seem to match the `#define`s below?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*! \brief Check if PSRAM is initialised
+ *  \ingroup hardware_psram
+ *
+ * \return true if PSRAM is initialised, false otherwise
+ */
+bool psram_is_available(void);
```
lurch: Should the description and the function both use "initialised" or "available", rather than mixing the two terms?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+ */
+bool psram_check_address(void* addr);
+
+/*! \brief Detect PSRAM size
+ *  \ingroup hardware_psram
+ *
+ * This will read the ID of the PSRAM chip and return the size based on the ID.
+ *
+ * You must configure the GPIO function for the CS pin before calling this function,
+ * and should also configure the GPIO in flash_devinfo to prevent toggling of the
```
lurch:
```suggestion
 * and should also configure the CS GPIO in flash_devinfo to prevent toggling of the
```
?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+ * previously configured GPIO (usually 0, so prints invalid characters to UART).
```
lurch:
```suggestion
 * previously configured GPIO (usually 0, so prints invalid characters to default UART).
```
?

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: Yes, but the default CS pins need to be defined for each platform

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+
+/*! \brief Re-initialise PSRAM
+ *  \ingroup hardware_psram
+ *
+ * This will re-initialise the PSRAM with the parameters set by \ref psram_configure_params.
+ *
+ * This calls \ref flash_start_xip internally, so will reset any QSPI pads changes you have made.
+ *
+ * \return PICO_OK on success, PICO_ERROR_PRECONDITION_NOT_MET if the PSRAM size is not set in
+ * flash_devinfo or the PSRAM parameters are not set by \ref psram_configure_params
```
lurch:
```suggestion
 * flash_devinfo or the PSRAM parameters are not set by \ref psram_configure_params or \ref psram_set_params
```

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+ * \param kgd Known Good Die
+ * \param eid EID
+ * \return size of PSRAM in bytes, or 0 if the KGD/EID is not recognised
+ */
+size_t psram_eid_to_size(uint8_t kgd, uint8_t eid);
+
+/*! \brief Provide a static PSRAM allocation, or malloc if PSRAM is not available
+ *  \ingroup hardware_psram
+ *
+ * This will allocate a static buffer in PSRAM and if available use that, otherwise it will use the heap.
```
lurch, line 244:
```suggestion
 * This will allocate a static buffer in PSRAM if available and use that, otherwise it will use the heap.
```

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,243 @@
...
+ * 
+ * This will fail to compile if PICO_PSRAM_SIZE_BYTES is not set
+ */
+#define psram_or_malloc(group, type, var, size) static type __psram_uninitialised(group) var##_psram[size]; type* var; if (psram_check_address(var##_psram + size)) { var = (type*)var##_psram; } else { var = (type*)malloc(size * sizeof(type)); }
```
lurch: Should there be a bunch of `\param` comments here?

will-v-pi: Yes, but I'm leaving fixing up this API until after it's decided if it's wanted, so it's easy to back out the one commit - I'm not sure how much utility it has

src/rp2040/pico_platform/BUILD.bazel
will-v-pi: I've added some more RP2040 incls, so this has been removed

src/rp2_common/hardware_flash/include/hardware/flash.h
will-v-pi: > Perhaps the earlier "Low-level function to execute a serial command on a flash device attached to the QSPI interface." wording needs to be updated to say something like "flash or PSRAM device" ? 🤷

I'll removel the reference to PSRAM, and remove the word "flash" from the top paragraph so it's just "device" and "chip"

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: Have gone for "available and initialised" in the docs, and left the function name the same

src/rp2_common/hardware_flash/flash.c
```diff
@@ -262,14 +275,15 @@ static void __no_inline_not_in_flash_func(flash_cs_force)(bool high) {
...
     } else {
-        hw_set_bits(&qmi_hw->direct_csr, QMI_DIRECT_CSR_ASSERT_CS0N_BITS);
+        hw_set_bits(&qmi_hw->direct_csr, cs == 0 ? QMI_DIRECT_CSR_ASSERT_CS0N_BITS : QMI_DIRECT_CSR_ASSERT_CS1N_BITS);
     }
 #endif
 }
 
-void __no_inline_not_in_flash_func(flash_do_cmd)(const uint8_t *txbuf, uint8_t *rxbuf, size_t count) {
+void __no_inline_not_in_flash_func(flash_do_cmd_cs)(const uint8_t *txbuf, uint8_t *rxbuf, size_t count, uint8_t cs) {
+    invalid_params_if(HARDWARE_FLASH, cs > 1);
```
lurch: If the doxygen for `flash_do_cmd_cs` is explicitly saying "On RP2040 the chip select index is ignored, as there is only one chip select." then should this `invalid_params_if` be wrapped in a `#if !PICO_RP2040` ?
Also, should the same `invalid_params_if` be added to `flash_cs_force` ?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,244 @@
...
+
+/** \file psram.h
+ *  \defgroup pico_psram pico_psram
+ *
+ * \brief Low level PSRAM setup functions
+ *
+ * Note some of these functions are *unsafe* if you are using both cores, and the other
+ * is executing from flash or psram concurrently with the operation. In this case, you
+ * must perform your own synchronisation to make sure that no XIP accesses take
+ * place during flash programming. One option is to use the
```
lurch: Is what is written here about "during flash programming" also true for "during psram writing"? Or are Flash and PSRAM sufficiently different that less precautions are needed with psram?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,244 @@
...
+#endif
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS, skip default GPIOs when auto-detecting psram chip select pin, type=bool, default=PICO_AUTO_DETECT_PSRAM_CS, group=hardware_psram
+#ifndef PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS
+#define PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS PICO_AUTO_DETECT_PSRAM_CS
+#endif
+
+#if PICO_AUTO_DETECT_PSRAM_CS
+#if PICO_RP2350
+#define PICO_AVAILABLE_CS1_GPIOS {0, 8, 19, 47}
```
lurch: Should this be made conditional on RP2350A versus RP2350B? (as GPIO 47 obviously only exists on RP2350B)

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,244 @@
...
+ *  \ingroup hardware_psram
+ *
+ * \return true if the address is in available PSRAM, false otherwise
+ */
+bool psram_check_address(void* addr);
+
+/*! \brief Detect PSRAM size
+ *  \ingroup hardware_psram
+ *
+ * This will read the ID of the PSRAM chip and return the size based on the ID.
```
lurch, line 160:
```suggestion
 * This will read the EID of the PSRAM chip and return the size based on the EID.
```
(this is just a hunch based on the doxygen for `psram_eid_to_size` :wink: )

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -101,6 +101,30 @@ if (NOT TARGET pico_standard_link)
         set_property(TARGET ${TARGET} APPEND PROPERTY PICO_TARGET_LINKER_SCRIPT_OVERRIDE_PATHS "${PATH}")
     endfunction()
 
+    # pico_override_flash_size(TARGET FLASH_SIZE)
+    # \brief\ Override the FLASH size for a given target
+    #
+    # \param\ TARGET The target to override the FLASH size for
+    # \param\ FLASH_SIZE The FLASH size to set for the target
+    function(pico_override_flash_size TARGET FLASH_SIZE)
```
lurch, line 109: It looks like this is only used/useful for testing, so perhaps it makes sense to just add it to `kitchen_sink` rather than `pico_standard_link` ? :shrug: 
(unintentional rhyme :joy: )

will-v-pi: It could be useful for creating universal binaries, if you wanted to force a smaller flash size to ensure your code always fits in that? Or if you wanted to build for multiple flash sizes in a single build directory?

I mostly added it because I was adding the PSRAM one and it was simple to add flash too.

lurch: I'm showing my lack of CMake knowledge here, but do we _need_ a specific override function for this, given that the board-header files all seem to follow the
```c
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (16 * 1024 * 1024)
#endif
```
pattern? Or are the override functions only needed because of the way that these values interact with the linker-scripts?

----

Hmmm, although that also raises an interesting side-question... are there some `#define`s that `check_board_header.py` ought to error about if they _aren't_ guarded by an `#ifndef` clause? Because whilst it checks that the `#ifndef` matches the enclosed `#define`, so that this would error:
```c
#ifndef PICO_FLASH_SIZE_BYTE
#define PICO_FLASH_SIZE_BYTES (16 * 1024 * 1024)
#endif
```
currently both
```c
pico_board_cmake_set_default(PICO_FLASH_SIZE_BYTES, (16 * 1024 * 1024))
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (16 * 1024 * 1024)
#endif
```
and
```c
pico_board_cmake_set_default(PICO_FLASH_SIZE_BYTES, (16 * 1024 * 1024))
#define PICO_FLASH_SIZE_BYTES (16 * 1024 * 1024)
```
are considered equally valid. (And presumably the latter version wouldn't work properly with your `pico_override_flash_size` function? :thinking: )

will-v-pi: It's needed to override the `pico_flash_region.ld` file, which sets the flash size in the linker script - the linker doesn't use `PICO_FLASH_SIZE_BYTES`.

I think that second version without `#ifndef` should be marked as invalid - definitely for any define with a `pico_board_cmake_set_default`, but more generally anything that wants to be overridable. Anything that is definitely fixed can just be a `#define` (eg the boot2 choice, rp2350 variant), but everything else should be overridable (eg I want to use an external LED because my on board one is broken). But that can probably come in a separate PR to develop, with specific exclusions for the defines which don't need to be `#ifndef`ed.

lurch: > But that can probably come in a separate PR to develop, with specific exclusions for the defines which don't need to be #ifndefed.

Yup, that's what I was planning to do, rather than polluting #2924 (and apologies for the off-topic questions in this PR).

EDIT: This `#ifndef` checking has been done in #2927

src/rp2_common/hardware_flash/flash.c
will-v-pi: I'll move it into `flash_cs_force` as that handles both of these issues

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: It's a typo from copying - will use "while running these functions" instead

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: Annoyingly that ruins the utility of this for me, as I'm compiling for RP2350A but running on an RP2350B board with PSRAM on CS 47 - but looking at the code, using GPIO 47 on RP2350A will trigger an assertion when `PARAM_ASSERTIONS_ENABLED_HARDWARE_GPIO` is enabled, so you are correct that this needs to be conditional.

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: EID is part of the ID - it also uses the KGD part, and returns 0 if that doesn't match what is expected, so I will leave this description as-is

src/rp2_common/hardware_flash/include/hardware/flash.h
```diff
@@ -138,16 +137,44 @@ void flash_get_unique_id(uint8_t *id_out);
  * it is recommended that this function only be used to extract flash metadata
  * during startup, before the main application begins to run: see the
  * implementation of pico_get_unique_id() for an example of this.
+ * 
+ * \if rp2040_specific
+ * On RP2040 the chip select index is ignored, as there is only one chip select.
+ * \endif
  *
  *  \param txbuf Pointer to a byte buffer which will be transmitted to the flash
  *  \param rxbuf Pointer to a byte buffer where data received from the flash will be written. txbuf and rxbuf may be the same buffer.
```
lurch: Should these also be updated to say "device" rather than "flash"?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,248 @@
...
+ * vector table in flash or psram, so you must disable interrupts before calling in
+ * this case.
+ *
+ * The unsafe functions are:
+ * - \ref psram_reinitialise
+ * - \ref psram_detect_cs_and_size
+ * - \ref psram_detect_size
+ *
+ * \subsection psram_example Example
+ * \include psram_program.c
```
lurch: Presumably there'll be a separate PR adding this program to `pico-examples` ?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * \return size of PSRAM, or 0 if none found
+ */
+size_t psram_detect_size(void);
+
+/*! \brief Detect PSRAM chip select pin and size
+ *  \ingroup hardware_psram
+ *
+ * This runs \ref psram_detect_size() for each CS GPIO in the array in turn,
+ * and returns the size as soon as a PSRAM chip is detected.
```
lurch, line 174: This may be a stupid question, but would it be useful for the code calling this function to be able to determine _which_ of the CS GPIOs got selected?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,410 @@
...
+        psram_size = 1024 * 1024; // 1 MiB
+        uint8_t size_id = eid >> 5;
+        if (size_id == 4) { // == 4 is for ISSI PSRAM
+            psram_size *= 16; // 16 MiB
+        } else if (eid == 0x26 || size_id == 2 || size_id == 3) { // == 3 is for ISSI PSRAM
+            psram_size *= 8; // 8 MiB
+        } else if (size_id == 0) {
+            psram_size *= 2; // 2 MiB
+        } else if (size_id == 1) {
+            psram_size *= 4; // 4 MiB
```
lurch: Looks like it might make sense to swap these?
```suggestion
        } else if (size_id == 1) {
            psram_size *= 4; // 4 MiB
        } else if (size_id == 0) {
            psram_size *= 2; // 2 MiB
```
That way the sizes in the comments would be in the order: 1 MiB, 16 MiB, 8 MiB, 4 MiB, 2 MiB

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: It can read the value from flash_devinfo with `flash_devinfo_get_cs_gpio(1)`

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,410 @@
...
+#define PSRAM_NOOP_CMD 0xFF
+
+
+size_t __weak psram_eid_to_size(uint8_t kgd, uint8_t eid) {
+    // Weak function to allow overriding if other PSRAM chips
+    // have different EID to size mapping
+    // Currently supports APS6404 and ISSI PSRAM
+    size_t psram_size = 0;
+
+    if (kgd == PICO_DEFAULT_PSRAM_ID) { // Known Good Die - expects 0x5D
```
lurch, line 31: I know nothing about PSRAM, but how likely is it that the code inside this `if` block would still be correct if somebody changes `PICO_DEFAULT_PSRAM_ID` to a value other than `0x5D` ?

will-v-pi: I have no idea, as the only chips I've seen so far have had the same KGD value - if it doesn't work they can just write their own `psram_eid_to_size` as it is a weak function

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,410 @@
...
+    flash_devinfo_set_cs_size(1, prev_size);
+
+    uint8_t kgd = rxbuffer[5];
+    uint8_t eid = rxbuffer[6];
+
+    return psram_eid_to_size(kgd, eid);
+}
+
+size_t psram_detect_cs_and_size(uint8_t *cs_gpios, size_t num) {
+    gpio_function_t prev_funcs[4] = {};
```
lurch: I guess this ought to check that `num <= 4` to prevent line 76 causing a buffer overflow?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,410 @@
...
+    #if PICO_RP2350_A2_SUPPORTED
+        // Workaround for RP2350-E14, where the bootrom only does this for GPIO 0 instead of the correct CS pin
+        hw_clear_bits(&pads_bank0_hw->io[gpio], PADS_BANK0_GPIO0_ISO_BITS);
+    #endif
+        size_t psram_size = psram_detect_size();
+        if (psram_size > 0) {
+            // CS GPIO found
+            return psram_size;
+        }
+        // Restore previous function to GPIO
```
lurch: Hmm, I guess if you pass in 3 GPIO pins to this function, and the PSRAM CS is on the second GPIO pin, then the immediate `return` above means that the GPIO function would be restored on the first GPIO, but **not** restored on the third GPIO?

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: I'll remove as this doesn't have examples yet - was copied from flash.h

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,248 @@
...
+#endif
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS, automatically detect psram chip select pin, type=bool, default=PICO_AUTO_DETECT_PSRAM, group=hardware_psram
+#ifndef PICO_AUTO_DETECT_PSRAM_CS
+#define PICO_AUTO_DETECT_PSRAM_CS PICO_AUTO_DETECT_PSRAM
+#elif PICO_AUTO_DETECT_PSRAM_CS && !PICO_AUTO_DETECT_PSRAM_SIZE
+#error "PICO_AUTO_DETECT_PSRAM_SIZE must be set to use PICO_AUTO_DETECT_PSRAM_CS"
+#endif
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS, skip default GPIOs when auto-detecting psram chip select pin, type=bool, default=PICO_AUTO_DETECT_PSRAM_CS, group=hardware_psram
```
lurch: Having read through the code I can now see what `PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS` is doing, but IMHO the "skip default GPIOs" description here is _far_ too vague.

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,410 @@
...
+        #ifdef CYW43_DEFAULT_PIN_WL_CLOCK
+            cs_gpios[i] == CYW43_DEFAULT_PIN_WL_CLOCK ||
+        #endif
+        #ifdef CYW43_DEFAULT_PIN_WL_CS
+            cs_gpios[i] == CYW43_DEFAULT_PIN_WL_CS ||
+        #endif
+            false // so the || works
+        ) {
+            // Replace with last valid GPIO in the array
+            cs_gpios[i] = cs_gpios[new_num-1];
```
lurch: Is it possible that this reorders the entries in the (trimmed down) `cs_gpios` array? Does that matter?

will-v-pi: Yes it reorders them, but I don't think that matters

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: Happy for a suggestion of better wording - I agree "skip default GPIOs" is too vague, it's skipping anything that is defined as a `PICO_DEFAULT_XXX_PIN` or a `CYW43_DEFAULT_PIN_XXX`, although that list can probably be trimmed down as I mentioned in the PR description

src/rp2_common/hardware_psram/psram.c
will-v-pi: Yes, I will fix that and restore all GPIOs (the bootrom will then put the chosen GPIO back to CS1 when configuring PSRAM)

src/rp2_common/hardware_psram/psram.c
will-v-pi: Actually, I can just use `gpio_function_t prev_funcs[num];` to avoid that issue

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * \return size of PSRAM, or 0 if none found
+ */
+size_t psram_detect_size(void);
+
+/*! \brief Detect PSRAM chip select pin and size
+ *  \ingroup hardware_psram
+ *
+ * This runs \ref psram_detect_size() for each CS GPIO in the array in turn,
+ * and returns the size as soon as a PSRAM chip is detected.
```
lurch, line 174: Perhaps this description needs updating now that the implementation has changed :wink: 
```suggestion
 * stopping as soon as a PSRAM chip is detected.
```

Also, maybe this function-description should warn that it will "wiggle" the CS GPIOs?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * This will setup the CS GPIO using flash_devinfo if PSRAM is found.
```
lurch, line 176: I think this might be clearer like:
```suggestion
 * If PSRAM is found, the CS GPIO will be configured in flash_devinfo.
```

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * \param cs_gpios Array of CS GPIOs to try
+ * \param num Number of CS GPIOs in the array
+ * \return size of PSRAM, or 0 if none found
+ */
+size_t psram_detect_cs_and_size(uint8_t *cs_gpios, size_t num);
```
lurch, line 182: Maybe this function should assert if `PICO_PSRAM_CS_PIN` has been set? :thinking:

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+    for (size_t i=0; i < num; i++) {
+        uint8_t gpio = cs_gpios[i];
+        flash_devinfo_set_cs_gpio(1, gpio);
+    #if PICO_RP2350_A2_SUPPORTED
+        // Workaround for RP2350-E14, where the bootrom only does this for GPIO 0 instead of the correct CS pin
+        hw_clear_bits(&pads_bank0_hw->io[gpio], PADS_BANK0_GPIO0_ISO_BITS);
+    #endif
+        psram_size = psram_detect_size();
+        if (psram_size > 0) {
+            // CS GPIO found, so will be left configured in flash_devinfo
```
lurch, line 85: If **no** PSRAM was found, should we also restore flash_devinfo to the original value?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+        uint8_t gpio = cs_gpios[i];
+        prev_funcs[i] = gpio_get_function(gpio);
+        gpio_set_function(gpio, GPIO_FUNC_NULL);
+    }
+    for (size_t i=0; i < num; i++) {
+        uint8_t gpio = cs_gpios[i];
+        flash_devinfo_set_cs_gpio(1, gpio);
+    #if PICO_RP2350_A2_SUPPORTED
+        // Workaround for RP2350-E14, where the bootrom only does this for GPIO 0 instead of the correct CS pin
+        hw_clear_bits(&pads_bank0_hw->io[gpio], PADS_BANK0_GPIO0_ISO_BITS);
```
lurch: Should this also be saved and then restored, in case external code is relying on the isolation behaviour?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+                        QMI_M1_WFMT_PREFIX_LEN_VALUE_8 << QMI_M1_WFMT_PREFIX_LEN_LSB |
+                        QMI_M1_WFMT_DUMMY_LEN_VALUE_NONE << QMI_M1_WFMT_DUMMY_LEN_LSB |
+                        QMI_M1_WFMT_SUFFIX_LEN_VALUE_NONE << QMI_M1_WFMT_SUFFIX_LEN_LSB);
+    qmi_hw->m[1].wcmd = PSRAM_QUAD_WRITE_CMD << QMI_M1_WCMD_PREFIX_LSB;
+
+    // Enable writes to PSRAM
+    hw_set_bits(&xip_ctrl_hw->ctrl, XIP_CTRL_WRITABLE_M1_BITS);
+}
+
+static bool has_psram = false;
```
lurch: Perhaps it makes sense to rename this to e.g. `psram_initialized` ?

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: If you enable auto-detection it ignores `PICO_PSRAM_CS_PIN`, so you can force auto-detection even if `PICO_PSRAM_CS_PIN` is defined - all of that is handled in `runtime_init_setup_psram`.

These individual functions should be callable regardless of the compile definitions, as you may want to run manual auto-detection, so they don't look at the `PICO_PSRAM_XXX` defines.

src/rp2_common/hardware_psram/psram.c
will-v-pi: We haven't set the size, so the bootrom will be ignoring the flash_devinfo CS1 GPIO anyway, so I think it's fine

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+
+size_t psram_get_size(void) {
+    if (!has_psram) {
+        return 0;
+    }
+    return flash_devinfo_size_to_bytes(flash_devinfo_get_cs_size(1));
+}
+
+bool psram_check_address(void* addr) {
+    uint32_t offset = (uint32_t)addr - (XIP_BASE + flash_devinfo_size_to_bytes(FLASH_DEVINFO_SIZE_MAX));
```
lurch, line 323: Could `addr` ever be _less_ than `(XIP_BASE + flash_devinfo_size_to_bytes(FLASH_DEVINFO_SIZE_MAX))` ?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+    }
+}
+
+#if !PICO_RUNTIME_NO_INIT_PSRAM
+void runtime_init_setup_psram(void) {
+    // Setup flash_devinfo from compile definitions if it is unset
+    #if defined(PICO_PSRAM_CS_PIN) && !PICO_AUTO_DETECT_PSRAM_CS
+    if (flash_devinfo_get_cs_size(1) == FLASH_DEVINFO_SIZE_NONE) {
+        flash_devinfo_set_cs_gpio(1, PICO_PSRAM_CS_PIN);
+    }
```
lurch: Why is setting the `PICO_PSRAM_CS_PIN` conditional on the value returned by `flash_devinfo_get_cs_size` ?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+        return true;
+    } else {
+        return false;
+    }
+}
+
+#if !PICO_RUNTIME_NO_INIT_PSRAM
+void runtime_init_setup_psram(void) {
+    // Setup flash_devinfo from compile definitions if it is unset
+    #if defined(PICO_PSRAM_CS_PIN) && !PICO_AUTO_DETECT_PSRAM_CS
```
lurch, line 349: Maybe `#if defined(PICO_PSRAM_CS_PIN) && PICO_AUTO_DETECT_PSRAM_CS` ought to be a compile-time error? Or is there a reason why that configuration would be useful?

Same question also for `#if defined(PICO_PSRAM_SIZE_BYTES) && PICO_AUTO_DETECT_PSRAM_SIZE`

src/rp2_common/hardware_psram/psram.c
will-v-pi: Actually, I think I should use `gpio_set_function(gpio, GPIO_FUNC_XIP_CS1);` here instead, rather than manually changing the isolation

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+
+    has_psram = flash_devinfo_get_cs_size(1) != FLASH_DEVINFO_SIZE_NONE;
+
+    static_assert(FLASH_DEVINFO_SIZE_MAX == FLASH_DEVINFO_SIZE_16M, "expected max region size of 16M");
+    extern uint32_t __psram_start__;
+    extern uint32_t __psram_end__;
+    uint32_t psram_words = (uint32_t)(&__psram_end__ - &__psram_start__);
+    if (psram_words > (flash_devinfo_size_to_bytes(flash_devinfo_get_cs_size(1)) >> 2)) { // >>2 is /4, for words
+        // Setup to bus fault for variables that don't fit in available PSRAM
+        flash_devinfo_size_t psram_size = flash_devinfo_get_cs_size(1);
```
lurch: `flash_devinfo_get_cs_size(1)` appears in this function _multiple_ times - is it worth caching the value into a local variable?

src/rp2_common/hardware_psram/psram.c
will-v-pi: Yes, if you're passing it a flash or bootrom address - but in that case it will still return false as `offset` will be very large

src/rp2_common/hardware_psram/psram.c
will-v-pi: So that if flash_devinfo is already configured by OTP that is always used - there isn't an equivalent NONE value for the cs_gpio as 0 is valid, so checking the cs_size is the correct thing to do

src/rp2_common/hardware_psram/psram.c
will-v-pi: Could be useful if you're compiling for a board with PSRAM, but want to enable auto-detection to work on other boards with different CS pins, or boards without PSRAM - enabling auto-detection overrides whatever is defined in the board header, see my priority list in the PR description

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,414 @@
...
+        int clear_start = 8; // Clear no regions by default
```
lurch: I found this code pretty hard to read. Instead of setting a `clear_start` value and then iterating from `clear_start` to `7`, would it make more sense to have a `clear_regions` value and then iterate from `(8 - clear_regions)` to `7`? (or similar, if I've got the maths slightly wrong)

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,245 @@
...
+
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS, skip default GPIOs when auto-detecting psram chip select pin, type=bool, default=PICO_AUTO_DETECT_PSRAM_CS, group=hardware_psram
+#ifndef PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS
+#define PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS PICO_AUTO_DETECT_PSRAM_CS
+#endif
+
+#if PICO_AUTO_DETECT_PSRAM_CS
+#if PICO_RP2350
+#if PICO_RP2350A
+#define PICO_AVAILABLE_CS1_GPIOS {0, 8, 19}
```
lurch: Might be worth adding a comment that these values are only used by `runtime_init_setup_psram` and are simply passed as a parameter to `psram_detect_cs_and_size`? When I first read this header file I incorrectly assumed that `psram_detect_cs_and_size` was doing some kind of filtering based on these values.

But that raises an interesting question - what happens if `psram_detect_cs_and_size` is given a GPIO number (e.g. 2) which isn't valid as a PSRAM CS pin?

src/rp2_common/hardware_psram/include/hardware/psram.h
lurch: Maybe something like `skip any _DEFAULT_ GPIOs defined in the board-header when auto-detecting psram chip select pin` ?

src/rp2_common/hardware_psram/psram.c
lurch: Ahh yes, I keep forgetting about that behaviour of unsigned arithmetic :joy: (too much time spent writing Python)

src/rp2_common/hardware_psram/psram.c
lurch: Ahhh, so if (upon entering this function) `flash_devinfo_get_cs_size(1) != FLASH_DEVINFO_SIZE_NONE` then I guess that must mean that the PSRAM was configured in OTP, and so large chunks of this function are effectively redundant, regardless of what other `#define`s have been set? Might be worth adding a comment about that.

src/rp2_common/hardware_psram/psram.c
lurch: > see my priority list in the PR description

Thanks for the expanded PR description, feels like some of that might be useful in `psram.h`

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: I can move them into `psram.c` so they're only available there

If you pass an invalid GPIO it will still try, but it will always fail to find PSRAM, as it will be configuring the pin for a different function (GPIO_FUNC_CORESIGHT_TRACE, GPIO_FUNC_GPCK, or just an invalid function) - same as if you passed an invalid function to `gpio_set_function`

src/rp2_common/hardware_psram/psram.c
will-v-pi: It gets modified at the start, but I can cache it for the last section

src/rp2_common/hardware_psram/include/hardware/psram.h
lurch: Ahh, so no harmful side-effects; good to know :slightly_smiling_face:

src/rp2_common/hardware_psram/psram.c
```diff
@@ -348,25 +358,26 @@ void runtime_init_setup_psram(void) {
     }
     #endif
 
-    has_psram = flash_devinfo_get_cs_size(1) != FLASH_DEVINFO_SIZE_NONE;
+    flash_devinfo_size_t psram_flash_devinfo_size = flash_devinfo_get_cs_size(1);
```
lurch, line 385: Looks like you could do something similar for `flash_devinfo_size_to_bytes(psram_flash_devinfo_size)` :wink:

src/rp2_common/hardware_psram/psram.c
will-v-pi: I've rearranged the function to make it clearer what's going on

src/rp2_common/hardware_psram/psram.c
will-v-pi: Have added lots of it there

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,429 @@
...
+    #elif defined(PICO_PSRAM_CS_PIN)
+    gpio_set_function(PICO_PSRAM_CS_PIN, GPIO_FUNC_XIP_CS1);
+    #endif
+    if (flash_devinfo_not_configured) {
+        // Attempt to auto-detect the PSRAM size
+        #if PICO_AUTO_DETECT_PSRAM_CS
+        uint8_t cs_gpios[] = PICO_AVAILABLE_CS1_GPIOS;
+        size_t num_cs_gpios = count_of(cs_gpios);
+        #if PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS
+        num_cs_gpios = remove_defaults_from_cs_gpios(cs_gpios, count_of(cs_gpios));
```
lurch: Seems like this could just be:
```suggestion
        num_cs_gpios = remove_defaults_from_cs_gpios(cs_gpios, num_cs_gpios);
```
?

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,429 @@
...
+    size_t psram_word_size = flash_devinfo_size_to_bytes(psram_flash_devinfo_size) >> 2; // >>2 is /4, for words
+    psram_initialized = psram_flash_devinfo_size != FLASH_DEVINFO_SIZE_NONE;
+
+    static_assert(FLASH_DEVINFO_SIZE_MAX == FLASH_DEVINFO_SIZE_16M, "expected max region size of 16M");
+    extern uint32_t __psram_start__;
+    extern uint32_t __psram_end__;
+    uint32_t psram_words = (uint32_t)(&__psram_end__ - &__psram_start__);
+    if (psram_words > psram_word_size) {
+        // Setup to bus fault for variables that don't fit in available PSRAM
+        int clear_regions = 0; // Clear no regions by default
```
lurch, line 395: This is more readable now thanks, and I can now see that this "Clear no regions by default" applies when the PSRAM size is 16MB :slightly_smiling_face:

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+
+/** \file psram.h
+ *  \defgroup pico_psram pico_psram
+ *
+ * \brief Low level PSRAM setup functions
+ *
+ * When using the runtime_init initialisation, this can initialise PSRAM in 3 ways,
+ * listed from highest to lowest priority:
+ *   1. If flash_devinfo is setup (e.g. configured in OTP), it will initialise PSRAM with
+ *      the flash_devinfo CS1 size and GPIO
```
lurch: I was confused by this line at first, and thought it was missing a comma after "CS1" :wink:  But now that I see what it's saying, would this be clearer?
```suggestion
 *      the flash_devinfo size and GPIO for CS1
```

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+ *   2. If `PICO_AUTO_DETECT_PSRAM` is set it will attempt to detect PSRAM size and CS GPIO
+ *      on CS1. This will attempt to use all available CS GPIOs as chip selects, so they
```
lurch:
```suggestion
 *      on CS1. This will attempt to use all available QMI CS1n GPIOs as chip selects, so they
```
might be _slightly_ clearer?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+ *      will be wiggled. By default, it will skip over some which are defined in the board
+ *      header (see `PICO_AUTO_DETECT_PSRAM_CS_SKIP_DEFAULTS`).
```
lurch, line 26: Hmm, I wonder if there's a use-case for being able to say "never attempt to probe this list-of-GPIOs for PSRAM CS" if there are certain pins which the user doesn't want to get wiggled, but which _aren't_ covered by any of the board-header `_DEFAULT_` defines? :thinking:

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+ *     - If the CS GPIO is know and set in `PICO_PSRAM_CS_PIN`, you can just enable
```
lurch:
```suggestion
 *     - If the CS GPIO is known and set in `PICO_PSRAM_CS_PIN`, you can just enable
```

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+ *       `PICO_AUTO_DETECT_PSRAM_SIZE` to only detect the size. Some board headers use
+ *       this behaviour if they have variants both with and without PSRAM fitted
+ *       (e.g. adafruit_feather_rp2350)
+ *   3. If `PICO_PSRAM_SIZE_BYTES` and `PICO_PSRAM_CS_PIN` are set (e.g. configured in the
+ *      board header, or with \ref pico_override_psram_size) it will initialise PSRAM with
+ *      that size and CS GPIO
+ * 
+ * Only the `PICO_AUTO_DETECT_PSRAM` method will verify that PSRAM is present before
```
lurch: Should this be
```suggestion
 * Only the `PICO_AUTO_DETECT_PSRAM` and `PICO_AUTO_DETECT_PSRAM_SIZE` methods will verify that PSRAM is present before
```
?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+ * using it.
+ * 
+ * Variables can be placed in PSRAM using __psram or __psram_uninitialised macros, and
+ * you can also write directly to the memory addresses.
+ * 
+ * If there are variables placed in PSRAM, it will setup XIP to cause bus faults on any
+ * access to PSRAM addresses greater than the size available. The \ref psram_check_address
+ * function should be used before accessing variables in PSRAM when auto-detection is on,
+ * to prevent these bus faults.
```
lurch, line 44: Should / could the kitchen_sink example demonstrate doing this?

src/rp2_common/hardware_psram/include/hardware/psram.h
```diff
@@ -0,0 +1,260 @@
...
+ *       (e.g. adafruit_feather_rp2350)
+ *   3. If `PICO_PSRAM_SIZE_BYTES` and `PICO_PSRAM_CS_PIN` are set (e.g. configured in the
+ *      board header, or with \ref pico_override_psram_size) it will initialise PSRAM with
+ *      that size and CS GPIO
+ * 
+ * Only the `PICO_AUTO_DETECT_PSRAM` method will verify that PSRAM is present before
+ * using it.
+ * 
+ * Variables can be placed in PSRAM using __psram or __psram_uninitialised macros, and
+ * you can also write directly to the memory addresses.
```
lurch:
```suggestion
 * you can also read/write the memory addresses directly.
```
? (I assume that PSRAM isn't write-only :wink: )

lurch: This might be one of my stupid questions, but if you wanted to have variables placed in PSRAM and **also** read/write PSRAM directly, is there a way of knowing which addresses in PSRAM _aren't_ used by the variables placed in PSRAM?
(or maybe that's something you'd never need to do in practice?)

lurch:
Should this PR also be adding `hardware_psram` to `docs/index.h` ?
Should line 15 of `src/rp2_common/hardware_psram/include/hardware/psram.h` be declaring `hardware_psram` (as used on e.g. line 133) instead of `pico_psram`?

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: Yes, it already tests this using `psram_or_malloc` which bus faults in the tiny PSRAM version - a demonstration would be more suited to a pico-example

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: You can read `extern uint32_t __psram_end__;` to get the end of the region allocated to variables

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: I'll go for this, as it's a sub-method, rather than a different method
```
 * Only the `PICO_AUTO_DETECT_PSRAM` methods (including `PICO_AUTO_DETECT_PSRAM_SIZE`) will
 * verify that PSRAM is present before using it.
```

src/rp2_common/hardware_psram/include/hardware/psram.h
will-v-pi: Maybe, and that could be a better way of including the default pins

src/rp2_common/hardware_psram/psram.c
```diff
@@ -97,70 +97,91 @@ size_t psram_detect_cs_and_size(uint8_t *cs_gpios, size_t num) {
...
+#if defined(CYW43_DEFAULT_PIN_WL_HOST_WAKE) && is_cs_gpio(CYW43_DEFAULT_PIN_WL_HOST_WAKE)
+    CYW43_DEFAULT_PIN_WL_HOST_WAKE,
+#endif
+#if defined(CYW43_DEFAULT_PIN_WL_CLOCK) && is_cs_gpio(CYW43_DEFAULT_PIN_WL_CLOCK)
+    CYW43_DEFAULT_PIN_WL_CLOCK,
+#endif
+#if defined(CYW43_DEFAULT_PIN_WL_CS) && is_cs_gpio(CYW43_DEFAULT_PIN_WL_CS)
+    CYW43_DEFAULT_PIN_WL_CS,
+#endif
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PINS, comma separated list of extra pins to not check when auto-detecting psram chip select pin, type=bool, default=undefined, group=hardware_psram
```
lurch: "comma separated list..." "type=bool"
:thinking: :laughing: 

Also, might be better to just remove the `default=...`, rather than saying `default=undefined`

src/rp2_common/hardware_psram/psram.c
```diff
@@ -97,70 +97,91 @@ size_t psram_detect_cs_and_size(uint8_t *cs_gpios, size_t num) {
...
+#define is_cs_gpio(gpio) (gpio == 0) || (gpio == 8) || (gpio == 19)
+#else
+#define is_cs_gpio(gpio) (gpio == 0) || (gpio == 8) || (gpio == 19) || (gpio == 47)
+#endif
+#else
+// Default to assume all GPIOs are CS GPIOs
+#define is_cs_gpio(gpio) true
+#endif
+
+static const uint8_t default_pin_cs_gpios[] = {
```
lurch, line 112: Depending on what other `#defines` are set, I guess this might end up as e.g.
```c
static const uint8_t default_pin_cs_gpios[] = {
    PICO_DEFAULT_UART_TX_PIN,
};
```
(i.e. with nothing after the last comma). Is that allowed in C?

will-v-pi: Yes, trailing commas are fine in C

src/rp2_common/hardware_psram/psram.c
will-v-pi:
```suggestion
// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PINS, comma separated list of extra pins to not check when auto-detecting psram chip select pin, type=list, group=hardware_psram
```

lurch: Unfortunately [extract_configs.py](https://github.com/raspberrypi/pico-sdk/blob/master/tools/extract_configs.py#L101) only supports `type=int`, `type=bool` and `type=enum` :confused: 

Although given that there are so few QMI CS pins, maybe we could just use e.g.
```
// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PIN_0
// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PIN_8
// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PIN_19
// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PIN_47
```
(all with `type=bool`)
:shrug:

will-v-pi: The [extract_cmake_configs.py](https://github.com/raspberrypi/pico-sdk/blob/master/tools/extract_cmake_configs.py#L120) supports list, so I've copied that support in, and it seems to work fine

src/rp2_common/hardware_psram/psram.c
```diff
@@ -0,0 +1,450 @@
...
+#if defined(CYW43_DEFAULT_PIN_WL_HOST_WAKE) && is_cs_gpio(CYW43_DEFAULT_PIN_WL_HOST_WAKE)
+    CYW43_DEFAULT_PIN_WL_HOST_WAKE,
+#endif
+#if defined(CYW43_DEFAULT_PIN_WL_CLOCK) && is_cs_gpio(CYW43_DEFAULT_PIN_WL_CLOCK)
+    CYW43_DEFAULT_PIN_WL_CLOCK,
+#endif
+#if defined(CYW43_DEFAULT_PIN_WL_CS) && is_cs_gpio(CYW43_DEFAULT_PIN_WL_CS)
+    CYW43_DEFAULT_PIN_WL_CS,
+#endif
+// PICO_CONFIG: PICO_AUTO_DETECT_PSRAM_CS_SKIP_PINS, comma separated list of extra pins to not check when auto-detecting psram chip select pin, type=list, group=hardware_psram
```
lurch: IIRC we normally put the `PICO_CONFIG:` entries in the header file? (although this isn't a hard'n'fast rule as we currently have 16 exceptions)

src/rp2_common/hardware_psram/psram.c
lurch: Nice.

kilograham (approved):

PR #2921, opened 2026-05-01
Updates to RISC-V compiler selection (prefer gcc-riscv32-pico-elf and maximum -march opts)

kilograham:
Make gcc-riscv32-pico-elf the preferred compiler, and add support for  trying different compiler flags until supported ones are found (we try more aggressive -march for RISC-V then less aggressive)

### DIFF
--- a/cmake/preload/toolchains/util/find_compiler.cmake
+++ b/cmake/preload/toolchains/util/find_compiler.cmake
@@ -41,7 +41,7 @@ function(pico_find_compiler_with_triples compiler_path triples compiler_suffix)
 endfunction()
 
 # If no single compiler flags variable is set, then choose first working compiler flags from a list.
-# This is done by doing a test compile via execute_proces, as we can't use check_c_compiler_flag
+# This is done by doing a test compile via execute_process, as we can't use check_c_compiler_flag
 # prior to project_setup.
 # Assuming the single flag variable is not already set, each of the flag_var_list flags are tried in order.
 # If the flags don't cause an error the single flag variable is set to the good flag value. If no valid flags
@@ -68,7 +68,7 @@ function(pico_choose_compiler_flags compiler_path common_lang_flags_var common_l
                     return()
                 endif()
             endforeach()
-            message(FATAL_ERROR "No compiler that supported required compiler flags")
+            message(FATAL_ERROR "No compiler found that supports required compiler flags")
         else()
             if (${common_lang_flags_var_list})
                 message(FATAL_ERROR "Either ${common_lang_flags_var} or ${common_lang_flags_var_list} must be set")
--- a/cmake/preload/toolchains/util/find_compiler.cmake
+++ b/cmake/preload/toolchains/util/find_compiler.cmake
@@ -39,3 +39,40 @@ function(pico_find_compiler_with_triples compiler_path triples compiler_suffix)
     pico_find_compiler(${compiler_path} "${triples}")
     set(${compiler_path} ${${compiler_path}} PARENT_SCOPE)
 endfunction()
+
+# If no single compiler flags variable is set, then choose first working compiler flags from a list.
+# This is done by doing a test compile via execute_proces, as we can't use check_c_compiler_flag
+# prior to project_setup.
+# Assuming the single flag variable is not already set, each of the flag_var_list flags are tried in order.
+# If the flags don't cause an error the single flag variable is set to the good flag value. If no valid flags
+# are found, a fatal error is raised.
+function(pico_choose_compiler_flags compiler_path common_lang_flags_var common_lang_flags_var_list)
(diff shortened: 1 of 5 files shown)


cmake/preload/toolchains/util/find_compiler.cmake
```diff
@@ -39,3 +39,40 @@ function(pico_find_compiler_with_triples compiler_path triples compiler_suffix)
     pico_find_compiler(${compiler_path} "${triples}")
     set(${compiler_path} ${${compiler_path}} PARENT_SCOPE)
 endfunction()
+
+# If no single compiler flags variable is set, then choose first working compiler flags from a list.
+# This is done by doing a test compile via execute_proces, as we can't use check_c_compiler_flag
```
lurch:
```suggestion
# This is done by doing a test compile via execute_process, as we can't use check_c_compiler_flag
```

cmake/preload/toolchains/util/find_compiler.cmake
```diff
@@ -39,3 +39,40 @@ function(pico_find_compiler_with_triples compiler_path triples compiler_suffix)
...
+                        RESULT_VARIABLE COMPILE_FAILED
+                        ERROR_QUIET
+                )
+
+                if (NOT COMPILE_FAILED)
+                    set(${common_lang_flags_var} ${flags} PARENT_SCOPE)
+                    return()
+                endif()
+            endforeach()
+            message(FATAL_ERROR "No compiler that supported required compiler flags")
```
lurch:
```suggestion
            message(FATAL_ERROR "No compiler found that supports required compiler flags")
```
?

cmake/preload/toolchains/util/find_compiler.cmake
```diff
@@ -39,3 +39,40 @@ function(pico_find_compiler_with_triples compiler_path triples compiler_suffix)
...
+    # simplify logic by not complaining if both set, since CMake calls compiler setup repeatedly
+    if (NOT ${common_lang_flags_var})
+        if (${common_lang_flags_var_list})
+            if(CMAKE_HOST_WIN32)
+                set(NULL_DEVICE "NUL")
+            else()
+                set(NULL_DEVICE "/dev/null")
+            endif()
+            foreach(flags IN LISTS ${common_lang_flags_var_list})
+                separate_arguments(COMPILER_CMD NATIVE_COMMAND "${compiler_path} ${flags} -x c -c ${NULL_DEVICE} -o ${NULL_DEVICE}")
```
will-v-pi: This doesn't work with the pico-sdk-tools compiler on Windows - I'm getting the error:
```
riscv32-unknown-elf-gcc.exe: fatal error: input file 'NUL' is the same as output file
```

A fix that works is to compile a dummy.c file instead
```suggestion
                separate_arguments(COMPILER_CMD NATIVE_COMMAND "${compiler_path} ${flags} -x c -c \"${CMAKE_CURRENT_LIST_DIR}/dummy.c\" -o ${NULL_DEVICE}")
```

will-v-pi:
You can fix the bazel checks by adding `*/empty.c` to the `IGNORED_FILE_PATTERNS` in tools/check_source_files_in_bazel_build.py - unfortunately `cmake/**` doesn't work because `match` doesn't support the recursive wildcard, and `full_match` was only added in python 3.13

cmake/preload/toolchains/util/find_compiler.cmake
```diff
@@ -39,3 +39,40 @@ function(pico_find_compiler_with_triples compiler_path triples compiler_suffix)
...
+    # simplify logic by not complaining if both set, since CMake calls compiler setup repeatedly
+    if (NOT ${common_lang_flags_var})
+        if (${common_lang_flags_var_list})
+            if(CMAKE_HOST_WIN32)
+                set(NULL_DEVICE "NUL")
+            else()
+                set(NULL_DEVICE "/dev/null")
+            endif()
+            foreach(flags IN LISTS ${common_lang_flags_var_list})
+                separate_arguments(COMPILER_CMD NATIVE_COMMAND "${compiler_path} ${flags} -x c -c ${CMAKE_CURRENT_LIST_DIR}/empty.c -o ${NULL_DEVICE}")
```
will-v-pi: The path needs quoting as it may contain spaces, which then throws the error:
```
riscv32-unknown-elf-gcc.exe: fatal error: cannot specify '-o' with '-c', '-S' or '-E' with multiple files
```

```suggestion
                separate_arguments(COMPILER_CMD NATIVE_COMMAND "${compiler_path} ${flags} -x c -c \"${CMAKE_CURRENT_LIST_DIR}/empty.c\" -o ${NULL_DEVICE}")
```

will-v-pi (approved):

PR #2924, opened 2026-05-06
Tweak check_board_header.py to also validate the PSRAM settings

lurch:
This relates to the changes in #2919 and is effectively a "sub-PR" of that one.


tools/check_board_header.py
```diff
@@ -37,10 +37,20 @@ def __init__(self, message, errors):
...
+compulsory_cmake_default_settings = frozenset(['PICO_FLASH_SIZE_BYTES'])
+
+# names for which pico_board_cmake_set_default (if present) needs a matching #define (and vice-versa)
+matching_cmake_default_settings = frozenset(['PICO_FLASH_SIZE_BYTES', 'PICO_RP2350_A2_SUPPORTED', 'PICO_PSRAM_SIZE_BYTES'])
+
+# names for which we MUST have a #define
+compulsory_defines = frozenset(['PICO_FLASH_SIZE_BYTES'])
+
+# groups of names where if one  #define is present, then all must be present
+linked_defines = frozenset([frozenset(['PICO_PSRAM_SIZE_BYTES', 'PICO_PSRAM_CS_PIN'])])
```
will-v-pi: There is a scenario where you could have `PICO_PSRAM_CS_PIN` defined but not `PICO_PSRAM_SIZE_BYTES`, for boards with a place to solder on PSRAM but none fitted from the factory. Can this check be made uni-directional instead?

tools/check_board_header.py
```diff
@@ -549,6 +565,24 @@ def read_defines_from(header_file, defines_dict):
                 if not any(func_pin in defines for func_pin in expected_function_pins):
                     errors.append(Exception("{}:{}  {} is defined but none of {} are defined".format(board_header, define.lineno, name, list_to_string_with(expected_function_pins, "or"))))
 
+    # check for invalid PSRAM CS pin
+    if name == "PICO_PSRAM_CS_PIN":
+        interface = "QMI"
```
will-v-pi, line 573: I think the interface should be QSPI (it's like SPI, but Quad). QMI stands for "QSPI Memory Interface".

tools/check_board_header.py
lurch: Ahh, I hadn't considered that. I'll update...

tools/check_board_header.py
lurch: I was going along with what Table 3 (on pages 18-20) of the RP2350 datasheet says. But I now realise that this is actually the `CS1N` pin of the (only) QMI interface, and not the `CSN` pin of the second QMI interface :facepalm: so I'll fix that.

tools/check_board_header.py
lurch: Fixed; you can now define `PICO_PSRAM_CS_PIN` without defining `PICO_PSRAM_SIZE_BYTES`, but if you define `PICO_PSRAM_SIZE_BYTES` then you have to also define `PICO_PSRAM_CS_PIN`.

lurch:
PR updated, rebased and force-pushed.

lurch:
Looking at https://github.com/raspberrypi/pico-sdk/pull/2919/commits/2afd108fc8b44090e1edd38f25f5f43ef1dbcc8f , should I also update this to error if `PICO_AUTO_DETECT_PSRAM_SIZE` is set and `PICO_PSRAM_CS_PIN` is missing?

lurch:
...although saying that, _perhaps_ it makes sense to check for `exists(PICO_PSRAM_CS_PIN) or exists(PICO_AUTO_DETECT_PSRAM_CS)` rather than just `exists(PICO_PSRAM_CS_PIN)`? :thinking:

will-v-pi:
I don't think there's a use case for `PICO_AUTO_DETECT_PSRAM_CS` in a board header, as if you know what board you're building for, you should know what pin CS is wired to - so I think just adding an error when `PICO_AUTO_DETECT_PSRAM_SIZE` is set and `PICO_PSRAM_CS_PIN` is missing is fine

lurch:
> so I think just adding an error when `PICO_AUTO_DETECT_PSRAM_SIZE` is set and `PICO_PSRAM_CS_PIN` is missing is fine

Done :+1:

will-v-pi (approved):

PR #2927, opened 2026-05-06
Update check_board_header.py to check for missing ifndefs

lurch:
Any board `#define`s with a `pico_board_cmake_set_default` counterpart, or that start with `PICO_DEFAULT_`, should be wrapped in an `#ifndef` guard so that they're easy to override.

Also fixes a few board-header errors that this detected.

(as discussed with @will-v-pi in #2919 )


will-v-pi (approved):

lurch:
Note that this has a merge-conflict with the `check_board_header.py` changes that are (now) in #2919 , so whichever PR gets merged first, the other one will need rebasing.

PR #2660, opened 2025-09-11
Add support for using XIP cache as SRAM

will-v-pi:
Draft for now, as it includes changes from https://github.com/raspberrypi/pico-sdk/pull/2645 and https://github.com/raspberrypi/pico-sdk/pull/2137, requires the [`xip-sram`](https://github.com/raspberrypi/picotool/tree/xip-sram/) branch of picotool (otherwise UF2 creation and signing fails), and is still WIP - just raising so it can be viewed early, and maybe some of these bits can be added to https://github.com/raspberrypi/pico-sdk/pull/2137

This should fix https://github.com/raspberrypi/pico-sdk/issues/2653 for both RP2040 and RP2350

This adds new binary types `xip_sram` (whole binary in XIP SRAM), and `copy_to_ram_using_xip_ram`/`no_flash_using_xip_ram` (time critical functions placed in XIP SRAM)

Also adds parsing of `addressmap.h` so variables in it can be accessed by using `pico_get_addressmap_value`, to remove hardcoded SRAM addresses from the CMake

PR #2932, opened 2026-05-13
Add support for using XIP cache as SRAM

### LINKED ISSUE #2653: Facilitate RP2040 XIP-cache-as-RAM feature, reclaiming bus bandwidth for time critical instructions in COPY_TO_RAM and NO_FLASH builds
The pico-sdk and RP2040 hardware provide a few facilities that improve performance by moving runtime code and data into SRAM:
1. "pico/platform/sections.h" currently provides the "__not_in_flash", "__not_in_flash_func", and "__time_critical_func" macros for placing runtime code and data into SRAM by assigning them linker section names in the source code.
2. The pico-sdk CMake scripts allow any of four binary types to be selected with similarly named project properties for the RP2040: "default", "blocked_ram", "copy_to_ram", or "no_flash"
3. The RP2040's eXecute-In-Place (XIP) cache has its own connection to the main AHB bus and provides SRAM speeds on cache hits when retrieving runtime code and data from flash

But this regime isn't perfect. The 16kB of XIP cache and its connection to the main AHB bus go mostly unused for PICO_COPY_TO_RAM and PICO_NO_FLASH binary type builds, leaving some performance opportunities unrealized in their implementations.

The RP2040's XIP cache can be disabled by clearing its CTRL.EN bit which allows its 16kB of memory to be used as SRAM directly. 
 
I'd like to update the pico-sdk to support the following:
1. Use the "__time_critical_func" macro to place runtime code into XIP RAM when it's available
2. Add a couple new "copy_to_ram_using_xip_ram" and "no_flash_using_xip_ram" binary type builds for the RP2040 that support this XIP RAM code placement
3. Add a new "PICO_USE_XIP_CACHE_AS_RAM" CMake property to enable the XIP cache's use as RAM for time critical instructions with PICO_COPY_TO_RAM and PICO_NO_FLASH binary type builds
4. Add a couple new CMake functions "pico_sections_not_in_flash(TARGET [list_of_sources])" and "pico_sections_time_critical(TARGET [list_of_sources])" that target selected source file content or a whole CMake build target's list of source files for placement into RAM and/or XIP RAM

I've put together a PR that achieves these 4 goals. I've only tested it with CMake based builds on the RP2040 hardware that I have. That said, I have made an effort to fail fast when CMake configuration properties are incompatible, and I've also made an effort to stay compatible with preexisting linker scripts by building on the preexisting section naming conventions. 

Please consider my PR and let me know if my changes are acceptable, or what I can/should change to make them  acceptable.

Thanks

will-v-pi:
This adds a new `xip_ram` binary type (RP2350 only) which puts everything (code, data, and stack) in XIP SRAM, and a `pico_place_time_critical_functions` CMake function to put functions marked as time_critical in specified memory (eg XIP SRAM or scratch) instead of regular SRAM. The `pico_place_time_critical_functions(TARGET xip_ram)` is only useful for `copy_to_ram` and `no_flash` binary types - it will still work for `default` binary types, but makes everything slower instead of faster.

Adds tests of both - currently the `pico_xip_sram_test` doesn't fit on Clang or Bazel so is skipped on both, and `pico_place_time_critical_functions` is not implemented in Bazel, so `pico_critical_xip_sram_test` sets the compile definitions directly, for the same effect.

Supercedes #2660, and fixes #2653


src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
steffenyount: This description text should probably be updated too, it's no longer 100% true to say:
"Currently this macro is identical in implementation to `__not_in_flash_func`"

steffenyount:
Hi, it's nice to see some progress on this, thanks!

In the description it says: "This adds a new xip_ram binary type (RP2350 only) which puts all code & data in XIP SRAM".

Can you clarify the intent and use case for this new xip_ram binary type?

By "all code & data" are you intending to put both .data and .text code sections into the eXecute In Place (XIP) SRAM, not just the eXecutable .text sections? That seems suboptimal. Is this some kind of provision to support External flash and PSRAM with the XIP hardware that I'm not familiar with?

Also, in the memmap_xip_ram.incl file, am I reading it correctly that all SRAM locations, including the usual SCRATCH X/Y locations, where the CPUs typically maintain their stack data, are being remapped into the XIP SRAM address space with the new xip_ram binary type?

From our previous conversations in #2660 and #2932 I had attempted the PR #2677 to add my two new build script functions separately, but I ran into supported CMAKE version limitations that precluded their inclusion. The build script functions were attempting to promote whole source file's contents like the "not_in_flash" and "time_critical" macros would, but at source file granularity: 
pico_sections_time_critical(TARGET [SOURCES])
pico_sections_not_in_flash(TARGET [SOURCES])

Is the new xip_ram binary type an attempt to do something similar? I'm still focused in RP2040 development, and think a new "copy_to_xip_and_ram" binary type option, that is basically equivalent to the existing "copy_to_ram" binary type option, except that it places all the TARGET eXecutable .text and .time_critical.text content into XIP SRAM could be useful.

Maybe in addition to the new pico_use_xip_sram_for_time_critical(TARGET) build script function we could also add a new pico_use_xip_sram_for_functions(TARGET) build script function that places all the TARGET eXecutable .text and .time_critical.text content into the XIP SRAM? That would make my proposed "copy_to_xip_and_ram" binary type option redundant. Also, the "pico_use_xip_sram_for_time_critical()" could be renamed to "pico_use_xip_sram_for_time_critical_functions()" making the two function's relation and their instruction focus more obvious.

Cheers!

will-v-pi:
> Can you clarify the intent and use case for this new xip_ram binary type?

The new `xip_ram` binary type is not for speed, it's to make a binary that doesn't use Main SRAM at all, so that Main SRAM can be used for something else - a use case for this could be a bootloader. It is entirely separate from all the changes we've discussed before.

The stuff we discussed before is now replaced with the `pico_use_xip_sram_for_time_critical` function, to avoid the need for new binary types. This places only the time critical functions in XIP SRAM.

> Is the new xip_ram binary type an attempt to do something similar? I'm still focused in RP2040 development, and think a new "copy_to_xip_and_ram" binary type option, that is basically equivalent to the existing "copy_to_ram" binary type option, except that it places all the TARGET eXecutable .text and .time_critical.text content into XIP SRAM could be useful.

Placing all of `.text` in XIP SRAM could be an extension to this in a separate PR (it would be more complex), but I think a better extension is your #2677 to allow placing whole files in `.time_critical.text` (if it can be made to work for older CMake).

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
will-v-pi: Thanks, I have updated the description

steffenyount, line 318: I believe this line should be updated too:
`#define RAM_SECTION_NAME(x) .time_critical.text.##x`

src/rp2040/pico_platform/memmap_default.ld
```diff
@@ -9,6 +9,7 @@ INCLUDE "memmap_default.incl"
  * └── memmap_default.incl                               rp2_common/pico_standard_link
  *     ├── set_memory_locations.incl                     rp2_common/pico_standard_link
  *     ├── memory_flash.incl                             rp2_common/pico_standard_link
+ *     ├── memory_xip_ram.incl                           rp2_common/pico_standard_link
```
lurch, line 13: Maybe I'm missing something obvious here, but these `*.ld` files all reference `memory_xip_ram.incl`, but this PR actually seems to be adding a `memmap_xip_ram.incl` file?

will-v-pi: The `memory_xip_ram.incl` file is already present, it just wasn’t being used

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
```diff
@@ -10,62 +10,102 @@
...
+#endif
 
 // PICO_CONFIG: PICO_SPINLOCK_ID_OS1, First Spinlock ID reserved for use by low level OS style software, min=0, max=31, default=14, group=hardware_sync
 #ifndef PICO_SPINLOCK_ID_OS1
+#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS
+#define PICO_SPINLOCK_ID_OS1 18
+#else
 #define PICO_SPINLOCK_ID_OS1 14
 #endif
+#endif
```
lurch, line 72: For the PICO_SPINLOCK_IDs defined before this one, the newly-added `#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS` sections are declaring a lower-numbered spinlock (than the old values), but for the PICO_SPINLOCK_IDs defined _after_ this one, the newly-added `#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS` sections are declaring a higher-numbered spinlock? :thinking:

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
```diff
@@ -10,62 +10,102 @@
...
 
 // PICO_CONFIG: PICO_SPINLOCK_ID_ATOMIC, Spinlock ID for atomics, min=0, max=31, default=13, group=hardware_sync
 #ifndef PICO_SPINLOCK_ID_ATOMIC
+#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS
+#define PICO_SPINLOCK_ID_ATOMIC 11
+#else
 #define PICO_SPINLOCK_ID_ATOMIC 13
 #endif
+#endif
```
lurch, line 63: Old spinlock numbering was `9, 10, 11, 12, 13`, but the `#if PICO_RP2350 && !PICO_USE_SW_SPIN_LOCKS` numbering is `5, 6, 7, 10, 11`. What happened to 9?

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
```diff
@@ -162,21 +162,24 @@
  *  \ingroup pico_platform
  *
  * Decorates a function name, such that the function will execute from RAM (assuming it is not inlined
- * into a flash function by the compiler) to avoid possible flash latency. Currently this macro is identical
+ * into a flash function by the compiler) to avoid possible flash latency. By default, this macro is identical
  * in implementation to `__not_in_flash_func`, however the semantics are distinct and a `__time_critical_func`
- * may in the future be treated more specially to reduce the overhead when calling such function from a flash
- * function.
+ * can be treated more specially to reduce the overhead when calling such function.
```
lurch:
```suggestion
 * can be treated more specially to reduce the overhead when calling such a function.
```

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
```diff
@@ -162,21 +162,24 @@
...
+ * 
+ * For binaries that are not executing from flash (eg copy_to_ram and no_flash), one special treatment can be
+ * using the \ref`pico_use_xip_sram_for_time_critical` CMake function to place them in XIP RAM instead, as the
+ * XIP ports would be otherwise unused.
```
lurch: What does "ports" here mean? Would "as the XIP RAM would be otherwise unused" also work?

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
will-v-pi: See https://github.com/raspberrypi/pico-sdk/pull/2645

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
will-v-pi: See https://github.com/raspberrypi/pico-sdk/pull/2645

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
will-v-pi: They're bus ports, see 2.1 in both the datasheets. My choice of wording was intended to clarify that the XIP ports of the bus fabric are unused when running from RAM, so can be used for time critical functions, as the main SRAM ports may be more congested.

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
will-v-pi: Thanks - added

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
lurch: Aha, thanks! :laughing: 
Maybe the PR-description should be updated to say that this PR also supersedes #2645 ?

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
lurch: Oh right, the APB / AHB stuff. _Maybe_ it's worth clarifying "XIP ports" as "XIP AHB ports" ? :man_shrugging:

will-v-pi:
```suggestion
 * XIP AHB ports would be otherwise unused. 
```

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -477,7 +500,7 @@ _call_xip_setup:
 
     // In a NO_FLASH binary, don't perform .data etc copy, since it's loaded
     // in-place by the SRAM load. Still need to clear .bss
-#if !PICO_NO_FLASH
+#if !PICO_CRT0_NO_DATA_COPY
```
kilograham, line 523: i guess this #ifdef is still useful to avoid wasting space, but for my own edification, does the "src == dest" take care of everything  (i.e. in NO_FLASH binaries, was it the case that src == dest in all case)

kilograham (commented):
Generally looks good

1. I'm a bit confused as to what the RP2040/RP2350 options are - is this strictly xip_cache as ram (separate from using PSRAM as per the other PR)
2. given the above, perhaps `xip_ram` is a poor name (not we also use xip_sram elsewhere), but:
3. All the existing names are bad, so maybe this would be a good time to rationilze and maybe even add aliases for the old ones (or rather make the old ones aliases)

no_flash => not sure (ram, ram_stored?) - note i wonder if PICO_NO_FLASH isn't strictly the binary type define; it means there is nothing in flash, right, but that could be set for a binary which uses sram & xip_ram - maybe no_flash IS the base binary type, and you use the other controls to turn on XIP_RAM additionally?
blocked_ram => not sure i care
copy_to_ram => i guess this is ok
xip_ram -> not sure

Just think it is worth enumerating all the possibilites (i.e. maybe there are subtypes of no_flash); how does PSRAM play with the above etc?

src/rp2_common/pico_crt0/crt0.S
will-v-pi: Currently yes, in NO_FLASH binaries src == dest for everything, as there is no copying performed - although new PSRAM variables will end up with src != dest, but that's handled in runtime_init not here.

These changes aren't actually needed anymore to make this PR work, but I think they're still useful to allow creating no_flash binaries that do copy their data around (eg if you need to load a sparse binary over the UART bootloader, squashing it together makes the transfer faster).

will-v-pi:
Just to clarify regarding 1, this is two separate ideas combined into one PR, because they both use XIP SRAM

- For RP2040 & RP2350, this adds the ability for `copy_to_ram`/`no_flash` binaries to put all their time critical functions into XIP SRAM, as the XIP AHB ports would otherwise be unused
- For RP2350 only, this adds the new `xip_ram` binary type to create XIP SRAM only binaries, where all the code & data is in XIP SRAM - these could be useful for bootloaders

This does not add the ability to place arbitrary data in XIP SRAM - I can add that if it would be useful

kilograham:
yeah, i mean historically i guess we needed different binary types for different linker scripts, but that is no longer strictly the case.

it probably makes sense to continue to have binary types for "where stuff gets put by the linker script", but up til now that is mostly "where all the stuff goes".

If we have main ram and xip ram, it is unlikely the main code/data will be split between, so you're likely to just want to access the xip-cache; which can(/may already be) something you can specify via cmake. By this i guess I'm talking about the enabling of XIP RAM (pinning) vs the storing of code/data there (though the latter requires the former not vice versa)

Maybe it is just me, but i find xip-ram confusing since PSRAM is also via XIP. I don't know if xip-cache is better, since it isn't the cache when being used as ram!

steffenyount:
Hi @will-v-pi

I took a swing at adding that support for placing a whole file's functions into XIP SRAM. I think it turned out better than I anticipated, and it should be compatible with CMake 3.13.

Please consider adding my changes into your PR, and let me know what you think of it.

(the tests section probably needs some work)

you can find it [here](https://github.com/will-v-pi/pico-sdk/pull/1)

Thanks,
-Steffen

will-v-pi:
> yeah, i mean historically i guess we needed different binary types for different linker scripts, but that is no longer strictly the case.
> 
> it probably makes sense to continue to have binary types for "where stuff gets put by the linker script", but up til now that is mostly "where all the stuff goes".

I think it's still useful to have "standard" binary types which have their own memmap (and it in necessary for `memmap_xip_ram.incl` as I've currently written it), because picking a "standard" binary type is used to set the compile definitions. This is more necessary for Bazel, where the compile definitions have to be set using constraints, so it needs explicit binary types for that.

> If we have main ram and xip ram, it is unlikely the main code/data will be split between, so you're likely to just want to access the xip-cache; which can(/may already be) something you can specify via cmake. By this i guess I'm talking about the enabling of XIP RAM (pinning) vs the storing of code/data there (though the latter requires the former not vice versa)

The first point of this PR (see "For RP2040 & RP2350 ..." in my comment above) is to split code between main ram and xip ram - it puts time critical function code in xip ram, and all other code/data in main ram.

Pinning can already be enabled using the `PICO_CRT0_PIN_XIP_SRAM` compile definition added by the low power PR for RP2350 - this PR adds a new `PICO_USE_XIP_CACHE_AS_RAM` which works on RP2040 too (on RP2350 it just enables pinning). I've also just added a check to the low power stuff to skip un-pinning if `PICO_USE_XIP_CACHE_AS_RAM` is set, so this can work with low power.

> Maybe it is just me, but i find xip-ram confusing since PSRAM is also via XIP. I don't know if xip-cache is better, since it isn't the cache when being used as ram!

It's called "XIP_SRAM" in the `addressmap.h` files, and "Cache-as-SRAM" in the datasheet - I prefer calling it `xip_ram`, but happy to change if there's a consensus for something else.

will-v-pi:
Here's a full description of binary types, as I think they will be after this PR merged, if you want to propose new names for any of them:

- `no_flash` - a binary that does not contain any code/data in Flash
  - storage address and runtime address of everything is the same (ie `> XXX AT> XXX`)
    - the exception is PSRAM, which is stored in main SRAM as A2 bootrom cannot load directly into PSRAM due to RP2350-E14
  - can use main SRAM, XIP_SRAM and PSRAM for code/data
  - by default, all code/data is in main SRAM
- `copy_to_ram` - a binary that starts in Flash, but main code/data is copied to main SRAM for execution
  - storage address is flash for everything (ie `> XXX AT> FLASH`)
  - can use main SRAM, XIP_SRAM, PSRAM and Flash for code/data
  - by default, all code/data is in main SRAM, except for `crt0.S` which is in Flash
- `default` - a binary that executes most code/data from Flash
  - storage address is flash for everything (ie `> XXX AT> FLASH`)
  - can use main SRAM, PSRAM and Flash for code/data, and XIP_SRAM with performance penalties to code/data in Flash
  - by default, all code is in Flash and all data is in main SRAM
- `xip_ram` - a binary that does not contain any code/data in Flash or main SRAM (not available on RP2040)
  - storage address and runtime address of everything is the same (ie `> XXX AT> XXX`)
    - the exception is PSRAM, which is stored in XIP_SRAM as A2 bootrom cannot load directly into PSRAM due to RP2350-E14
  - can use XIP_SRAM and PSRAM for code/data - XIP_SRAM is used as if it were main SRAM, for the purposes of the macros below
  - by default, all code/data is in XIP_SRAM
- `blocked_ram` - identical to `default`, but uses non-striped memory for main SRAM instead of striped (RP2040 only)

For binaries that can use main SRAM, they can place other code/data there using `__not_in_flash` and similar macros

For binaries that can use Flash, they can place other data there using the `__in_flash` macro

For binaries that can use PSRAM, they can place other data there using `__in_psram` and `__uninitialized_psram` macros

For binaries that can use XIP_SRAM (does not include `xip_ram`, as that uses XIP_SRAM as if it were main SRAM), they can place other code there using the `__time_critical_func` macro, provided they call `pico_use_xip_sram_for_time_critical` in their CMakeLists.txt

Any other placements (eg code in PSRAM, data in XIP_SRAM) currently require custom linker script sections and custom section attributes.

src/rp2_common/pico_low_power/CMakeLists.txt
```diff
@@ -34,13 +34,19 @@ pico_promote_common_scope_vars()
 # It also sets PICO_CRT0_PIN_XIP_SRAM=1 to pin the XIP_SRAM,
 # if the persistent data is stored in XIP_SRAM.
 #
-# \param\ PERSISTENT_DATA_LOC The persistent data location to set
+# \param\ PERSISTENT_DATA_LOC The persistent data location to set - optionally pass the memory region name (supports xip_ram)
```
will-v-pi:
```suggestion
# \param\ PERSISTENT_DATA_LOC The persistent data location to set (e.g. 0x20040000), or the memory region name (currently supports xip_ram)
```

steffenyount:
With memory_psram.incl and memory_xip_ram.incl providing integrated region support out-of-the-box, the comment/example text provided in memory_extra.incl file seems less relevant, and should probably be updated.

src/rp2_common/pico_standard_link/CMakeLists.txt
steffenyount: PICO_XIP_RAM should probably be integrated for PICO_DEFAULT_BINARY_TYPE and PICO_XIP_RAM resolution and exclusion, like all the other built-in binary types being supported above.

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -234,23 +241,12 @@ if (NOT TARGET pico_standard_link)
...
-            set(PICO_USE_BLOCKED_RAM 1)
-        endif()
-        if (PICO_DEFAULT_BINARY_TYPE STREQUAL copy_to_ram)
-            set(PICO_COPY_TO_RAM 1)
-        endif()
     endif()
+    # finally check multiple legacy variables haven't been set
     if ((PICO_NO_FLASH AND PICO_USE_BLOCKED_RAM) OR
         (PICO_USE_BLOCKED_RAM AND PICO_COPY_TO_RAM) OR
         (PICO_COPY_TO_RAM AND PICO_NO_FLASH))
```
lurch, line 286: In some C macros where we wanted to ensure that only one `#define` was set to one and the others were set to zero, @kilograham did a neat trick of `#if (DEFINE1 + DEFINE2 + DEFINE3) > 1` - is something similar possible in CMake?

will-v-pi: I don’t think so as CMake has lots of truthy values, and it would probably look more complicated as CMake requires a separate `math` function

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -307,23 +303,22 @@ if (NOT TARGET pico_standard_link)
...
     # PICO_BUILD_DEFINE: PICO_USE_BLOCKED_RAM, whether this is a 'blocked_ram' build, type=bool, default=0, but dependent on CMake options, group=pico_standard_link
-    target_compile_definitions(pico_standard_link INTERFACE PICO_USE_BLOCKED_RAM=$<IF:$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>,use_blocked_ram>,1,$<AND:$<BOOL:${PICO_USE_BLOCKED_RAM}>,$<STREQUAL:,$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>>>>)
-    # PICO_COPY_TO_RAM will be set based on PICO_TARGET_BUILD_TYPE target property being in COPY_TO_RAM_BINARY_TYPES if set, otherwise to the value of the PICO_COPY_TO_RAM cmake variable unless PICO_TARGET_TYPE is set to something else
+    target_compile_definitions(pico_standard_link INTERFACE PICO_USE_BLOCKED_RAM=$<IF:$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>,blocked_ram>,1,$<AND:$<STREQUAL:${PICO_DEFAULT_BINARY_TYPE},blocked_ram>,$<STREQUAL:,$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>>>>)
+    # PICO_COPY_TO_RAM will be set based on PICO_TARGET_BINARY_TYPE target property being in COPY_TO_RAM_BINARY_TYPES if PICO_TARGET_BINARY_TYPE is set, otherwise it is set based on PICO_DEFAULT_BINARY_TYPE being in COPY_TO_RAM_BINARY_TYPES
```
lurch, line 354: I'm not entirely sure of the semantics here, but I **think** this is saying "PICO_COPY_TO_RAM will be set based on PICO_TARGET_BINARY_TYPE or PICO_DEFAULT_BINARY_TYPE being in COPY_TO_RAM_BINARY_TYPES" ? :thinking:

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -307,23 +303,22 @@ if (NOT TARGET pico_standard_link)
...
     # PICO_BUILD_DEFINE: PICO_NO_FLASH, whether this is a 'no_flash' build, type=bool, default=0, but dependent on CMake options, group=pico_standard_link
-    target_compile_definitions(pico_standard_link INTERFACE PICO_NO_FLASH=$<IF:$<IN_LIST:$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>,${NO_FLASH_BINARY_TYPES}>,1,$<AND:$<BOOL:${PICO_NO_FLASH}>,$<STREQUAL:,$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>>>>)
-    # PICO_USE_BLOCKED_RAM will be set based on PICO_TARGET_BUILD_TYPE target property being equal to use_blocked_ram if set, otherwise to the value of the PICO_USE_BLOCKED_RAM cmake variable unless PICO_TARGET_TYPE is set to something else
+    target_compile_definitions(pico_standard_link INTERFACE PICO_NO_FLASH=$<IF:$<IN_LIST:$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>,${NO_FLASH_BINARY_TYPES}>,1,$<AND:$<IN_LIST:${PICO_DEFAULT_BINARY_TYPE},${NO_FLASH_BINARY_TYPES}>,$<STREQUAL:,$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>>>>)
+    # PICO_USE_BLOCKED_RAM will be set based on PICO_TARGET_BINARY_TYPE target property being equal to blocked_ram if PICO_TARGET_BINARY_TYPE is set, otherwise it is set based on PICO_DEFAULT_BINARY_TYPE being equal to blocked_ram
```
lurch, line 351: It looks like this is the only variant which is looking for the PICO_TARGET_BINARY_TYPE target property being equal to a specific value, rather than being in some other list?

src/rp2_common/pico_standard_link/CMakeLists.txt
will-v-pi: The ordering has always been a bit weird, but I think that was so the sentence starts with the variable name for clarity. If you split it at the comma, the first half applies “if PICO_TARGET_BINARY_TYPE is set”, and the second half applies if not. I.e if PICO_TARGET_BINARY_TYPE is set, it ignores PICO_DEFAULT_BINARY_TYPE

src/rp2_common/pico_standard_link/CMakeLists.txt
```diff
@@ -332,10 +327,10 @@ if (NOT TARGET pico_standard_link)
...
     # boot_stage2 will be linked if PICO_NO_FLASH would be defined to 0; note if boot_stage2 headers not present, then boot_stage2 is omitted from build anyway
     if (TARGET boot_stage2_headers)
-        target_link_libraries(pico_standard_link INTERFACE $<$<NOT:$<IF:$<STREQUAL:$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>,no_flash>,1,$<AND:$<BOOL:${PICO_NO_FLASH}>,$<STREQUAL:,$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>>>>>:$<IF:$<BOOL:$<TARGET_PROPERTY:PICO_TARGET_BOOT_STAGE2>>,$<TARGET_PROPERTY:PICO_TARGET_BOOT_STAGE2>,bs2_default>_library>)
+        target_link_libraries(pico_standard_link INTERFACE $<$<NOT:$<IF:$<IN_LIST:$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>,${NO_FLASH_BINARY_TYPES}>,1,$<AND:$<IN_LIST:${PICO_DEFAULT_BINARY_TYPE},${NO_FLASH_BINARY_TYPES}>,$<STREQUAL:,$<TARGET_PROPERTY:PICO_TARGET_BINARY_TYPE>>>>>:$<IF:$<BOOL:$<TARGET_PROPERTY:PICO_TARGET_BOOT_STAGE2>>,$<TARGET_PROPERTY:PICO_TARGET_BOOT_STAGE2>,bs2_default>_library>)
```
lurch, line 371: Hard to tell if the `>>>>>` here is valid CMake or the result of a git merge conflict :joy:

src/rp2_common/pico_standard_link/CMakeLists.txt
will-v-pi: It’s legacy (and RP2040 only), and the define isn’t actually used anywhere in the SDK (it wasn’t actually being set before, because use_blocked_ram isn’t a binary type, it’s just blocked_ram), so I left it using specific values as we are very unlikely to add more binary types that use it

src/rp2_common/pico_standard_link/CMakeLists.txt
lurch: Lets just hope that we don't get more binary types in future then... :wink:

will-v-pi: These are the legacy defines, so we won’t be adding more (eg PICO_XIP_RAM isn’t added here) - new method is just to set PICO_DEFAULT_BINARY_TYPE

src/rp2_common/pico_standard_link/CMakeLists.txt
lurch: Ahhh, so it's kinda like
`PICO_COPY_TO_RAM = (defined(PICO_TARGET_BINARY_TYPE) ? PICO_TARGET_BINARY_TYPE : PICO_DEFAULT_BINARY_TYPE) in COPY_TO_RAM_BINARY_TYPES`
?

kilograham:
ok, i'm reasonably cool with all this, but 

> Any other placements (eg code in PSRAM, data in XIP_SRAM) currently require custom linker script sections and custom section attributes.

A bit confusing as there is data in XIP_SRAM describe just above that

I think i'm still a bit weirded out about the cmake function to put time_critical elsewhere

I wonder if we should 

1. make sure we have the full gamut of in_ram, in_blah functions (including adding in_scratch_x as an alias for scratch_x)
2. make .time_critical, .not_in_flash definitions based on that (can still default to .time_critical I guess), but we could just make it target in_ram, or whatever if you want to specify a different location (same for not_in_flash i guess)

src/rp2_common/pico_standard_link/CMakeLists.txt
will-v-pi: Yes, that's it

will-v-pi:
Following those suggestions, I've refactored this to no longer use linker script generation, and just use compile definitions instead. I've also added the new macros in `sections.h` (`__in_ram`, `__in_scratch_x/y`, `__in_xip_ram`) to place data in those sections, with `__in_xip_ram` throwing an error if `PICO_USE_XIP_CACHE_AS_RAM` isn't set.

This means the new function is `pico_place_time_critical_functions(TARGET MEMORY)`, which can be used to place the time critical functions in `xip_ram`, or `scratch_x/y`. All this does is set the `PICO_TIME_CRITICAL_PLACEMENT` compile definition, along with setting `PICO_USE_XIP_CACHE_AS_RAM=1` if `xip_ram` is used.

I've also added a `pico_use_xip_cache_as_ram` function, which just sets `PICO_USE_XIP_CACHE_AS_RAM=1`.

I've left in the `pico_include_in_generated_section` function, as it could be useful in the future.

kilograham (changes requested):
So I think it is worth being clear about thigns; if i understand correctly

1. section .time_critical is now in RAM always (we should probably point out clearly that this is the case - maybe you already did - we don't want to change the name for histortical reasons in the linker script, but worth being verbose wherever we see it that it isn't striclty tied to the time_criitical macro any more

2. __not_in_flash() should probably default to in_ram() but possibly be overridable like time_critical

3. __time_critical (and __not_in_flash) can then i think default to in_ram (which would be defined as section ,time_critical)

it is probably worth adding doxygen to each of the new macros (and linking to the other ones where synojyms)

steffenyount:
> Generally looks good
> 
> 1. I'm a bit confused as to what the RP2040/RP2350 options are - is this strictly xip_cache as ram (separate from using PSRAM as per the other PR)
> 2. given the above, perhaps `xip_ram` is a poor name (not we also use xip_sram elsewhere), but:
> 3. All the existing names are bad, so maybe this would be a good time to rationilze and maybe even add aliases for the old ones (or rather make the old ones aliases)
> 
> no_flash => not sure (ram, ram_stored?) - note i wonder if PICO_NO_FLASH isn't strictly the binary type define; it means there is nothing in flash, right, but that could be set for a binary which uses sram & xip_ram - maybe no_flash IS the base binary type, and you use the other controls to turn on XIP_RAM additionally? blocked_ram => not sure i care copy_to_ram => i guess this is ok xip_ram -> not sure
> 
> Just think it is worth enumerating all the possibilites (i.e. maybe there are subtypes of no_flash); how does PSRAM play with the above etc?

Perhaps renaming `xip_ram` to something like `no_flash_using_xip_ram_as_ram` would be more descriptive, since it is re-using the linker `.incl` section definitions from the `no_flash` binary type?

Or maybe implementing the `xip_ram` functionality as a link option rather than a whole binary type strategy (E.g: `PICO_LINK_XIP_RAM_AS_RAM`) that is compatible with the `no_flash` binary type would be cleaner? 

Relatedly, there is precedent for the `xip_ram` kind of almost entire section definition re-use, as modeled by the `blocked_ram` binary type definition. In that case the only difference is seen in the `memmap_blocked_ram.ld` file where `RAM_ORIGIN = 0x21000000;` is being set before deferring with `INCLUDE "memmap_default.incl"`.

For the `blocked_ram` binary type, has anyone considered eliminating the `memmap_blocked_ram.ld` file, setting the linker script symbol directly via parameter like `-Wl,--defsym=RAM_ORIGIN=0x21000000`, and then just using the vanilla `memmap_default.ld` when the `blocked_ram` binary type is set or when `PICO_USE_BLOCKED_RAM` is set?

Or alternatively pushing the flag evaluation down into the `default_locations.ld` file by replacing the line:
`RAM_ORIGIN_DEFAULT = 0x20000000;`
with something like:
`RAM_ORIGIN_DEFAULT = (DEFINED(PICO_USE_BLOCKED_RAM) ? PICO_USE_BLOCKED_RAM : 0) ? 0x21000000 : 0x20000000;`

It looks like the `xip_ram` binary type, could similarly have all of its custom symbol values (listed below) sent to the linker directly via parameter, thus eliminating the need for the `memmap_xip_ram.ld` file, by using the `memmap_no_flash.ld` file directly:
```RAM_ORIGIN
RAM_LENGTH
SCRATCH_X_ORIGIN
SCRATCH_X_LENGTH
SCRATCH_Y_ORIGIN
SCRATCH_Y_LENGTH
```

With this alternate approach, the sdk could cleanly focus on its three base binary layouts: `default`, `copy_to_ram`, and `no_flash` with `blocked_ram` and `xip_ram` binary types being implemented basically as short hand notation for configuration options on the `default` and `no_flash` base binary layouts respectively.
E.g: `blocked_ram` could be equivalent to setting:
```
PICO_DEFAULT_BINARY_TYPE=default
PICO_USE_BLOCKED_RAM=1
```
E.g: `xip_ram` could be equivalent to setting:
```
PICO_DEFAULT_BINARY_TYPE=no_flash
PICO_USE_XIP_CACHE_AS_XIP_RAM=1
PICO_LINK_XIP_RAM_AS_RAM=1
```
E.g: the `FLASH` and `PSRAM` configuration could be reimagined to operate similarly, with their `__in_flash()` and `__in_psram()` section macros throwing an error when the corresponding hardware isn't available, like when the `PICO_USE_*_SIZE_BYTES` parameter is `0` or undefined.
```
PICO_DEFAULT_BINARY_TYPE=default
PICO_USE_FLASH_SIZE_BYTES=8388608
```
```
PICO_DEFAULT_BINARY_TYPE=copy_to_ram
PICO_USE_PSRAM_SIZE_BYTES=8388608
PICO_LINK_PSRAM_AS_RAM
```
E.g: meanwhile for a larger performance oriented application running from ram, I might want to set the following options to selectively place whole sections (E.g: time_critical, not_in_flash, or .text*) of functions/instructions into XIP_RAM to take advantage of the dedicated AHB ports for their instruction bandwidth:
```
PICO_DEFAULT_BINARY_TYPE=copy_to_ram
PICO_USE_XIP_CACHE_AS_XIP_RAM=1
PICO_PLACE_TIME_CRITICAL_FUNCTIONS_IN_XIP_RAM=
PICO_PLACE_ALL_FUNCTIONS_IN_XIP_RAM='main.c.o;example.c.o'
```

Note: I'm trying out the following parameter prefix notation here to differentiate their effects, what do you think?
`PICO_USE_*` - parameters that reflect hardware constraints and affect memory region availability. E.g: the availability of a FLASH region, a PSRAM region, or the XIP_RAM region.
`PICO_LINK_*` - parameters that affect the linker script's general memory region assignments
`PICO_PLACE_*` - parameters that affect the linker script's section assignments into those memory regions,

will-v-pi:
@kilograham I have re-arranged `sections.h` to make it clearer what everything does:
 - All `__in_xxx` macros place directly into linker sections which are located in the relevant memory
 - `__scratch_x` and `__scratch_y` are kept as aliases to `__in_scratch_x` and `__in_scratch_y`
 - `__not_in_flash` is placed using `PICO_NOT_IN_FLASH_PLACEMENT` which defaults to `__in_ram`, and the same for `__time_critical_func` using `PICO_TIME_CRITICAL_PLACEMENT` which also defaults to `__in_ram`

@steffenyount Thanks, I have changed `memmap_xip_ram.incl` to include `memmap_no_flash.incl`, now that they are identical. I think I prefer to keep the `blocked_ram` and `xip_ram` customisations in separate binary types, otherwise the address setting would need to be duplicated between CMake and Bazel. The rest of what you're talking about sounds like a much larger refactor, which would be outside the scope of this PR, especially as this PR is now making minimal linker script modifications and doing the re-assignment through macros instead.

Regarding the ability to place all functions somewhere else, I think that definitely belongs in a different PR, and I think I'd want to see a use case which can't be solved by just annotating the functions with the `__no_inline_not_in_flash_func` or `__time_critical_func` macros?

kilograham (approved):

PR #2700, opened 2025-10-15
Use multilib concept with Clang/ATfE

rgrr:
These are corrections to the SDK to make my project compile with clang/ATfE 21.1.1.  This should be generic.

- using clang multilib/sysroot concept which makes life much easier
- corrected llvm_libc files to make it compile and run with 21.1.1
- always use ld for linking.  lld optimizes too much away for rp2040

Final result is:

```
#       gcc                14.2.1
#              llvm_libc -  FAIL      (mixture of newlib/llvm_libc)
#              picolibc  -  FAIL      (FDEV_SETUP_STREAM is unknown, mixture if newlib/picolibc)
#              newlib    -   OK       (same as -DPICO_CLIB=)

#       clang              19.1.5   21.1.1
#              llvm_libc -  FAIL      OK       (19.1.5: llvm_libc has some incompatibilities)
#              picolibc  -   OK       OK
#              newlib    -   OK       OK       (same as -DPICO_CLIB=)
```

"Issue" can be found in #2701.

Closes #2701

PR #2934, opened 2026-05-14
newlib/picolibc auto detection

### LINKED ISSUE #2933: Debian testing provides only picolibc but no newlib resulting in missing "stdout"
which means, that linking fails because the wrong newlib_interface.c is used (instead of the correct picolibc_interface.c if either using PICO_CLIB=newlib or empty.

This results in a linker error because "stdout" is missing.

Would be nice to have some kind of auto-detect for this.

rgrr:
For Debian testing there exists no longer newlib, instead picolibc is provided.

If the user does not explicitly selects picolibc with PICO_CLIB=picolibc the linking stage returns an undefined symbol "stdout" because the wrong newlib_interface.c had been compiled in.

This PR checks in case of gcc toolchain if newlib is installed and if not tries to use picolibc instead.  The logic should be generic also for other development  environments than Debian.

If another toolchain is selected (like clang), the auto-detect mechanism is omitted.  In this case the clang toolchain logic must catch this case.  See #2700 ;-)

Fixes #2933

### DIFF
--- a/src/rp2_common/pico_clib_interface/CMakeLists.txt
+++ b/src/rp2_common/pico_clib_interface/CMakeLists.txt
@@ -21,6 +21,10 @@ if (NOT TARGET pico_clib_interface)
     #            PICO_STDIO_SHORT_CIRCUIT_CLIB_FUNCS=0
     #)
 
+    target_include_directories(pico_picolibc_interface SYSTEM INTERFACE
+            ${CMAKE_CURRENT_LIST_DIR}/include/picolibc
+    )
+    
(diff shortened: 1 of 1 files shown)


kilograham:
can you please try #2935 which is my alternate implementation

rgrr:
I'll give it a try on sunday.

H.

rgrr:
> can you please try #2935 which is my alternate implementation

It is working, see my PR comment

kilograham:
replaced by #2935

PR #2935, opened 2026-05-14
Add detection of newlib vs picolibc if not specified

### LINKED ISSUE #2933: Debian testing provides only picolibc but no newlib resulting in missing "stdout"

kilograham:
fixes #2933

### DIFF
--- a/src/rp2_common/pico_clib_interface/CMakeLists.txt
+++ b/src/rp2_common/pico_clib_interface/CMakeLists.txt
@@ -54,7 +54,7 @@ if (NOT TARGET pico_clib_interface)
         file(WRITE "${CLIB_DETECT_SOURCE}" "${CLIB_DETECT_CODE}")
 
         set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
-        foreach(CLIB picolibc newlib)
+        foreach(CLIB newlib picolibc)
             # Attempt to compile the source file
             try_compile(CLIB_MATCHED
                     "${CMAKE_CURRENT_BINARY_DIR}"              # Directory for temporary build files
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_clib_interface/CMakeLists.txt
```diff
@@ -36,8 +36,43 @@ if (NOT TARGET pico_clib_interface)
...
+            #error not using wanted c library
+            #endif
+        ")
+
+        # Create a temporary file for the source
+        set(CLIB_DETECT_SOURCE "${CMAKE_CURRENT_BINARY_DIR}/clib_check.c")
+        file(WRITE "${CLIB_DETECT_SOURCE}" "${CLIB_DETECT_CODE}")
 
+        set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+        foreach(CLIB picolibc newlib)
```
rgrr: I think order should be "newlib picolibc", because newlib was default before

rgrr (commented):
Tested it with my gcc/picolibc installation on Debian testing which no longer has newlib for arm-none-eabi

src/rp2_common/pico_clib_interface/CMakeLists.txt
kilograham:
```suggestion
        foreach(CLIB newlib picolibc)
```

kilograham: good point

will-v-pi (approved):

PR #2936, opened 2026-05-14
ignore cmake/ directory in bazel source file checks

kilograham:
cc @armandomontanez

### DIFF
--- a/tools/check_source_files_in_bazel_build.py
+++ b/tools/check_source_files_in_bazel_build.py
@@ -166,6 +166,7 @@ def git_ls_files_by_extension(file_suffixes: Iterable[str]) -> Iterable[Path]:
     ).stdout
 
     bazel_files = [Path(f) for f in bazel_file_list.splitlines()]
+    bazel_files = [f for f in bazel_files if not f.is_relative_to("cmake")]
     return bazel_files
 
 


will-v-pi (commented):
This doesn't fix the issue you're seeing in #2921 - you also need to change `git_ls_files_by_extension` to only return files in these paths.

Also, it's not just files under `src` and `test` that should be checked, it should also check `tools/pioasm`, so I think excluding `cmake` is possibly the better option than explicitly specifying paths to search?

A fix could be to revert this change and just add
```python
bazel_files = [f for f in bazel_files if not f.is_relative_to("cmake")]
```
before `return bazel_files` in `git_ls_files_by_extension`, to exclude the files under `cmake`

kilograham:
i have force pushed your change.

if you are feeling keen @will-v-pi can you fix it to not complain if you have board definitions headers which aren't checked it ;-)

will-v-pi:
> i have force pushed your change.
> 
> if you are feeling keen @will-v-pi can you fix it to not complain if you have board definitions headers which aren't checked it ;-)

I think it already shouldn't complain about that, as the use of `git ls-files` means it only complains about files which are tracked in git?

PR #2937, opened 2026-05-14
Detect PICO_COMPILER automatically when PICO_TOOLCHAIN_PATH is set

kilograham:
In passing with other compiler/clib related changes, this PR removes the need to specify `PICO_COMPILER` when already specifying `PICO_TOOLCHAIN_PATH`

Thus `PICO_DEFAULT_COMPILER` is now only used if you have neither specified `PICO_COMPILER` or a toolchain file/path in which case you're basically looking in the users path environment at which point it's not a bad idea to pick one to look for (we could even help there, but that is a more involved change)


cmake/pico_pre_load_toolchain.cmake
```diff
@@ -1,4 +1,7 @@
 # PICO_CMAKE_CONFIG: PICO_TOOLCHAIN_PATH, Path to search for compiler, type=string, default=none (i.e. search system paths), group=build, docref=cmake-toolchain-config
+if (NOT PICO_TOOLCHAIN_PATH AND DEFINED ENV{PICO_TOOLCHAIN_PATH})
+    set(PICO_TOOLCHAIN_PATH $ENV{PICO_TOOLCHAIN_PATH})
```
will-v-pi, line 3: Should have a message, like the other bits like this (eg `PICO_SDK_PATH` in `pico_sdk_import.cmake`)
```suggestion
    set(PICO_TOOLCHAIN_PATH $ENV{PICO_TOOLCHAIN_PATH})
    message("Using PICO_TOOLCHAIN_PATH from environment ('${PICO_TOOLCHAIN_PATH}')")
```

cmake/pico_pre_load_toolchain.cmake
```diff
@@ -19,13 +22,30 @@ if (CMAKE_BUILD_TYPE STREQUAL "Default")
 endif()
 
 if (NOT (DEFINED PICO_COMPILER OR DEFINED CMAKE_TOOLCHAIN_FILE))
-    if (DEFINED PICO_DEFAULT_COMPILER)
-        pico_message("Defaulting compiler (PICO_COMPILER) to '${PICO_DEFAULT_COMPILER}' since not specified.")
+    if (PICO_TOOLCHAIN_PATH)
+        if (PICO_DEFAULT_GCC_COMPILER)
+            FILE(GLOB _COMPILER "${PICO_TOOLCHAIN_PATH}/bin/*gcc*")
```
will-v-pi: We currently support `PICO_TOOLCHAIN_PATH` being set to the `bin` folder of the toolchain, so this should also search for `${PICO_TOOLCHAIN_PATH}/*gcc*`
```suggestion
            file(GLOB _COMPILER "${PICO_TOOLCHAIN_PATH}/bin/*gcc*" "${PICO_TOOLCHAIN_PATH}/*gcc*")
```

(also lowercase `file` for consistency as that's what we use elsewhere)

cmake/pico_pre_load_toolchain.cmake
```diff
@@ -19,13 +22,30 @@ if (CMAKE_BUILD_TYPE STREQUAL "Default")
...
+            if (_COMPILER)
+                set(PICO_COMPILER ${PICO_DEFAULT_GCC_COMPILER})
+                message("Detected GCC so defaulting PICO_COMPILER to `${PICO_COMPILER}`")
+            endif()
+        endif()
+        if (NOT PICO_COMPILER AND PICO_DEFAULT_CLANG_COMPILER)
+            FILE(GLOB _COMPILER "${PICO_TOOLCHAIN_PATH}/bin/*llvm*")
```
will-v-pi: Same as above
```suggestion
            file(GLOB _COMPILER "${PICO_TOOLCHAIN_PATH}/bin/*llvm*" "${PICO_TOOLCHAIN_PATH}/*llvm*")
```

cmake/pico_pre_load_toolchain.cmake
kilograham: nah, not in this csae as these functions are called at weird times and we use the env var to pass the value around

cmake/pico_pre_load_toolchain.cmake
kilograham: not sure i knew that!

will-v-pi (approved):

cmake/pico_pre_load_toolchain.cmake
will-v-pi: Maybe not officially supported, but it works because we use `PATH_SUFFIXES` with `find_program` in `pico_find_compiler`, which searches the root `PICO_TOOLCHAIN_PATH` in addition to the `bin` suffix

cmake/pico_pre_load_toolchain.cmake
```diff
@@ -19,13 +22,30 @@ if (CMAKE_BUILD_TYPE STREQUAL "Default")
 endif()
 
 if (NOT (DEFINED PICO_COMPILER OR DEFINED CMAKE_TOOLCHAIN_FILE))
-    if (DEFINED PICO_DEFAULT_COMPILER)
-        pico_message("Defaulting compiler (PICO_COMPILER) to '${PICO_DEFAULT_COMPILER}' since not specified.")
+    if (PICO_TOOLCHAIN_PATH)
+        if (PICO_DEFAULT_GCC_COMPILER)
+            file(GLOB _COMPILER "${PICO_TOOLCHAIN_PATH}/bin/*gcc*" "${PICO_TOOLCHAIN_PATH}/*gcc*")
```
lurch, line 27: I know very little about compilers and toolchain paths, but would this produce false-positives if `PICO_TOOLCHAIN_PATH` was inadvertently set to `/usr/bin` ? (Or is that something that would never happen?)

kilograham: yeah it probably would, but we're still better than today, and it would be a kinda silly thing to do

Issue #2630, opened 2025-08-17
Colored status led not fully working


tazounet:
Hello,

I was using the code from "pio/ws2812" without problems and then I switch to the new "colored status led" library with the sdk 2.2.0.

1) "colored_status_led_set_on_with_color" don't change the color if the LED is already ON. Why not but it's not describe in the documentation.
2) I created a "colored_status_led_change_color" function to handle color change, but it was not working until I add a sleep between "colored_status_led_set_state" and "colored_status_led_set_on_with_color". This timing constraint is not in the documentation. Is it a bug ?
3) Adding a "colored_status_led_change_color" directly in the library could be a good addition :)

The working code:
```
bool colored_status_led_change_color(uint32_t color)
{
    colored_status_led_set_state(false);
    sleep_us(100);
    return colored_status_led_set_on_with_color(color);
}
```
Thanks

lurch:
Rather than adding a new `colored_status_led_change_color` function, does the original code work as expected if you change https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_status_led/status_led.c#L83 from
```c
        if (led_on && !colored_status_led_on) {
```
to
```c
        if (led_on) {
```
?

peterharperuk:
I think you need a delay after setting a ws2812 led (50us?). I'm not sure I like the idea of the API doing this. So setting the led color while it's already on might not work. We could document this however.

tazounet:
@lurch yes with this modification we can change the color without the need to set the LED off.

@peterharperuk 50us was not working, 100us seems reliable.

kilograham:
`colored_status_led_set_on_with_color()` should change the color; that is its intent; if it doesn't that is a bug.

it is also a requirement of the library to take care of this, as the way the LED behaves is its own problem.

Note that, a different PIO program or implementation could fix this (e.g. if the PIO program continually updated the color with the required delay)

kilograham:
fix merged into develop

PR #2938, opened 2026-05-14
Fix rapid changes to colored_status_led

kilograham:
WS2812 style LEDs require a reset time before setting the color againprior to this fix, setting color/on/off

in rapid succession would just send the updated values "down the string"

Whilst it would be nice to have the PIO program take care of updating the latest FIFO value after the relevant delay, that costs a fair amount of PIO instrution space, so I have decided to "fix" the issue on the CPU side.

Since the delays in question are non trivial (e.g. 50us), i've decided not to have the call block,  but instead set a timer when necessary (using the default alarm pool) - if that isn't available or the alarm add fails, then the caller blocks instead.

Note the implementation uses a spin lock, because it needs to work with multicore, but also to be fair you might expect to be able to set an LED from an IRQ

### DIFF
--- a/src/rp2_common/pico_status_led/status_led.c
+++ b/src/rp2_common/pico_status_led/status_led.c
@@ -44,7 +44,7 @@ static bool colored_status_led_on;
 #define PICO_COLORED_STATUS_LED_WS2812_FREQ 800000
 #endif
 
-// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=50, group=pico_status_led
+// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=100, group=pico_status_led
 #ifndef PICO_COLORED_STATUS_LED_RESET_DELAY_US
 #define PICO_COLORED_STATUS_LED_RESET_DELAY_US 100
 #endif
--- a/src/rp2_common/pico_status_led/status_led.c
+++ b/src/rp2_common/pico_status_led/status_led.c
@@ -44,9 +44,9 @@ static bool colored_status_led_on;
 #define PICO_COLORED_STATUS_LED_WS2812_FREQ 800000
 #endif
 
-// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=100, group=pico_status_led
+// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=50, group=pico_status_led
 #ifndef PICO_COLORED_STATUS_LED_RESET_DELAY_US
-#define PICO_COLORED_STATUS_LED_RESET_DELAY_US 100
+#define PICO_COLORED_STATUS_LED_RESET_DELAY_US 50
 #endif
 
 #ifndef PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
@@ -65,7 +65,7 @@ static int8_t alarm_pending;
 #define alarm_pending false
 #endif
 
-#define COLOR_STATUS_LED_UPDATE_TIME_US (1 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
+#define COLOR_STATUS_LED_UPDATE_TIME_US (2 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
 
 // Extract from 0xWWRRGGBB
 #define RED(c) (((c) >> 16) & 0xff)
@@ -118,6 +118,7 @@ static void set_ws2812(uint32_t value) {
 #else
         if (alarm_pending) {
             // we defer the set to the already waiting alarm
+            spin_unlock(spin_lock, save);
             break;
         } else {
             uint64_t now = time_us_64();
@@ -250,7 +251,7 @@ void status_led_deinit(void) {
         cancel_alarm(alarm_id);
         alarm_id = 0;
     }
-    alarm_pending = false;
+    alarm_pending = 0;
 #endif
     if (pio) {
         pio_remove_program_and_unclaim_sm(&ws2812_program, pio, sm, offset);
--- a/src/rp2_common/pico_status_led/status_led.c
+++ b/src/rp2_common/pico_status_led/status_led.c
@@ -44,7 +44,7 @@ static bool colored_status_led_on;
 #define PICO_COLORED_STATUS_LED_WS2812_FREQ 800000
 #endif
 
-// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED , type=int, default=50, group=pico_status_led
+// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=50, group=pico_status_led
 #ifndef PICO_COLORED_STATUS_LED_RESET_DELAY_US
 #define PICO_COLORED_STATUS_LED_RESET_DELAY_US 50
 #endif
@@ -75,7 +75,7 @@ static int8_t alarm_pending;
 
 static void unsafe_set_ws2812(uint32_t value, uint64_t now) {
     if (pio) {
-        pio_sm_drain_tx_fifo(pio, sm); // want to jump passed any previous queued values
+        pio_sm_drain_tx_fifo(pio, sm); // want to jump past any previous queued values
 #if PICO_COLORED_STATUS_LED_USES_WRGB
         // Convert to 0xWWGGRRBB
         pio_sm_put_blocking(pio, sm, WHITE(value) << 24 | GREEN(value) << 16 | RED(value) << 8 | BLUE(value));
--- a/src/rp2_common/pico_status_led/status_led.c
+++ b/src/rp2_common/pico_status_led/status_led.c
@@ -6,6 +6,8 @@
 
 #include "pico/status_led.h"
 
+#include "hardware/sync/spin_lock.h"
+
 #if PICO_STATUS_LED_AVAILABLE && defined(CYW43_WL_GPIO_LED_PIN) && !defined(PICO_DEFAULT_LED_PIN)
 #define STATUS_LED_USING_WL_GPIO 1
 #else
@@ -33,7 +35,8 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
 static bool colored_status_led_on;
 
 #if COLORED_STATUS_LED_USING_WS2812_PIO
-#include <hardware/pio.h>
+#include "hardware/pio.h"
+#include "pico/time.h"
 #include "ws2812.pio.h"
 
 // PICO_CONFIG: PICO_COLORED_STATUS_LED_WS2812_FREQ, Frequency per bit for the WS2812 colored status LED, type=int, default=800000, group=pico_status_led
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,35 +35,101 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+#include "hardware/pio.h"
+#include "pico/time.h"
 #include "ws2812.pio.h"
 
 // PICO_CONFIG: PICO_COLORED_STATUS_LED_WS2812_FREQ, Frequency per bit for the WS2812 colored status LED, type=int, default=800000, group=pico_status_led
 #ifndef PICO_COLORED_STATUS_LED_WS2812_FREQ
 #define PICO_COLORED_STATUS_LED_WS2812_FREQ 800000
 #endif
 
+// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED , type=int, default=50, group=pico_status_led
```
lurch:
```suggestion
// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=50, group=pico_status_led
```

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,35 +35,101 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
 // Extract from 0xWWRRGGBB
 #define RED(c) (((c) >> 16) & 0xff)
 #define GREEN(c) (((c) >> 8) & 0xff)
 #define BLUE(c) (((c) >> 0) & 0xff)
 #define WHITE(c) (((c) >> 24) && 0xff)
 
-bool set_ws2812(uint32_t value) {
+static void unsafe_set_ws2812(uint32_t value, uint64_t now) {
     if (pio) {
+        pio_sm_drain_tx_fifo(pio, sm); // want to jump passed any previous queued values
```
lurch:
```suggestion
        pio_sm_drain_tx_fifo(pio, sm); // want to jump past any previous queued values
```

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,35 +35,101 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+                //
+                // note we use alarm_pending separate from alarm_id, as alarm_id may be returned even if the
+                // alarm fires during the add_alarm_at. and don't use a boolean because if we fail
+                // to add the alarm, we don't know what has happened in between since we unlock the spin lock
+                // before adding the alarm since that is a slowish call
+                alarm_pending++;
+                spin_unlock(spin_lock, save);
+#if PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
+                alarm_id = add_alarm_at(next_safe_set_time, deferred_set_ws2812, NULL, true);
+                if (alarm_id > 0) break;
```
lurch: I guess the `break` here means that there's a code-path where the `spin_lock_blocking` on line 127 won't get hit. Does that matter?

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,35 +35,101 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
 #include "ws2812.pio.h"
 
 // PICO_CONFIG: PICO_COLORED_STATUS_LED_WS2812_FREQ, Frequency per bit for the WS2812 colored status LED, type=int, default=800000, group=pico_status_led
 #ifndef PICO_COLORED_STATUS_LED_WS2812_FREQ
 #define PICO_COLORED_STATUS_LED_WS2812_FREQ 800000
 #endif
 
+// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED , type=int, default=50, group=pico_status_led
+#ifndef PICO_COLORED_STATUS_LED_RESET_DELAY_US
+#define PICO_COLORED_STATUS_LED_RESET_DELAY_US 50
```
lurch: This define doesn't seem to actually be used anywhere??

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,35 +35,101 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
 #else
         // Convert to 0xGGRRBB00
         pio_sm_put_blocking(pio, sm, GREEN(value) << 24 | RED(value) << 16 | BLUE(value) << 8);
 #endif
-        return true;
+        next_safe_set_time = now + MINIMUM_WS2812_DELAY_US;
     }
+}
+
+static int64_t deferred_set_ws2812(__unused alarm_id_t id, __unused  void *user_data) {
```
lurch: Looks like this function could also be wrapped by `#if PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL` ?

src/rp2_common/pico_status_led/status_led.c
will-v-pi: Yes, the calculated `MINIMUM_WS2812_DELAY_US` seems to be used instead.

If it wants to be configurable, could use `#define PICO_COLORED_STATUS_LED_RESET_CYCLES` which defaults to `PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24`, and then use `PICO_COLORED_STATUS_LED_RESET_CYCLES` in the `MINIMUM_WS2812_DELAY_US` calculation?

Equally, could just remove this define, if the value is expected to be fixed

lurch:
I guess the PR addresses [this](https://github.com/raspberrypi/pico-sdk/issues/2630#issuecomment-3195923095) comment :slightly_smiling_face:

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,35 +35,101 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+static uint32_t next_value;
+static uint64_t next_safe_set_time;
+#if PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
+static alarm_id_t alarm_id;
+static int8_t alarm_pending;
+#else
+#define alarm_id 0
+#endif
+
+#define MINIMUM_WS2812_DELAY_US (1 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
```
will-v-pi: All the boards I have on my desk require 3x this delay value to work correctly - where did this value come from?

My test is just a modified color_blink example to send colours as often as possible - when working it is white (ie cycling through red-green-blue), when not working it is just red (the first colour set)
```c
#include "pico/stdlib.h"
#include "pico/status_led.h"

int main() {
    status_led_init();
    uint32_t count = 0;
    while (true) {
        // flash red then green then blue
        uint32_t color = PICO_COLORED_STATUS_LED_COLOR_FROM_RGB(count % 3 == 0 ? 0xaa : 0, count % 3 == 1 ? 0xaa : 0, count % 3 == 2 ? 0xaa : 0);
        colored_status_led_set_on_with_color(color);
        count++;
    }
    status_led_deinit();
}
```

src/rp2_common/pico_status_led/status_led.c
kilograham: oops

src/rp2_common/pico_status_led/status_led.c
kilograham: weird ok - i mean 50us is what the WS2812 datasheet says - i'm fine with 150

src/rp2_common/pico_status_led/status_led.c
kilograham: suresure

src/rp2_common/pico_status_led/status_led.c
kilograham: actually i changed it to 100 as we weren't including the actual color setting time before

src/rp2_common/pico_status_led/status_led.c
kilograham: yeah, was a bug - they are indeed supposed to be added together

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -165,6 +245,13 @@ void status_led_deinit(void) {
     status_led_context = NULL;
 #endif
 #if COLORED_STATUS_LED_USING_WS2812_PIO
+#if PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
+    if (alarm_id > 0) {
+        cancel_alarm(alarm_id);
+        alarm_id = 0;
+    }
+    alarm_pending = false;
```
lurch:
```suggestion
    alarm_pending = 0;
```
? (It's declared as `static int8_t alarm_pending;`)

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,36 +35,114 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+static uint32_t next_value;
+static uint64_t next_safe_set_time;
+#if PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
+static alarm_id_t alarm_id;
+static int8_t alarm_pending;
+#else
+#define alarm_pending false
+#endif
+
+#define COLOR_STATUS_LED_UPDATE_TIME_US (1 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
```
lurch: Super-minor nitpick: this could be `COLORED_STATUS_LED_UPDATE_TIME_US` for consistency.

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,36 +35,114 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+            break;
+        } else {
+            spin_unlock(spin_lock, save);
+            busy_wait_until(next_safe_set_time);
+            save = spin_lock_blocking(spin_lock);
+        }
+#else
+        if (alarm_pending) {
+            // we defer the set to the already waiting alarm
+            break;
```
will-v-pi: This code path never calls `spin_unlock(spin_lock, save);`, and it gets hit if you call this a second time before `next_safe_set_time`, as `alarm_pending` is already set.
```suggestion
            spin_unlock(spin_lock, save);
            break;
```

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,36 +35,114 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
 
 // PICO_CONFIG: PICO_COLORED_STATUS_LED_WS2812_FREQ, Frequency per bit for the WS2812 colored status LED, type=int, default=800000, group=pico_status_led
 #ifndef PICO_COLORED_STATUS_LED_WS2812_FREQ
 #define PICO_COLORED_STATUS_LED_WS2812_FREQ 800000
 #endif
 
+// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=100, group=pico_status_led
+#ifndef PICO_COLORED_STATUS_LED_RESET_DELAY_US
+#define PICO_COLORED_STATUS_LED_RESET_DELAY_US 100
+#endif
```
will-v-pi: Setting this to 51 is working alright for me, now it's actually being added in - 50 works some of the time, but not consistently. Maybe this should be 50, and then add 1us extra to `COLOR_STATUS_LED_UPDATE_TIME_US`

```suggestion
// PICO_CONFIG: PICO_COLORED_STATUS_LED_RESET_DELAY_US, Required reset delay in microseconds for the WS2812 colored status LED, type=int, default=50, group=pico_status_led
#ifndef PICO_COLORED_STATUS_LED_RESET_DELAY_US
#define PICO_COLORED_STATUS_LED_RESET_DELAY_US 50
#endif
```

src/rp2_common/pico_status_led/status_led.c
will-v-pi: See above
```suggestion
#define COLOR_STATUS_LED_UPDATE_TIME_US (2 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
```

src/rp2_common/pico_status_led/status_led.c
kilograham: oops - yeah i moved the unlocks into the code paths and missed one, thank you

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,36 +35,115 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+static uint32_t next_value;
+static uint64_t next_safe_set_time;
+#if PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
+static alarm_id_t alarm_id;
+static int8_t alarm_pending;
+#else
+#define alarm_pending false
+#endif
+
+#define COLOR_STATUS_LED_UPDATE_TIME_US (2 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
```
kilograham:
```suggestion
// 2 + is empirical and allows WS2812 reset delay of 50us to work 
#define COLORED_STATUS_LED_UPDATE_TIME_US (2 + (1000000 * (PICO_COLORED_STATUS_LED_USES_WRGB ? 32 : 24)) / PICO_COLORED_STATUS_LED_WS2812_FREQ)
```

src/rp2_common/pico_status_led/status_led.c
```diff
@@ -33,36 +35,115 @@ static uint32_t colored_status_led_on_color = PICO_DEFAULT_COLORED_STATUS_LED_ON
...
+        pio_sm_drain_tx_fifo(pio, sm); // want to jump past any previous queued values
 #if PICO_COLORED_STATUS_LED_USES_WRGB
         // Convert to 0xWWGGRRBB
         pio_sm_put_blocking(pio, sm, WHITE(value) << 24 | GREEN(value) << 16 | RED(value) << 8 | BLUE(value));
 #else
         // Convert to 0xGGRRBB00
         pio_sm_put_blocking(pio, sm, GREEN(value) << 24 | RED(value) << 16 | BLUE(value) << 8);
 #endif
-        return true;
+        next_safe_set_time = now + COLOR_STATUS_LED_UPDATE_TIME_US + PICO_COLORED_STATUS_LED_RESET_DELAY_US;
```
kilograham:
```suggestion
        next_safe_set_time = now + COLORED_STATUS_LED_UPDATE_TIME_US + PICO_COLORED_STATUS_LED_RESET_DELAY_US;
```

will-v-pi (approved):
All working now - thanks

PR #2942, opened 2026-05-15
Small test fixups

will-v-pi:
Tests with stdio_usb need the final output to be flushed before exiting (which is done by `stdio_deinit_all`), otherwise they don't always print `PASSED`

`hardware_sync_spin_lock_test` takes too long on RP2040, so increase the timeout as it was close to timing out on RP2350 anyway

### DIFF
--- a/test/pico_test/include/pico/test.h
+++ b/test/pico_test/include/pico/test.h
@@ -55,10 +55,16 @@ but not sure that is implemented yet.
                                     }
 
 // Calls stdio_deinit_all before exiting, to avoid losing final output
-#define PICOTEST_END_TEST()       if (picotest_error_code != 0)                                     \
-                                      {printf("%s: Failed\n", picotest_description); stdio_deinit_all(); return -1;}  \
-                                  else                                                              \
-                                      {printf("%s: Success\n", picotest_description); puts("PASSED"); stdio_deinit_all(); return 0;}
+#define PICOTEST_END_TEST()         if (picotest_error_code != 0) {                                 \
+                                        printf("%s: Failed\n", picotest_description);               \
+                                        stdio_deinit_all();                                         \
+                                        return -1;                                                  \
+                                    } else {                                                        \
+                                        printf("%s: Success\n", picotest_description);              \
+                                        puts("PASSED");                                             \
+                                        stdio_deinit_all();                                         \
+                                        return 0;                                                   \
+                                    }
 
 
 #endif
--- a/test/pico_test/include/pico/test.h
+++ b/test/pico_test/include/pico/test.h
@@ -54,10 +54,17 @@ but not sure that is implemented yet.
                                         return -1;                                                  \
                                     }
 
+#if LIB_PICO_STDIO_USB  // Must deinit before exiting, to avoid losing final output
+#define PICOTEST_END_TEST()       if (picotest_error_code != 0)                                     \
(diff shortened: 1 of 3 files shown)


test/pico_test/include/pico/test.h
```diff
@@ -54,10 +54,17 @@ but not sure that is implemented yet.
                                         return -1;                                                  \
                                     }
 
+#if LIB_PICO_STDIO_USB  // Must deinit before exiting, to avoid losing final output
```
peterharperuk: Why not always call stdio_deinit_all even when LIB_PICO_STDIO_USB=0?

will-v-pi: Added

peterharperuk (dismissed):

test/pico_test/include/pico/test.h
will-v-pi: Ah, `host` doesn't have an implementation for `stdio_deinit_all` - I will add one that does nothing

test/pico_test/include/pico/test.h
```diff
@@ -54,10 +54,11 @@ but not sure that is implemented yet.
                                         return -1;                                                  \
                                     }
 
+// Calls stdio_deinit_all before exiting, to avoid losing final output
 #define PICOTEST_END_TEST()       if (picotest_error_code != 0)                                     \
-                                      {printf("%s: Failed\n", picotest_description); return -1;}  \
+                                      {printf("%s: Failed\n", picotest_description); stdio_deinit_all(); return -1;}  \
                                   else                                                              \
-                                      {printf("%s: Success\n", picotest_description); puts("PASSED"); return 0;}
+                                      {printf("%s: Success\n", picotest_description); puts("PASSED"); stdio_deinit_all(); return 0;}
```
lurch: Now that there's multiple function calls, perhaps it's worth splitting out over multiple lines, e.g. like is done with `PICOTEST_CHECK_AND_ABORT` ?
I.e.
```c
#define PICOTEST_END_TEST()       if (picotest_error_code != 0) {                    \
                                      printf("%s: Failed\n", picotest_description);  \
                                      stdio_deinit_all();                            \
                                      return -1;                                     \
                                  } else {                                           \
                                      printf("%s: Success\n", picotest_description); \
                                      puts("PASSED");                                \
                                      stdio_deinit_all();                            \
                                      return 0;                                      \
                                  }
```

lurch:
IIRC (when I was looking at one of Graham's PRs) there are also some tests that call `puts` directly rather than using the `PICOTEST_` macros. Do those tests also need to be updated to do an explicit `stdio_deinit_all()`?

Ahh, see my comment here https://github.com/raspberrypi/pico-sdk/pull/2897#issuecomment-4305356543

will-v-pi:
> IIRC (when I was looking at one of Graham's PRs) there are also some tests that call `puts` directly rather than using the `PICOTEST_` macros. Do those tests also need to be updated to do an explicit `stdio_deinit_all()`?
> 
> Ahh, see my comment here [#2897 (comment)](https://github.com/raspberrypi/pico-sdk/pull/2897#issuecomment-4305356543)

Those tests don't use `stdio_usb`, so don't need to call `stdio_deinit_all()`, so I won't update them

test/pico_test/include/pico/test.h
will-v-pi: Done

peterharperuk (approved):

kilograham (approved):

PR #2943, opened 2026-05-16
Fix incorrect doc for stdio_put_string

### LINKED ISSUE #2818: Docs: unclear description for `stdio_put_string`
The function doc states "This method returns as soon as input is available, but more characters may be returned up to the end of the buffer", but that doesn't make sense as this is an output-only function, I believe it might have been mistakenly pasted from the documentation of an input procedure such as `stdio_get_until`.

The function implementation looks like it won't return early if buffers are full, so it would be good to make that clear, since there is a returned `len`.

anoshyn:
## Summary

The doxygen for `stdio_put_string` in `src/rp2_common/pico_stdio/include/pico/stdio.h` reads *"This method returns as soon as input is available, but more characters may be returned up to the end of the buffer."* That text was copy-pasted from an input-reading helper (e.g. `stdio_get_until`); `stdio_put_string` is output-only and never returns early on its own. The wording is actively misleading because the function has an `int` return value, so a reader could plausibly assume some short-write behavior.

Rewrite the description to match the implementation in `stdio.c:92-114`:

- writes the buffer to every registered driver in turn (no early return);
- documents the `len == -1` → `strlen` convention;
- documents `cr_translation` behavior (applies to any appended newline too);
- documents the `0`-return edge case when `PICO_STDIO_IGNORE_NESTED_STDOUT` is set and a concurrent stdout write is in progress.

No behavior change; doxygen only.

Fixes #2818.

## Test plan

- [x] `git diff` — only `src/rp2_common/pico_stdio/include/pico/stdio.h` is touched; only the doxygen block for `stdio_put_string`.
- [ ] CI builds pass (header still compiles cleanly).
- [x] Doxygen renders without warnings for `stdio_put_string`.

### DIFF
--- a/src/rp2_common/pico_stdio/include/pico/stdio.h
+++ b/src/rp2_common/pico_stdio/include/pico/stdio.h
@@ -172,17 +172,19 @@ void stdio_set_chars_available_callback(void (*fn)(void*), void *param);
  */
 int stdio_get_until(char *buf, int len, absolute_time_t until);
 
-/*! \brief Prints a buffer to stdout with optional newline and carriage return insertion
+/*! \brief Prints a buffer to stdout, optionally appending a newline
  * \ingroup pico_stdio
  *
- * This method returns as soon as input is available, but more characters may
- * be returned up to the end of the buffer.
+ * Writes \p s to every registered stdio driver. The call returns once
+ * each driver has accepted the bytes; it may block if a driver applies
+ * back-pressure.
  *
  * \param s the characters to print
- * \param len the length of s
- * \param newline true if a newline should be added after the string
- * \param cr_translation true if line feed to carriage return translation should be performed
- * \return the number of characters written
+ * \param len the length of \p s, or `-1` to compute it with `strlen`
+ * \param newline true if a newline should be appended after the string
+ * \param cr_translation true if line-feed to carriage-return translation should be performed
+ * \return the number of characters from \p s that were written (excluding any appended newline);
+ *         `0` if a concurrent stdout write was in progress and `PICO_STDIO_IGNORE_NESTED_STDOUT` is set
  */
 int stdio_put_string(const char *s, int len, bool newline, bool cr_translation);
 
--- a/src/rp2_common/pico_stdio/include/pico/stdio.h
+++ b/src/rp2_common/pico_stdio/include/pico/stdio.h
@@ -176,8 +176,7 @@ int stdio_get_until(char *buf, int len, absolute_time_t until);
  * \ingroup pico_stdio
  *
  * Writes \p s to every registered stdio driver. The call returns once
- * each driver has accepted the bytes; it may block if a driver applies
- * back-pressure.
+ * each driver has accepted the bytes
  *
  * \param s the characters to print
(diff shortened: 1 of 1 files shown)


anoshyn:
@kilograham @lurch when you have a moment, would you mind taking a look? This is a doxygen-only change to `stdio_put_string` that fixes the description copy-paste flagged in #2818. Happy to adjust wording if you'd prefer different phrasing. Thanks!

lurch:
Hi @anoshyn , as requested in other PRs, if you could update **all** of your pico-sdk PRs to remove the "Co-Authored-By: Claude" lines, that'd be helpful :+1:

src/rp2_common/pico_stdio/include/pico/stdio.h
```diff
@@ -175,14 +175,20 @@ int stdio_get_until(char *buf, int len, absolute_time_t until);
 /*! \brief Prints a buffer to stdout with optional newline and carriage return insertion
  * \ingroup pico_stdio
  *
- * This method returns as soon as input is available, but more characters may
- * be returned up to the end of the buffer.
+ * Writes the first \p len characters of \p s to every registered stdio driver in
+ * turn. If \p len is `-1`, the length is taken from `strlen(s)`. When \p newline
+ * is true, a single line-feed character is appended after the buffer. When
+ * \p cr_translation is true, each line-feed written (including any appended
+ * newline) is converted to a carriage-return + line-feed pair. This call does
```
lurch: Feels like much of this is just duplicating information that's already covered in the `\param` descriptions below?

anoshyn:
Force-pushed: dropped the `Co-Authored-By` trailer, and trimmed the prose body so it no longer restates the `\param` lines - now just a brief plus two short sentences covering what isn't in the params (full-buffer write, may block under back-pressure). New HEAD: c2a8ef1e.

kilograham (dismissed):

src/rp2_common/pico_stdio/include/pico/stdio.h
```diff
@@ -172,17 +172,19 @@ void stdio_set_chars_available_callback(void (*fn)(void*), void *param);
  */
 int stdio_get_until(char *buf, int len, absolute_time_t until);
 
-/*! \brief Prints a buffer to stdout with optional newline and carriage return insertion
+/*! \brief Prints a buffer to stdout, optionally appending a newline
  * \ingroup pico_stdio
  *
- * This method returns as soon as input is available, but more characters may
- * be returned up to the end of the buffer.
+ * Writes \p s in full to every registered stdio driver. The call returns once
```
lurch: I think I'd be tempted to do:
```suggestion
 * Writes \p s to every registered stdio driver. The call returns once
```
as you could give a `len` shorter than `s` if you didn't want to write all of it. (But this is optional, given that Graham has already approved this PR.)

anoshyn: Good catch — applied in 3dc5e04. "in full" did contradict the `\param len` line right below it.

kilograham (approved):

PR #2944, opened 2026-05-16
Make boot/picobin.h self-contained

### LINKED ISSUE #2797: Including "boot/picobin.h" triggers include ordering error it caused itself
Hey guys, this one just needs a one-line fix somewhere:

#### Problem

When using 
```c
#include "boot/picobin.h"
```

before any other includes, this compilation error is emitted:
```
/usr/local/pico-sdk/src/rp2350/pico_platform/include/pico/platform.h:20:2: error: #error pico/platform.h should not be included directly; include pico.h instead
   20 | #error pico/platform.h should not be included directly; include pico.h instead
```

This is initially confusing, however, because the user obviously didn't try to include `pico/platform.h` directly. Furthermore, including `pico.h` directly instead (the solution indicated) does not actually cause the inclusion of `boot/picobin.h`.

#### Expected Outcome

If `boot/picobin.h` is intended to be included directly by the user, it should include its own dependencies. (As it's included directly [here](https://github.com/raspberrypi/pico-examples/blob/4c3a3dc0196dd426fddd709616d0da984e027bab/flash/partition_info/partition_info.c#L10) I'm assuming that's the case.) Otherwise, it should emit an error identifying a header that can be included that includes `boot/picobin.h`, after including its dependencies.

anoshyn:
## Summary

`#include "boot/picobin.h"` before any other SDK header errors out with:
```
pico/platform.h:20: #error pico/platform.h should not be included directly; include pico.h instead
```

`boot/picobin.h` was pulling in `pico/platform.h` directly (gated only by `NO_PICO_PLATFORM`), which works only when something else (typically `pico.h`) has already set `_PICO_H`. The diagnostic blamed `pico/platform.h`, while the actual offender was the non-obvious indirect include via `boot/picobin.h` — a confusing experience flagged in #2797.

The header only uses one platform.h facility: the `_u()` literal-suffix macro. The same `_u()` fallback already existed in the `NO_PICO_PLATFORM` branch of this header. Promote that fallback to the unconditional definition and drop the `pico/platform.h` include. Add `<stdint.h>` next to the existing `<stdbool.h>` in the C-only block so `uint32_t` is available standalone.

This keeps `boot/picobin.h` at the lightweight "common" tier (no new heavy SDK dependency) and makes it includable in any order.

Fixes #2797.

## Test plan

- [x] Repro reproduces on develop: a TU containing only `#include "boot/picobin.h"` fails to compile, citing the `pico/platform.h should not be included directly` error.
- [x] After the change, the same TU compiles.
- [x] `__ASSEMBLER__`-mode preprocessing still works (`_u(x)` falls through to the asm branch).
- [x] Full SDK build remains clean (`cmake --build build`) for the default `PICO_PLATFORM=rp2040 PICO_BOARD=pico` config. Existing internal consumers (`pico_cyw43_driver`, `pico_crt0`, `pico_bootrom`, `cyw43_partition_firmware`) include picobin.h alongside `pico.h`/`pico/bootrom.h`, so they already pull in platform.h transitively and are unaffected.
- [ ] CI to confirm other board / RP2350 / RISC-V configurations.

### DIFF
--- a/src/common/boot_picobin_headers/include/boot/picobin.h
+++ b/src/common/boot_picobin_headers/include/boot/picobin.h
@@ -7,15 +7,11 @@
 #ifndef _BOOT_PICOBIN_H
 #define _BOOT_PICOBIN_H
 
-#ifndef NO_PICO_PLATFORM
-#include "pico/platform.h"
-#else
 #ifndef _u
 #ifdef __ASSEMBLER__
 #define _u(x) x
 #else
 #define _u(x) x ## u
-#endif
 #endif
 #endif
 
@@ -149,6 +145,7 @@
 #ifndef __ASSEMBLER__
 
 #include <stdbool.h>
+#include <stdint.h>
 
 typedef struct {
     // these must all be word aligned


anoshyn:
@kilograham @lurch when you have a moment, would you mind taking a look? This is a small standalone-include fix for `boot/picobin.h` flagged in #2797. Verified that the existing internal consumers still build clean. Happy to take a different approach if you'd prefer including `pico.h` instead. Thanks!

kilograham:
please don't tag people in issues

will-v-pi (approved):
LGTM, and still works with picotool

src/common/boot_picobin_headers/include/boot/picobin.h
```diff
@@ -7,17 +7,16 @@
 #ifndef _BOOT_PICOBIN_H
 #define _BOOT_PICOBIN_H
 
-#ifndef NO_PICO_PLATFORM
-#include "pico/platform.h"
-#else
+// boot/picobin.h is intended to be includable standalone (e.g. before pico.h),
+// so it does not pull in pico/platform.h. The only facility it needs from
+// platform.h is the _u() literal-suffix macro, which we provide locally below.
```
lurch: This maybe feels like a comment that belongs in the commit-message, rather than in the file itself? :thinking: 
(i.e. if you encountered this comment _without_ the context of this PR, you might wonder what it was talking about.)

will-v-pi (changes requested):
Please remove the `Co-Authored-By: Claude` from the commit message

anoshyn:
Force-pushed: dropped the `Co-Authored-By` trailer, and per the inline note removed the standalone-include rationale from picobin.h itself - the context now lives only in the commit message body. New HEAD: 636f2abf.

will-v-pi (approved):

will-v-pi:
This issue also occurs with boot/picoboot.h

PR #2945, opened 2026-05-16
hardware_pio: clarify pio_sm_restart docs (PC, running state)

### LINKED ISSUE #2844: C SDK Doc: pio_sm_restart, pc is not mentioned
In C SDK at the `pio_sm_restart()` function description doesn't mention PC and running state of the SM.
In the datasheet of 2040 it is stated that PC is not affected. There is nothing in 2350 datasheet about the the PC. Do we know anything about the PC at RP2350 when the function is called?

None of the document mentions the running state of the state machine (the function name ends with `_restart` so I think it would be a safe bet assuming the statemachine will be running after the call - but it won't).

If PC is not affected at RP2350 either, then:
- please syncronize the 2350 and 2040 datasheets at the PIO:List of registers:CTRL Register:SM_RESTART
- update the description in the C SDK (in document and at code comment)

If running state also not affected, please mention that also in the C SDK document, source comments and in Datasheet also. That would be helpful.

anoshyn:
## Summary

Per #2844, the doxygen for `pio_sm_restart()` doesn't mention the program counter or the state machine's enable/running state, and includes a couple of inaccuracies.

This PR rewrites the doxygen for `pio_sm_restart` and its multi-SM sibling `pio_restart_sm_mask` against the authoritative wording from the `PIO_CTRL.SM_RESTART` field description in `hardware/regs/pio.h` (same text on both RP2040 and RP2350):

**Fixed inaccuracies:**
- The clock divider counter is NOT cleared by `SM_RESTART` — that's a separate bit (`CLKDIV_RESTART`, exposed via `pio_sm_clkdiv_restart`). The old doc claimed it was cleared.
- "ISR" was ambiguous (input shift register vs interrupt service routine); now spelled out.
- The `@param mask` description on `pio_restart_sm_mask` said "modify the enabled state of", but the function does not touch `SM_ENABLE`.

**Filled in:**
- Explicit note that the state machine's enable/running state is preserved (the issue's main ask — the `_restart` suffix is misleading without this).
- Explicit note that the program counter is preserved, with a pointer to `pio_sm_exec` for the "set PC after restart" use case.

PC behavior is documented as "not affected" on RP2040 in the datasheet; the RP2350 datasheet's SM_RESTART description omits the PC from its not-affected list, but the underlying hardware bit is identical and the rest of the description matches. Happy to soften the wording if maintainers know of a behavior difference on RP2350.

Fixes #2844.

## Test plan

- [x] `cmake --build build --target docs` produces no new warnings; specifically nothing on the changed block in `src/rp2_common/hardware_pio/include/hardware/pio.h`.
- [x] Rendered HTML inspected: cross-references to `pio_sm_set_enabled`, `pio_sm_clkdiv_restart`, `pio_sm_exec` resolve; backtick code spans render as `<code>`.
- [x] Full SDK build still succeeds (`cmake --build build`); no code change.
- [ ] CI to confirm broader configurations (RP2350 / RISC-V / etc.).

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -1108,11 +1108,19 @@ static inline void pio_set_sm_multi_mask_enabled(PIO pio, uint32_t mask_prev, ui
 }
 #endif
 
-/*! \brief Restart a state machine with a known state
+/*! \brief Restart a state machine with a known internal state
  *  \ingroup hardware_pio
  *
- * This method clears the ISR, shift counters, clock divider counter
- * pin write flags, delay counter, latched EXEC instruction, and IRQ wait condition.
+ * Clears the following on the selected state machine: input and output shift
+ * counters, the contents of the input shift register, the delay counter, the
+ * waiting-on-IRQ state, any stalled instruction written to `SMx_INSTR` or run
+ * by `OUT/MOV EXEC`, and any pin write left asserted due to `OUT_STICKY`.
+ *
+ * Not affected: the state machine's enable/running state (see
+ * \ref pio_sm_set_enabled), its program counter, the contents of its output
+ * shift register, and its X and Y scratch registers. To restart the clock
+ * divider use \ref pio_sm_clkdiv_restart; to set the program counter to a
+ * specific value, follow this call with a `JMP` via \ref pio_sm_exec.
  *
  * \param pio The PIO instance; e.g. \ref pio0, \ref pio1 etc.
  * \param sm State machine index (0..3)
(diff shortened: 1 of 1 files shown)


anoshyn:
Force-pushed to drop the `Co-Authored-By` trailer. New HEAD: 392cd87c.

lurch:
The change here matches what is stated for the `PIO: CTRL.SM_RESTART` field in the RP2040 Datasheet:
Screenshot from 2026-05-27 13-16-38

and RP2350 Datasheet:
Screenshot from 2026-05-27 13-19-21

but Luke probably ought to have the final-approval on this PR.

anoshyn:
@lurch Thanks for checking this against both datasheets. I’ll wait for Luke’s final call.

Wren6991 (approved):
I reviewed the new doxygen against the hardware implementation and it all seems to match. Presumably that's where I looked when I wrote the register descriptions too.

Thanks, this is a useful correction/clarification.

PR #2951, opened 2026-05-16
pico_btstack: allow pico_btstack_make_gatt_header to be called more than once per target

### LINKED ISSUE #2829: Can't call pico_btstack_make_gatt_header more than once for a target
Fix...

set(TARGET_GATT "${TARGET_LIB}_${GATT_NAME}_gatt_header")

anoshyn:
## Summary

Fixes #2829. `pico_btstack_make_gatt_header()` (in `src/rp2_common/pico_btstack/CMakeLists.txt`) builds its internal custom-target name as `"${TARGET_LIB}_gatt_header"`, which collides on the second call with the same `TARGET_LIB`. Configure fails with:

```
CMake Error at .../pico_btstack/CMakeLists.txt:336 (add_custom_target):
  add_custom_target cannot create target "${TARGET_LIB}_gatt_header" because
  another target with the same name already exists.
```

This prevents a single executable / library from owning more than one `.gatt` file. The reporter even proposed the one-line fix - disambiguate the target name with the `.gatt` file's base name (already extracted into `GATT_NAME` a few lines above):

```diff
-set(TARGET_GATT "${TARGET_LIB}_gatt_header")
+set(TARGET_GATT "${TARGET_LIB}_${GATT_NAME}_gatt_header")
```

`GATT_BINARY_DIR` and the include directory added via `target_include_directories` are derived from `TARGET_GATT`, so each `.gatt` file also gets its own output directory automatically.

## End-to-end verification

Out-of-tree consumer (rp2350 / pico2_w) that calls the function twice on a single executable target:

```cmake
add_executable(gatt_double main.c)
target_link_libraries(gatt_double PRIVATE pico_stdlib)
pico_btstack_make_gatt_header(gatt_double PRIVATE a.gatt)
pico_btstack_make_gatt_header(gatt_double PRIVATE b.gatt)
```

| | Result |
|---|---|
| `origin/develop` (no fix) | configure fails with the duplicate-target error above |
| this branch (fix) | configure succeeds; building `gatt_double_a_gatt_header` and `gatt_double_b_gatt_header` runs `compile_gatt.py` twice, producing `generated/gatt_double_a_gatt_header/a.h` and `generated/gatt_double_b_gatt_header/b.h` with the respective `.gatt` contents |

## Compatibility

The custom-target name and its derived `GATT_BINARY_DIR` change shape, but neither is part of the public API - they're internal to the function and only consumed by `target_include_directories` and `add_dependencies` inside the same function body. The generated header name (`${GATT_NAME}.h`) is unchanged, so `#include \"foo.h\"` continues to work the same way.

## Test plan

- [x] Repro the failure on `origin/develop` with an out-of-tree consumer that calls the function twice (got the duplicate-target error).
- [x] Apply the fix; same consumer configures cleanly and both `_gatt_header` sub-targets build, producing two distinct `*.h` files.
- [x] Full SDK build clean: `cmake --build build` for `PICO_PLATFORM=rp2040 PICO_BOARD=pico` (no in-tree consumer of this function, so this confirms there is no syntax regression).
- [ ] CI to confirm broader configurations.


peterharperuk:
Change looks ok. Please remove "Co-Authored-By: Claude" line from commit message

anoshyn:
Force-pushed to drop the `Co-Authored-By` trailer. New HEAD: 9779911e.

kilograham (approved):

kilograham:
@peterharperuk please merge if u r haopy

peterharperuk (approved):

PR #2954, opened 2026-05-19
Add Soldered NULA Ethernet W55RP20 Board definition

### LINKED ISSUE #2953: Add Soldered NULA Ethernet W55RP20
Add pin definitions for the upcoming Soldered NULA Ethernet W55RP20 board

JosipKuci:
Fixes #2953 
Adds a pin definition header file for the upcoming Soldered NULA Ethernet W55RP20 board.
As the board is currently in development, i will be attaching an image of the device as well as its schematic below (note that both the image and schematic refer to it as Soldered NULA Ether, this will be changed to Ethernet before release):

[NULA Ethernet W55RP20 Schematic.pdf](https://github.com/user-attachments/files/28012540/NULA.Ethernet.W55RP20.Schematic.pdf)


lurch:
I think this is the first board-header PR we've had using the W55RP20 chip :smiley: 
According to the datasheet that I was able to find at https://docs.wiznet.io/Product/Chip/MCU/W55RP20/documents_md (unfortunately the datasheet link at https://wiznet.io/products/ionic-chips/w55rp20 isn't working), it uses GPIOs 20 to 25 inclusive on the RP2040 die to communicate with the W5500 die. Perhaps these pins ought to be defined in this board-header, to make it clearer that those GPIO pins aren't available for any other purpose?

JosipKuci:
> I think this is the first board-header PR we've had using the W55RP20 chip 😃 According to the datasheet that I was able to find at https://docs.wiznet.io/Product/Chip/MCU/W55RP20/documents_md (unfortunately the datasheet link at https://wiznet.io/products/ionic-chips/w55rp20 isn't working), it uses GPIOs 20 to 25 inclusive on the RP2040 die to communicate with the W5500 die. Perhaps these pins ought to be defined in this board-header, to make it clearer that those GPIO pins aren't available for any other purpose?

Thats great to hear 😁! I have added the defines in the newest commit, let me know if a different naming scheme is required!

src/boards/include/boards/soldered_nula_ethernet_w55rp20.h
```diff
@@ -0,0 +1,124 @@
...
+#ifndef PICO_SD_DAT3_PIN
+#define PICO_SD_DAT3_PIN 13 // DAT3 of the SD card is the chip select pin
+#endif
+#ifndef PICO_SD_DAT_PIN_COUNT
+#define PICO_SD_DAT_PIN_COUNT 1
+#endif
+
+// The following pins are used internally between 
+// the RP2040 and W5500 chip and should NOT be used
+#ifndef PICO_INTERNAL_SPI_CS
```
lurch: Maybe `W55RP20_INTERNAL_SPI_CS` is a better name? Although the datasheet says that this is a SPI bus implemented by PIO rather than a hardware SPI interface, so perhaps `W55RP20_INTERNAL_PIO_SPI_CS` would be better? (because GPIO 20 is hardware `SPI0 RX` rather than a SPI CS pin)

...although thinking on this some more, I guess this is _quite_ similar to the CYW43 chip used on [Pico W](https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/pico_w.h#L123) and [Pico 2 W](https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/pico2_w.h#L123) (which is also connected via a PIO-implemented SPI interface), so perhaps it makes sense to follow that naming scheme? (As I guess this would also allow the same pin-naming scheme to be used for an RP2040 connected to an external W5500 chip :slightly_smiling_face: )
So maybe something like `W5500_PIN_ETH_DYNAMIC`, `W5500_DEFAULT_PIN_ETH_DATA_OUT`, `W5500_DEFAULT_PIN_ETH_DATA_IN`, `W5500_DEFAULT_PIN_ETH_CLOCK`, `W5500_DEFAULT_PIN_ETH_CS`, etc.?

This is just a suggestion; you have better familiarity with the W55RP20 chip than I do, so feel free to choose different names if they'd make more sense.

JosipKuci: Changed them to what i think would best describe them and keep it modular so that it could be used for boards which dont have them internally. Kept the SPI naming scheme since that is also used in the datasheet. Hope its ok!

src/boards/include/boards/soldered_nula_ethernet_w55rp20.h
```diff
@@ -0,0 +1,124 @@
...
+#endif
+#ifndef PICO_SD_DAT3_PIN
+#define PICO_SD_DAT3_PIN 13 // DAT3 of the SD card is the chip select pin
+#endif
+#ifndef PICO_SD_DAT_PIN_COUNT
+#define PICO_SD_DAT_PIN_COUNT 1
+#endif
+
+// The following pins are used internally between 
+// the RP2040 and W5500 chip and should NOT be used
```
lurch:
```suggestion
// the RP2040 and W5500 chip and should NOT be used for other purposes
```

lurch:
Small comment suggestion, but otherwise this LGTM! (and will also work with the `_DEFAULT_PIN_` change that I've made in #2959 )

JosipKuci:
> Small comment suggestion, but otherwise this LGTM! (and will also work with the `_DEFAULT_PIN_` change that I've made in #2959 )

Thank you for the fast responses! The clearer comment has been added in the newest commit.

lurch (approved):

JosipKuci:
Thank you all for the fast process! Its always a joy contributing here :)

Best regards,
JosipKuci

PR #2955, opened 2026-05-19
Fix reversed p0/p1 description in REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL doc

### LINKED ISSUE #2949: rom_reset_usb_boot_extra appears to pass p0/p1 to rom_reboot in the opposite order from the documented BOOTSEL ABI
## What's wrong

There is a contradiction between (a) the documented `REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL` parameter ABI and (b) what the SDK's `rom_reset_usb_boot_extra()` wrapper actually passes to `rom_reboot()`.

### Documented ABI

`src/common/boot_picoboot_headers/include/boot/picoboot_constants.h:15`:
```c
#define REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL      0x2 // param0 = gpio_pin_number, param1 = flags
```

`src/rp2_common/pico_bootrom/include/pico/bootrom.h:469-475`:
```
 * REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL - reboot into BOOTSEL mode.
 *  p0 - the GPIO number to use as an activity indicator (enabled by flag in p1).
 *  p1 - a set of flags:
 *   0x01 : DISABLE_MSD_INTERFACE
 *   0x02 : DISABLE_PICOBOOT_INTERFACE
 *   0x10 : GPIO_PIN_ACTIVE_LOW
 *   0x20 : GPIO_PIN_ENABLED
```

So per the docs: **p0 = GPIO number, p1 = flags**.

### What the SDK does

`src/rp2_common/pico_bootrom/bootrom.c:46` (RP2040 fallback path):
```c
rom_reboot(REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS, 10, flags, usb_activity_gpio_pin_mask);
//                                                                                   ^p0    ^p1
```

`src/rp2_common/pico_bootrom/bootrom.c:66` (newer path, used on RP2350):
```c
rom_reboot(REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS, 10, flags, (uint)usb_activity_gpio_pin);
//                                                                                   ^p0    ^p1
```

Both pass **p0 = flags, p1 = GPIO** — the opposite of the documented ABI.

`rom_reboot()` itself just forwards `(flags, delay_ms, p0, p1)` to `rom_func_lookup_inline(ROM_FUNC_REBOOT)`, so the wrapper isn't doing any swapping under the hood.

## How I noticed

While preparing PR #2947 (fix for the wValue bit-overlap in `pico_stdio_usb/reset_interface.c`, issue #2713), I tried an end-to-end test on a Pico 2 (RP2350) using the bootsel-reset USB request to specify GPIO 25 as the activity-LED with `active_low=true`:

```
wValue = (25 << 10) | 0x200 | 0x100 = 0x6700
```

The expected observable behaviour during BOOTSEL: with `active_low=true` the bootrom should hold the activity GPIO **HIGH** during idle (so the Pico 2's onboard LED on GPIO 25 lights solid). What I actually observed: LED **steady off**, regardless of `active_low` value, in both `active_low=false` and `active_low=true` runs and with MSD enabled (`disable_mask=0`) so there was clear MSD enumeration / I/O activity.

The wValue parsing in `reset_interface.c` is correct (verified by reading the code), so the breakage is downstream of that — at the `rom_reset_usb_boot_extra` → `rom_reboot` call boundary, exactly where the apparent p0/p1 swap is.

## Which side is wrong?

I'm not in a position to say with certainty which of the two is the source of truth (the bootrom's actual behaviour, vs. the SDK doc):

* If the **doc is correct** and the bootrom genuinely takes `(p0=gpio, p1=flags)`, then `bootrom.c:46` and `:66` need to swap the last two args, and the activity-LED feature has been silently broken whenever `usb_activity_gpio_pin >= 0` was passed. That matches what I observed.

* If the **SDK code is correct** and the bootrom actually takes `(p0=flags, p1=gpio)`, then the doxygen in `bootrom.h` and the comment in `picoboot_constants.h` are stale and should be updated. In that case, the activity LED not lighting on my hardware would have a different explanation, which I'd want to understand before closing this.

Picotool's `PC_REBOOT2` call (raspberrypi/picotool `main.cpp`) only passes `dParam0=0, dParam1=0`, so it doesn't exercise the activity-LED path and can't disambiguate.

## Repro / environment

* SDK: latest `develop` (`origin/develop`, includes commit 2492b1c).
* Board: Pico 2 (RP2350), PID `2e8a:0009` in app mode, `2e8a:000f` in BOOTSEL.
* Firmware: minimal app linking `pico_stdlib` with `pico_enable_stdio_usb(... 1)` and `PICO_STDIO_USB_RESET_BOOTSEL_FIXED_ACTIVITY_LED=0`.
* Host trigger: `libusb` control transfer, `bmRequestType=0x21`, `bRequest=0x01` (RESET_REQUEST_BOOTSEL), `wValue=0x6700`, `wIndex=2` (reset interface number from `lsusb -v`).
* Observed: device reboots into BOOTSEL (PID `2e8a:000f`), MSD enumerates, files browsable; activity-LED on GPIO 25 remains off throughout, including during sustained MSD reads.

Happy to gather any additional traces (e.g. a UART-stdio variant of the same firmware that prints `gpio` and `active_low` right before calling `rom_reset_usb_boot_extra`) if that would help triage.

anoshyn:
## Summary

Two places in the SDK described the bootrom's BOOTSEL reboot ABI with `p0` and `p1` swapped relative to what the bootrom actually accepts:

- `src/common/boot_picoboot_headers/include/boot/picoboot_constants.h:15`
- `src/rp2_common/pico_bootrom/include/pico/bootrom.h:469-475`

`rom_reset_usb_boot_extra()` has always passed `(p0 = flags, p1 = gpio_pin_number)`. The doc said the opposite (`p0 = gpio, p1 = flags`). Swap the descriptions; no code change.

## Why we know which side is right

When I filed #2949 I suspected the SDK was passing the args in the wrong order. Hardware testing on RP2350 today flipped that conclusion:

A minimal firmware calling

```c
rom_reboot(REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS,
           10, /* p0 */ 25, /* p1 */ 0x30 /* GPIO_PIN_SPECIFIED|ACTIVE_LOW per doc */)
```

re-enters BOOTSEL with the **MSD interface disabled** (no `/media/$USER/RP2350/`, no `/dev/sd*`), only the PICOBOOT interface up.

- Under the doc-correct interpretation, those args mean `gpio = 25` (valid) and `flags = 0x30` with **no disable bits** — MSD should be up.
- Under the SDK-correct interpretation, they mean `flags = 25 = 0x19` — bit 0 is `DISABLE_MSD_INTERFACE`. That matches what's actually observed.

The only way the bootrom could disable MSD from those args is by reading `p0` as flags. So the SDK code is right; the doc is reversed. Reproduced with a second test (`p1 = 0x20`, same `p0 = 25`, identical observable outcome).

Full reasoning and observations: https://github.com/raspberrypi/pico-sdk/issues/2949#issuecomment-4491368535

## Test plan

- [x] Verified by inspecting both files that the swapped pairs are internally consistent (the `0x10 GPIO_PIN_ACTIVE_LOW` cross-reference in `bootrom.h` now points at `p1`, where the GPIO actually lives).
- [x] Hardware test as described above (RP2350 / Pico 2).
- [x] (Out of scope of this PR) Followed up on the "activity LED never lights up" subplot separately — it's a real, independent bug. Filed as #2957.

Closes #2949.

### DIFF
--- a/src/rp2_common/pico_bootrom/include/pico/bootrom.h
+++ b/src/rp2_common/pico_bootrom/include/pico/bootrom.h
@@ -467,12 +467,12 @@ static inline int rom_set_bootrom_stack(bootrom_stack_t *stack) {
  * \ref REBOOT2_FLAG_REBOOT_TYPE_NORMAL - reboot into the normal boot path.
  * 
  * \ref REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL - reboot into BOOTSEL mode.
- *  p0 - the GPIO number to use as an activity indicator (enabled by flag in p1).
- *  p1 - a set of flags:
+ *  p0 - a set of flags:
  *   0x01 : DISABLE_MSD_INTERFACE - Disable the BOOTSEL USB drive (see <<section_bootrom_mass_storage>>)
  *   0x02 : DISABLE_PICOBOOT_INTERFACE - Disable the {picoboot} interface (see <<section_bootrom_picoboot>>).
- *   0x10 : GPIO_PIN_ACTIVE_LOW - The GPIO in p0 is active low.
+ *   0x10 : GPIO_PIN_ACTIVE_LOW - The GPIO in p1 is active low.
  *   0x20 : GPIO_PIN_ENABLED - Enable the activity indicator on the specified GPIO.
+ *  p1 - the GPIO number to use as an activity indicator (enabled by GPIO_PIN_ENABLED flag in p0).
  * 
  * \ref REBOOT2_FLAG_REBOOT_TYPE_RAM_IMAGE - reboot into an image in RAM. The region of RAM or XIP RAM is searched for an image to run. This is the type
  * of reboot used when a RAM UF2 is dragged onto the BOOTSEL USB drive.
--- a/src/rp2_common/pico_bootrom/include/pico/bootrom.h
+++ b/src/rp2_common/pico_bootrom/include/pico/bootrom.h
@@ -470,8 +470,8 @@ static inline int rom_set_bootrom_stack(bootrom_stack_t *stack) {
  *  p0 - a set of flags:
  *   0x01 : DISABLE_MSD_INTERFACE - Disable the BOOTSEL USB drive (see <<section_bootrom_mass_storage>>)
  *   0x02 : DISABLE_PICOBOOT_INTERFACE - Disable the {picoboot} interface (see <<section_bootrom_picoboot>>).
- *   0x10 : GPIO_PIN_ACTIVE_LOW - The GPIO in p1 is active low.
- *   0x20 : GPIO_PIN_ENABLED - Enable the activity indicator on the specified GPIO.
+ *   0x10 : GPIO_PIN_ACTIVE_LOW - The GPIO specified in p1 is active low (GPIO_PIN_SPECIFIED must also be set).
+ *   0x20 : GPIO_PIN_SPECIFIED - Enable the activity indicator on the GPIO specified in p1.
  *  p1 - the GPIO number to use as an activity indicator (enabled by GPIO_PIN_ENABLED flag in p0).
  * 
  * \ref REBOOT2_FLAG_REBOOT_TYPE_RAM_IMAGE - reboot into an image in RAM. The region of RAM or XIP RAM is searched for an image to run. This is the type
--- a/src/common/boot_picoboot_headers/include/boot/picoboot_constants.h
+++ b/src/common/boot_picoboot_headers/include/boot/picoboot_constants.h
@@ -12,7 +12,7 @@
 // --- note these match BOOT_TYPE_ constants in pico/bootrom_constants.h
 // values 0-7 are secure/non-secure
 #define REBOOT2_FLAG_REBOOT_TYPE_NORMAL       0x0 // param0 = diagnostic partition
-#define REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL      0x2 // param0 = gpio_pin_number, param1 = flags
+#define REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL      0x2 // param0 = flags, param1 = gpio_pin_number
 #define REBOOT2_FLAG_REBOOT_TYPE_RAM_IMAGE    0x3 // param0 = image_region_base, param1 = image_region_size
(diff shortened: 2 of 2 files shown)


kilograham (approved):

kilograham (approved):

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -467,12 +467,12 @@ static inline int rom_set_bootrom_stack(bootrom_stack_t *stack) {
...
- *  p0 - the GPIO number to use as an activity indicator (enabled by flag in p1).
- *  p1 - a set of flags:
+ *  p0 - a set of flags:
  *   0x01 : DISABLE_MSD_INTERFACE - Disable the BOOTSEL USB drive (see <<section_bootrom_mass_storage>>)
  *   0x02 : DISABLE_PICOBOOT_INTERFACE - Disable the {picoboot} interface (see <<section_bootrom_picoboot>>).
- *   0x10 : GPIO_PIN_ACTIVE_LOW - The GPIO in p0 is active low.
- *   0x20 : GPIO_PIN_ENABLED - Enable the activity indicator on the specified GPIO.
+ *   0x10 : GPIO_PIN_ACTIVE_LOW - The GPIO specified in p1 is active low (GPIO_PIN_SPECIFIED must also be set).
+ *   0x20 : GPIO_PIN_SPECIFIED - Enable the activity indicator on the GPIO specified in p1.
+ *  p1 - the GPIO number to use as an activity indicator (enabled by GPIO_PIN_ENABLED flag in p0).
```
lurch, line 475: "(enabled by GPIO_PIN_SPECIFIED flag in p0)."
:wink: 
(I'll submit a PR tomorrow, unless you beat me to it)

lurch: Fix is in #2960

PR #2956, opened 2026-05-19
Add support for LLVM Embedded Toolchain for Arm 21

kilograham:
they moved headers and libraries around a bit

also removed the `picolibc` detection which is now handled better in `pico_clib_interface` (so depends on #2935)

### DIFF
--- a/cmake/preload/toolchains/util/pico_arm_clang_common.cmake
+++ b/cmake/preload/toolchains/util/pico_arm_clang_common.cmake
@@ -43,39 +43,34 @@ endforeach()
 
 list(APPEND CMAKE_TRY_COMPILE_PLATFORM_VARIABLES PICO_CLIB)
 
-foreach(PICO_CLANG_RUNTIME IN LISTS PICO_CLANG_RUNTIMES)
-    # LLVM embedded-toolchain for ARM style
-    find_path(PICO_COMPILER_SYSROOT NAMES include/stdio.h
-            HINTS
-            ${PICO_COMPILER_DIR}/../lib/clang-runtimes/arm-none-eabi/${PICO_CLANG_RUNTIME}
-            ${PICO_COMPILER_DIR}/../lib/clang-runtimes/${PICO_CLANG_RUNTIME}
-    )
+if (NOT PICO_COMPILER_SYSROOT)
+    # Note we explicitly look for the correct libraries, because we don't necessarily trust multilib (even
+    # in newer Arm LLVM ETfA to correctly pick the right one
+    foreach(PICO_CLANG_RUNTIME IN LISTS PICO_CLANG_RUNTIMES)
+        # LLVM embedded-toolchain for ARM style
+        find_path(PICO_COMPILER_SYSROOT NAMES lib
+                HINTS
+                ${PICO_COMPILER_DIR}/../lib/clang-runtimes/arm-none-eabi/${PICO_CLANG_RUNTIME}
+                ${PICO_COMPILER_DIR}/../lib/clang-runtimes/${PICO_CLANG_RUNTIME}
+        )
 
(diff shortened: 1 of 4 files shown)


cmake/preload/toolchains/util/pico_arm_clang_common.cmake
```diff
@@ -43,39 +43,34 @@ endforeach()
...
-foreach(PICO_CLANG_RUNTIME IN LISTS PICO_CLANG_RUNTIMES)
-    # LLVM embedded-toolchain for ARM style
-    find_path(PICO_COMPILER_SYSROOT NAMES include/stdio.h
-            HINTS
-            ${PICO_COMPILER_DIR}/../lib/clang-runtimes/arm-none-eabi/${PICO_CLANG_RUNTIME}
-            ${PICO_COMPILER_DIR}/../lib/clang-runtimes/${PICO_CLANG_RUNTIME}
-    )
+if (NOT PICO_COMPILER_SYSROOT)
+    # Note we explicitly look for the correct libraries, because we don't necessarily trust multilib (even
+    # in newer Arm LLVM ETfA to correctly pick the right one
```
will-v-pi: Isn't is ATfE (https://github.com/arm/arm-toolchain/releases), not ETfA?
```suggestion
    # in newer Arm LLVM ATfE to correctly pick the right one
```

cmake/preload/toolchains/util/pico_arm_clang_common.cmake
```diff
@@ -91,9 +86,22 @@ if (PICO_CLIB STREQUAL "llvm_libc")
...
-        message(FATAL_ERROR "Could not find an llvm runtime for '${PICO_CLANG_RUNTIME}'")
+        message(FATAL_ERROR "Could not find an llvm runtime")
     endif()
-
     set(PICO_COMMON_LANG_FLAGS "${PICO_COMMON_LANG_FLAGS} --sysroot ${PICO_COMPILER_SYSROOT}")
+
+    find_path(_CLANG_HEADERS_DIR NAMES include/stdio.h
+            HINTS
+            ${PICO_COMPILER_SYSROOT}
+            ${PICO_COMPILER_SYSROOT}/.. # new style LLVM ETfA (21+)
```
will-v-pi:
```suggestion
            ${PICO_COMPILER_SYSROOT}/.. # new style LLVM ATfE (21+)
```

cmake/preload/toolchains/util/pico_arm_clang_common.cmake
```diff
@@ -91,9 +86,22 @@ if (PICO_CLIB STREQUAL "llvm_libc")
...
+    )
+    if (NOT _CLANG_HEADERS_DIR)
+        message(FATAL_ERROR "Could not find llvm headers (searched ${PICO_COMPILER_SYSROOT}/include and ${PICO_COMPILER_SYSROOT}/../include)")
+    else()
+        if (NOT _CLANG_HEADERS_DIR STREQUAL PICO_COMPILER_SYSROOT)
+            # LLVM ETfA moves the headers out of the multilib so we need to explicitly add C++/C headers (in that order)
```
will-v-pi: This should also mention 21+
```suggestion
            # LLVM ATfE 21+ moves the headers out of the multilib so we need to explicitly add C++/C headers (in that order)
```

will-v-pi (dismissed):
Couple of comment fixes (I think), but other than that looks good - although definitely shouldn't be merged before #2935, otherwise LLVM builds using ATfE will break, as they need the picolibc auto-detection

lurch:
Should we add this to https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/multi-gcc.yml ?

PR #2958, opened 2026-05-19
pio_encoding fixes + added PIO_VERSION to platform_defs.h

kilograham:
* `pio_encode_mov(pio_pindirs)` was wrong because the `mov` destination for pindirs is a different value
* Noticed that `pio_instructions.h` is included before `PICO_PIO_VERSION` is set; didn't want to make a circular include, and previously pio_instructions.h is somewhat standalone include-able, so decided to add `PIO_VERSION` to `platform_defs.h`... TBH it should have been there already as the init of `PICO_PIO_VERSION` was a bit of a hack. I doubt anyone wants to override PICO_PIO_VERSION anyway
* I saw that we have `pio_exec_mov` and `pio_exec_out` - kept these for backwards compatibility, but decided to just allow `pio_exec` and have `pio_encode_mov` take care of it. as i imagine most of uses of `pio_encode_mov` take a constant arg, this isn't much of an overhead (the alternative would have been to encode the two different values in the enum value, which is fine, but likely going to lead to more code if done at runtime vs just the ifs)

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
@@ -66,12 +66,13 @@ enum pio_src_dest {
     pio_x = 1u,
     pio_y = 2u,
     pio_null = 3u | _PIO_INVALID_SET_DEST | _PIO_INVALID_MOV_DEST,
-#if PIO_VERSION > 0
+#if PICO_PIO_VERSION > 0
     pio_pindirs_mov = 3u | _PIO_INVALID_IN_SRC | _PIO_INVALID_MOV_SRC,
     pio_pindirs = 4u | _PIO_INVALID_IN_SRC | _PIO_INVALID_MOV_SRC,
 #else
     pio_pindirs = 4u | _PIO_INVALID_IN_SRC | _PIO_INVALID_MOV_SRC | _PIO_INVALID_MOV_DEST,
 #endif
+    pio_pindirs_out = 4u | _PIO_INVALID_IN_SRC | _PIO_INVALID_MOV_SRC | _PIO_INVALID_MOV_DEST,
     pio_exec_mov = 4u | _PIO_INVALID_IN_SRC | _PIO_INVALID_OUT_DEST | _PIO_INVALID_SET_DEST | _PIO_INVALID_MOV_SRC,
     pio_status = 5u | _PIO_INVALID_IN_SRC | _PIO_INVALID_OUT_DEST | _PIO_INVALID_SET_DEST | _PIO_INVALID_MOV_DEST,
     pio_pc = 5u | _PIO_INVALID_IN_SRC | _PIO_INVALID_SET_DEST | _PIO_INVALID_MOV_SRC,
@@ -316,7 +317,7 @@ static inline uint pio_encode_wait_irq(bool polarity, bool relative, uint irq) {
     return _pio_encode_instr_and_args(pio_instr_bits_wait, 2u | (polarity ? 4u : 0u), _pio_encode_irq(relative, irq));
 }
 
-#if PICO_PIO_VERSION > 0
+#if PICO_PICO_PIO_VERSION > 0
 /*! \brief Encode a WAIT for jmppin instruction
  *  \ingroup pio_instructions
  *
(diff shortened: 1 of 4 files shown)


kilograham:
fixes #2754

lurch:
> I doubt anyone wants to override PICO_PIO_VERSION anyway

Maybe it's worth removing the `PICO_CONFIG: PICO_PIO_VERSION` line? :thinking:

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
```diff
@@ -66,13 +66,19 @@ enum pio_src_dest {
     pio_x = 1u,
     pio_y = 2u,
     pio_null = 3u | _PIO_INVALID_SET_DEST | _PIO_INVALID_MOV_DEST,
+#if PIO_VERSION > 0
```
lurch: For consistency, should this be using `PICO_PIO_VERSION`?
Or should some of the places in `pio_instructions.h`, `pio.h` and `pio.c` that are currently doing `#if PICO_PIO_VERSION > 0` and `#if PICO_PIO_VERSION == 0` be updated to use `PIO_VERSION` instead?

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
```diff
@@ -398,6 +404,11 @@ static inline uint pio_encode_pull(bool if_empty, bool block) {
 static inline uint pio_encode_mov(enum pio_src_dest dest, enum pio_src_dest src) {
     valid_params_if(PIO_INSTRUCTIONS, !(dest & _PIO_INVALID_MOV_DEST));
     valid_params_if(PIO_INSTRUCTIONS, !(src & _PIO_INVALID_MOV_SRC));
+    // generic constants aren't the same as the MOV ones
+    if (dest == pio_exec) dest = pio_exec_mov;
```
lurch, line 411: I'm not really awake-enough yet to figure out what this code is doing, but (for symmetry) should `pio_encode_out` be doing `if (dest == pio_exec) dest = pio_exec_out;`?

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
will-v-pi: I would say either the `PICO_PIO_VERSION` define should be moved into `pio_instructions.h` and used everywhere, or it should be removed entirely and `PIO_VERSION` should be used everywhere?

Or is it the case that the version used in `pio_instructions.h` must match the chip being used (hence using `PIO_VERSION` directly), but the rest of the code can be compiled for version 0 on both RP2040 and RP2350?

If we're keeping `PICO_PIO_VERSION ` then there should also be a check that `PICO_PIO_VERSION <= PIO_VERSION`, otherwise you get lots of compile errors when trying to build RP2040 code with `PICO_PIO_VERSION=1`.

kilograham: I guess I'll just rename the one in platform_defs to PICO_PIO_VERSION (I had been avoiding that because other h/w stuff doesn't have a PICO_) we have PICO_PIO_VERSION in the build too, so I think it is probably confusing to introduce another

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
kilograham: i could add an assert i guess!

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
```diff
@@ -310,7 +317,7 @@ static inline uint pio_encode_wait_irq(bool polarity, bool relative, uint irq) {
     return _pio_encode_instr_and_args(pio_instr_bits_wait, 2u | (polarity ? 4u : 0u), _pio_encode_irq(relative, irq));
 }
 
-#if PICO_PIO_VERSION > 0
+#if PICO_PICO_PIO_VERSION > 0
```
lurch:
```suggestion
#if PICO_PIO_VERSION > 0
```

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
lurch: Just because I was curious - f4e041b is what this would look like with `PICO_PIO_VERSION` replaced by `PIO_VERSION`

will-v-pi (approved):

src/rp2_common/hardware_pio/include/hardware/pio_instructions.h
```diff
@@ -398,6 +407,11 @@ static inline uint pio_encode_pull(bool if_empty, bool block) {
 static inline uint pio_encode_mov(enum pio_src_dest dest, enum pio_src_dest src) {
     valid_params_if(PIO_INSTRUCTIONS, !(dest & _PIO_INVALID_MOV_DEST));
     valid_params_if(PIO_INSTRUCTIONS, !(src & _PIO_INVALID_MOV_SRC));
+    // generic constants aren't the same as the MOV ones
+    if (dest == pio_exec) dest = pio_exec_mov;
+#if PICO_PIO_VERSION > 0
+    if (dest == pio_pindirs) dest = pio_pindirs_mov;
```
lurch, line 413: Probably doesn't matter now, but I guess this could have been:
```c
    else if (dest == pio_pindirs) dest = pio_pindirs_mov;
```

PR #2961, opened 2026-05-20
pico_bootrom: fix activity LED not lighting in BOOTSEL on RP2350

### LINKED ISSUE #2957: BOOTSEL activity-LED (`REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL` + `GPIO_PIN_SPECIFIED`) does not drive the configured GPIO on RP2350
## What's wrong

On RP2350 (Pico 2, current `develop` bootrom), the BOOTSEL activity-LED feature exposed via `REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL` / `rom_reset_usb_boot_extra()` does not appear to drive the specified GPIO at all, regardless of arguments.

## Repro

Three test firmwares, all run on the same Pico 2 board. Each call to `rom_reboot` is preceded by three visible blinks on GPIO 25 (driven from user code) to confirm the firmware actually ran. Built against `develop` at fdd65e6c.

### Test 1 — sanity check (user code only)

```c
gpio_init(25);
gpio_set_dir(25, GPIO_OUT);
gpio_put(25, 1);
while (1) tight_loop_contents();
```

**Result:** LED solid on. Confirms GPIO 25 reaches the LED, polarity is active-high, and the pin can be driven from user code.

### Test 2 — SDK-correct args, no `ACTIVE_LOW`

```c
rom_reboot(
    REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS,
    10,
    /* p0 = flags */ BOOTSEL_FLAG_GPIO_PIN_SPECIFIED,   // 0x20
    /* p1 = gpio  */ 25);
```

Equivalent to `rom_reset_usb_boot_extra(25, 0, false)`.

**Expected:** MSD up, activity-LED idle-low (LED off idle, briefly on during MSD activity since this LED is active-high).

**Observed:** MSD up and serving reads correctly. LED stays off throughout, including during sustained reads driven from the host:

```
for i in $(seq 1 200); do cat /media/$USER/RP2350/INFO_UF2.TXT > /dev/null; done
```

### Test 3 — SDK-correct args, `ACTIVE_LOW` set (the diagnostic that nails it)

```c
rom_reboot(
    REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS,
    10,
    /* p0 = flags */ BOOTSEL_FLAG_GPIO_PIN_SPECIFIED | BOOTSEL_FLAG_GPIO_PIN_ACTIVE_LOW,  // 0x30
    /* p1 = gpio  */ 25);
```

Equivalent to `rom_reset_usb_boot_extra(25, 0, true)`.

**Expected:** with `ACTIVE_LOW`, the bootrom holds the pin idle **high**. On this active-high LED that means LED solid on during BOOTSEL idle, briefly dipping off during activity.

**Observed:** LED stays off the entire time BOOTSEL is up — no idle drive, no activity blinks.

## Why Test 3 is the smoking gun

Test 2 alone could be explained as "bootrom is driving the pin but the reads I'm generating don't register as observable activity pulses on the host." Test 3 rules that out: it requires no activity dependence — the bootrom should be holding the pin **statically high** for the whole BOOTSEL session — and the pin stays low. So the bootrom isn't driving GPIO 25 at all under any tested combination, even with `BOOTSEL_FLAG_GPIO_PIN_SPECIFIED` clearly set.

## Environment

- SDK: `origin/develop` at fdd65e6c
- Board: Raspberry Pi RP2350 (Pico 2), `Board-ID: RP2350`, `INFO_UF2.TXT` reports `UF2 Bootloader v1.0`
- USB ID in BOOTSEL: `2e8a:000f`
- Host: Linux 6.12, libusb-based control-transfer triggers ruled out; tests call `rom_reboot` directly
- Toolchain: `arm-none-eabi-gcc 14.2.1`

## Related

- Today's testing also resolved the related argument-order question (originally #2949): the SDK's `rom_reset_usb_boot_extra()` passes `(p0 = flags, p1 = gpio)` and the bootrom does interpret them in that order. The doc comments were reversed; PR #2955 fixes them. So the wrong arg order is **not** the cause of this bug — my tests above use the correct order and still see no LED drive.
- PR #2947 fixes a separate `wValue` bit-overlap in `src/rp2_common/pico_stdio_usb/reset_interface.c` that confounded my earlier USB-control-transfer-based BOOTSEL tests; today's tests bypass that path entirely.

## What I haven't ruled out

- Whether the activity-LED feature works under some specific bootrom revision or specific host-side trigger I haven't tested. If there's a known-good repro pointing at a working configuration on RP2350, I'd appreciate the pointer.
- Whether there's an additional enable bit I'm missing. `src/common/boot_picoboot_headers/include/boot/picoboot_constants.h:27-30` only documents `DISABLE_MSD_INTERFACE`, `DISABLE_PICOBOOT_INTERFACE`, `GPIO_PIN_ACTIVE_LOW`, `GPIO_PIN_SPECIFIED`. The `bootrom.h` doxygen still uses a stale `GPIO_PIN_ENABLED` name for the 0x20 flag — same value, just an inconsistent name (orthogonal to this bug).
- Whether GPIO 25 specifically is reserved by the bootrom on RP2350. Worth retrying with a different GPIO if anyone has a Pico 2 with an LED wired to e.g. GPIO 2 they can flash and report.

Happy to send the firmware sources / a self-contained reproducer repo if useful.

anoshyn:
## Summary

`rom_reset_usb_boot()` and `rom_reset_usb_boot_extra()` in `src/rp2_common/pico_bootrom/bootrom.c` set `BOOTSEL_FLAG_GPIO_PIN_SPECIFIED` in the BOOTSEL flags word (`p0` of `rom_reboot`) but not the corresponding bit in the reboot-flags word (the first argument to `rom_reboot`). On RP2350 the bootrom requires bit `0x20` to be set in **both** positions before it will honour the GPIO number in `p1` as an activity indicator — without the bit in the reboot-flags word the LED stays dark even though every other parameter is correct.

This makes the activity-LED parameter on `rom_reset_usb_boot[_extra]()` a no-op on RP2350. Fixing it is a two-line change at the call sites.

## Why `0x20` in the reboot-flags word

`0x20` in the reboot-flags word is already defined in `picoboot_constants.h` as `REBOOT2_FLAG_REBOOT_TO_RISCV`. The bit is overloaded by reboot type: for `REBOOT_TYPE_NORMAL`/`RAM_IMAGE`/`FLASH_UPDATE` it selects the RISC-V core, but for `REBOOT_TYPE_BOOTSEL` it means "GPIO info is valid in `p0`/`p1`", matching the meaning of `BOOTSEL_FLAG_GPIO_PIN_SPECIFIED` in `p0`.

The patch names the bit locally as `_REBOOT2_BOOTSEL_GPIO_INFO_VALID` (leading underscore to flag that the BOOTSEL-context meaning is empirically derived, not from a published bootrom header). A maintainer may want to promote it into `picoboot_constants.h` once the official name is known.

## How this was found

Originally filed in #2957 after observing that the LED never lit during BOOTSEL even with `rom_reset_usb_boot_extra(25, 0, false)`. A later [comment on that issue](https://github.com/raspberrypi/pico-sdk/issues/2957#issuecomment-4499389416) pointed out that `BOOTSEL_FLAG_GPIO_PIN_SPECIFIED` needed to be set in **both** the reboot-flags word and `p0`. Confirmed on a Pico 2 with two tests:

1. Direct `rom_reboot(REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS | 0x20, 10, BOOTSEL_FLAG_GPIO_PIN_SPECIFIED, 25)` — LED flickers on MSD activity. Without the `| 0x20` it doesn't.
2. With the patch applied, the high-level wrapper `rom_reset_usb_boot_extra(25, 0, false)` produces the same visible flicker. Unpatched, the same call leaves the LED dark.

## Test plan

- [x] Built and flashed `rom_reset_usb_boot_extra(25, 0, false)` on Pico 2 (RP2350) against the patched SDK — LED visibly flickers on `cat /media/$USER/RP2350/INFO_UF2.TXT > /dev/null`.
- [x] Behaviour without the patch reproduced as solid dark LED in the same test.
- [ ] (Out of scope) RP2040 path is unchanged — RP2040 uses the dedicated `ROM_FUNC_RESET_USB_BOOT` and never enters the `#elif defined(ROM_FUNC_REBOOT)` branch.

Closes #2957.

### DIFF
--- a/src/rp2_common/pico_bootrom/bootrom.c
+++ b/src/rp2_common/pico_bootrom/bootrom.c
@@ -32,18 +32,28 @@ bool rom_funcs_lookup(uint32_t *table, unsigned int count) {
 }
 
 
+// When REBOOT_TYPE is BOOTSEL, the bootrom requires bit 0x20 to be set in
+// the reboot-flags word (the first argument to rom_reboot) for the GPIO
+// number in p1 to be honoured for the activity indicator. Numerically this
+// is the same bit as REBOOT2_FLAG_REBOOT_TO_RISCV, which is overloaded by
+// reboot type and means "BOOTSEL GPIO info valid" in this context — see
+// pico-sdk issue #2957 for the empirical confirmation on RP2350.
+#define _REBOOT2_BOOTSEL_GPIO_INFO_VALID 0x20u
+
 void __attribute__((noreturn)) rom_reset_usb_boot(uint32_t usb_activity_gpio_pin_mask, uint32_t disable_interface_mask) {
 #ifdef ROM_FUNC_RESET_USB_BOOT
     rom_reset_usb_boot_fn func = (rom_reset_usb_boot_fn) rom_func_lookup(ROM_FUNC_RESET_USB_BOOT);
     func(usb_activity_gpio_pin_mask, disable_interface_mask);
 #elif defined(ROM_FUNC_REBOOT)
+    uint32_t reboot_flags = REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS;
     uint32_t flags = disable_interface_mask;
     if (usb_activity_gpio_pin_mask) {
         flags |= BOOTSEL_FLAG_GPIO_PIN_SPECIFIED;
+        reboot_flags |= _REBOOT2_BOOTSEL_GPIO_INFO_VALID;
         // the parameter is actually the gpio number, but we only care if BOOTSEL_FLAG_GPIO_PIN_SPECIFIED
         usb_activity_gpio_pin_mask = (uint32_t)__builtin_ctz(usb_activity_gpio_pin_mask);
     }
-    rom_reboot(REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL | REBOOT2_FLAG_NO_RETURN_ON_SUCCESS, 10, flags, usb_activity_gpio_pin_mask);
+    rom_reboot(reboot_flags, 10, flags, usb_activity_gpio_pin_mask);
     __builtin_unreachable();
 #else
     panic_unsupported();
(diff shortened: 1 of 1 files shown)


kilograham:
closing this AI slop

kilograham:
the original reasoning is wrong, and the fix is ugfy, and all the reasoning is still wrong

PR #2963, opened 2026-05-20
Fix pico_status_led CRLF and missing `from_us_since_boot` which broke tests on some compilers

kilograham:


src/rp2_common/pico_status_led/status_led.c
```diff
@@ -1,266 +1,266 @@
...
+    while (true) {
+#if !PICO_COLORED_STATUS_LED_USE_DEFAULT_ALARM_POOL
+        uint64_t now = time_us_64();
+        if (now >= next_safe_set_time) {
+            unsafe_set_ws2812(value, now);
+            spin_unlock(spin_lock, save);
+            break;
+        } else {
+            spin_unlock(spin_lock, save);
+            busy_wait_until(from_ms_since_boot(next_safe_set_time));
```
lurch, line 116: Should this have been `from_us_since_boot` ?
`next_safe_set_time` is set on line 87 and those all appear to be us values rather than ms values? :thinking:

lurch: Oh my, `from_ms_since_boot` doesn't even exist in the SDK! :facepalm:  I guess the fact that the CI passed must just be a quirk of the `#ifdef`s meaning that this chunk of the code isn't even _seen_ by the compiler.

lurch: Fix is in #2965

lurch:
This is a follow-up to #2938

PR #2965, opened 2026-05-21
Fix from_ms_since_boot / from_us_since_boot mixup

lurch:
Follow-up to #2963


will-v-pi (approved):

will-v-pi:
Yep, I ran into this too - this bit only gets built when building for a board with a WS2812 on it (eg `adafruit_feather_rp2350`), which is why CI didn't notice it

PR #2526, opened 2025-06-16
remove unnecessary escaping

markhermeling:
Fixes #2520 my removing unnecessary escape '\' characters.

PR #2968, opened 2026-05-26
Parse included board files for pico_cmake_set declarations

### LINKED ISSUE #2112: "// pico_cmake_set" not seen in included header files.

will-v-pi:
Currently if you have a `#include` within a board header file (eg in [vgaboard.h](https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/vgaboard.h)) any `pico_cmake_set` and `pico_cmake_set_default` declarations in the included file aren't parsed (#2112)

This fixes #2112 by parsing those files for these comments. It only parses includes which start with `boards/`, as it needs to extract the `.h` file directly for use with `pico_find_in_paths`.

The inclusion order matches what would be expected (I think) - ie:
```c
// pico_cmake_set PICO_PLATFORM = rp2350

#include "boards/pico.h"
```
gives `PICO_PLATFORM = rp2040` from `pico.h`,
```c
#include "boards/pico.h"

// pico_cmake_set PICO_PLATFORM = rp2350
```
gives `PICO_PLATFORM = rp2350`,
```c
// pico_cmake_set_default PICO_FLASH_SIZE_BYTES = (8 * 1024 * 1024)
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (8 * 1024 * 1024)
#endif

#include "boards/pico.h"
```
gives `PICO_FLASH_SIZE_BYTES = (8 * 1024 * 1024)` (matching the pre-processor defines), and
```c
#include "boards/pico.h"

// pico_cmake_set_default PICO_FLASH_SIZE_BYTES = (8 * 1024 * 1024)
#ifndef PICO_FLASH_SIZE_BYTES
#define PICO_FLASH_SIZE_BYTES (8 * 1024 * 1024)
#endif
```
gives `PICO_FLASH_SIZE_BYTES = (2 * 1024 * 1024)` from `pico.h` (again matching the pre-processor defines)

This is a duplicate of #2113, which I accidentally closed


cmake/generic_board.cmake
```diff
@@ -42,6 +42,13 @@ if (EXISTS ${PICO_BOARD_HEADER_FILE})
                 list(APPEND PICO_BOARD_CMAKE_OVERRIDES ${CMAKE_MATCH_1})
             endif()
         endif()
+        if (LINE MATCHES "^[ \t\]*#include[ \t\]*\"boards\/(.+)\"")
```
kilograham: seems like this could be an `elseif` - not i don't particularly care, but since I'm gonna suggest you also prevent foot-gunning on recursive includes (which might otherwise be fixed by include guards in the headers themselves)

kilograham (changes requested):
seems like this could be an elseif

cmake/generic_board.cmake
```diff
@@ -41,10 +41,10 @@ if (EXISTS ${PICO_BOARD_HEADER_FILE})
             else()
                 list(APPEND PICO_BOARD_CMAKE_OVERRIDES ${CMAKE_MATCH_1})
             endif()
-        endif()
-        if (LINE MATCHES "^[ \t\]*#include[ \t\]*\"boards\/(.+)\"")
+        elseif (LINE MATCHES "^[ \t\]*#include[ \t\]*\"boards\/(.+)\"")
```
lurch: I'm no CMake expert, but why is this regex doing `[ \t\]*` when `[ \t]*` is being used on lines 35 and 38?

will-v-pi: Because this PR was originally created before https://github.com/raspberrypi/pico-sdk/pull/2526 - I will remove the unnecessary escapes

cmake/generic_board.cmake
```diff
@@ -14,6 +14,7 @@ pico_find_in_paths(PICO_BOARD_HEADER_FILE PICO_BOARD_HEADER_DIRS ${PICO_BOARD}.h
 if (EXISTS ${PICO_BOARD_HEADER_FILE})
     message("Using board configuration from ${PICO_BOARD_HEADER_FILE}")
     list(APPEND PICO_CONFIG_HEADER_FILES ${PICO_BOARD_HEADER_FILE})
+    list(APPEND PICO_BOARD_INCLUDED_FILES ${PICO_BOARD_HEADER_FILE})
```
lurch, line 17: Should `PICO_INCLUDED_BOARD_HEADER_FILE` also be appended to `PICO_CONFIG_HEADER_FILES`? :thinking:  And if so, could you just use `PICO_CONFIG_HEADER_FILES` instead of `PICO_BOARD_INCLUDED_FILES` ? (Apologies in advance if this is a stupid suggestion!)

lurch:
Is a similar fix to this needed for the Bazel side of things? Or is this only needed for CMake?

cmake/generic_board.cmake
will-v-pi: No, it doesn't need to be added to `PICO_CONFIG_HEADER_FILES` (used to make `config_autogen.h`), as it's being included by the base board file already in the C/C++ build

will-v-pi:
> Is a similar fix to this needed for the Bazel side of things? Or is this only needed for CMake?

Bazel handles boards differently to CMake, and doesn't use the `pico_board_cmake_set`/`pico_board_cmake_set_default` values, so this fix isn't needed there

kilograham (approved):

PR #2972, opened 2026-05-27
Add crt0 support for booting no_flash binaries using the bootrom

will-v-pi:
Add crt0 support for running through the bootrom when using no_flash binaries with a debugger, by programming the watchdog scratch for a RAM image boot before calling into the bootrom.

This allows using load_maps to load code under the debugger (eg xip_sram pinning, or copying around stuff), just as they would work from any other boot. No support is added for RP2040, as it doesn't have magic watchdog reboot types, and the existing entry_point works fine anyway.

This PR enables this by default, but that can be changed if desired - there is code size decrease on Arm (0x8) as it would need to setup RCP and SP anyway, but an increase on Risc-V (0x2a)

### DIFF
--- a/src/rp2_common/pico_crt0/crt0.S
+++ b/src/rp2_common/pico_crt0/crt0.S
@@ -440,7 +440,7 @@ _entry_point:
         // scratch 5 = magic number xored with -(magic number) == -2
         subs r5, r6, #5 // 3 - 5 = -2
         // load the watchdog registers
-        stm r1, {r2, r3, r4, r5, r6, r7}
+        stm r1!, {r2, r3, r4, r5, r6, r7}
     #endif
 
     // Go back through bootrom to finish the boot.
--- a/src/rp2_common/pico_crt0/crt0.S
+++ b/src/rp2_common/pico_crt0/crt0.S
@@ -21,12 +21,12 @@
 #define PICO_CRT0_NEAR_CALLS 0
 #endif
 
-// PICO_CONFIG: PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM, Whether crt0 should always reset through the bootrom when loaded using a debugger - not supported on RP2040, default=1, type=bool, advanced=true, group=pico_crt0
+// PICO_CONFIG: PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM, Whether crt0 should reset through the bootrom when loaded using a debugger - defining this variable is not supported on RP2040, default=1, type=bool, advanced=true, group=pico_crt0
 #if PICO_RP2040
-#define PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM 0
+#define PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM (!PICO_NO_FLASH)
 #else
-#ifndef PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM
-#define PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM 1
+#ifndef PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM
+#define PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM 1
 #endif
 #endif
 
@@ -392,7 +392,7 @@ binary_info_header:
 .global _entry_point
 _entry_point:
 
-#if PICO_NO_FLASH && !PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM
+#if !PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM
     // in the NO_FLASH case, we do not do a rest thru bootrom below, so the RCP may or may not have been initialized:
     //
     // in the normal case (UF2 download, etc.), we will have passed thru bootrom initialization, but if
@@ -424,8 +424,10 @@ _entry_point:
     ldr r0, =__vectors
     // Vector through our own table (SP, VTOR will not have been set up at
     // this point). Same path for debugger entry and bootloader entry.
-#else // PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM or !PICO_NOFLASH
-    #if PICO_NO_FLASH // PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM is always 0 on RP2040, so this code doesn't support RP2040
+#else // PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM
+    #if PICO_NO_FLASH
+        // PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM is always !PICO_NO_FLASH on RP2040, so this code doesn't support RP2040,
+        // because RP2040 doesn't have special watchdog boot types
         // Launch this no_flash binary using a RAM image boot
         ldr r1, =(WATCHDOG_BASE + WATCHDOG_SCRATCH2_OFFSET) // registers start from here
         // scratch 2 = __logical_binary_start
@@ -440,7 +442,7 @@ _entry_point:
         // scratch 5 = magic number xored with -(magic number) == -2
         subs r5, r6, #5 // 3 - 5 = -2
         // load the watchdog registers
-        stm r1!, {r2, r3, r4, r5, r6, r7}
+        stmia r1!, {r2, r3, r4, r5, r6, r7}
     #endif
 
     // Go back through bootrom to finish the boot.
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_crt0/crt0.S
```diff
@@ -414,9 +424,28 @@ _entry_point:
...
+        // scratch 4&7 = magic number
+        ldr r0, =0xb007c0d3
+        str r0, [r1, #8]
+        str r0, [r1, #20]
+        // scratch 5 = magic number xored with -(magic number) == -2
+        ldr r0, =(-2)
+        str r0, [r1, #12]
+        // scratch 6 = ram image boot == 3
+        ldr r0, =3
+        str r0, [r1, #16]
```
kilograham: note you could save some space (well 10 bytes) by using a stmia with 5 different regs

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -382,7 +392,7 @@ binary_info_header:
 .global _entry_point
 _entry_point:
 
-#if PICO_NO_FLASH
+#if PICO_NO_FLASH && !PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM
     // on the NO_FLASH case, we do not do a rest thru bootrom below, so the RCP may or may not have been initialized:
```
lurch:
```suggestion
    // in the NO_FLASH case, we do not do a rest thru bootrom below, so the RCP may or may not have been initialized:
```

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -382,7 +392,7 @@ binary_info_header:
...
     //
     // in the normal (e.g. UF2 download etc. case) we will have passed thru bootrom initialization, but if
```
lurch:
```suggestion
    // in the normal case (UF2 download, etc.), we will have passed thru bootrom initialization, but if
```

src/rp2_common/pico_crt0/crt0.S
will-v-pi: Thanks, that prompted a few other size reductions, so this new code is now slightly smaller than the old code was

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -414,10 +424,31 @@ _entry_point:
...
+        // scratch 6 = ram image boot == 3
+        movs r6, #3
+        // scratch 5 = magic number xored with -(magic number) == -2
+        subs r5, r6, #5 // 3 - 5 = -2
+        // load the watchdog registers
+        stm r1!, {r2, r3, r4, r5, r6, r7}
+    #endif
+
+    // Go back through bootrom to finish the boot.
+    #if BOOTROM_VTABLE_OFFSET < 255 // fits in 8-bit immediate
```
kilograham: arguably this should be < 256 however BOOTROM_VTABLE_OFFSET is always 0 on arm, so i'd just do a movs, and we'll get an error if that were ever to change (same below)

src/rp2_common/pico_crt0/crt0_riscv.S
```diff
@@ -289,11 +294,31 @@ binary_info_header:
 .global _entry_point
 _entry_point:
 
-#if PICO_NO_FLASH
+#if PICO_NO_FLASH && !PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM
```
kilograham: while you are there, you could comment this line out and add // fall thru to _reset_hanlder

kilograham (dismissed):
once you've made fix two comment fixes go ahead and merge

src/rp2_common/pico_crt0/crt0_riscv.S
will-v-pi: Comment out which line? This line is required for RP2040 and to allow enabling the old behaviour on RP2350

kilograham: i meant the `j _reset_handler`

kilograham: as that is what follows anyway

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -414,10 +424,27 @@ _entry_point:
...
+        ldr r3, =(SRAM_END - SRAM_BASE) // assume maximum size for block loop
+        // scratch 4&7 = magic number
+        ldr r4, =0xb007c0d3
+        mov r7, r4
+        // scratch 6 = ram image boot == 3
+        movs r6, #3
+        // scratch 5 = magic number xored with -(magic number) == -2
+        subs r5, r6, #5 // 3 - 5 = -2
+        // load the watchdog registers
+        stm r1!, {r2, r3, r4, r5, r6, r7}
```
kilograham: also can you change this to stmia - which is the 16 bit instruction

kilograham: and i don't know if this will make one of our weirder compilers unhappy

src/rp2_common/pico_crt0/crt0_riscv.S
will-v-pi: Ah, I missed this was the Risc-V file - done

will-v-pi:
> once you've made fix two comment fixes go ahead and merge

It needs re-approval

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -20,6 +21,15 @@
 #define PICO_CRT0_NEAR_CALLS 0
 #endif
 
+// PICO_CONFIG: PICO_CRT0_ALWAYS_RESET_THROUGH_BOOTROM, Whether crt0 should always reset through the bootrom when loaded using a debugger - not supported on RP2040, default=1, type=bool, advanced=true, group=pico_crt0
```
kilograham: I think this name is confusing, since it isn't ALWAYS, it is when entered thru _entry_point

maybe `PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM` or similar

kilograham (changes requested):

src/rp2_common/pico_crt0/crt0.S
kilograham: at which point i wonder if this should just default based on NO_FLASH, and allow you to set it to 0 if you really want for a flash binary

kilograham: i.e. line 395 would be just `#if !PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM`

src/rp2_common/pico_crt0/crt0.S
will-v-pi: Yep, I can change it, although it does get compiled to the 16-bit instruction. Note there are already cases of `stm !` in `bss_fill_loop` and `data_cpy_loop`, so it probably doesn't cause issues with other compilers.

src/rp2_common/pico_crt0/crt0.S
will-v-pi: I will change this - note that it defaults to `1`, but on RP2040 it is forced to be `!PICO_NO_FLASH` by not guarding the `#define` with `#ifndef`, because RP2040 doesn't have special watchdog boot types.

kilograham (approved):

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -382,10 +392,10 @@ binary_info_header:
 .global _entry_point
 _entry_point:
 
-#if PICO_NO_FLASH
-    // on the NO_FLASH case, we do not do a rest thru bootrom below, so the RCP may or may not have been initialized:
+#if !PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM
+    // in the NO_FLASH case, we do not do a rest thru bootrom below, so the RCP may or may not have been initialized:
```
lurch, line 396: Oops, just noticed there's a small "rest" -> "reset" typo here.

src/rp2_common/pico_crt0/crt0.S
```diff
@@ -20,6 +21,15 @@
 #define PICO_CRT0_NEAR_CALLS 0
 #endif
 
+// PICO_CONFIG: PICO_CRT0_DEBUG_ENTRY_RESETS_VIA_BOOTROM, Whether crt0 should reset through the bootrom when loaded using a debugger - defining this variable is not supported on RP2040, default=1, type=bool, advanced=true, group=pico_crt0
```
lurch, line 24: Would there be a use-case for somebody wanting to explicitly set this to 0 for RP2040 flash builds?
(Also, I guess this is a "setting" rather than a "variable" :wink: )

PR #2974, opened 2026-05-28
pico_cyw43_arch FreeRTOS+lwIP: route errno through newlib (LWIP_ERRNO_STDINCLUDE)

### LINKED ISSUE #2973: pico_cyw43_arch_lwip_sys_freertos: LWIP_PROVIDE_ERRNO=1 conflicts with newlib's thread-local errno
## Summary

`src/rp2_common/pico_cyw43_arch/CMakeLists.txt:70` unconditionally sets `LWIP_PROVIDE_ERRNO=1` on the `pico_cyw43_arch_lwip_sys_freertos` variant. Under FreeRTOS+newlib (the only libc bundled with the SDK's FreeRTOS port), this causes a silent storage divergence: lwIP writes errno values to its own `extern int errno` global, while any POSIX-style code reading `errno` via `<errno.h>` hits newlib's thread-local `(*__errno())` macro. The two are different memory locations.

The bug is silent in code that uses lwIP's own socket API (which reads lwIP's global directly), but **fatal for code that follows the standard mbedTLS BIO pattern** — i.e. mbedTLS's stock `library/net_sockets.c`, or any port of it. That code does non-blocking `recv()`, reads `errno`, and translates `EAGAIN`/`EWOULDBLOCK` into `MBEDTLS_ERR_SSL_WANT_READ` so the TLS state machine retries. With the storage mismatch, the shim reads `0` from newlib's untouched thread-local and treats every empty poll as a hard error → session tear-down.

## Affected configuration

- `pico_cyw43_arch_lwip_sys_freertos` (Pico W / Pico 2W under FreeRTOS)
- lwIP + mbedTLS + any TCP-bytestream client that uses the stock mbedTLS BIO (or an equivalent shim reading errno via `<errno.h>`)
- The bug surfaces at the first non-blocking `recv()` after the TLS handshake completes — typical observable: TLS handshake succeeds, application sends e.g. MQTT `CONNECT`, server queues `CONNACK`, client immediately FINs the connection without reading the response.

## Root cause

`pico_cyw43_arch/CMakeLists.txt`:

```cmake
target_compile_definitions(pico_cyw43_arch_lwip_sys_freertos_headers INTERFACE
        CYW43_LWIP=1
        LWIP_PROVIDE_ERRNO=1   # <-- declares lwIP's own global errno
        ...
)
```

`LWIP_PROVIDE_ERRNO=1` makes `lwip/errno.h` declare `extern int errno;` and define the `E*` constants itself. That's the correct choice for **bare-metal** builds where no libc owns `errno`. For FreeRTOS+newlib, newlib already provides a thread-local errno (with all the same `E*` constants), and lwIP's global silently diverges from it.

## Fix proposal

Switch the FreeRTOS+lwIP variant to `LWIP_ERRNO_STDINCLUDE=1`, which keeps lwIP's errno-constants behaviour but routes through `<errno.h>` — so lwIP and the rest of the program share newlib's thread-local storage:

```diff
 target_compile_definitions(pico_cyw43_arch_lwip_sys_freertos_headers INTERFACE
         CYW43_LWIP=1
-        LWIP_PROVIDE_ERRNO=1
+        LWIP_ERRNO_STDINCLUDE=1
         ...
)
```

This is behaviour-preserving for users who don't touch `errno` directly (lwIP's `E*` constants still resolve, just through `<errno.h>`). It is **correcting** for users on the stock mbedTLS BIO pattern.

The `threadsafe_background` (bare-metal) variant intentionally does **not** set `LWIP_PROVIDE_ERRNO=1` in the current code, so no change there is needed.

## Workarounds available today

A user can override the SDK's `-D` by defining `LWIP_ERRNO_STDINCLUDE` and `#undef LWIP_PROVIDE_ERRNO` in their own `lwipopts.h`. This works but the symptom is hard to diagnose without first understanding the storage divergence — the visible failure mode (TLS connection that handshakes successfully then dies on first `recv()`) leads most developers to look at TLS config, mbedTLS flags, cert chains, etc. first.

## PR

PR with the proposed change against `develop`: https://github.com/raspberrypi/pico-sdk/pull/2974

skycmoon:
Fixes #2973.

## What

Swap `LWIP_PROVIDE_ERRNO=1` for `LWIP_ERRNO_STDINCLUDE=1` on the `pico_cyw43_arch_lwip_sys_freertos` variant. Single hunk in `src/rp2_common/pico_cyw43_arch/CMakeLists.txt`.

## Why

`LWIP_PROVIDE_ERRNO=1` is correct for bare-metal builds (no libc owns `errno`), but the FreeRTOS+lwIP variant always ships with newlib, which already provides a thread-local `errno` via `<errno.h>`. The unconditional `LWIP_PROVIDE_ERRNO=1` makes lwIP declare its own `extern int errno` global, diverging from newlib's storage. Any code reading `errno` via `<errno.h>` — including the standard mbedTLS `library/net_sockets.c` BIO — misses the `EWOULDBLOCK` lwIP wrote, treats every non-blocking `recv()` poll as fatal, and tears the TLS session down right after handshake.

`LWIP_ERRNO_STDINCLUDE=1` keeps lwIP's `E*` constants available but routes them via `<errno.h>`, so lwIP and the rest of the program share newlib's thread-local storage. Issue #2973 has the full diagnosis and an observable symptom description (TLS handshake completes → application sends e.g. MQTT `CONNECT` → server queues `CONNACK` → client immediately FINs without reading the response → loop forever at the application's backoff cadence).

## Scope

- Only `pico_cyw43_arch_lwip_sys_freertos_headers` is touched.
- `pico_cyw43_arch_lwip_threadsafe_background` (bare-metal) does not currently set `LWIP_PROVIDE_ERRNO=1`, so no change needed there.
- Comment block added inline explaining the choice for future maintainers.

## Compatibility

Behaviour-preserving for users who don't touch `errno` directly — lwIP's `E*` constants still resolve, just through `<errno.h>`. Correcting for users on the stock mbedTLS BIO pattern (or any equivalent POSIX-style errno shim).

## Test

I verified the equivalent fix locally as a project-side override in `lwipopts.h`: cancelling the SDK's `-D LWIP_PROVIDE_ERRNO=1` and setting `LWIP_ERRNO_STDINCLUDE=1` resolved a persistent MQTT `CONNECT/CONNACK/close` storm on Pico 2W (FreeRTOS + lwIP + mbedTLS + coreMQTT). This PR moves that fix from a per-project workaround to the SDK default for the variant.

### DIFF
--- a/src/rp2_common/pico_cyw43_arch/CMakeLists.txt
+++ b/src/rp2_common/pico_cyw43_arch/CMakeLists.txt
@@ -67,7 +67,13 @@ if (PICO_CYW43_SUPPORTED) # set by BOARD=pico-w
                     pico_cyw43_arch_sys_freertos)
             target_compile_definitions(pico_cyw43_arch_lwip_sys_freertos_headers INTERFACE
                     CYW43_LWIP=1
-                    LWIP_PROVIDE_ERRNO=1
+                    # FreeRTOS+lwIP uses newlib, which provides a thread-local errno
+                    # via <errno.h>. LWIP_PROVIDE_ERRNO=1 would declare lwIP's own
+                    # `extern int errno` global, diverging from newlib's storage and
+                    # silently breaking mbedTLS BIO shims (and any code reading errno
+                    # via <errno.h>) that miss the EWOULDBLOCK lwIP wrote.
+                    # LWIP_ERRNO_STDINCLUDE=1 routes lwIP through <errno.h> instead.
+                    LWIP_ERRNO_STDINCLUDE=1
                     # now the default
                     #PICO_LWIP_CUSTOM_LOCK_TCPIP_CORE=1 # we want to override the lwip locking mechanism to use our mutex
                     )


peterharperuk (approved):

src/rp2_common/pico_cyw43_arch/CMakeLists.txt
```diff
@@ -67,7 +67,13 @@ if (PICO_CYW43_SUPPORTED) # set by BOARD=pico-w
                     pico_cyw43_arch_sys_freertos)
             target_compile_definitions(pico_cyw43_arch_lwip_sys_freertos_headers INTERFACE
                     CYW43_LWIP=1
-                    LWIP_PROVIDE_ERRNO=1
+                    # FreeRTOS+lwIP uses newlib, which provides a thread-local errno
```
kilograham, line 70: um, who says it uses newlib? surely it is dependent on the compiler/toolchain?

kilograham (changes requested):
i'm not convinced by this yet - as i say, who says that newlib is used? and this is non-overridable now as implemented

peterharperuk:
Yes. I think it was a mistake to add this to the library. It should have been left to be defined in lwipopts.h

PR #2977, opened 2026-06-01
resolve inconsistent use of PICO_PIO_VERSION > 0 and PICO_PIO_USE_GPIO_BASE

### LINKED ISSUE #2712: Inconsistency on the use of the PICO_RP2350A directive
I'm implementing a new design with the RP2350B and have found an inconsistency on the usage of the `PICO_RP2350A` directive.

After setting the PIO base to channel 16 I found that all the GPIOs were in the incorrect offset. I used GPIO20 as the base pin with `sm_config_set_in_pins`, then on the PIO I read 32 bits with `in pins 32`, so I expected to have what is in GPIO20 present at bit 0 of the result, but it was present in bit 16.

Checking the SDK cource code I can see that on `pio_set_gpio_base` and `pio_set_gpio_base_unsafe` the only check that is made is to check if the PIO version is higher than 0, what is set to 1 for any version of the RP2350:

```
static int pio_set_gpio_base_unsafe(PIO pio, uint gpio_base) {
    invalid_params_if_and_return(PIO, gpio_base != 0 && (!PICO_PIO_VERSION || gpio_base != 16), PICO_ERROR_BAD_ALIGNMENT);
#if PICO_PIO_VERSION > 0
    uint32_t used_mask = _used_instruction_space[pio_get_index(pio)];
    invalid_params_if_and_return(PIO, used_mask, PICO_ERROR_INVALID_STATE);
    pio->gpiobase = gpio_base;
#else
    ((void)pio);
    ((void)gpio_base);
#endif
    return PICO_OK;
}
```

But then, checking functions like `pio_sm_set_in_pins` or `sm_config_set_in_pins` we see this:

```
static inline void pio_sm_set_in_pins(PIO pio, uint sm, uint in_base) {
    check_pio_param(pio);
    check_sm_param(sm);
#if PICO_PIO_USE_GPIO_BASE
    in_base -= pio_get_gpio_base(pio);
#endif
    valid_params_if(HARDWARE_PIO, in_base < 32);
    pio->sm[sm].pinctrl = (pio->sm[sm].pinctrl & ~PIO_SM0_PINCTRL_IN_BASE_BITS) |
                 (in_base << PIO_SM0_PINCTRL_IN_BASE_LSB);
}
```

Instead of checking if `PICO_PIO_VERSION` is greater than zero it is being checked if `PICO_PIO_USE_GPIO_BASE` is set, and this is derived from the definition of `PICO_RP2350A`.

This leads to inconsistencies, in my case I did not defined a proper board file, is a custom board and didn't knew that defining `PICO_RP2350A` to zero was a requirement, and as the `pio_set_gpio_base` in fact did changed the base pin, it made me think that it was an error in my code somewhere else and not that I would need to define something else in order to be able to use the pio base offset properly.

As my point of view, this should be consistent, if `pio_set_gpio_base` allows to change the base GPIO for any variant of the RP2350 it should also always check the base set when the sm functions are used, or in the other way (and I think is the proper way) if `PICO_RP2350A` is not set to zero then do not apply the GPIO base offset and leave it to zero, in this way all is consistent using the check in base to the `PICO_RP2350A` definition.

Also, (maybe this is my fault) I haven't been able to find anythin in the documentation of the SDK about this behavior, in `pio_set_gpio_base` is documented that setting the value does not change the side effects of this setting, but in the `sm` functions there is no mention to the effects of having the version defined as RP2350B or not.

kilograham:
`PICO_PIO_VERSION > 0` applies to any RP2350, however `PICO_PIO_USE_GPIO_BASE=1` is only set by default for variants with >32 GPIOs

When `PICO_PIO_USE_GPIO_BASE= 0` we omit any setting/logic related to gpio_base (i.e. the SDK assumes it is always 0, and never sets it), so we replace old uses of `PICO_PIO_VERSION > 0` with `PICO_PUO_USE_GPIO_BASE` when gating gpio_base code

fixes #2712

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -151,6 +151,8 @@ typedef pio_hw_t *PIO;
 // PICO_CONFIG: PICO_PIO_USE_GPIO_BASE, Enable code for handling more than 32 PIO pins, type=bool, default=true when supported and when the device has more than 32 pins, group=hardware_pio
 #define PICO_PIO_USE_GPIO_BASE ((NUM_BANK0_GPIOS) > 32)
 #endif
+#else
+#define PICO_PIO_USE_GPIO_BASE 0
 #endif
 
 /**
@@ -784,7 +786,7 @@ static inline pio_sm_config pio_get_default_sm_config(void) {
  * \return the current GPIO base for the PIO instance
   */
 static inline uint pio_get_gpio_base(PIO pio) {
-#if PICO_PIO_VERSION > 0
+#if PICO_PIO_USE_GPIO_BASE
     return pio->gpiobase;
 #else
     ((void)pio);
--- a/src/rp2_common/hardware_pio/pio.c
+++ b/src/rp2_common/hardware_pio/pio.c
@@ -83,7 +83,7 @@ static int find_offset_for_program(PIO pio, const pio_program_t *program) {
 
 static int pio_set_gpio_base_unsafe(PIO pio, uint gpio_base) {
     invalid_params_if_and_return(HARDWARE_PIO, gpio_base != 0 && (!PICO_PIO_VERSION || gpio_base != 16), PICO_ERROR_BAD_ALIGNMENT);
-#if PICO_PIO_VERSION > 0
+#if PICO_PIO_USE_GPIO_BASE
     uint32_t used_mask = _used_instruction_space[pio_get_index(pio)];
     invalid_params_if_and_return(HARDWARE_PIO, used_mask, PICO_ERROR_INVALID_STATE);
     pio->gpiobase = gpio_base;
@@ -96,7 +96,7 @@ static int pio_set_gpio_base_unsafe(PIO pio, uint gpio_base) {
 
 int pio_set_gpio_base(PIO pio, uint gpio_base) {
     int rc = PICO_OK;
-#if PICO_PIO_VERSION > 0
+#if PICO_PIO_USE_GPIO_BASE
     uint32_t save = hw_claim_lock();
     rc = pio_set_gpio_base_unsafe(pio, gpio_base);
     hw_claim_unlock(save);
@@ -108,7 +108,7 @@ int pio_set_gpio_base(PIO pio, uint gpio_base) {
 }
 
 static bool is_gpio_compatible(PIO pio, uint32_t used_gpio_ranges) {
-#if PICO_PIO_VERSION > 0
+#if PICO_PIO_USE_GPIO_BASE
     bool gpio_base = pio_get_gpio_base(pio);
     return !((gpio_base && (used_gpio_ranges & 1)) ||
              (!gpio_base && (used_gpio_ranges & 4)));


will-v-pi (approved):

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -151,6 +151,8 @@ typedef pio_hw_t *PIO;
 // PICO_CONFIG: PICO_PIO_USE_GPIO_BASE, Enable code for handling more than 32 PIO pins, type=bool, default=true when supported and when the device has more than 32 pins, group=hardware_pio
 #define PICO_PIO_USE_GPIO_BASE ((NUM_BANK0_GPIOS) > 32)
```
lurch, line 152: It might make sense to re-order:
```c
#if PICO_PIO_VERSION > 0
#ifndef PICO_PIO_USE_GPIO_BASE
// PICO_CONFIG: PICO_PIO_USE_GPIO_BASE, Enable code for handling more than 32 PIO pins, type=bool, default=true when supported and when the device has more than 32 pins, group=hardware_pio
#define PICO_PIO_USE_GPIO_BASE ((NUM_BANK0_GPIOS) > 32)
#endif
#else
#define PICO_PIO_USE_GPIO_BASE 0
#endif
```
to
```c
// PICO_CONFIG: PICO_PIO_USE_GPIO_BASE, Enable code for handling more than 32 PIO pins, type=bool, default=true when supported and when the device has more than 32 pins, group=hardware_pio
#ifndef PICO_PIO_USE_GPIO_BASE
#if PICO_PIO_VERSION > 0
#define PICO_PIO_USE_GPIO_BASE ((NUM_BANK0_GPIOS) > 32)
#else
#define PICO_PIO_USE_GPIO_BASE 0
#endif
```
?

will-v-pi: I think the whole point is it forces `PICO_PIO_USE_GPIO_BASE` to 0 if `PICO_PIO_VERSION=0`, because if you have `PICO_PIO_VERSION=0` and set `PICO_PIO_USE_GPIO_BASE=1`, your code won't compile

lurch: Ahh of course :facepalm:

src/rp2_common/hardware_pio/pio.c
```diff
@@ -83,7 +83,7 @@ static int find_offset_for_program(PIO pio, const pio_program_t *program) {
 
 static int pio_set_gpio_base_unsafe(PIO pio, uint gpio_base) {
     invalid_params_if_and_return(HARDWARE_PIO, gpio_base != 0 && (!PICO_PIO_VERSION || gpio_base != 16), PICO_ERROR_BAD_ALIGNMENT);
```
lurch, line 85: I guess this also ought to be using `PICO_PIO_USE_GPIO_BASE` rather than `PICO_PIO_VERSION` ?

lurch:
Hmmm, if `PICO_PIO_USE_GPIO_BASE == 0` then `is_gpio_compatible()` **always** returns true. I _think_ this means that if there's a PIO program which is specifically writing to pins above 32, and the user tries to load that program on a RP2350A, then it'll be reported as "compatible"?
Should `is_gpio_compatible()` return false if `used_gpio_ranges & 4` when `PICO_PIO_USE_GPIO_BASE == 0`?

src/rp2_common/hardware_pio/pio.c
kilograham: i will re-review in the light of the other comment - it isn't clear without looking deeper what the contract of this particular method is when PICO_PIO_USE_GPIO_BASE=0 - probably as you say it shouldn't be called to change the gpio base at all (indeed it probably just shouldn't exist)

kilograham:
oops - i merged this by mistake as it was approved, and i got trigger happy, i'll go revist in another PR

kilograham:
opened #2989

PR #2978, opened 2026-06-01
make queue_init return a bool success

### LINKED ISSUE #2851: How to detect queue_init memory allocation failure ?
queue_init() does not return a value so the API doesn't provide a way to detect an error condition.

I solved a recent problem where increasing the size of the queue (from 1024 to 2048 elements of 16 bytes) caused a core crash when adding an item to the queue with queue_try_add(). I can only think this was a consequence of a failure to allocate memory. (On RP2040).

Absent a change to the API, is it reasonable to peek inside the queue_t struct to see if the data member is NULL, as one would expect it to be if the calloc() failed ?

Thanks.

kilograham:
fixes #2851

### DIFF
--- a/src/common/pico_util/include/pico/util/queue.h
+++ b/src/common/pico_util/include/pico/util/queue.h
@@ -48,8 +48,9 @@ typedef struct {
  * \param element_size Size of each value in the queue
  * \param element_count Maximum number of entries in the queue
  * \param spinlock_num The spin ID used to protect the queue
+ * \return true if the queue was initialized; false if it couldn't be (e.g. malloc failed)
  */
-void queue_init_with_spinlock(queue_t *q, uint element_size, uint element_count, uint spinlock_num);
+bool queue_init_with_spinlock(queue_t *q, uint element_size, uint element_count, uint spinlock_num);
 
 /*! \brief Initialise a queue, allocating a (possibly shared) spinlock
  *  \ingroup queue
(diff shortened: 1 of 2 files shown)


will-v-pi (dismissed):

src/common/pico_util/include/pico/util/queue.h
```diff
@@ -48,18 +48,20 @@ typedef struct {
  * \param element_size Size of each value in the queue
  * \param element_count Maximum number of entries in the queue
  * \param spinlock_num The spin ID used to protect the queue
```
lurch:
```suggestion
 * \param spinlock_num The spinlock ID used to protect the queue
```
?

src/common/pico_util/include/pico/util/queue.h
```diff
@@ -48,18 +48,20 @@ typedef struct {
...
+ * \return true if the queue was initialized; false if it couldn't be (e.g. malloc failed)
```
lurch: Does it matter that it says malloc in `queue.h` but uses calloc in `queue.c` ?

src/common/pico_util/queue.c
```diff
@@ -8,13 +8,14 @@
 #include <string.h>
 #include "pico/util/queue.h"
 
-void queue_init_with_spinlock(queue_t *q, uint element_size, uint element_count, uint spinlock_num) {
+bool queue_init_with_spinlock(queue_t *q, uint element_size, uint element_count, uint spinlock_num) {
     lock_init(&q->core, spinlock_num);
-    q->data = (uint8_t *)calloc(element_count + 1, element_size);
     q->element_count = (uint16_t)element_count;
```
lurch, line 13: Would it make sense to set `q->element_count` to 0 if the `calloc` fails?

src/common/pico_util/include/pico/util/queue.h
kilograham: not really, but i could change it to alloc (I did think about this as i was typing it)

src/common/pico_util/queue.c
kilograham: seems like overkill (and extra code)

PR #2980, opened 2026-06-03
Fix queue.h line endings

will-v-pi:


kilograham (approved):

lurch:
This is a follow-up to #2978
It seems like using the "Suggestion" feature in GitHub PRs sometimes causes GitHub to (unexpectedly) switch the line-endings on an entire file :cry:

PR #2983, opened 2026-06-05
Fix sections.h doxygen

will-v-pi:
Was missing docs for `__time_critical_func` and `__persistent_data`, and picking up `PICO_NOT_IN_FLASH_PLACEMENT` instead of `__not_in_flash`, mostly because of the `PREDEFINED` stuff in the Doxyfile.in

I believe the Doxyfile.in changes are fine, because the docs are always in the .h file but the function name wrapped in `__time_critical_func` etc is in the .c file. I couldn't see any adverse effects when checking other functions marked `__time_critical_func` etc.

### DIFF
--- a/src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
+++ b/src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
@@ -155,7 +155,7 @@
  *              garbage-collected independently
  */
 #ifndef __in_xip_ram
-#if PICO_USE_XIP_CACHE_AS_RAM
+#if PICO_USE_XIP_CACHE_AS_RAM || PICO_COMBINED_DOCS
 #define __in_xip_ram(group) __attribute__((section(".xip_ram." group)))
 #elif PICO_XIP_RAM
 #define __in_xip_ram(group) __in_ram(group)
@@ -198,7 +198,7 @@
  *
  * \param name  the name of the variable to place in the section
  */
-#if LIB_PICO_LOW_POWER && HAS_POWMAN_TIMER
+#if (LIB_PICO_LOW_POWER && HAS_POWMAN_TIMER) || PICO_COMBINED_DOCS
 #ifndef __persistent_data
 #define __persistent_data(name) __attribute__((section(".persistent_data." #name))) name
 #endif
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
```diff
@@ -199,6 +198,7 @@
  *
  * \param name  the name of the variable to place in the section
  */
+#if LIB_PICO_LOW_POWER && HAS_POWMAN_TIMER
```
kilograham: what is the purpose of this addition (which doesn't seem to correspond with a new #endif)

i'm not sure this will cause it to be included correctly in doxygen if that is the intent?

will-v-pi: It’s been moved down from line 186 so that this does get included in the doxygen - when this define is not met (which is the case in the docs build), it sees the .bss define below, and therefore still includes `__persistent_data` in the doxygen. Before this change it wasn’t included in the doxygen.

kilograham (dismissed):

src/rp2_common/pico_platform_sections/include/pico/platform/sections.h
kilograham: i'm still a bit confused tho; without running the doxygen - isn't in weirded out by the doxygen comment being outside the #if, and the #if is presumably false for doxygen?

will-v-pi: It builds fine - the previous issue was the doxygen comment was inside the `#if`, so wasn't being included - now the doxygen comment is outside the `#if` and the `#if` turns into `#define __persistent_data(name) __attribute__((section(".bss." #name))) name` which the doxygen picks up. It is displaying the wrong version, as it picks up the `.bss` version not the `.persistent_data` version, but at least it is included now

will-v-pi: I've added `|| PICO_COMBINED_DOCS` to this define, so the docs build picks up the `.persistent_data` define

kilograham (approved):

PR #2987, opened 2026-06-06
fix compiler warnings

kilograham:


src/rp2_common/hardware_pio/pio.c
```diff
@@ -438,13 +438,13 @@ bool pio_claim_free_sm_and_add_program_for_gpio_range(const pio_program_t *progr
             // We need to claim an SM on the PIO
             int8_t sm_index[NUM_PIO_STATE_MACHINES];
             // on second pass, if there is one, we try and claim all the state machines so that we can change the GPIO base
-            uint num_claimed;
-            for(num_claimed = 0; num_claimed < (pass ? NUM_PIO_STATE_MACHINES : 1u) ; num_claimed++) {
+            int num_claimed;
+            for(num_claimed = 0; num_claimed < (pass ? (int)NUM_PIO_STATE_MACHINES : 1) ; num_claimed++) {
```
lurch, line 442: Hmm, looks like `num_claimed` can't ever be negative? :thinking:

kilograham: true, but i don't have to use an unsigned int just for that reason, and it is used in signed arithmetic below

will-v-pi (approved):

PR #2989, opened 2026-06-06
fix PIO gpio compatibility checks

kilograham:


src/rp2_common/hardware_pio/pio.c
```diff
@@ -107,22 +107,24 @@ int pio_set_gpio_base(PIO pio, uint gpio_base) {
     return rc;
 }
 
+// note this function checks compatibility of GPIO ranges even on PIO_VERSION 0;
+// i.e. it will fail if you try to ask for pins 32-47 on RP2040
```
lurch: Perhaps?
```suggestion
// i.e. it will fail if you try to ask for pins 32-47 on RP2040 (or RP2350A)
```

src/rp2_common/hardware_pio/pio.c
```diff
@@ -107,22 +107,24 @@ int pio_set_gpio_base(PIO pio, uint gpio_base) {
...
 static bool is_gpio_compatible(PIO pio, uint32_t used_gpio_ranges) {
 #if PICO_PIO_USE_GPIO_BASE
     bool gpio_base = pio_get_gpio_base(pio);
     return !((gpio_base && (used_gpio_ranges & 1)) ||
              (!gpio_base && (used_gpio_ranges & 4)));
 #else
     ((void)pio);
-    ((void)used_gpio_ranges);
-    return true;
+    return used_gpio_ranges < 4; // only using pins 0--31
```
lurch:
```suggestion
    return used_gpio_ranges < 4; // only using pins 0-31
```

src/rp2_common/hardware_pio/pio.c
```diff
@@ -107,22 +107,24 @@ int pio_set_gpio_base(PIO pio, uint gpio_base) {
...
 #endif
 }
 
 static bool is_program_gpio_compatible(PIO pio, const pio_program_t *program) {
 #if PICO_PIO_VERSION > 0
     return is_gpio_compatible(pio, program->used_gpio_ranges);
 #else
```
lurch, line 118: At first I wondered why this was gated on `PICO_PIO_VERSION` rather than `PICO_PIO_USE_GPIO_BASE`, but then I realised that this matches https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_pio/include/hardware/pio.h#L941 :+1:

src/rp2_common/hardware_pio/pio.c
```diff
@@ -429,7 +431,12 @@ bool pio_claim_free_sm_and_add_program_for_gpio_range(const pio_program_t *progr
     uint32_t required_gpio_ranges;
     if (gpio_count) required_gpio_ranges = (1u << (gpio_start >> 4)) | (1u << ((gpio_start + gpio_count - 1) >> 4));
     else            required_gpio_ranges = 0;
+#if PICO_PIO_USE_GPIO_BASE
     int passes = set_gpio_base ? 2 : 1;
+#else
+    ((void)set_gpio_base); // silently ignore parameter rather than error
```
lurch: Is it worth adding a comment to `pio.h` that this parameter is ignored on non-RP2350B ?

lurch:
While digging into this yesterday, I noticed that [`pio.h`](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_pio/include/hardware/pio.h#L76) says:
`If PICO_PIO_USE_GPIO_BASE != 1 then ...` and `If PICO_PIO_USE_GPIO_BASE == 1 then...`
However the condition is always checked as `#if PICO_PIO_USE_GPIO_BASE` or `#if !PICO_PIO_USE_GPIO_BASE`, so I guess it'd be more accurate to say:
`If PICO_PIO_USE_GPIO_BASE == 0 then ...` and `If PICO_PIO_USE_GPIO_BASE != 0 then...`
?

src/rp2_common/hardware_pio/pio.c
```diff
@@ -429,7 +431,12 @@ bool pio_claim_free_sm_and_add_program_for_gpio_range(const pio_program_t *progr
...
+    int passes = 1;
+#endif
```
will-v-pi: Is this addition even needed - it sets `set_gpio_base = 0;` above when `#if !PICO_PIO_USE_GPIO_BASE`, which should have the same effect

src/rp2_common/hardware_pio/pio.c
will-v-pi: I agree, but the comment should be more specific that it's ignored when `PICO_PIO_USE_GPIO_BASE=0`, as `PICO_PIO_USE_GPIO_BASE` is a `PICO_CONFIG` so can be overridden by the user on RP2350A to be 1

src/rp2_common/hardware_pio/pio.c
kilograham: yaeh, but this produces a lot less code when using <=32 pins

src/rp2_common/hardware_pio/pio.c
kilograham: fixed - note github seems to have started adding ^M even from the apply batch tab :-(

src/rp2_common/hardware_pio/pio.c
will-v-pi: In that case, should the `gpio_count` stuff above also be included in this `#ifdef` to produce even less code?

I can't do a proper suggestion because of where I commented, but something like
```suggestion
#if PICO_PIO_USE_GPIO_BASE
    uint32_t required_gpio_ranges;
    if (gpio_count) required_gpio_ranges = (1u << (gpio_start >> 4)) | (1u << ((gpio_start + gpio_count - 1) >> 4));
    else            required_gpio_ranges = 0;
    int passes = set_gpio_base ? 2 : 1;
#else
    ((void)gpio_count);
    ((void)set_gpio_base); // silently ignore parameter rather than error
    uint32_t required_gpio_ranges = 0;
    int passes = 1;
#endif
```

Then the:
```
#if !PICO_PIO_USE_GPIO_BASE
    // short-circuit some logic when not using GPIO_BASE
    set_gpio_base = 0;
    gpio_count = 0;
#endif
```
above can be removed entirely

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -2043,8 +2043,9 @@ bool pio_claim_free_sm_and_add_program(const pio_program_t *program, PIO *pio, u
  * \param gpio_start the lowest GPIO number required (0-47 on RP2350B, 0-31 otherwise)
  * \param gpio_count the count of consecutive GPIOs required
  * \param set_gpio_base if there is no free SM on a PIO instance with the right GPIO base, and there IS an unused PIO
- *                      instance, then that PIO will be reconfigured so that this method can succeed
- *
+ *                      instance, then that PIO will be reconfigured so that this method can succeed. Note
+ *                      this parameter is ignored when PICO_PIO_USE_GPIO_BASE=0; i.e. by default on anything other than RP2350B
+0 *
```
```suggestion
 *
```

src/rp2_common/hardware_pio/pio.c
lurch: It looks like the `if (gpio_count)` was only needed because it was explicitly being set to 0 in the earlier `#if !PICO_PIO_USE_GPIO_BASE` section?

On a slight side-note, presumably there'd be no point calling `pio_claim_free_sm_and_add_program_for_gpio_range()` with an actual `gpio_count` paramter of 0, so maybe it's worth adding
```c
invalid_params_if(HARDWARE_PIO, !gpio_count);
```
?

will-v-pi: With all my compilers (arm-none-eabi-gcc 12.2.1; riscv32-unknown-elf-gcc 15.1.0; clang 20.1.0, 21.1.1 and 22.1.0) doing both Release and Debug builds, this addition makes no difference to code size - the compiler is already removing the unecessary code due to the `set_gpio_base = 0;` and `gpio_count = 0;` above.

What compiler & build are you using that produces more code without this change?

kilograham: not sure which change you are refering to now

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -2043,10 +2043,14 @@ bool pio_claim_free_sm_and_add_program(const pio_program_t *program, PIO *pio, u
...
  * \param gpio_count the count of consecutive GPIOs required
  * \param set_gpio_base if there is no free SM on a PIO instance with the right GPIO base, and there IS an unused PIO
- *                      instance, then that PIO will be reconfigured so that this method can succeed
+ *                      instance, then that PIO will be reconfigured so that this method can succeed. Note
+ *                      this parameter is ignored when PICO_PIO_USE_GPIO_BASE=0; i.e. by default on anything other than RP2350B
  *
  * \return true on success, false otherwise
  * \see pio_remove_program_and_unclaim_sm
+ *
+ * \note on RP2050 or RP2350A (strictly when PICO_PIO_USE_GPIO_BASE == 0), gpio_start + gpio_count must be <= 32
```
will-v-pi:
```suggestion
 * \note on RP2040 or RP2350A (strictly when PICO_PIO_USE_GPIO_BASE == 0), gpio_start + gpio_count must be <= 32
```

src/rp2_common/hardware_pio/pio.c
```diff
@@ -107,22 +104,19 @@ int pio_set_gpio_base(PIO pio, uint gpio_base) {
...
-#else
-    ((void)pio);
-    ((void)used_gpio_ranges);
-    return true;
-#endif
 }
+#endif
 
 static bool is_program_gpio_compatible(PIO pio, const pio_program_t *program) {
 #if PICO_PIO_VERSION > 0
```
will-v-pi, line 116: I think this is what's causing the CI failure - this should be

```suggestion
#if PICO_PIO_USE_GPIO_BASE
```

to match the wrap around `is_gpio_compatible`

src/rp2_common/hardware_pio/pio.c
```diff
@@ -412,17 +405,30 @@ void pio_sm_drain_tx_fifo(PIO pio, uint sm) {
...
+            int offset_or_error = pio_add_program(*pio, program);
+            if (offset_or_error >= 0) {
+                *sm = sm_or_error;
+                *offset = offset_or_error;
+                return true;
+            }
+            pio_sm_unclaim(*pio, sm_or_error);
+        }
+    }
+    return false;
```
lurch, line 423: I guess this should also do the `*pio = NULL;` that `pio_claim_free_sm_and_add_program_for_gpio_range` does?

src/rp2_common/hardware_pio/pio.c
```diff
@@ -412,17 +405,30 @@ void pio_sm_drain_tx_fifo(PIO pio, uint sm) {
...
 #if !PICO_PIO_USE_GPIO_BASE
-    // short-circuit some logic when not using GPIO_BASE
-    set_gpio_base = 0;
-    gpio_count = 0;
-#endif
+    if (gpio_start + gpio_count > 32) return false;
+    (void)set_gpio_base;
+    return pio_claim_free_sm_and_add_program(program, pio, sm, offset);
+#else
+    invalid_params_if(HARDWARE_PIO, (gpio_start + gpio_count) > NUM_BANK0_GPIOS);
```
lurch: Should this assertion be moved back up above the `#if !PICO_PIO_USE_GPIO_BASE` line (as it was previously), as if you're debugging I guess an assertion is easier to spot than a function returning false?

src/rp2_common/hardware_pio/pio.c
```diff
@@ -412,17 +405,30 @@ void pio_sm_drain_tx_fifo(PIO pio, uint sm) {
...
+                *offset = offset_or_error;
+                return true;
+            }
+            pio_sm_unclaim(*pio, sm_or_error);
+        }
+    }
+    return false;
 }
 
 bool pio_claim_free_sm_and_add_program_for_gpio_range(const pio_program_t *program, PIO *pio, uint *sm, uint *offset, uint gpio_start, uint gpio_count, bool set_gpio_base) {
```
lurch: As I suggested in an earlier comment, should any function that takes a `gpio_count` parameter do
```c
invalid_params_if(HARDWARE_PIO, gpio_count == 0);
```
(because I can't think of a legitimate reason why you'd do that?)

src/rp2_common/hardware_pio/pio.c
lurch: ...although `program->used_gpio_ranges` is gated on `#if PICO_PIO_VERSION > 0` so perhaps it makes sense to restore `is_gpio_compatible` to an earlier version in this PR where it worked for RP2350A?

src/rp2_common/hardware_pio/pio.c
lurch: Oh, and I guess if you did move the `invalid_params_if` back up to where it was, it'd assert for somebody trying to do `gpio_start=29, gpio_count=2` on RP2040 :wink:

src/rp2_common/hardware_pio/pio.c
lurch: This thread is now invalid after the latest refactoring.

lurch:
Random thought: It's not something that I've tested, but if you explicitly set `PICO_PIO_USE_GPIO_BASE=1` on an RP2350A and then did `pio_set_gpio_base(pio, 16)` would / should all the pin-setting methods then only allow to access pins 16 to 29 inclusive? :wink:   (Not that you'd ever do that of course, but perhaps it'd be a good test of the logic in this PR?)

src/rp2_common/hardware_pio/pio.c
kilograham: yeah, i mean this is indeed gated on requiring the member in the struct

src/rp2_common/hardware_pio/pio.c
kilograham: yeah, but i have changed it to an invalid_params_if

src/rp2_common/hardware_pio/pio.c
kilograham: i have changed both to only touch the output parameters on success

kilograham:
> Random thought: It's not something that I've tested, but if you explicitly set `PICO_PIO_USE_GPIO_BASE=1` on an RP2350A and then did `pio_set_gpio_base(pio, 16)` would / should all the pin-setting methods then only allow to access pins 16 to 29 inclusive? 😉 (Not that you'd ever do that of course, but perhaps it'd be a good test of the logic in this PR?)

if you explicitly do that, I think it'll just give you 32 pins but with the top-bit swapped. (i.e. pin 0 <> pin 16, pin 1 <> pin 17)

src/rp2_common/hardware_pio/pio.c
kilograham: what earlier comment? the set_out_pin_count etc can take 0 - this one can actually use 0 to to mean "don't care" and i will doc that

kilograham: oh yeah, i see what you mean - now that this method is not used for antyhing else - it doesn't make much sense for it to take 0

src/rp2_common/hardware_pio/pio.c
```diff
@@ -411,66 +404,79 @@ void pio_sm_drain_tx_fifo(PIO pio, uint sm) {
...
-
+bool pio_claim_free_sm_and_add_program_for_gpio_range(const pio_program_t *program, PIO *pio_out, uint *sm_out, uint *offset_out, uint gpio_start, uint gpio_count, bool set_gpio_base) {
 #if !PICO_PIO_USE_GPIO_BASE
-    // short-circuit some logic when not using GPIO_BASE
-    set_gpio_base = 0;
-    gpio_count = 0;
-#endif
+    invalid_params_if(HARDWARE_PIO, (gpio_start + gpio_count) > 32);
+    (void)set_gpio_base;
+    return pio_claim_free_sm_and_add_program(program, pio, sm, offset);
```
lurch: Oops, you forgot to update these
```suggestion
    return pio_claim_free_sm_and_add_program(program, pio_out, sm_out, offset_out);
```

src/rp2_common/hardware_pio/pio.c
lurch: Sorry for nit-picking, but I believe these are actually two separate (but similar) issues?
I think that _outside_ of any `#if` you ought to be doing
```c
invalid_params_if(HARDWARE_PIO, (gpio_start + gpio_count) > NUM_BANK0_GPIOS);
```
(because this will always be true, regardless of what else is going on)

and then specifically in the `#if !PICO_PIO_USE_GPIO_BASE` case you should do
```c
invalid_params_if_and_return(HARDWARE_PIO, (gpio_start + gpio_count) > 32, false);
```
(to match the `\note` you added in `pio.h`)
?

src/rp2_common/hardware_pio/pio.c
```diff
@@ -107,22 +104,19 @@ int pio_set_gpio_base(PIO pio, uint gpio_base) {
...
     bool gpio_base = pio_get_gpio_base(pio);
     return !((gpio_base && (used_gpio_ranges & 1)) ||
              (!gpio_base && (used_gpio_ranges & 4)));
-#else
-    ((void)pio);
-    ((void)used_gpio_ranges);
-    return true;
-#endif
 }
+#endif
```
lurch, line 113: I think that checking for **both** of `PICO_PIO_VERSION` and `PICO_PIO_USE_GPIO_BASE` on the same line will just lead to later confusion? Should this instead be:
```c
#if PICO_PIO_VERSION > 0
static bool is_gpio_compatible(PIO pio, uint32_t used_gpio_ranges) {
#if PICO_PIO_USE_GPIO_BASE
    bool gpio_base = pio_get_gpio_base(pio);
    return !((gpio_base && (used_gpio_ranges & 1)) ||
             (!gpio_base && (used_gpio_ranges & 4)));
#else // RP2350A
    ((void)pio);
    return used_gpio_ranges < 4; // only using pins 0--31
#endif
}
#endif
```
(a slight adaptation of the code that appeared in https://github.com/raspberrypi/pico-sdk/pull/2989/commits/0b0b3a436f9e7bae60735d4c0754ca77d51522e4 )

src/rp2_common/hardware_pio/pio.c
lurch: > the set_out_pin_count etc can take 0 - this one can actually use 0 to to mean "don't care" and i will doc that

Ah right, so that'd be saying "I don't want to use any Out pins"?

src/rp2_common/hardware_pio/pio.c
kilograham: i reworked again - i want it to allow up to 32 on RP2040, and its fine for now on RP2350B as that is 48 (which is a handy multiple of 16)

src/rp2_common/hardware_pio/pio.c
kilograham: nope; the function is only used if those conditions are true

lurch: Okay - I've double-checked and I think my mistake was assuming that `pio_get_gpio_base()` wasn't defined when `PICO_PIO_USE_GPIO_BASE==0`, and therefore this would go wrong on RP2350A if a PIO v1 program tries to use a pin above 32.

But when `PICO_PIO_USE_GPIO_BASE==0` then `pio_get_gpio_base()` just returns 0, and so on a RP2350A (i.e `PICO_PIO_VERSION==1` and `PICO_PIO_USE_GPIO_BASE==0`) then `is_gpio_compatible()` will effectively collapse down to
```c
    return !(used_gpio_ranges & 4);
```
:slightly_smiling_face:

src/rp2_common/hardware_pio/pio.c
lurch: Yeah I see that you've cleverly combined both my suggestions above into a single `invalid_params_if_and_return()` :grinning:

src/rp2_common/hardware_pio/pio.c
kilograham: yeah except this is a stale dicussion - as 0 isn't allowed any more

src/rp2_common/hardware_pio/pio.c
```diff
@@ -411,66 +404,79 @@ void pio_sm_drain_tx_fifo(PIO pio, uint sm) {
...
-    invalid_params_if(HARDWARE_PIO, (gpio_start + gpio_count) > NUM_BANK0_GPIOS);
-
+bool pio_claim_free_sm_and_add_program_for_gpio_range(const pio_program_t *program, PIO *pio_out, uint *sm_out, uint *offset_out, uint gpio_start, uint gpio_count, bool set_gpio_base) {
+    invalid_params_if_and_return(HARDWARE_PIO, (gpio_start + gpio_count) > MAX(32, NUM_BANK0_GPIOS), false);
 #if !PICO_PIO_USE_GPIO_BASE
-    // short-circuit some logic when not using GPIO_BASE
-    set_gpio_base = 0;
-    gpio_count = 0;
-#endif
+    invalid_params_if(HARDWARE_PIO, (gpio_start + gpio_count) > 32);
```
lurch, line 429: Is this now redundant in light of the `invalid_params_if_and_return()` just two lines above?
Or is it needed in case somebody sets `PICO_PIO_USE_GPIO_BASE=0` on a RP2350B?

PR #2991, opened 2026-06-08
Small clarifications to the PIO "pinhi" code

lurch:
I changed "5th bit" to "bit 5" to remove any ambiguity with zero-based indexing (personally I'd consider "bit 0" to be "the first bit").

I removed the "magic number" in `sm_config_set_jmp_pin`.

I added some explanatory comments.

I changed `used` from `uint` to an explicit `uint32_t`, because I believe the bit-shifting on line 848 won't work correctly if it's not 32-bits? (`pinhi` itself is also declared as `uint32_t`)

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -73,7 +73,7 @@
  * instance either addresses pins 0-31 or 16-47 based on \ref pio_set_gpio_base). The
  * `pio_sm_` methods that directly affect the hardware always take _real_ pin numbers in the full range, however:
  *
- * * If `PICO_PIO_USE_GPIO_BASE != 1` then the 5th bit of the pin number is ignored. This is done so
+ * * If `PICO_PIO_USE_GPIO_BASE != 1` then bit 5 of the pin number is ignored. This is done so
  *   that programs compiled for boards with RP2350A do not incur the extra overhead of dealing with higher pins that don't exist.
  *   Effectively these functions behave exactly like RP2040 in this case.
  *   Note that `PICO_PIO_USE_GPIO_BASE` is defaulted to 0 if `PICO_RP2350A` is 1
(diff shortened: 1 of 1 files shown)


src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -827,19 +826,25 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
     invalid_params_if_and_return(PIO, gpio_under_16 && gpio_base, PICO_ERROR_BAD_ALIGNMENT);
     invalid_params_if_and_return(PIO, gpio_over_32 && !gpio_base, PICO_ERROR_BAD_ALIGNMENT);
     // flip the top bit of any used (execctrl/pinctrl) values to turn:
     // bit6(32) + 0-15  -> base(16) + 16-31
     // bit6(0)  + 16-31 -> base(16) + 0-15
     static_assert(PINHI_EXECCTRL_LSB == 20, ""); // we use shifts to mask off bits below
-    pio->sm[sm].execctrl = config->execctrl ^ (gpio_base ? ((used >> 20) << (PIO_SM0_EXECCTRL_JMP_PIN_LSB + 4)) : 0);
+    pio->sm[sm].execctrl = config->execctrl ^ (gpio_base ? ((used >> PINHI_EXECCTRL_LSB) << (PIO_SM0_EXECCTRL_JMP_PIN_LSB + 4)) : 0);
+    // the "12" here is a result of "sizeof(used) - PINHI_EXECCTRL_LSB" (i.e. 32 - 20) and is used to shift off the execctrl bits
     pio->sm[sm].pinctrl = config->pinctrl ^ (gpio_base ? ((used << 12) >> 8) : 0);
```
kilograham, line 848: probably worth converting the magic numbers into logic based off PINHI_EXECCTRL_LSB then

lurch: I _could_ do that, but would
```c
pio->sm[sm].pinctrl = config->pinctrl ^ (gpio_base ? ((used & (PINHI_ALL_PINCTRL_LSBS * 0x1f)) << 4) : 0);
```
also be an acceptable approach? (with the comments obviously updated to suit)

Please double-check my maths; bit-twiddling isn't my strong point!

kilograham: i think i prefer the shifts - less code generated likely since no need to fabricate a weird constant

kilograham: note also the comment above should say "flip the top bits"

kilograham: actually that is not true!!

kilograham: this is a bug, it should just be flipping (PINHI_ALL_PINCTR_LSBS << 4u) on execctrl . i will merge your change and fix this and another issue

kilograham (approved):

src/rp2_common/hardware_pio/include/hardware/pio.h
lurch: > less code generated likely since no need to fabricate a weird constant

Obviously doesn't matter now, but I copied the `PINHI_ALL_PINCTRL_LSBS * 0x1f` from https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_pio/include/hardware/pio.h#L311 :slightly_smiling_face: (But you're probably right about the code-generation, as that's in a `static_assert` which will only be evaluated at compile-time?)

lurch:
Continued in #2993 ...

PR #2993, opened 2026-06-08
clarify pio_sm_set_config use of pinhi

kilograham:
I though there was a bug in the code, but there wasn't - it just was still rather unclear - this attempts to make it more understandable


src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -283,10 +283,15 @@ static_assert(DREQ_PIO2_RX0 == DREQ_PIO2_TX0 + NUM_PIO_STATE_MACHINES, "");
...
  *
- *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configre the PIO's
+ *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configure the PIO's
  *   GPIO base before calling the method, however you can use \ref pio_claim_free_sm_and_add_program_for_gpio_range
  *   to find/configure a PIO instance suitable for a particular GPIO range.
  *
+ * \note when `PICO_PIO_USE_GPIO_BASE == 1` \ref pio_sm_set_config ignores fields which haven't had the corresponding
+ * `sm_config_` pin function called, so that you don't have to move settings for unused pin sets into the correct
+ * pin range. Therefore it is always a best practice to explicitly configure a pin range starting at pon zero to 0,
+ * rather than relying on the default value being 0.
```
```suggestion
 * pin range. Therefore it is always best practice to explicitly configure a pin range starting at pin zero to 0,
 * rather than relying on the default value being 0.
```
Small typo, but I'm also not sure what this is trying to actually say?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -813,7 +818,7 @@ static inline void check_pio_pin_mask64(__unused PIO pio, __unused uint sm, __un
  *  \ingroup hardware_pio
  *
  * \if rp2350_specific
- * See \ref sm_config_pins "sm_config_ pins" for more detail on why this method might fail on RP2350B
+ * See \ref sm_config_pins "sm_config_pins" for more detail on why this method might fail on RP2350B
```
lurch: I'd been assuming that this `\ref sm_config_pins "sm_config_ pins"` (with a space inside the quotation marks) was deliberate, because it also occurs in a number of other places in this file? :man_shrugging:

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,34 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
-    uint32_t used = (~config->pinhi >> 4) & PINHI_ALL_PIN_LSBS; // checks if bit 5 of any field is 0
-    // configs that use pins 0-15
-    uint32_t gpio_under_16 = (~config->pinhi) & (~config->pinhi >> 1) & used; // checks if bits 0 and 1 of any field are both 0
-    // configs that use pins 32-47
-    uint32_t gpio_over_32 = (config->pinhi >> 1) & used; // checks if bit 1 of any field is 1
+
+    // boolean (in bit 0 of each field) 1=field_used
+    uint32_t field_used_flags = (~config->pinhi >> 4) & PINHI_ALL_PIN_LSBS;
+    // boolean (in bit 0 of each field) 1=no_used_pins_greather_than_15
+    // note: we assume that if the field is used and no pins are used in the range 16-47, then pins 0-15 must be used
```
lurch: I think this is part of the "definition" of `pinhi`, rather than an "assumption" ?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,34 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
+    invalid_params_if_and_return(PIO, gpio_over_32_flag && !gpio_base, PICO_ERROR_BAD_ALIGNMENT);
+    // flip bit 4 of used (execctrl/pinctrl) values if gpio_base is non-zero (i.e. 16), to turn:
+    // pin & 32 | pin & 0x1f || base | pin_value
+    //     0    +    16-31   -> 16   + 0-15
+    //    32    +    0-15    -> 16   + 16-31
+    //
+    // note, that for gpio_base of zero we have:
+    // pin & 32 | pin & 0x1f || base | pin_value
+    //     0    +    0-15    -> 0    + 0-15
+    //     0    +    16-31   -> 0    + 16-31
```
lurch: For people who don't dream in binary, perhaps this'd be slightly clearer as:
```suggestion
    // flip bit 4 of used (execctrl/pinctrl) values if gpio_base is non-zero (i.e. 16), to turn:
    // pin    pin & 32 | pin & 0x1f || base | pin_value
    // 16-31      0    +    16-31   -> 16   + 0-15
    // 32-47     32    +    0-15    -> 16   + 16-31
    //
    // note, that for gpio_base of zero we have:
    // pin    pin & 32 | pin & 0x1f || base | pin_value
    // 0-15       0    +    0-15    -> 0    + 0-15
    // 16-31      0    +    16-31   -> 0    + 16-31
```

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,34 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
     // 0b00010 - pin is in range 32-47
-    uint32_t used = (~config->pinhi >> 4) & PINHI_ALL_PIN_LSBS; // checks if bit 5 of any field is 0
-    // configs that use pins 0-15
-    uint32_t gpio_under_16 = (~config->pinhi) & (~config->pinhi >> 1) & used; // checks if bits 0 and 1 of any field are both 0
-    // configs that use pins 32-47
-    uint32_t gpio_over_32 = (config->pinhi >> 1) & used; // checks if bit 1 of any field is 1
+
+    // boolean (in bit 0 of each field) 1=field_used
+    uint32_t field_used_flags = (~config->pinhi >> 4) & PINHI_ALL_PIN_LSBS;
+    // boolean (in bit 0 of each field) 1=no_used_pins_greather_than_15
```
lurch: Perhaps
```suggestion
    // boolean (in bit 0 of each field) 1=some_used_pins_less_than_16
```
makes this more obvious? (as well as better matching the comment on `gpio_over_32_flag`)

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,34 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
+    // note: we assume that if the field is used and no pins are used in the range 16-47, then pins 0-15 must be used
+    uint32_t gpio_under_16_flags = (~config->pinhi) & (~config->pinhi >> 1) & field_used_flags;
+    // boolean (in bit 0 of each field) 1=some_used_pins_greather_than_31
+    uint32_t gpio_over_32_flag = (config->pinhi >> 1) & field_used_flags;
```
lurch: I guess this ought to be `gpio_over_31_flags` ? :wink:

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: actually the first is more correct

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: why?

kilograham: oh i see; i thought you just re-ordered the columns

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: i will clarify more

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: i improved

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -283,10 +283,15 @@ static_assert(DREQ_PIO2_RX0 == DREQ_PIO2_TX0 + NUM_PIO_STATE_MACHINES, "");
...
  *   or if the PIO's GPIO base (see \ref pio_set_gpio_base) does not allow access to the required pins.
  *
- *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configre the PIO's
+ *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configure the PIO's
  *   GPIO base before calling the method, however you can use \ref pio_claim_free_sm_and_add_program_for_gpio_range
  *   to find/configure a PIO instance suitable for a particular GPIO range.
  *
+ * \note when `PICO_PIO_USE_GPIO_BASE == 1` \ref pio_sm_set_config ignores fields which haven't had the corresponding
+ * `sm_config_` pin function called, so that you don't have to move settings for unused pin sets into the correct
+ * pin range. Therefore it is always a best practice to explicitly configure a pin range starting at pin zero to 0
```
lurch: Is this suggested tweak correct?
```suggestion
 * pin range. Therefore it is always a best practice to explicitly configure a pin range starting at pin zero to gpio_base 0
```

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,36 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
+    // note, that for gpio_base of zero we have:
+    //  pin  | pin & 32 | pin & 0x1f || base | pin_value
+    //       |          |  (stored)  ||      | (configured)
+    // 0-15  |     0    +    0-15    -> 0    + 0-15
+    // 16-31 |     0    +    16-31   -> 0    + 16-31
+
+    // note we already checked (above) via static_assert( (1u << PINHI_EXECCTRL_LSB) > (PINHI_ALL_PINCTRL_LSBS * 0x1f), "")
+    // that PINHI_EXECTTRL_LSB is above all the pinctrl fields...
+
+    // ... so shift pinctrl bits off top, and flip EXECCTRL_JMP_PIN_MSB if used
```
```suggestion
    // ... so shift pinctrl bits off the bottom, and flip EXECCTRL_JMP_PIN_MSB if used
```
?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,36 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
+    pio->sm[sm].execctrl = config->execctrl ^ (gpio_base ? ((field_used_flags >> PINHI_EXECCTRL_LSB) << PIO_SM0_EXECCTRL_JMP_PIN_MSB) : 0);
+    // ... so shift exectrl bits off the top, and flip pinctrl MSBs if used
```
```suggestion
    // ... so shift execctrl bits off the top, and flip pinctrl MSBs if used
```

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -831,21 +836,36 @@ static inline int pio_sm_set_config(PIO pio, uint sm, const pio_sm_config *confi
...
+    //  pin  | pin & 32 | pin & 0x1f || base | pin_value
+    //       |          |  (stored)  ||      | (configured)
+    // 0-15  |     0    +    0-15    -> 0    + 0-15
+    // 16-31 |     0    +    16-31   -> 0    + 16-31
+
+    // note we already checked (above) via static_assert( (1u << PINHI_EXECCTRL_LSB) > (PINHI_ALL_PINCTRL_LSBS * 0x1f), "")
+    // that PINHI_EXECTTRL_LSB is above all the pinctrl fields...
+
+    // ... so shift pinctrl bits off top, and flip EXECCTRL_JMP_PIN_MSB if used
+    pio->sm[sm].execctrl = config->execctrl ^ (gpio_base ? ((field_used_flags >> PINHI_EXECCTRL_LSB) << PIO_SM0_EXECCTRL_JMP_PIN_MSB) : 0);
```
lurch, line 876: Previously this was flipping `(PIO_SM0_EXECCTRL_JMP_PIN_LSB + 4)` and now it's flipping `PIO_SM0_EXECCTRL_JMP_PIN_MSB`. _Perhaps_ it's worth adding a
```c
static_assert(PIO_SM0_EXECCTRL_JMP_PIN_MSB == (PIO_SM0_EXECCTRL_JMP_PIN_LSB + 4), "");
```
? :shrug:

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: no

lurch: Well I guess I still don't understand what this is trying to say then :confused:

will-v-pi: I don't really understand this either - could you add a snippet of example code showing how to do this, which might make it clearer for the reader?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -283,10 +283,16 @@ static_assert(DREQ_PIO2_RX0 == DREQ_PIO2_TX0 + NUM_PIO_STATE_MACHINES, "");
...
- *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configre the PIO's
+ *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configure the PIO's
  *   GPIO base before calling the method, however you can use \ref pio_claim_free_sm_and_add_program_for_gpio_range
  *   to find/configure a PIO instance suitable for a particular GPIO range.
  *
+ * \note when `PICO_PIO_USE_GPIO_BASE == 1` \ref pio_sm_set_config ignores fields which haven't had the corresponding
+ * `sm_config_` pin function called, so that you don't have to move settings for unused pin sets into the correct
+ * pin range. Therefore, it is always a best practice to explicitly configure a pin range starting at pin zero to 0
+ * via the corresponding sm_config_ function (e.g. `sm_config_set_out_pin_base(config, 0)`), as the default
+ * values for pin ranges from \ref pio_get_default_sm_config are now `GPIO_BASE + 0` not 0 on RP2350B.
```
will-v-pi: I'm afraid I still don't understand what the "to 0" means - I thought `sm_config_set_out_pin_base(config, 0)` configures a pin range starting at pin 0 to `config`, not "to 0"?

lurch: I think it's saying something like "Don't assume that any unconfigured pin_bases will be set to 0 (because that might not be the case on RP2350B)".

But yeah, I guess the "to 0" could be removed?
"Therefore, it is always a best practice to explicitly configure a pin range starting at pin zero via the corresponding sm_config_ function"

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -748,10 +754,10 @@ static inline void sm_config_set_mov_status(pio_sm_config *c, enum pio_mov_statu
...
  * --------|--------
  * Clock Divider | 1
- * Out Pins | 32 starting at 0
- * Set Pins | 0 starting at 0
- * In Pins | 32 starting at 0
- * Side Set Pins (base) | 0
+ * Out Pins | 32 starting at 0 (see note below)
+ * Set Pins | 0 starting at 0 (see note below)
+ * In Pins | 32 starting at 0 (see note below)
+ * Side Set Pins (base) | 0 (see note below)
```
lurch, line 760: Should this "(see note below)" also be added to "Jmp Pin | 0" on line 765?

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -760,6 +766,15 @@ static inline void sm_config_set_mov_status(pio_sm_config *c, enum pio_mov_statu
  * Out Special | sticky=false, has_enable_pin=false, enable_pin_index=0
  * Mov Status | status_sel=STATUS_TX_LESSTHAN, n=0
  *
+ * \note on RP2350B with PICO_PIO_USE_GPIO_BASE = 1, the default out/set/in/side set pin
```
lurch: Probably better to capitalise these for consistency with the previous paragraph?
```suggestion
 * \note on RP2350B with PICO_PIO_USE_GPIO_BASE = 1, the default Out/Set/In/Side Set pin
```

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -760,6 +766,15 @@ static inline void sm_config_set_mov_status(pio_sm_config *c, enum pio_mov_statu
...
+ * bases are actually 0 relative to the GPIO_BASE of the PIO instance the pio_sm_config
+ * is applied to, so that the non-user-specified 0 is not included in logic related to
```
lurch: "the non-user-specified 0" is a bit confusing. After reading this sentence a few times, I think it's saying something like:
```suggestion
 * is applied to, so that any pin bases which aren't explicitly specified are not included in logic related to
```

src/rp2_common/hardware_pio/include/hardware/pio.h
kilograham: yes

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -283,10 +283,16 @@ static_assert(DREQ_PIO2_RX0 == DREQ_PIO2_TX0 + NUM_PIO_STATE_MACHINES, "");
...
  *
- *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configre the PIO's
+ *   To be clear, \ref pio_sm_set_config does not change the PIO's GPIO base for you; you must configure the PIO's
  *   GPIO base before calling the method, however you can use \ref pio_claim_free_sm_and_add_program_for_gpio_range
  *   to find/configure a PIO instance suitable for a particular GPIO range.
  *
+ * \note when `PICO_PIO_USE_GPIO_BASE == 1` \ref pio_sm_set_config ignores fields which haven't had the corresponding
+ * `sm_config_` pin function called, so that you don't have to move settings for unused pin sets into the correct
+ * pin range. Therefore, it is always a best practice to explicitly configure a pin range starting at pin zero to 0
+ * via the corresponding sm_config_ function (e.g. `sm_config_set_out_pin_base(config, 0)`), as the default
```
lurch: I still agree with @will-v-pi that the "pin zero to 0" here seems confusing?
```suggestion
 * pin range. Therefore, it is always a best practice to explicitly configure a pin range starting at pin zero
 * via the corresponding sm_config_ function (e.g. `sm_config_set_out_pin_base(config, 0)`), as the default
```

PR #3000, opened 2026-06-10
Allow response to be specified in makefsdata script

### LINKED ISSUE #2646: makefsdata.py captive portal support
This is an issue that's come out of work for this PR https://github.com/raspberrypi/pico-examples/pull/701

The current access point example generates a 404 response "HTTP/1.1 302 Redirect\nLocation: http://<gw address>/ledtest\n\n" The gw address is in CYW43_DEFAULT_IP_AP_ADDRESS I think.

The sdk includes makefsdata.py (in src/rp2_common/pico_lwip/tools) which is run via the cmake function pico_set_lwip_httpd_content. You can generate a 404 response by adding a 404.html file to your content but crucially you can't modify the http headers, only the html page that's returned. This means you can't fully implement a captive portal (I think).

Add a parameter to makefsdata.py that generates a specific header response, e.g.

--header_response "404=HTTP/1.1 302 Redirect\nLocation: http://192.168.4.1/ledtest\n\n"

Have a look at process_file in makefsdata.py and how it generates the content for a file. In this case we would just generate 404.html with just the header response given. All the '\\','n' sequences will have to be replaced with a real "\n". This will then need to be plumbed into cmake somehow via a new optional argument to the pico_set_lwip_httpd_content function.

Captive portal functionality might work then.

peterharperuk:
Currently the header contents are hardcoded.
This prevents using the makefsdata script to implement a captive portal as we have to send a redirect when a 404 result would be returned.

When loading file contents, look for a file with the same name with a "hdr" and use its contents for the header.

We can then use this for the access point example which can then be much simpler.

Fixes #2646

### DIFF
--- a/src/rp2_common/pico_lwip/tools/makefsdata.py
+++ b/src/rp2_common/pico_lwip/tools/makefsdata.py
@@ -18,8 +18,8 @@ LWIP_HTTPD_SSI_EXTENSIONS = [".shtml", ".shtm", ".ssi", ".xml", ".json"]
 def process_file(input_dir, file):
     results = []
 
-    # Ignore hdr files
-    if file.suffix == ".hdr":
+    # Ignore response files
+    if file.suffix == ".response":
         return None
 
     # Check content type
--- a/src/rp2_common/pico_lwip/tools/makefsdata.py
+++ b/src/rp2_common/pico_lwip/tools/makefsdata.py
@@ -18,6 +18,10 @@ LWIP_HTTPD_SSI_EXTENSIONS = [".shtml", ".shtm", ".ssi", ".xml", ".json"]
 def process_file(input_dir, file):
     results = []
 
+    # Ignore hdr files
+    if file.suffix == ".hdr":
+        return None
+
     # Check content type
     content_type, content_encoding = mimetypes.guess_type(file)
     if content_type is None:
@@ -37,9 +41,21 @@ def process_file(input_dir, file):
         if file.name.startswith(f"{response_id}."):
             response_type = response_id
             break
-    data = f"{response_types[response_type]}\r\n"
-    comment = f"\"{response_types[response_type]}\" ({len(data)} chars)"
+
+    # If we find a file with the same name and a hdr extension - use its contents for the header
+    header_file = file.with_suffix('.hdr')
+    if header_file.is_file():
+        data = header_file.read_text()
+        comment = f"\"{response_types[response_type]}\" ({len(data)} chars) from {header_file.name}"
+    else:
+        data = f"{response_types[response_type]}\r\n"
+        comment = f"\"{response_types[response_type]}\" ({len(data)} chars)"
     results.append({'data': bytes(data, "utf-8"), 'comment': comment});
+
+    # load file contents
+    file_contents = file.read_bytes()
+    if len(file_contents) == 0:
+        return results
 
     # user agent
     data = f"Server: {HTTPD_SERVER_AGENT}\r\n"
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_lwip/tools/makefsdata.py
```diff
@@ -37,10 +41,22 @@ def process_file(input_dir, file):
...
+        comment = f"\"{response_types[response_type]}\" ({len(data)} chars) from {header_file.name}"
+    else:
+        data = f"{response_types[response_type]}\r\n"
+        comment = f"\"{response_types[response_type]}\" ({len(data)} chars)"
     results.append({'data': bytes(data, "utf-8"), 'comment': comment});
 
+    # load file contents
+    file_contents = file.read_bytes()
+    if len(file_contents) == 0:
+        return results
```
lurch, line 57: I _think_ the assumption here is that if you want to set up the redirect as described in #2646 you'd have a `404.hdr` file with the contents:
```
HTTP/1.1 302 Redirect\nLocation: http:///ledtest\n\n
```
and then a 0-byte `404.html` file? (to trigger this early-return)

Would there ever be a use-case for having a 0-byte `.html` file when there _isn't_ a matching `.hdr` filename? (and if there's not, maybe that should raise an error?)

And looking at things from the other direction, would you ever have a custom `.hdr` file where the matching `.html` file _isn't_ a 0-byte file?

peterharperuk: > Would there ever be a use-case for having a 0-byte .html file when there isn't a matching .hdr filename

You can do that at the moment. It would return the standard header. I think perhaps the return should not happen in this case. I might fix that as I've changed the behaviour of the script.

> would you ever have a custom .hdr file where the matching .html file isn't a 0-byte file?

yes I wanted to support this.

src/rp2_common/pico_lwip/tools/makefsdata.py
```diff
@@ -37,10 +41,22 @@ def process_file(input_dir, file):
         if file.name.startswith(f"{response_id}."):
             response_type = response_id
             break
-    data = f"{response_types[response_type]}\r\n"
-    comment = f"\"{response_types[response_type]}\" ({len(data)} chars)"
+
+    # If we find a file with the same name and a hdr extension - use its contents for the header
+    header_file = file.with_suffix('.hdr')
```
peterharperuk: I think this should be changed - we're not actually specifying the header. We're specifying the response?

src/rp2_common/pico_lwip/tools/makefsdata.py
```diff
@@ -18,6 +18,10 @@
 def process_file(input_dir, file):
     results = []
 
+    # Ignore hdr files
+    if file.suffix == ".hdr":
```
lurch: Presumably this should be
```suggestion
    # Ignore response files
    if file.suffix == ".response":
```
?

peterharperuk: doh

src/rp2_common/pico_lwip/tools/makefsdata.py
peterharperuk: Actually, I think it's a bit daft that you have to include a zero size file. I'll see if I can fix this.

peterharperuk: actually - it's very messy. The lwip httpd code looks for the html file extension. I think it's better to leave it as it is.

peterharperuk:
@kilograham There are a couple of example PR's that need this...

https://github.com/raspberrypi/pico-examples/pull/754
https://github.com/raspberrypi/pico-examples/pull/701

kilograham (approved):

Issue #3001, opened 2026-06-11
Pico W BLE examples hang during CYW43 bring-up when cyw43_delay_ms uses async_context_wait_until


aaronduignan:
## Summary

Pico SDK 2.2.0 BLE examples for Pico W appear to hang during initial CYW43 Bluetooth bring-up when the CYW43 HAL delay hooks use `async_context_wait_until()`.

Changing `cyw43_delay_ms()` / `cyw43_delay_us()` to use `busy_wait_ms()` / `busy_wait_us()` allows the BLE example to continue and advertise.

## Environment

- Board: Raspberry Pi Pico W
- SDK: 2.2.0
- SDK commit: `a1438dff1d38bd9c65dbd693f0e5db4b9ae91779`
- `cyw43-driver`: `v1.1.0`
- `btstack`: `v1.6.2-1-g501e6d2b8`
- `PICO_BOARD=pico_w`
- `BTSTACK_EXAMPLE_TYPE=background`

## Reproduced with

- `picow_ble_temp_sensor`
- `picow_bt_example_gatt_counter_background`

## Actual behavior

The examples reach:

`hci_power_control(HCI_POWER_ON);`

but do not continue to the expected BTstack startup output, and the device does not advertise.

When stepping through the code, execution enters:

`hci_transport_cyw43_open()`
`cyw43_bluetooth_hci_init()`
`cyw43_ensure_bt_up()`
`cyw43_ensure_up()`

Instrumentation showed the code stops at the first CYW43 reset delay after `WL_REG_ON` is driven low:

`cyw43_delay_ms(20);`

## Expected behavior

The BLE examples should complete Bluetooth HCI power-on and begin advertising.

## Workaround / fix candidate

In `src/rp2_common/pico_cyw43_driver/cyw43_driver.c`, changing:

`cyw43_delay_ms()` and `cyw43_delay_us()`

from `async_context_wait_until(...)` to non-scheduling busy waits fixes BLE advertising for me:

`busy_wait_ms(ms);`
`busy_wait_us(us);`

peterharperuk:
How come I can't reproduce this? Can you attach a callstack of the hang?

peterharperuk:
Also, test develop as there's a timer fix on there https://github.com/raspberrypi/pico-sdk/pull/2838. Although I can't reproduce any hang on master so I'm not sure what's causing your problem.

aaronduignan:
[...]

PR #3002, opened 2026-06-11
Use non-scheduling CYW43 driver delays

aaronduignan:
Relates to #3001.

This changes the CYW43 HAL delay hooks to use `busy_wait_ms()` / `busy_wait_us()` instead of routing through `async_context_wait_until()`.

On Pico W with SDK 2.2.0, the BLE examples hang during initial CYW43 Bluetooth bring-up at the first `cyw43_delay_ms(20)` in `cyw43_ensure_up()`. With these hooks implemented as non-scheduling busy waits, the BLE GATT counter example completes bring-up and begins advertising.

Tested locally with:

- `PICO_BOARD=pico_w`
- `BTSTACK_EXAMPLE_TYPE=background`
- `picow_bt_example_gatt_counter_background`

### DIFF
--- a/src/rp2_common/pico_cyw43_driver/cyw43_driver.c
+++ b/src/rp2_common/pico_cyw43_driver/cyw43_driver.c
@@ -267,11 +267,15 @@ void cyw43_await_background_or_timeout_us(uint32_t timeout_us) {
 }
 
 void cyw43_delay_ms(uint32_t ms) {
-    async_context_wait_until(cyw43_async_context, make_timeout_time_ms(ms));
+    // The low-level CYW43 driver calls this while holding the async-context lock.
+    // Do not use async_context_wait_until()/sleep_ms() here; background work is
+    // intentionally blocked until the lock is released.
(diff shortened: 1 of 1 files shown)


src/rp2_common/pico_cyw43_driver/cyw43_driver.c
```diff
@@ -267,11 +267,15 @@ void cyw43_await_background_or_timeout_us(uint32_t timeout_us) {
 }
 
 void cyw43_delay_ms(uint32_t ms) {
-    async_context_wait_until(cyw43_async_context, make_timeout_time_ms(ms));
+    // The low-level CYW43 driver calls this while holding the async-context lock.
```
peterharperuk, line 270: This comment doesn't really help. async_context_wait_until doesn't need the context to be unlocked as far as I can see. If you're experiencing a hang it's important to understand the real reason.

aaronduignan: stale comment and you are right too.

PR #3003, opened 2026-06-11
Add a script to check for missing extern_c guards

### LINKED ISSUE #2865: Add a github commit test to check header files for missing extern "C"
I suggest something simple to be going on with...

```
grep -r -L '^extern "C"' . --include '*.h' | grep '/include/pico/'
```

Find header files in "/include/pico/" that don't contain extern "C". This will catch the worst cases?
Just throw an error if the list of offending headers changes?

peterharperuk:
Fixes #2865

example...
`python tools/check_extern_c_guards.py . --match "**/include/pico/**" --ignore-list .extern_c_ignore`

### DIFF
--- a/.extern_c_ignore
+++ b/.extern_c_ignore
@@ -1,0 +1,53 @@
+# .extern_c_ignore
+#
+# Headers that have been reviewed and confirmed NOT to need
+# #ifdef __cplusplus / extern "C" guards.
+#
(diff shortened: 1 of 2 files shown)


kilograham (approved):

kilograham:
I'm gonna believe (then try it out!!)

PR #3004, opened 2026-06-11
Add weak powman_timer_get_lposc_calib_freq function

will-v-pi:
By default this function reads from valid OTP if PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP is set (like before), but it is weak so can be overridden by the user (eg to use frequency_count_khz instead)

The existing behaviour of `powman_timer_set_1khz_tick_source_lposc` is unchanged, other than the addition of ignoring out-of-range OTP values on RP2350.

### DIFF
--- a/src/rp2_common/hardware_powman/include/hardware/powman.h
+++ b/src/rp2_common/hardware_powman/include/hardware/powman.h
@@ -32,25 +32,40 @@ extern "C" {
 #define PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP 1
 #endif
 
+/*! \brief Get the calibrated frequency of the low power oscillator
+ *  \ingroup hardware_powman
+ *
+ * The frequency returned is the value stored in the OTP, or 0 if there is no OTP value
+ * set, the OTP value is out of range, or PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP is 0
+ *
+ * \note This is a weak function, so can be overriden to provide a custom calibration
+ * method (for example using \ref frequency_count_khz)
+ *
+ * \return The LPOSC frequency, or 0 if unknown
+ */
+uint32_t powman_timer_get_lposc_calib_freq(void);
+
 /*! \brief Use the ~32KHz low power oscillator as the powman timer source
  *  \ingroup hardware_powman
- *  \note The frequency is set to the value stored in the OTP, or the reset value if
- *  there is no OTP value set or PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP is 0
+ *
+ * \note The frequency is set to the value returned by \ref powman_timer_get_lposc_calib_freq
  */
 void powman_timer_set_1khz_tick_source_lposc(void);
 
 /*! \brief Use the low power oscillator (specifying frequency) as the powman timer source
  *  \ingroup hardware_powman
- *  \param lposc_freq_hz specify an exact lposc freq to trim it, or 0 to use the reset values
+ *  \param lposc_freq_hz specify an exact lposc freq to trim it, or 0 to use the existing value
  */
 void powman_timer_set_1khz_tick_source_lposc_with_hz(uint32_t lposc_freq_hz);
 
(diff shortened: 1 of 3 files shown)


src/rp2_common/hardware_powman/powman.c
```diff
@@ -64,24 +64,33 @@ uint64_t powman_timer_get_ms(void) {
...
 
-void powman_timer_set_1khz_tick_source_lposc(void) {
+uint32_t __weak powman_timer_get_lposc_calib_freq(void) {
 #if PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP
     uint16_t* lposc_calib_data = (uint16_t*)OTP_DATA_BASE + OTP_DATA_LPOSC_CALIB_ROW;
     if (*lposc_calib_data == 0) {
-        powman_timer_set_1khz_tick_source_lposc_with_hz(0);
+        return 0;
+#if PICO_RP2350
+    } else if (*lposc_calib_data > (40 * 1000) || *lposc_calib_data < (26 * 1000)) {
```
peterharperuk: It might be nice to add #defines for these just in case they need to be changed.

src/rp2_common/hardware_powman/include/hardware/powman.h
```diff
@@ -32,25 +32,40 @@ extern "C" {
...
+ *
+ * The frequency returned is the value stored in the OTP, or 0 if there is no OTP value
+ * set, the OTP value is out of range, or PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP is 0
+ *
+ * \note This is a weak function, so can be overriden to provide a custom calibration
+ * method (for example using \ref frequency_count_khz)
+ *
+ * \return The LPOSC frequency, or 0 if unknown
+ */
+uint32_t powman_timer_get_lposc_calib_freq(void);
```
peterharperuk, line 46: G has commented about using uint rather than uint32_t in API's before - I don't care personally.

src/rp2_common/hardware_powman/include/hardware/powman.h
```diff
@@ -32,25 +32,40 @@ extern "C" {
...
- *  there is no OTP value set or PICO_POWMAN_CALIBRATE_LPOSC_FROM_OTP is 0
+ *
+ * \note The frequency is set to the value returned by \ref powman_timer_get_lposc_calib_freq
  */
 void powman_timer_set_1khz_tick_source_lposc(void);
 
 /*! \brief Use the low power oscillator (specifying frequency) as the powman timer source
  *  \ingroup hardware_powman
- *  \param lposc_freq_hz specify an exact lposc freq to trim it, or 0 to use the reset values
+ *  \param lposc_freq_hz specify an exact lposc freq to trim it, or 0 to use the existing value
```
peterharperuk, line 57: I guess this is ok because the register defaults to POWMAN_LPOSC_FREQ_KHZ_INT_RESET?

peterharperuk (commented):
Some comments / questions

src/rp2_common/hardware_powman/include/hardware/powman.h
will-v-pi: Yes, so the existing value will be the reset value by default

src/rp2_common/hardware_powman/include/hardware/powman.h
will-v-pi: I'll stick with `uint32_t` for now as that is what's passed to `powman_timer_set_1khz_tick_source_lposc_with_hz` below

src/rp2_common/hardware_powman/powman.c
will-v-pi: I've added `LPOSC_MIN_EXPECTED_HZ` and `LPOSC_MAX_EXPECTED_HZ` defines for these in `platform_defs.h`

peterharperuk (approved):

kilograham (approved):

PR #3008, opened 2026-06-12
xTimerDelete in freertos deinit

### LINKED ISSUE #2687: when running under FreeRTOS SMP, the timer task can attempt to free the CYW43 async context's statically allocated timer
hey there.  while running FreeRTOS SMP 11.2.0 on SDK version 2.2.0 with static allocation enabled, I ran into an issue where calling `cyw43_arch_deinit` from a FreeRTOS task would cause FreeRTOS to throw an assertion failure and hard fault:

```
assertion "( ( uint8_t * ) ( pxLink ) >= &( ucHeap[ 0 ] ) ) && ( ( uint8_t * ) ( pxLink ) <= &( ucHeap[ (128*1024) - 1 ] ) )" failed: file "/home/vogon/work/ais-epaper/build/freertos_kernel-src/portable/MemMang/heap_4.c", line 368, function: vPortFree
```

I think the series of events which precipitates this assertion failing are:

1) `async_context_freertos_deinit` calls `xTimerDelete` on the async context's associated timer, which enqueues an asynchronous timer delete command for the timer task.
2) `async_context_freertos_deinit` continues executing past the `memset` at the end of the function.
3) the timer task is scheduled and receives the command from `async_context_freertos_deinit` to delete the timer.
4) `prvProcessReceivedCommands` checks `pxTimer->ucStatus`, which has been cleared to 0, and decides the timer was dynamically allocated.
5) `vPortFree` notices that the statically-allocated timer isn't in a heap block and panics.

I've verified that commenting out the memset at the bottom of `async_context_freertos_deinit` (which seems to be redundant with a memset at the top of `async_context_freertos_init`) prevents the crash from happening.

I've uploaded a gist of the app I'm running (with all the commented-out code removed but otherwise not cleaned up), as well as my FreeRTOSConfig.h and CMakeLists.txt: https://gist.github.com/vogon/4c08fdd31a565609d02cb2534a4386bd

peterharperuk:
Fixes #2687

### DIFF
--- a/src/rp2_common/pico_async_context/async_context_freertos.c
+++ b/src/rp2_common/pico_async_context/async_context_freertos.c
@@ -190,7 +190,12 @@ void async_context_freertos_deinit(async_context_t *self_base) {
     if (self->task_complete_sem) {
         vSemaphoreDelete(self->task_complete_sem);
     }
+#if configSUPPORT_STATIC_ALLOCATION
+    // the timer task might still be running on the other core, so don't clear static memory
+    memset(self, 0, offsetof(async_context_freertos_t, lock_mutex_buf));
+#else
     memset(self, 0, sizeof(*self));
+#endif
 }
 
 void async_context_freertos_acquire_lock_blocking(async_context_t *self_base) {
--- a/src/rp2_common/pico_async_context/async_context_freertos.c
+++ b/src/rp2_common/pico_async_context/async_context_freertos.c
@@ -170,6 +170,11 @@ static uint32_t end_task_func(void *param) {
     return 0;
 }
 
+static void timer_delete_sync_helper(__unused void *param1, __unused uint32_t param2) {
+    TaskHandle_t xTaskToNotify = (TaskHandle_t)param1;
+    xTaskNotifyGive(xTaskToNotify);
+}
+
 void async_context_freertos_deinit(async_context_t *self_base) {
     async_context_freertos_t *self = (async_context_freertos_t *)self_base;
     if (self->task_handle) {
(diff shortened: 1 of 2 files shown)


src/rp2_common/pico_async_context/async_context_freertos.c
```diff
@@ -179,7 +179,12 @@ void async_context_freertos_deinit(async_context_t *self_base) {
         }
     }
     if (self->timer_handle) {
-        xTimerDelete(self->timer_handle, 0);
+        if (xTimerDelete(self->timer_handle, 0)) {
+            int max_iterations = 10;
+            while(xTimerIsTimerActive(self->timer_handle) && max_iterations-- > 0) {
```
kilograham: Not convinced that waiting 10 loops, is any better than doing nothing, if you really care maybe do the delete in an `xTimerPendFunctionCall`?

src/rp2_common/pico_async_context/async_context_freertos.c
```diff
@@ -190,7 +195,6 @@ void async_context_freertos_deinit(async_context_t *self_base) {
     if (self->task_complete_sem) {
         vSemaphoreDelete(self->task_complete_sem);
     }
```
kilograham, line 206: I mean it wasn't totally unnecessary; it made calling deinit multiple times ok (whether this is useful or not, i don't know)

src/rp2_common/pico_async_context/async_context_freertos.c
peterharperuk: No - seems a bit naff. I don't really understand why it isn't executed straight away.
SMP I guess.

src/rp2_common/pico_async_context/async_context_freertos.c
peterharperuk: true

peterharperuk: I've had a second attempt - just not clearing static memory.

kilograham:
i updated the code, but not sure what you are using to test @peterharperuk `

peterharperuk:
The freertos example in pico-examples. It exits after the counter gets to 10
later: Actually you have to add something like sleep_ms(1000) after the call to async_context_deinit to reproduce the crash.

peterharperuk:
This fixes the problem and looks good.

will-v-pi (approved):

PR #3010, opened 2026-06-16
Remove duplicate wrapper linker commands in pico_mem_ops

DeepPG:
Fix duplicated wrapper functions definitions

### DIFF
--- a/src/rp2_common/pico_bit_ops/BUILD.bazel
+++ b/src/rp2_common/pico_bit_ops/BUILD.bazel
@@ -7,8 +7,6 @@ cc_library(
     srcs = ["bit_ops_aeabi.S"],
     linkopts = select({
         "//bazel/constraint:rp2040": [
-            "-Wl,--wrap=__clzsi2",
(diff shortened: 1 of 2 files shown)


lurch:
Hmmm, looks like the same duplication is being done in https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_bit_ops/CMakeLists.txt too?

DeepPG:
Thank you for your observation, I fixed that too and sqaushed them, I hope that's ok.

lurch:
I'm kinda poking in the dark here, but looking at https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html and https://software-dl.ti.com/codegen/docs/tiarmclang/rel5_1_1_LTS/compiler_manual/c_cpp_language_implementation/intrinsics/index.html I wonder if the "correct fix" might have been to leave one of the `__clzsi2` entries in the `# gcc` section, and remove the one in the `# armclang` section?

kilograham (approved):

PR #3011, opened 2026-06-16
Remove unecessary bits from pad_checksum on RP2350

### LINKED ISSUE #2334: RP2350: pad_checksum appears to set incorrect cpu

will-v-pi:
Remove unecessary bits from pad_checksum (now just called pad), and fixup references and comments

Should now describe what actually happens with boot2 on RP2350

Also add a kitchen sink variant with PICO_EMBED_XIP_SETUP=1 to check it still works

Fixes #2334

### DIFF
--- a/src/rp2350/boot_stage2/pad
+++ b/src/rp2350/boot_stage2/pad
@@ -11,18 +11,11 @@ def any_int(x):
         raise argparse.ArgumentTypeError("expected an integer, not '{!r}'".format(x))
 
 
-def bitrev(x, width):
-    return int("{:0{w}b}".format(x, w=width)[::-1], 2)
-
-
 parser = argparse.ArgumentParser()
 parser.add_argument("ifile", help="Input file (binary)")
 parser.add_argument("ofile", help="Output file (assembly)")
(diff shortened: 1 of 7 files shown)


src/rp2350/boot_stage2/pad
```diff
@@ -11,18 +11,11 @@ def any_int(x):
...
-
 parser = argparse.ArgumentParser()
 parser.add_argument("ifile", help="Input file (binary)")
 parser.add_argument("ofile", help="Output file (assembly)")
-parser.add_argument("-p", "--pad", help="Padded size (bytes), including 4-byte checksum, default 256",
+parser.add_argument("-p", "--pad", help="Padded size (bytes), default 256",
                     type=any_int, default=256)
-parser.add_argument("-s", "--seed", help="Checksum seed value, default 0",
-                    type=any_int, default=0)
-parser.add_argument("-a", "--arch", default="arm", choices=["arm", "riscv"])
```
lurch, line 25: Why has this flag been removed? (If it's redundant, do you know why it was there initially?)

kilograham (approved):

PR #3012, opened 2026-06-16
Fixups for packaged no_flash and xip_ram binaries

will-v-pi:
This PR includes some fixups to packaged no_flash and xip_ram binaries:

- `sections_psram.incl` and `section_xip_ram.incl` are moved before `sections_default_data.incl`, in case `persistent_data` is placed in xip_ram - this is because initialised data must come before uninitialised. Without this change, binaries using xip_ram for persistent data and code will overwrite the persistent data with 0s every boot
- Add some more tests of packaging, including hashing and signing
- Fix `_entry_point` for xip_ram binaries, as the RAM image boot will fail if the end of the window is not in valid memory

It goes with the picotool PR https://github.com/raspberrypi/picotool/pull/325 which squashes the ELF files, to package them better


src/rp2_common/pico_standard_link/script_include/section_end.incl
```diff
@@ -17,6 +17,9 @@ SECTIONS
     __StackBottom = __StackTop - SIZEOF(.stack_dummy);
     PROVIDE(__stack = __StackTop);
 
+    /* provide the maximum ram window size for crt0 */
+    __ram_window_size = LENGTH(RAM);
```
kilograham: seems like this could just be called __ram_size, and we should make it clear in the assembly in crt0.S why we are using _logical_binary_start and __ram_size as the two values - perhaps its worth adding a __ram_start, or perhaps we want to use _logical_binary_end (or __logical_binary_end_4k which is rounded up) obv i guess you need space for the added block loops, hence I guess why u are using the end

will-v-pi: I've re-done this with a new `section_no_flash_end.incl` file that sets `__ram_window_start` and `__ram_window_size` to a window starting at `__logical_binary_start` and ending at the end of RAM, plus SCRATCH_X/Y if they are contiguous with RAM. This should give the maximum size that the block loop could fit into, and still work for binaries that use custom RAM and SCRATCH_X/Y origins and sizes.

`__logical_binary_start` is used instead of `ORIGIN(RAM)` in case you have override `section_no_flash_text.incl`, because there is an assertion that `__embedded_block_end` is within 4k of `__logical_binary_start` in `section_platform_end.incl`.

kilograham: ok cool; yeah i mean i guess i was just checking that we know that __logical_binary_start (and thus now __ram_window_start) is actually before, and in the same memory region as __ram_window_end, but i guess it is in all cases, and this would break (I hope) if it wasnt

will-v-pi: Yes, it should be in all cases that don't overwrite `section_no_flash_text.incl`, and the RAM boot will fail if they do end up in different regions

kilograham (approved):

PR #3013, opened 2026-06-17
fix -fanalyzer warnings (from Arm GCC 14.2/15 at least)

kilograham:
* note no plans to fix false positives from earlier versions

### DIFF
--- a/src/rp2_common/pico_rand/rand.c
+++ b/src/rp2_common/pico_rand/rand.c
@@ -224,7 +224,8 @@ static uint64_t capture_additional_rosc_samples(uint n) {
                 bit_done = true;
                 if (i == n - 1) {
                     // samples has our random bits, so let's mix them in now
-                    samples = rosc_samples = (rosc_samples << n) | samples;
+                    if (n < 64) samples |= rosc_samples << n;
+                    rosc_samples = samples;
                 }
             }
             spin_unlock(lock, save);
--- a/test/kitchen_sink/kitchen_sink.c
+++ b/test/kitchen_sink/kitchen_sink.c
@@ -194,13 +194,13 @@ int main(void) {
 
     extern uint32_t __load_start_overlay_second;
     extern uint32_t __load_stop_overlay_second;
-    stored_words = (uint32_t)(&__load_stop_overlay_second - &__load_start_overlay_second);
+    stored_words = ((uintptr_t)&__load_stop_overlay_second - (uintptr_t)&__load_start_overlay_second) / sizeof(uint32_t);
     memcpy(&__overlays_start__, &__load_start_overlay_second, 4 * stored_words);
     printf("overlay_second after load = %d, %d\n", overlay_second_one, overlay_second_two);
 
     extern uint32_t __load_start_overlay_first;
     extern uint32_t __load_stop_overlay_first;
-    stored_words = (uint32_t)(&__load_stop_overlay_first - &__load_start_overlay_first);
+    stored_words = ((uintptr_t)&__load_stop_overlay_first - (uintptr_t)&__load_start_overlay_first) / sizeof(uint32_t);
     memcpy(&__overlays_start__, &__load_start_overlay_first, 4 * stored_words);
     printf("overlay_first after load = %d\n", overlay_first);
     printf("overlay_second after overlay_first load = %d, %d\n", overlay_second_one, overlay_second_two);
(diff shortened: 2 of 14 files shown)


will-v-pi (commented):
I'm getting `-fanalyzer` warnings out of `sha256.c` on RP2350 with Arm GCC 15.2.Rel1 and Risc-V GCC 15.1:
```
src/rp2_common/pico_sha256/sha256.c:99:17: warning: stack-based buffer over-read [CWE-126] [-Wanalyzer-out-of-bounds]
```

kilograham:
I guess I should have said Arm... (the RISC-V does have that error, which is bogus)

`arm-none-eabi-gcc (Arm GNU Toolchain 15.2.Rel1 (Build arm-15.86)) 15.2.1 20251203`
 works fine for me; where did u add the -fanalyzer?

kilograham:
also - the only warning on the pico-gcc for RISC-V is in segger_rtt

src/rp2_common/pico_rand/rand.c
```diff
@@ -224,7 +224,8 @@ static uint64_t capture_additional_rosc_samples(uint n) {
                 bit_done = true;
                 if (i == n - 1) {
                     // samples has our random bits, so let's mix them in now
-                    samples = rosc_samples = (rosc_samples << n) | samples;
+                    if (n < 64) samples |= rosc_samples << n;
```
lurch, line 227: @andygpz11 Should this `capture_additional_rosc_samples` function also assert that `n <= 64` ?
(It gets called as `capture_additional_rosc_samples(PICO_RAND_ROSC_BIT_SAMPLE_COUNT)` and `PICO_CONFIG: PICO_RAND_ROSC_BIT_SAMPLE_COUNT` declares `max=64`)

test/kitchen_sink/kitchen_sink.c
```diff
@@ -194,13 +194,13 @@ int main(void) {
 
     extern uint32_t __load_start_overlay_second;
     extern uint32_t __load_stop_overlay_second;
-    stored_words = (uint32_t)(&__load_stop_overlay_second - &__load_start_overlay_second);
+    stored_words = ((uintptr_t)&__load_stop_overlay_second - (uintptr_t)&__load_start_overlay_second) / sizeof(uint32_t);
```
lurch, line 197: There's a few places in this PR where this pattern appears. I guess adding a common macro like:
```c
#define WORDS_BETWEEN(start_symbol, end_symbol) ((uintptr_t)&(end_symbol) - (uintptr_t)&(start_symbol)) / sizeof(uint32_t)
```
probably isn't worth the include-hassle?

src/rp2_common/pico_rand/rand.c
```diff
@@ -224,7 +224,8 @@ static uint64_t capture_additional_rosc_samples(uint n) {
                 bit_done = true;
                 if (i == n - 1) {
```
lurch, line 225: Hmmm... given that this happens at the end of the loop-body, and this `if (i == n - 1)` presumably means that this code-block only executes at the (end of) the last loop iteration, is there any reason that this couldn't just be moved to after the closing-brace of the `for` loop, and thus saving an `if`-check on each loop iteration?

will-v-pi:
> I guess I should have said Arm... (the RISC-V does have that error, which is bogus)
> 
> `arm-none-eabi-gcc (Arm GNU Toolchain 15.2.Rel1 (Build arm-15.86)) 15.2.1 20251203` works fine for me; where did u add the -fanalyzer?

I get the warning when adding `-fanalyzer` to `kitchen_sink_options` and building `kitchen_sink` (which throws an error because `-Werror` is set on `kitchen_sink`), and also when adding it to `pico_sha256_test`. My compiler is the same:

`arm-none-eabi-gcc (Arm GNU Toolchain 15.2.Rel1 (Build arm-15.86)) 15.2.1 20251203`

and my Risc-V compiler is from https://github.com/raspberrypi/pico-sdk-tools/releases/tag/v2.2.0-3 which has the same warning as Arm for me

EDIT: And it's a Release build

kilograham:
ah i was doing a debug build - maybe thats it

test/kitchen_sink/kitchen_sink.c
kilograham: nope; i don't care ;-)

src/rp2_common/pico_rand/rand.c
kilograham: its done under the lock

will-v-pi (approved):

src/rp2_common/pico_rand/rand.c
lurch: Ahhh. I guess re-acquiring the lock _after_ the for-loop, might be more expensive than doing this `if (i == n - 1)` during every loop-iteration.

PR #3014, opened 2026-06-17
remove nosys.specs from main build as pico_libc_interface provides syscalls

kilograham:


will-v-pi (approved):

projectgus:
FWIW I have always understood the reason for `nosys.specs` is to see a linker warning if you *haven't* implemented necessary syscalls yourself. If you don't pass this then it'll make no difference if custom syscalls are linked as intended, but it's now possible for a build to quietly link the newlib default syscall implementations instead - which might pull in a bunch of unintended dependencies or runtime behaviour.

(Maybe this doesn't matter, and also I'm not 100% sure my understanding is correct, but I thought I'd mention it.)

PR #3015, opened 2026-06-17
Add BLOCK_DEVICE_PARTITION_ID define in bootrom_constants.h

will-v-pi:
This is not used by the SDK yet, but will be used by `picotool bdev` in https://github.com/raspberrypi/picotool/pull/191

### DIFF
--- a/src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
+++ b/src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
@@ -133,6 +133,12 @@
 #define BOOT_TBYB_AND_UPDATE_FLAG_OTP_VERSION_APPLIED              0x2
 #define BOOT_TBYB_AND_UPDATE_FLAG_OTHER_ERASED                     0x4
 
+// PICO_CONFIG: BLOCK_DEVICE_PARTITION_ID, ID of block device partition which must match the ID used in the partition table JSON, type=int, default=0x626C6F636B646576, group=pico_bootrom
+#ifndef BLOCK_DEVICE_PARTITION_ID
+// The default 0x626C6F636B646576 value is the ASCII encoding of "blockdev"
+#define BLOCK_DEVICE_PARTITION_ID 0x626C6F636B646576
+#endif
+
 #ifndef __ASSEMBLER__
 // Limited to 3 arguments in case of varm multiplex hint (trashes Arm r3)
 typedef int (*bootrom_api_callback_generic_t)(uint32_t r0, uint32_t r1, uint32_t r2);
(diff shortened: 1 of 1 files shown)


src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
```diff
@@ -133,6 +133,12 @@
 #define BOOT_TBYB_AND_UPDATE_FLAG_OTP_VERSION_APPLIED              0x2
 #define BOOT_TBYB_AND_UPDATE_FLAG_OTHER_ERASED                     0x4
 
+// PICO_CONFIG: BLOCK_DEVICE_PARTITION_ID, ID of block device partition which must match the ID used in the partition table JSON, type=int, default=0x626C6F636B646576, group=pico_bootrom
```
lurch: `min=0, max=0xFFFFFFFFFFFFFFFF` ? :laughing:

kilograham:
as per the picotool comments; I don't think there is a single BLOCK_DEVICE_PARTITION_ID as you might want to use different IDs to mark different partitions - I'm fine with having a default id though

will-v-pi:
> as per the picotool comments; I don't think there is a single BLOCK_DEVICE_PARTITION_ID as you might want to use different IDs to mark different partitions - I'm fine with having a default id though

Yeah, I’ve added to support for specifying other IDs to picotool, but I think a default ID is nice for the basic use case

lurch:
> > as per the picotool comments; I don't think there is a single BLOCK_DEVICE_PARTITION_ID as you might want to use different IDs to mark different partitions - I'm fine with having a default id though
> 
> Yeah, I’ve added to support for specifying other IDs to picotool, but I think a default ID is nice for the basic use case

Perhaps that would be made clearer by renaming it to `BLOCK_DEVICE_PARTITION_DEFAULT_ID` or `BLOCK_DEVICE_DEFAULT_PARTITION_ID` ?

src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
```diff
@@ -133,6 +133,12 @@
 #define BOOT_TBYB_AND_UPDATE_FLAG_OTP_VERSION_APPLIED              0x2
 #define BOOT_TBYB_AND_UPDATE_FLAG_OTHER_ERASED                     0x4
 
+// PICO_CONFIG: BLOCK_DEVICE_DEFAULT_PARTITION_ID, default ID of block device partition which must match the ID used in the partition table JSON, type=int, default=0x626C6F636B646576, group=pico_bootrom
```
lurch, line 136: I have to confess that I've not fully grokked this, but if this is now a "default", should the "must match" be changed to "should match"?

kilograham (approved):

PR #3016, opened 2026-06-17
add .gitattributes for LF line endings, and fixup content

kilograham:


lurch (approved):
LGTM :+1: 
I ran a dirty shell one-liner, and that produced the same diff as you have here :slightly_smiling_face: 
`find . -type f -exec file {} \; | grep -E "(ASCII|Unicode) text" | cut -d: -f1 | xargs dos2unix`

will-v-pi (approved):
LGTM, although it’s annoying that GitHub now requires this for suggestions to work

PR #3017, opened 2026-06-17
Fix btstack flash bank in partitions

will-v-pi:
The code was writing to the untranslated addresses with `flash_range_program`, but reading from the translated `XIP_BASE`. This should fix that issue by reading from the untranslated alias.

A partition table that will reproduce this issue is [pt.uf2](https://github.com/user-attachments/files/29059300/pt.zip), generated from this JSON:
```json
{
    "$schema": "https://raw.githubusercontent.com/raspberrypi/picotool/master/json/schemas/partition-table-schema.json",
    "version": [1, 0],
    "unpartitioned": {
        "families": ["absolute"],
        "permissions": {
            "secure": "rw",
            "nonsecure": "rw",
            "bootloader": "rw"
        }
    },
    "partitions": [
        {
            "name": "A",
            "id": 0,
            "size": "2036K",
            "families": ["rp2350-arm-s", "rp2350-riscv"],
            "permissions": {
                "secure": "rw",
                "nonsecure": "rw",
                "bootloader": "rw"
            }
        },
        {
            "name": "B",
            "id": 1,
            "size": "2036K",
            "families": ["rp2350-arm-s", "rp2350-riscv"],
            "permissions": {
                "secure": "rw",
                "nonsecure": "rw",
                "bootloader": "rw"
            },
            "link": ["a", 0]
        }
    ]
}
```
The 2036K sizes are to ensure there is room for the flash bank after the partitions.

### DIFF
--- a/src/rp2_common/pico_btstack/btstack_flash_bank.c
+++ b/src/rp2_common/pico_btstack/btstack_flash_bank.c
@@ -90,7 +90,7 @@ static void pico_flash_bank_read(void *context, int bank, uint32_t offset, uint8
     if ((offset + size) > PICO_FLASH_BANK_SIZE) return;
 
     // Flash is xip
-    memcpy(buffer, (void *)(XIP_BASE + pico_flash_bank_get_storage_offset_func() + (PICO_FLASH_BANK_SIZE * bank) + offset), size);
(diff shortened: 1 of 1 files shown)


peterharperuk (approved):
Yes - it didn't get as far as loading firmware with that partition table. After your fix bluetooth works and it seems to remember its bonding details

PR #3019, opened 2026-06-18
Add PICO_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED (defaults to 1) to allow mutlicore_lockout to work before core 1 is started

### LINKED ISSUE #2639: rom_flash_op unusable when pico_multicore is included but core1 hasn't been launched yet
I've got a codebase that does some flash operations immediately on startup, but potentially launches a thread on the second core much much later.

While I can put `flash_safe_execute_core_init` into the thread functions no problem, while the second core hasn't been launched and that call hasn't been reached, `flash_safe_execute` invariably fails with PICO_ERROR_NOT_PERMITTED.

I can't use `PICO_FLASH_ASSUME_CORE1_SAFE`, because when the core is running, it is not safe, so it needs the lockout code to run.

I can write a custom `get_flash_safety_helper` that can assume core 1 is safe when I know it's not running, but I want to use the default handlers when core 1 *is* running, and they're all defined as static in pico-sdk so I have no access to the symbols. I can't even call the original `get_flash_safety_helper` to get the references since it's weak and by defining my own I've shadowed it.

The only real option I have left is to duplicate the implementations out of pico-sdk manually.

kilograham:
fixes #2639

### DIFF
--- a/src/rp2_common/pico_flash/flash.c
+++ b/src/rp2_common/pico_flash/flash.c
@@ -172,8 +172,8 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         // todo we may get preempted here, but I think that is OK unless what is pre-empts requires
         //      the other core to be running.
 #elif PICO_FLASH_SAFE_EXECUTE_SUPPORT_MULTICORE_LOCKOUT
-        // we cannot mix multicore_lockout and FreeRTOS RP2040 as they both use the multicore FIFO...
-        // the user, will have to roll their own mechanism in this case. FreeRTOS on RP2350 uses doorbells
+        // we cannot mix multicore_lockout and FreeRTOS on RP2040 as they both use the multicore FIFO...
+        // the user will have to roll their own mechanism in this case. FreeRTOS on RP2350 uses doorbells
 #if LIB_FREERTOS_KERNEL && PICO_RP2040
 #if PICO_FLASH_ASSERT_ON_UNSAFE
         assert(false); // we expect the other core to have been initialized via flash_safe_execute_core_init()
--- a/src/rp2_common/pico_flash/flash.c
+++ b/src/rp2_common/pico_flash/flash.c
@@ -6,7 +6,7 @@
 
 #include "pico/flash.h"
 #include "hardware/sync.h"
-#if PICO_FLASH_SAFE_EXECUTE_PICO_SUPPORT_MULTICORE_LOCKOUT
+#if PICO_FLASH_SAFE_EXECUTE_SUPPORT_MULTICORE_LOCKOUT
 #include "pico/multicore.h"
 #endif
 #include "pico/time.h"
@@ -30,7 +30,7 @@
 // 2. Regular pico_multicore - we need to use multicore lockout.
 // 3. FreeRTOS on core 0, no use of core 1 - we just want to disable IRQs
 // 4. FreeRTOS SMP on both cores - we need to schedule a high priority task on the other core to disable IRQs.
-// 5. FreeRTOS on one core, but application is using the other core. ** WE CANNOT SUPPORT THIS TODAY ** without
+// 5. FreeRTOS on one core, but application is using the other core. ** WE CANNOT SUPPORT THIS TODAY on RP2040 ** without
 //    the equivalent PICO_FLASH_ASSUME_COREx_SAFE (i.e. the user making sure the other core is fine)
 
 static bool default_core_init_deinit(bool init);
@@ -93,7 +93,7 @@ static bool default_core_init_deinit(__unused bool init) {
 #if PICO_FLASH_SAFE_EXECUTE_USE_FREERTOS_SMP
     return true;
 #endif
-#if PICO_FLASH_SAFE_EXECUTE_PICO_SUPPORT_MULTICORE_LOCKOUT
+#if PICO_FLASH_SAFE_EXECUTE_SUPPORT_MULTICORE_LOCKOUT
     if (!init) {
         return false;
     }
@@ -171,10 +171,10 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         }
         // todo we may get preempted here, but I think that is OK unless what is pre-empts requires
         //      the other core to be running.
-#elif PICO_FLASH_SAFE_EXECUTE_PICO_SUPPORT_MULTICORE_LOCKOUT
-        // we cannot mix multicore_lockout and FreeRTOS as they both use the multicore FIFO...
-        // the user, will have to roll their own mechanism in this case.
-#if LIB_FREERTOS_KERNEL
+#elif PICO_FLASH_SAFE_EXECUTE_SUPPORT_MULTICORE_LOCKOUT
+        // we cannot mix multicore_lockout and FreeRTOS RP2040 as they both use the multicore FIFO...
+        // the user, will have to roll their own mechanism in this case. FreeRTOS on RP2350 uses doorbells
+#if LIB_FREERTOS_KERNEL && PICO_RP2040
 #if PICO_FLASH_ASSERT_ON_UNSAFE
         assert(false); // we expect the other core to have been initialized via flash_safe_execute_core_init()
                        // unless PICO_FLASH_ASSUME_COREX_SAFE is set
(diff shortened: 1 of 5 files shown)


kilograham:
it slightly irks me that this adds code to multicore_launch_core1 even if multicore_lockout functions aren't called (this was true before actually, its just now two places)

kilograham:
oops, converting to draft, as I need to add a "multicore_lockout_can_be_called()" or whatever

src/rp2_common/pico_multicore/multicore.c
```diff
@@ -28,7 +28,22 @@
 // and is a no-op if set. We DO have a new `multicore_lockout_victim_deinit()` method, which can be called in a pinch after
 // the reset before calling `multicore_lockout_victim_init()` again, so that is good. We will reset the flag
 // for core1 in `multicore_reset_core1()` though as a convenience since most people will use that to reset core 1.
-static bool lockout_victim_initialized[NUM_CORES];
+
+#undef PICO_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED
```
peterharperuk: I'm confused why this is being set to zero here? And 1 above?

peterharperuk:
It still asserts in default_enter_safe_zone_timeout_ms. configNUMBER_OF_CORES=1, running freertos on core 0 which calls flash_safe_execute_core_init. Core 1 not started. Is that right?

I've been using this example to reproduce the issue https://github.com/raspberrypi/pico-examples/pull/762

src/rp2_common/pico_multicore/multicore.c
kilograham: ah; i was testing!

kilograham:
> It still asserts in default_enter_safe_zone_timeout_ms. configNUMBER_OF_CORES=1, running freertos on core 0 which calls flash_safe_execute_core_init. Core 1 not started. Is that right?
> 
> I've been using this example to reproduce the issue [raspberrypi/pico-examples#762](https://github.com/raspberrypi/pico-examples/pull/762)

yeah - as per draft comment, it is missing a bit to make it work; i'll try the example... i'm going back and forth as to whether i want this, or to force people to call a method to setup core 1!

src/rp2_common/pico_multicore/multicore.c
```diff
@@ -345,7 +368,17 @@ void multicore_lockout_end_blocking(void) {
 }
 
 bool multicore_lockout_victim_is_initialized(uint core_num) {
-    return lockout_victim_initialized[core_num];
+    return core_status[core_num] == CORE_STATUS_LOCKOUT_ENABLED;
+}
+
+bool multicore_lockout_ready() {
```
will-v-pi:
```suggestion
bool multicore_lockout_ready(void) {
```

src/rp2_common/pico_runtime/CMakeLists.txt
will-v-pi, line 71: I pushed a commit to fix this, as both `CRT0_FAR_CALLS` and `MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED` weren't implemented correctly, so were always excluded

src/rp2_common/pico_multicore/multicore.c
kilograham: hmm i thought i fixed that!

src/rp2_common/pico_runtime/CMakeLists.txt
kilograham: thx

will-v-pi (dismissed):
LGTM - tested it works with the `flash_progam` example. If you link `pico_multicore`, without this PR it panics, and with this PR it now works fine.

src/rp2_common/pico_flash/flash.c
```diff
@@ -171,17 +171,17 @@ static int default_enter_safe_zone_timeout_ms(__unused uint32_t timeout_ms) {
         }
         // todo we may get preempted here, but I think that is OK unless what is pre-empts requires
         //      the other core to be running.
-#elif PICO_FLASH_SAFE_EXECUTE_PICO_SUPPORT_MULTICORE_LOCKOUT
-        // we cannot mix multicore_lockout and FreeRTOS as they both use the multicore FIFO...
-        // the user, will have to roll their own mechanism in this case.
-#if LIB_FREERTOS_KERNEL
+#elif PICO_FLASH_SAFE_EXECUTE_SUPPORT_MULTICORE_LOCKOUT
+        // we cannot mix multicore_lockout and FreeRTOS RP2040 as they both use the multicore FIFO...
+        // the user, will have to roll their own mechanism in this case. FreeRTOS on RP2350 uses doorbells
```
lurch:
```suggestion
        // we cannot mix multicore_lockout and FreeRTOS on RP2040 as they both use the multicore FIFO...
        // the user will have to roll their own mechanism in this case. FreeRTOS on RP2350 uses doorbells
```

src/rp2_common/pico_multicore/include/pico/multicore.h
```diff
@@ -38,6 +38,11 @@ extern "C" {
 #endif
 #endif
 
+// PICO_CONFIG: PICO_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED, Allow multicore_lockout functions called from core 0 to succeed if core1 has not been started, default=1, group=pico_multicore
```
lurch:
```suggestion
// PICO_CONFIG: PICO_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED, Allow multicore_lockout functions called from core 0 to succeed if core1 has not been started, type=bool, default=1, group=pico_multicore
```
?

src/rp2_common/pico_multicore/include/pico/multicore.h
```diff
@@ -38,6 +38,11 @@ extern "C" {
...
+#ifndef PICO_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED
```
lurch, line 42: I guess this _could_ be `PICO_SUPPORT_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED` but perhaps that's a bit long :wink:

src/rp2_common/pico_multicore/include/pico/multicore.h
```diff
@@ -440,6 +445,12 @@ static inline uint multicore_doorbell_irq_num(uint doorbell_num) {
  * \ref multicore_lockout_end_timeout_us to release the lockout.
  *
  * \note Because multicore lockout uses the intercore FIFOs, the FIFOs <b>cannot</b> be used for any other purpose
+ *
+ * \note By default, for convenience, multicore_lockout_start_ functions will succeed on core 0, if core 1 has either not been started
+ * via multicore_launch_core1 functions, or has subsequently been reset via multicore_reset_core1. Therefore, it is not safe to
+ * (though equally not very likely that you would) call multicore_launch1 while core 0 is inside of a multicore_lockout_ function.
```
lurch:
```suggestion
 * (though equally not very likely that you would) call multicore_launch_core1 while core 0 is inside of a multicore_lockout_ function.
```

will-v-pi (dismissed):

src/rp2_common/pico_multicore/multicore.c
```diff
@@ -28,7 +28,19 @@
 // and is a no-op if set. We DO have a new `multicore_lockout_victim_deinit()` method, which can be called in a pinch after
 // the reset before calling `multicore_lockout_victim_init()` again, so that is good. We will reset the flag
 // for core1 in `multicore_reset_core1()` though as a convenience since most people will use that to reset core 1.
-static bool lockout_victim_initialized[NUM_CORES];
+
+#define CORE_STATUS_NOT_RUNNING         ((uint8_t)0)
+// not needed when PICO_MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED==0
```
lurch: Not clear which line this comment is referring to? :confused:

src/rp2_common/pico_runtime/CMakeLists.txt
```diff
@@ -63,6 +63,8 @@ endforeach()
 # PANIC                 - default panic impl which brings in stdio;
 # AUTO_INIT_MUTEX       - auto init mutexes, without this you get no printf mutex either;
 # CRT0_FAR_CALLS        - use blx not bl for calls from crt0 to user overridable functions;
+# THREAD_LOCAL          - support for thread locals;
+# MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED - allow multicore_lockout_start to be called on core 0 before core 1 is startes
```
lurch:
```suggestion
# MULTICORE_LOCKOUT_BEFORE_CORE1_STARTED - allow multicore_lockout_start to be called on core 0 before core 1 is started
```

will-v-pi (approved):

PR #3021, opened 2026-06-20
minor cleanup

kilograham:
common stuff

### DIFF
--- a/src/rp2_common/pico_double/include/pico/double.h
+++ b/src/rp2_common/pico_double/include/pico/double.h
@@ -238,7 +238,7 @@ extern "C" {
     int32_t double2int_z(double d);
     //! \brief Convert a double to an unsigned 32-bit integer, rounding towards zero
     //! On Arm this conversion is saturating (to UINT32_MAX/UINT32_MIN) for out of range input except when using `pico_double_compiler`
-    int32_t double2uint_z(double d);
+    uint32_t double2uint_z(double d);
 #endif
 #endif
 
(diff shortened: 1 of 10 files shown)


src/rp2_common/pico_double/include/pico/double.h
```diff
@@ -350,7 +350,7 @@ uint64_t double2ufix64(double f, int e);
 
 #if PICO_DOUBLE_HAS_POWINT
 #if !__PICO_DOUBLE_ARM_OPTIMIZED && __has_builtin(__builtin_powi)
-    static __force_inline double powint(double f, int32_t p) {
+    static __force_inline double powint(double d, int32_t p) {
         return __builtin_powi(f, p);
```
will-v-pi:
```suggestion
        return __builtin_powi(d, p);
```

src/rp2_common/pico_runtime_init/runtime_init.c
```diff
@@ -109,18 +109,17 @@ PICO_RUNTIME_INIT_FUNC_HW(runtime_init_usb_power_down, PICO_RUNTIME_INIT_USB_POW
 
 #if !PICO_RUNTIME_NO_INIT_PER_CORE_ENABLE_COPROCESSORS
 #include "hardware/gpio.h" // PICO_USE_GPIO_COPROCESSOR is defined here
-#include "hardware/structs/m33.h"
```
will-v-pi, line 112: Does this need to be replaced with `#include "pico/platform/cpu_regs.h"`?

test/pico_low_power_test/CMakeLists.txt
will-v-pi, line 1: The indentation in this file has gone a bit odd - specifically the end brackets for `add_executable` are now 1 level less indented that before, and most of the `if (NOT PICO_RISCV)` hasn't been indented correctly

src/rp2_common/pico_double/include/pico/double.h
```diff
@@ -230,30 +230,30 @@ extern "C" {
...
+    static inline int32_t double2uint_z(double d) { return (uint32_t)d; }
 #else
     //! \brief Convert a double to a signed 32-bit integer, rounding towards zero.
     //! On Arm this conversion is saturating (to INT32_MAX/INT32_MIN) for out of range input except when using `pico_double_compiler`
-    int32_t double2int_z(double f);
+    int32_t double2int_z(double d);
     //! \brief Convert a double to an unsigned 32-bit integer, rounding towards zero
     //! On Arm this conversion is saturating (to UINT32_MAX/UINT32_MIN) for out of range input except when using `pico_double_compiler`
-    int32_t double2uint_z(double f);
+    int32_t double2uint_z(double d);
```
lurch: Comment above the code says "Convert a double to an unsigned 32-bit integer" - does this mean that this _actually_ ought to be:
```suggestion
    uint32_t double2uint_z(double d);
```
?
(same thing is also true for various other `double2uint`, `double2uint64`, `float2uint`, `float2uint64` functions too)

Or is it the code-comment that's wrong / misleading?

src/rp2_common/pico_runtime_init/runtime_init.c
kilograham: it is already included indirectly, but i've added it back for clarity

src/rp2_common/pico_double/include/pico/double.h
kilograham: goodt catch; thanks

test/pico_low_power_test/CMakeLists.txt
kilograham: i've fixed the RISC_V - the add_executable i like as it is now!

will-v-pi (approved):

PR #3022, opened 2026-06-21
add some struct doxygen

kilograham:
thanks claude!

### DIFF
--- a/src/common/boot_picobin_headers/include/boot/picobin.h
+++ b/src/common/boot_picobin_headers/include/boot/picobin.h
@@ -147,16 +147,26 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+/*! \brief A single entry in a PICOBIN load map
+ *  \ingroup boot_picobin_headers
+ *
+ * Describes the mapping of one contiguous region from its storage location to its runtime address.
+ */
 typedef struct {
     // these must all be word aligned
-    uint32_t storage_address_rel;
-    uint32_t runtime_address;
-    uint32_t size;
+    uint32_t storage_address_rel; ///< Storage address of the region, relative to the load map base
+    uint32_t runtime_address;     ///< Target runtime address to which the region is loaded
+    uint32_t size;                ///< Size of the region in bytes
 } picobin_load_map_entry;
 
+/*! \brief PICOBIN load map describing regions to copy from storage to RAM
+ *  \ingroup boot_picobin_headers
+ *
+ * Contains a header word followed by a variable-length array of load map entries.
+ */
 typedef struct {
-    uint32_t header;
-    picobin_load_map_entry entries[];
+    uint32_t header;                  ///< Encodes the entry count and whether addresses are relative
+    picobin_load_map_entry entries[]; ///< Array of load map entries
 } picobin_load_map;
 
 static inline unsigned int picobin_load_map_entry_count(const picobin_load_map *lm) {
--- a/src/common/boot_picobin_headers/include/boot/picobin.h
+++ b/src/common/boot_picobin_headers/include/boot/picobin.h
@@ -154,9 +154,9 @@
  */
 typedef struct {
     // these must all be word aligned
-    uint32_t storage_address_rel; ///< Storage address of the region, relative to the load map base
+    uint32_t storage_address_rel; ///< Storage address of the region - if absolute then an absolute address, otherwise relative to the address of the LOAD_MAP item containing this entry
     uint32_t runtime_address;     ///< Target runtime address to which the region is loaded
-    uint32_t size;                ///< Size of the region in bytes
+    uint32_t size;                ///< If absolute then the end of the runtime address region, otherwise the size of the region in bytes
 } picobin_load_map_entry;
 
 /*! \brief PICOBIN load map describing regions to copy from storage to RAM
@@ -165,7 +165,7 @@ typedef struct {
  * Contains a header word followed by a variable-length array of load map entries.
  */
 typedef struct {
-    uint32_t header;                  ///< Encodes the entry count and whether addresses are relative
+    uint32_t header;                  ///< Encodes the item type, item size, entry count, and whether addresses are absolute or relative
     picobin_load_map_entry entries[]; ///< Array of load map entries
 } picobin_load_map;
 
--- a/src/common/boot_picoboot_headers/include/boot/picoboot.h
+++ b/src/common/boot_picoboot_headers/include/boot/picoboot.h
@@ -82,23 +82,21 @@ enum picoboot_status {
     PICOBOOT_UNSUPPORTED_MODIFICATION = 17,
 };
 
-/*! \brief Parameters for a reboot command
+/*! \brief Parameters for a reboot command (RP2040 only)
  *  \ingroup boot_picoboot_headers
  *
  * Sent as the argument payload of a PC_REBOOT command.
  */
 struct __packed picoboot_reboot_cmd {
-    uint32_t dPC; ///< Program counter to reboot to; 0 means reset into the regular boot path
-    uint32_t dSP; ///< Stack pointer value at reboot
+    uint32_t dPC; ///< Program counter to reboot to; 0 means reset into the regular boot path, otherwise must be a RAM address
+    uint32_t dSP; ///< Stack pointer value at reboot; ignored unless dPC is a RAM address
     uint32_t dDelayMS; ///< Delay in milliseconds before rebooting
 };
 
-/*! \brief Parameters for an extended reboot command (RP2350 only)
+/*! \brief Parameters for an extended reboot command (not available on RP2040)
  *  \ingroup boot_picoboot_headers
  *
- * Sent as the argument payload of a PC_REBOOT2 command. Has the same
- * layout as picoboot_reboot_cmd at the dSP/dDelayMS offsets but adds
- * a flags word and two extra parameters.
+ * Sent as the argument payload of a PC_REBOOT2 command.
  */
 struct __packed picoboot_reboot2_cmd {
     uint32_t dFlags; ///< Reboot flags controlling the boot path
@@ -107,11 +105,11 @@ struct __packed picoboot_reboot2_cmd {
     uint32_t dParam1; ///< Second reboot parameter (interpretation depends on flags)
 };
 
-// used for EXEC, VECTORIZE_FLASH
-/*! \brief Parameters for a command that operates on a single address
+/*! \brief Parameters for a command that operates on a single address (RP2040 only)
  *  \ingroup boot_picoboot_headers
  *
- * Sent as the argument payload of PC_EXEC and PC_VECTORIZE_FLASH commands.
+ * Sent as the argument payload of PC_EXEC and PC_VECTORIZE_FLASH commands,
+ * which are not supported on RP2350.
  */
 struct __packed picoboot_address_only_cmd {
     uint32_t dAddr; ///< Target address
@@ -128,17 +126,18 @@ struct __packed picoboot_range_cmd {
     uint32_t dSize; ///< Size of the range in bytes
 };
 
-/*! \brief Placeholder command retained for backwards compatibility with RP2350 bootrom builds
- *  \ingroup boot_picoboot_headers
- */
+// remains defined for backwards compatibility with RP2350 bootrom builds
 struct __packed picoboot_exec2_cmd {
-    uint32_t dummy; ///< Unused; present for backwards compatibility
+    uint32_t dummy;
 };
 
+/*! \brief Exclusivity level for a PC_EXCLUSIVE_ACCESS command
+ *  \ingroup boot_picoboot_headers
+ */
 enum picoboot_exclusive_type {
-    NOT_EXCLUSIVE = 0,
-    EXCLUSIVE,
-    EXCLUSIVE_AND_EJECT
+    NOT_EXCLUSIVE = 0,   ///< No restriction on USB Mass Storage operation
+    EXCLUSIVE,           ///< Disable USB Mass Storage writes (any active UF2 download will be aborted)
+    EXCLUSIVE_AND_EJECT  ///< Lock out USB Mass Storage by marking the drive media as not present (eject the drive)
 };
 
 /*! \brief Parameters for an exclusive-access command
@@ -150,7 +149,7 @@ struct __packed picoboot_exclusive_cmd {
     uint8_t bExclusive; ///< Exclusivity level; one of the picoboot_exclusive_type values
 };
 
-/*! \brief Parameters for an OTP read or write command (RP2350 only)
+/*! \brief Parameters for an OTP read or write command (not available on RP2040)
  *  \ingroup boot_picoboot_headers
  *
  * Sent as the argument payload of PC_OTP_READ and PC_OTP_WRITE commands.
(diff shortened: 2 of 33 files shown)


src/common/boot_picobin_headers/include/boot/picobin.h
```diff
@@ -147,16 +147,26 @@
...
+ * Describes the mapping of one contiguous region from its storage location to its runtime address.
+ */
 typedef struct {
     // these must all be word aligned
-    uint32_t storage_address_rel;
-    uint32_t runtime_address;
-    uint32_t size;
+    uint32_t storage_address_rel; ///< Storage address of the region, relative to the load map base
+    uint32_t runtime_address;     ///< Target runtime address to which the region is loaded
+    uint32_t size;                ///< Size of the region in bytes
```
will-v-pi: This should also include descriptions for absolute load maps - also, should probably use the wording more similar to the datasheet, for consistency (eg `relative to the address of this LOAD_MAP item` rather than `relative to the load map base`)

src/common/boot_picobin_headers/include/boot/picobin.h
```diff
@@ -147,16 +147,26 @@
...
 
+/*! \brief PICOBIN load map describing regions to copy from storage to RAM
+ *  \ingroup boot_picobin_headers
+ *
+ * Contains a header word followed by a variable-length array of load map entries.
+ */
 typedef struct {
-    uint32_t header;
-    picobin_load_map_entry entries[];
+    uint32_t header;                  ///< Encodes the entry count and whether addresses are relative
```
will-v-pi: Also encodes the item type and the size of the item

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
+ *  \ingroup boot_picoboot_headers
+ *
+ * Sent as the argument payload of a PC_REBOOT command.
+ */
 struct __packed picoboot_reboot_cmd {
-    uint32_t dPC; // 0 means reset into regular boot path
-    uint32_t dSP;
-    uint32_t dDelayMS;
+    uint32_t dPC; ///< Program counter to reboot to; 0 means reset into the regular boot path
+    uint32_t dSP; ///< Stack pointer value at reboot
```
will-v-pi: This should note that `dPC` can only be 0 or a RAM address, and that `dSP` is ignored unless booting into a RAM address - maybe using similar wording to the datasheet?

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
     PICOBOOT_UNSUPPORTED_MODIFICATION = 17,
 };
 
+/*! \brief Parameters for a reboot command
```
will-v-pi: This should be RP2040 only

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
 };
 
-
-// note this (with pc_sp) union member has the same layout as picoboot_reboot_cmd except with extra dFlags
+/*! \brief Parameters for an extended reboot command (RP2350 only)
+ *  \ingroup boot_picoboot_headers
+ *
+ * Sent as the argument payload of a PC_REBOOT2 command. Has the same
+ * layout as picoboot_reboot_cmd at the dSP/dDelayMS offsets but adds
+ * a flags word and two extra parameters.
```
will-v-pi: This information probably isn't very helpful - the fields for `REBOOT2` seem unrelated to the fields for `REBOOT`?

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
-    uint32_t dSP;
-    uint32_t dDelayMS;
+    uint32_t dPC; ///< Program counter to reboot to; 0 means reset into the regular boot path
+    uint32_t dSP; ///< Stack pointer value at reboot
+    uint32_t dDelayMS; ///< Delay in milliseconds before rebooting
 };
 
-
-// note this (with pc_sp) union member has the same layout as picoboot_reboot_cmd except with extra dFlags
+/*! \brief Parameters for an extended reboot command (RP2350 only)
```
will-v-pi: Might be better documented as `(not available on RP2040)`?

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
-    uint32_t dParam0;
-    uint32_t dParam1;
+    uint32_t dFlags; ///< Reboot flags controlling the boot path
+    uint32_t dDelayMS; ///< Delay in milliseconds before rebooting
+    uint32_t dParam0; ///< First reboot parameter (interpretation depends on flags)
+    uint32_t dParam1; ///< Second reboot parameter (interpretation depends on flags)
 };
 
 // used for EXEC, VECTORIZE_FLASH
+/*! \brief Parameters for a command that operates on a single address
```
will-v-pi: I think this is RP2040 only, as `EXEC` and `VECTORIZE_FLASH` aren't supported on RP2350

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
-    uint32_t dAddr;
-    uint32_t dSize;
+    uint32_t dAddr; ///< Start address of the range
+    uint32_t dSize; ///< Size of the range in bytes
 };
 
-// remains defined for backwards compatibility with RP2350 bootrom builds
+/*! \brief Placeholder command retained for backwards compatibility with RP2350 bootrom builds
+ *  \ingroup boot_picoboot_headers
+ */
```
will-v-pi: This probably shouldn't appear in the docs - it's only here to preserve backwards compatibility with bootrom builds

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
 struct __packed picoboot_exec2_cmd {
-    uint32_t dummy;
-};
-
-enum picoboot_exclusive_type {
-    NOT_EXCLUSIVE = 0,
-    EXCLUSIVE,
-    EXCLUSIVE_AND_EJECT
```
will-v-pi, line 119: This enum shouldn't be deleted - it should be documented as it's referenced below

src/common/boot_picoboot_headers/include/boot/picoboot.h
will-v-pi: Same for other `(RP2350 only)` commands

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,132 @@ enum picoboot_status {
...
+ * Sent as the argument payload of a PC_GET_INFO command.
+ */
 struct __packed picoboot_get_info_cmd {
-    uint8_t bType;
-    uint8_t bParam;
-    uint16_t wParam;
-    uint32_t dParams[3];
+    uint8_t bType; ///< Info type selector
+    uint8_t bParam; ///< Primary parameter for the selected info type
+    uint16_t wParam; ///< Secondary parameter for the selected info type
```
will-v-pi: These aren't documented in the datasheet, and don't seem to be used by `picotool` - are these dummy parameters?

src/common/boot_uf2_headers/include/boot/uf2.h
```diff
@@ -45,18 +45,24 @@
...
-    uint32_t block_no;
-    uint32_t num_blocks;
+    uint32_t magic_start0; ///< First magic number identifying a UF2 block (UF2_MAGIC_START0)
+    uint32_t magic_start1; ///< Second magic number identifying a UF2 block (UF2_MAGIC_START1)
+    uint32_t flags;        ///< Block flags (e.g. UF2_FLAG_FAMILY_ID_PRESENT)
+    uint32_t target_addr;  ///< Address in flash or memory at which to write this block's payload
+    uint32_t payload_size; ///< Number of valid data bytes in the payload field
+    uint32_t block_no;     ///< Zero-based index of this block within the UF2 file
+    uint32_t num_blocks;   ///< Total number of blocks in the UF2 file
     uint32_t file_size; // or familyID;
```
will-v-pi: Should this be documented like the other fields? Presumbaly this should always be the family ID for RP2 UF2 blocks

src/common/pico_base_headers/include/pico/types.h
```diff
@@ -37,7 +37,7 @@ typedef unsigned int uint;
  */
 #if PICO_OPAQUE_ABSOLUTE_TIME_T
 typedef struct {
-    uint64_t _private_us_since_boot;
+    uint64_t _private_us_since_boot; ///< Private implementation detail
```
will-v-pi: Does this need documenting? I'm pretty sure the doxygen build won't pick it up anyway, as `PICO_OPAQUE_ABSOLUTE_TIME_T` won't be set

will-v-pi (commented):
I have a few comments - generally, I think the picobin/picoboot docs should use similar wording to the datasheet for consistency, either changing the datasheet to match these descriptions, or changing these descriptions to align with the datasheet

will-v-pi:
Is this PR supposed to update `btstack` (which is causing the `other-bazel-checks` failures), or should that be in a separate PR? I think the deleted `picoboot_exclusive_type` enum is causing the other CI failures.

src/common/boot_picobin_headers/include/boot/picobin.h
kilograham: can you please update/suggest

will-v-pi:
```suggestion
    uint32_t storage_address_rel; ///< Storage address of the region - if absolute then an absolute address, otherwise relative to the address of the LOAD_MAP item containing this entry
    uint32_t runtime_address;     ///< Target runtime address to which the region is loaded
    uint32_t size;                ///< If absolute then the end of the runtime address region, otherwise the size of the region in bytes
```

will-v-pi: Have pushed an update which addresses most of my comments - also written using claude, but I gave it the datasheets so it had a source of truth to write the comments from

src/common/boot_picoboot_headers/include/boot/picoboot.h
will-v-pi: @kilograham this question still needs answering - are these unused parameters (and should therefore be documented as such), or are these used by some versions of `get_info`? As far as I can tell, `picotool` only uses the `dParams` below, and these `bParam` and `wParam` are unused.

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,137 @@ enum picoboot_status {
...
+ *
+ * Sent as the argument payload of a PC_GET_INFO command.
+ */
 struct __packed picoboot_get_info_cmd {
-    uint8_t bType;
-    uint8_t bParam;
-    uint16_t wParam;
-    uint32_t dParams[3];
+    uint8_t bType; ///< Info type selector
+    uint8_t bParam; ///< Primary parameter for the selected info type
```
kilograham:
```suggestion
    uint8_t bParam; ///< Unused
```

src/common/boot_picoboot_headers/include/boot/picoboot.h
```diff
@@ -82,87 +82,137 @@ enum picoboot_status {
...
+    uint16_t wParam; ///< Secondary parameter for the selected info type
```
kilograham:
```suggestion
    uint16_t wParam; ///< Unused
```

will-v-pi (approved):

src/common/pico_binary_info/include/pico/binary_info/structure.h
```diff
@@ -64,104 +64,153 @@ typedef struct _binary_info_core binary_info_t;
...
+        uint32_t len;                        ///< Maximum length of the string buffer
 } binary_info_ptr_string_with_name_t;
 
+/*! \brief Binary info entry describing a block device in the binary
+ *  \ingroup pico_binary_info
+ */
 typedef struct __packed _binary_info_block_device {
-        struct _binary_info_core core;
+        struct _binary_info_core core;       ///< Common binary info header
         bi_ptr_of(const char) name; // optional static name (independent of what is formatted)
```
lurch, line 148: Should the existing `// ` style comment here have been converted into a `///< ` style comment? (same also on line 151)

src/common/pico_binary_info/include/pico/binary_info/structure.h
```diff
@@ -64,104 +64,153 @@ typedef struct _binary_info_core binary_info_t;
...
 typedef struct __packed _binary_info_pins_with_func {
-    struct _binary_info_core core;
+    struct _binary_info_core core;  ///< Common binary info header
     // p4_5 : p3_5 : p2_5 : p1_5 : p0_5 : func_4 : 010_3 //individual pins p0,p1,p2,p3,p4 ... if fewer than 5 then duplicate p
     //                    phi_5 : plo_5 : func_4 : 001_3 // pin range plo-phi inclusive
-    uint32_t pin_encoding;
+    uint32_t pin_encoding;          ///< Encoded pin numbers and function (see BI_PINS_ENCODING_* for format)
 } binary_info_pins_with_func_t;
 
+/*! \brief Binary info entry describing one or more GPIO pins (up to 64) and their assigned function
```
lurch, line 168: Unlike `_binary_info_pins64_with_name` (which has a `uint64_t pin_mask`) I _don't_ think that this is limited to just "up to 64" pins? The comments on lines 173 and 174 implies that the pin numbers are encoded with 8 bits, so it seems like this could actually encode up to 256 pins?

src/common/pico_time/include/pico/timeout_helper.h
```diff
@@ -13,9 +13,15 @@
 extern "C" {
 #endif
 
+/*! \brief State used to track a timeout condition
+ *  \ingroup pico_time
+ *
+ * Holds the deadline and an optional per-iteration parameter used by
+ * the check_timeout_fn callbacks returned by the init functions.
+ */
```
lurch, line 21: Would have been nice to add doxygen for `init_single_timeout_until` and `init_per_iteration_timeout_us` too :wink:

src/common/pico_util/include/pico/util/queue.h
```diff
@@ -29,15 +29,22 @@ extern "C" {
 
 #include "pico/lock_core.h"
 
+/*! \brief A multi-core and IRQ safe queue instance
+ *  \ingroup queue
+ *
+ * Stores a fixed number of elements of a fixed size. All operations are
```
lurch, line 35: The "Stores a fixed number of elements" seems to be in contrast with the `queue_try_add` and `queue_try_remove` functions? :wink: 
Perhaps this was supposed to say "Stores a maximum number of elements"?

src/rp2_common/hardware_interp/include/hardware/interp.h
```diff
@@ -304,10 +311,17 @@ static inline void interp_set_force_bits(interp_hw_t *interp, uint lane, uint bi
...
+ * Holds a snapshot of all interpolator registers so that the interpolator
+ * state can be saved and restored around code that needs to use it for a
+ * different purpose.
+ */
 typedef struct {
-    uint32_t accum[2];
-    uint32_t base[3];
-    uint32_t ctrl[2];
+    uint32_t accum[2]; ///< Saved accumulator values for lanes 0 and 1
+    uint32_t base[3];  ///< Saved base register values for lanes 0, 1 and 2
```
lurch, line 323: This "for lanes 0, 1 and 2" comment implies that there are 3 lanes, which isn't the case? (From a quick look at the RP2040 Datasheet, it looks like BASE2 is just a "mix" of both lanes?)

lurch:
Please ignore my post-merge comments here - I've now created #3038 instead.

PR #3024, opened 2026-06-22
add pio_set_input_sync_bypass_with_mask and pio_set_input_sync_bypass_with_mask64

### LINKED ISSUE #2986: Add function to hardware_pio for modifying input_sync_bypass (to help RP2350B users)
Looking at the changes in https://github.com/raspberrypi/pico-examples/pull/727/files it seems slightly unfortunate that when dealing with PIO pin numbers > 32 on a RP2350B, "most" of the functionality is nicely wrapped up by the 64-bit pin functions, but the user still has to manually figure out the pin-offsets when setting the `input_sync_bypass` register (as this is only exposed via https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_structs/include/hardware/structs/pio.h and https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_structs/include/hardware/structs/pio.h and has no wrapper-functions in https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/hardware_pio/include/hardware/pio.h).

Maybe it'd be helpful if we had e.g. `pio_set_input_sync_bypass_with_mask(PIO pio, uint32_t pin_mask)` and `pio_set_input_sync_bypass_with_mask64(PIO pio, uint64_t pin_mask)` functions?
(and if adding that, maybe it _also_ makes sense to add functions for reading `dbg_padout` and `dbg_padoe` into 32-bit and 64-bit uints too?)

kilograham:
fixes #2986

### DIFF
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -2005,7 +2005,7 @@ int pio_sm_set_consecutive_pindirs(PIO pio, uint sm, uint pins_base, uint pin_co
  * \param bypass_enables the values to set - 1 = enable bypass, 0 = disable bypass (if the corresponding bit in pin_mask is set)
  * \param pin_mask a bit for each pin to indicate whether the corresponding bypass enable bit for that pin should be updated.
  */
-void pio_set_input_sync_bypass_with_mask(PIO pio, uint32_t bypass_enables, uint32_t pin_mask);
+void pio_sm_set_input_sync_bypass_with_mask(PIO pio, uint32_t bypass_enables, uint32_t pin_mask);
 
 /*! \brief Enable/disable the input synchronizer bypass for multiple pins for the PIO instance
  *  \ingroup hardware_pio
@@ -2014,7 +2014,7 @@ void pio_set_input_sync_bypass_with_mask(PIO pio, uint32_t bypass_enables, uint3
  * \param bypass_enables the values to set - 1 = enable bypass, 0 = disable bypass (if the corresponding bit in pin_mask is set)
  * \param pin_mask a bit for each pin to indicate whether the corresponding bypass enable bit for that pin should be updated.
  */
-void pio_set_input_sync_bypass_with_mask64(PIO pio, uint64_t bypass_enables, uint64_t pin_mask);
+void pio_sm_set_input_sync_bypass_with_mask64(PIO pio, uint64_t bypass_enables, uint64_t pin_mask);
 
 /*! \brief Mark a state machine as used
  *  \ingroup hardware_pio
--- a/src/rp2_common/hardware_pio/include/hardware/pio.h
+++ b/src/rp2_common/hardware_pio/include/hardware/pio.h
@@ -2005,7 +2005,7 @@ int pio_sm_set_consecutive_pindirs(PIO pio, uint sm, uint pins_base, uint pin_co
  * \param bypass_enables the values to set - 1 = enable bypass, 0 = disable bypass (if the corresponding bit in pin_mask is set)
  * \param pin_mask a bit for each pin to indicate whether the corresponding bypass enable bit for that pin should be updated.
  */
-void pio_sm_set_input_sync_bypass_with_mask(PIO pio, uint32_t bypass_enables, uint32_t pin_mask);
+void pio_set_input_sync_bypass_with_mask(PIO pio, uint32_t bypass_enables, uint32_t pin_mask);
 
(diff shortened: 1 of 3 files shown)


peterharperuk (dismissed):

src/rp2_common/hardware_pio/pio.c
```diff
@@ -341,6 +342,33 @@ void pio_sm_set_pindirs_with_mask64(PIO pio, uint sm, uint64_t pindirs, uint64_t
...
+}
+#endif
+
+void pio_sm_set_input_sync_bypass_with_mask64(PIO pio, uint64_t bypass_enables, uint64_t pin_mask) {
+    check_pio_pin_mask64(pio, 0, pin_mask);
+#if PICO_PIO_USE_GPIO_BASE
+    uint gpio_base = pio_get_gpio_base(pio);
+    bypass_enables >>= gpio_base;
+    pin_mask >>= gpio_base;
+#endif
```
lurch, line 368: This "pattern" appears a number of times in this file, so I guess you _could_ add a
```c
#if PICO_PIO_USE_GPIO_BASE
#define shift_pin_values_and_mask(pio, values, mask) \
    uint gpio_base = pio_get_gpio_base(pio); \
    values >>= gpio_base; \
    mask >>= gpio_base;
#else
#define shift_pin_values_and_mask(pio, values, mask)
#endif
```
macro, but that's possibly too much indirection?

lurch (commented):
I've not tested this, but it looks like it works in a very similar style to e.g. `pio_sm_set_pindirs_with_mask64`.
Thanks Graham :slightly_smiling_face:

src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -1998,6 +1998,24 @@ void pio_sm_set_pindirs_with_mask64(PIO pio, uint sm, uint64_t pin_dirs, uint64_
...
 int pio_sm_set_consecutive_pindirs(PIO pio, uint sm, uint pins_base, uint pin_count, bool is_out);
 
+/*! \brief Enable/disable the input synchronizer bypass for multiple pins for the PIO instance
+ *  \ingroup hardware_pio
+ *
+ * \param pio The PIO instance; e.g. \ref pio0, \ref pio1 etc.
+ * \param bypass_enables the values to set - 1 = enable bypass, 0 = disable bypass (if the corresponding bit in pin_mask is set)
+ * \param pin_mask a bit for each pin to indicate whether the corresponding bypass enable bit for that pin should be updated.
+ */
+void pio_set_input_sync_bypass_with_mask(PIO pio, uint32_t bypass_enables, uint32_t pin_mask);
```
peterharperuk: Do you have the function name wrong in the header files? Should be pio_sm_set not pio_set

lurch: Nope, because input_sync_bypass applies at the PIO-level, not at the StateMachine-level?

peterharperuk: good point. Then the source file is wrong

lurch: Oh yeah. Whoops.

peterharperuk (approved):

src/rp2_common/hardware_pio/pio.c
```diff
@@ -341,6 +342,33 @@ void pio_sm_set_pindirs_with_mask64(PIO pio, uint sm, uint64_t pindirs, uint64_t
     pio_sm_set_pindirs_with_mask_internal(pio, sm, (uint32_t)pindirs, (uint32_t)pin_mask);
 }
 
+void pio_set_input_sync_bypass_with_mask_internal(PIO pio, uint32_t bypass_enables, uint32_t pin_mask) {
+    check_pio_param(pio);
+    hw_xor_bits(&pio->input_sync_bypass, (pio->input_sync_bypass ^ bypass_enables) & pin_mask);
```
lurch, line 347: Is this doing a read-modify-write? I thought the whole point of `hw_xor_bits` was that it modified things atomically? (Or have I misunderstood, and this is just the way that things need to be done? Hmmm, I guess the only other way to do this would be via separate calls to `hw_clear_bits` and `hw_set_bits`, which would probably be slower than what's being done here.)
Ah! Would it be worth using `hw_write_masked` here, rather than manually doing these bit-manipulations?

peterharperuk: It does look like hw_write_masked would work. But I'll let @kilograham comment as he wrote it.

PR #3025, opened 2026-06-22
Update extern c ignore

peterharperuk:
Add some missing extern_c wrappers to some header files.
Update the script to make checking the exclude list easier.


lurch:
Should we (also) add some kind of GitHub CI workflow for this, so that it gets run for every PR (and thus preventing any "bad" headers from getting merged in future) ?

peterharperuk:
That's the plan

kilograham (approved):

lurch:
This is a follow-up to #3003 and fixes #2865

PR #3029, opened 2026-06-24
Use `unsigned int` instead of `uint` in `BINARY_INFO_MAKE_TAG`

### LINKED ISSUE #3028: `BINARY_INFO_MAKE_TAG` uses non-portable `uint` typedef
The [`BINARY_INFO_MAKE_TAG`](https://github.com/raspberrypi/pico-sdk/blob/develop/src/common/pico_binary_info/include/pico/binary_info/structure.h#L45) macro in the `binary_info` module relies on `uint` being defined, which is only true when building as part of the pico SDK (or if we happened to already have this typedef'd ourselves). 

In [`pico/binary_info/structure.h`](https://github.com/raspberrypi/pico-sdk/blob/develop/src/common/pico_binary_info/include/pico/binary_info/structure.h#L10) it explicitly states:
> NOTE: This file may be included by non SDK code, so does not use SDK includes

I ran into this issue when using these headers in my own library, it seems like the most portable solution here is to just use `unsigned int` directly.

I'll follow up with a PR.

loopj:
Adjust the [BINARY_INFO_MAKE_TAG](https://github.com/raspberrypi/pico-sdk/blob/develop/src/common/pico_binary_info/include/pico/binary_info/structure.h#L45) macro in the binary_info module to use `unsigned int` instead of `uint`, making it more portable without users having to typedef `uint` themselves, since this isn't a standard stdint type.

Makes the `binary_info` headers more portable (as intended) and is backwards compatible.

Fixes #3028

### DIFF
--- a/src/common/pico_binary_info/include/pico/binary_info/structure.h
+++ b/src/common/pico_binary_info/include/pico/binary_info/structure.h
@@ -42,7 +42,7 @@ typedef struct _binary_info_core binary_info_t;
 // note plan is to reserve c1 = 0->31 for "collision tags"; i.e.
 // for which you should always use random IDs with the binary_info,
 // giving you 4 + 8 + 32 = 44 bits to avoid collisions
-#define BINARY_INFO_MAKE_TAG(c1, c2) ((((unsigned int)c2&0xffu)<<8u)|((unsigned int)c1&0xffu))
+#define BINARY_INFO_MAKE_TAG(c1, c2) ((uint16_t)((((c2)&0xffu)<<8u)|((c1)&0xffu)))
 
 // Raspberry Pi defined. do not use
 #define BINARY_INFO_TAG_RASPBERRY_PI BINARY_INFO_MAKE_TAG('R','P')
--- a/src/common/pico_binary_info/include/pico/binary_info/structure.h
+++ b/src/common/pico_binary_info/include/pico/binary_info/structure.h
@@ -42,7 +42,7 @@ typedef struct _binary_info_core binary_info_t;
 // note plan is to reserve c1 = 0->31 for "collision tags"; i.e.
 // for which you should always use random IDs with the binary_info,
 // giving you 4 + 8 + 32 = 44 bits to avoid collisions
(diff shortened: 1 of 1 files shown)


src/common/pico_binary_info/include/pico/binary_info/structure.h
```diff
@@ -42,7 +42,7 @@ typedef struct _binary_info_core binary_info_t;
 // note plan is to reserve c1 = 0->31 for "collision tags"; i.e.
 // for which you should always use random IDs with the binary_info,
 // giving you 4 + 8 + 32 = 44 bits to avoid collisions
-#define BINARY_INFO_MAKE_TAG(c1, c2) ((((uint)c2&0xffu)<<8u)|((uint)c1&0xffu))
+#define BINARY_INFO_MAKE_TAG(c1, c2) ((((unsigned int)c2&0xffu)<<8u)|((unsigned int)c1&0xffu))
```
lurch: If both `c1` and `c2` are being ANDed with 0xff, I guess that `unsigned char` would be sufficient? :thinking:

loopj: I was aiming to make the diff as trivial as possible. I can change to `unsigned char` if preferred, they generate the same result either way.

The generated assembly is identical for both at O1 and greater so there isn't really a real world benefit, so it just comes down to a style preference.

lurch: This is probably a question for @kilograham then. Although note that:
```bash
$ git grep BINARY_INFO_MAKE_TAG
src/common/pico_binary_info/include/pico/binary_info/structure.h:#define BINARY_INFO_MAKE_TAG(c1, c2) ((((uint)c2&0xffu)<<8u)|((uint)c1&0xffu))
src/common/pico_binary_info/include/pico/binary_info/structure.h:#define BINARY_INFO_TAG_RASPBERRY_PI BINARY_INFO_MAKE_TAG('R','P')
test/kitchen_sink/kitchen_sink.c:                           BINARY_INFO_MAKE_TAG('K', 'S'),
```

kilograham (dismissed):

kilograham (approved):

will-v-pi (approved):

PR #3031, opened 2026-06-25
Add some rough power consumption values to pico_low_power

will-v-pi:
Add some rough power consumption values to the description at the top of `low_power.h`, measured both when powered from VSYS (to avoid the diode and potential divider power drains that occur when powering from VBUS) and from 3V3 (to avoid the conversion losses in the 3.3V regulator, as battery powered devices may use a different 3.3V regulator).

Also add RP2040 needing external clock for dormant to the note, and fix incorrect `invalid_params_if` in `low_power_go_dormant`.

### DIFF
--- a/src/rp2_common/pico_low_power/include/pico/low_power.h
+++ b/src/rp2_common/pico_low_power/include/pico/low_power.h
@@ -57,11 +57,13 @@ extern "C" {
  * Some rough power consumption values when going to low power modes using timers, measured on Pico-series boards
  * (powered either from VSYS at 5.2V, or from 3V3 at 3.3V, running low_power_test_simple):
  *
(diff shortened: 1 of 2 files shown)


peterharperuk (dismissed):
Someone will probably ask why these numbers don't match 4.1.1 in the pico 2 datasheet

will-v-pi:
> Someone will probably ask why these numbers don't match 4.1.1 in the pico 2 datasheet

I've updated them so they more closely match those values. Note that this testing did not use a USB cable, as the Pstates are more likely to be useful on battery powered devices without a USB cable connected, so I didn't need to enable USB pull-downs to get the lowest power as there was no USB cable connected.

kilograham (approved):

Issue #2732, opened 2025-11-27
Should we increase shared handler count if using stdusb?


peterharperuk:
I just hit this assertion and couldn't immediately understand why.

ssertion "irq_handler_chain_free_slot_head >= 0"

We should add a comment to the code that the number of shared handlers is > PICO_MAX_SHARED_IRQ_HANDLERS

But the reason I suddenly hit the problem was that I used a program which enabled stdusb which uses 2 shared handlers and my code tries to use 3 shared handlers.

Should we increase PICO_MAX_SHARED_IRQ_HANDLERS if we link to stdusb?
Generally 4 seems a bit stingy to me - especially if you have a RP2350B?

What do you think @kilograham ?

kilograham:
yeah, lets up the handlers for stdio USB (and maybe add a few more on RP2350)

kilograham:
merged into develop (we add +2 when using stdio_usb)

PR #3037, opened 2026-06-27
add 2 shared irq handlers when using stdio_usb

kilograham:
note i didn't add a new config option to turn this off, since you can just set PICO_MAX_SHARED_IRQ_HANDLERS if you actually really care

### DIFF
--- a/src/rp2_common/hardware_irq/include/hardware/irq.h
+++ b/src/rp2_common/hardware_irq/include/hardware/irq.h
@@ -10,7 +10,12 @@
 // These two config items are also used by assembler, so keeping separate
 // PICO_CONFIG: PICO_MAX_SHARED_IRQ_HANDLERS, Maximum number of shared IRQ handlers, default=4, advanced=true, group=hardware_irq
 #ifndef PICO_MAX_SHARED_IRQ_HANDLERS
-#define PICO_MAX_SHARED_IRQ_HANDLERS 4
+#define _PICO_MAX_SHARED_IRQ_HANDLERS 4
+#if LIB_PICO_STDIO_USB
+#define PICO_MAX_SHARED_IRQ_HANDLERS (_PICO_MAX_SHARED_IRQ_HANDLERS + 2)
+#else
+#define PICO_MAX_SHARED_IRQ_HANDLERS _PICO_MAX_SHARED_IRQ_HANDLERS
+#endif
 #endif
 
 // PICO_CONFIG: PICO_DISABLE_SHARED_IRQ_HANDLERS, Disable shared IRQ handlers, type=bool, default=0, group=hardware_irq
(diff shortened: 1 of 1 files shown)


kilograham:
note i didn't add more on rp2350 by default, as no one else has complained

kilograham:
fixes #2732

src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -10,7 +10,12 @@
 // These two config items are also used by assembler, so keeping separate
 // PICO_CONFIG: PICO_MAX_SHARED_IRQ_HANDLERS, Maximum number of shared IRQ handlers, default=4, advanced=true, group=hardware_irq
```
lurch:
```suggestion
// PICO_CONFIG: PICO_MAX_SHARED_IRQ_HANDLERS, Maximum number of shared IRQ handlers, default=6 if using stdio_usb otherwise 4, advanced=true, group=hardware_irq
```

will-v-pi: It might be better phrased something like “4 available for use plus 2 used by stdio_usb if that is in use”? Probably with less clunky wording than mine

lurch: Hmmm, maybe "6 if using stdio_usb (with 2 of those used by stdio_usb itself) otherwise 4" ? :shrug: 
(I'm just kinda guessing here as I don't really understand this shared IRQ handlers stuff)

src/rp2_common/hardware_irq/include/hardware/irq.h
```diff
@@ -8,9 +8,14 @@
 #define _HARDWARE_IRQ_H
 
 // These two config items are also used by assembler, so keeping separate
-// PICO_CONFIG: PICO_MAX_SHARED_IRQ_HANDLERS, Maximum number of shared IRQ handlers, default=4, advanced=true, group=hardware_irq
+// PICO_CONFIG: PICO_MAX_SHARED_IRQ_HANDLERS, Maximum number of shared IRQ handlers, default=6 if using stdio_usb otherwise 4, advanced=true, group=hardware_irq
```
will-v-pi, line 11: Maybe this could do with some better clarification that you're not getting extra shared IRQ handlers to use when you're using stdio_usb, the number is just being increased so stdio_usb still leaves you with 4 handlers to use?

will-v-pi (approved):
One comment, but approving anyway, as it's good enough without addressing that

peterharperuk (approved):

src/rp2_common/hardware_irq/include/hardware/irq.h
kilograham: feel free to PR

PR #3038, opened 2026-06-27
Small doxygen tweaks

lurch:
Small followup to #3022

### DIFF
--- a/src/common/pico_binary_info/include/pico/binary_info/structure.h
+++ b/src/common/pico_binary_info/include/pico/binary_info/structure.h
@@ -155,7 +155,7 @@ typedef struct __packed _binary_info_block_device {
 #define BI_PINS_ENCODING_RANGE 1
 #define BI_PINS_ENCODING_MULTI 2
 
-/*! \brief Binary info entry describing one or more GPIO pins and their assigned function
+/*! \brief Binary info entry describing one or more GPIO pins and their assigned function, supporting pin numbers <32, and up to 5 pins
  *  \ingroup pico_binary_info
  */
 typedef struct __packed _binary_info_pins_with_func {
--- a/src/common/pico_binary_info/include/pico/binary_info/structure.h
+++ b/src/common/pico_binary_info/include/pico/binary_info/structure.h
@@ -165,7 +165,7 @@ typedef struct __packed _binary_info_pins_with_func {
     uint32_t pin_encoding;          ///< Encoded pin numbers and function (see BI_PINS_ENCODING_* for format)
 } binary_info_pins_with_func_t;
 
-/*! \brief Binary info entry describing one or more GPIO pins (more than 32) and their assigned function
+/*! \brief Binary info entry describing one or more GPIO pins and their assigned function, supporting pin numbers <256, and up to 7 pins
  *  \ingroup pico_binary_info
  */
 typedef struct __packed _binary_info_pins64_with_func {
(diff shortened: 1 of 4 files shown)


src/common/pico_binary_info/include/pico/binary_info/structure.h
will-v-pi: Maybe add something like this here

```suggestion
/*! \brief Binary info entry describing one or more GPIO pins and their assigned function, supporting pin numbers <32, and up to 5 pins
```

src/common/pico_binary_info/include/pico/binary_info/structure.h
```diff
@@ -160,19 +160,19 @@ typedef struct __packed _binary_info_block_device {
...
 typedef struct __packed _binary_info_pins_with_func {
     struct _binary_info_core core;  ///< Common binary info header
-    // p4_5 : p3_5 : p2_5 : p1_5 : p0_5 : func_4 : 010_3 //individual pins p0,p1,p2,p3,p4 ... if fewer than 5 then duplicate p
+    // p4_5 : p3_5 : p2_5 : p1_5 : p0_5 : func_4 : 010_3 //individual pins p0,p1,p2,p3,p4 ... if fewer than 5 then duplicate last pin
     //                    phi_5 : plo_5 : func_4 : 001_3 // pin range plo-phi inclusive
     uint32_t pin_encoding;          ///< Encoded pin numbers and function (see BI_PINS_ENCODING_* for format)
 } binary_info_pins_with_func_t;
 
-/*! \brief Binary info entry describing one or more GPIO pins (up to 64) and their assigned function
+/*! \brief Binary info entry describing one or more GPIO pins (more than 32) and their assigned function
```
will-v-pi:
```suggestion
/*! \brief Binary info entry describing one or more GPIO pins and their assigned function, supporting pin numbers <256, and up to 7 pins
```

src/rp2_common/hardware_interp/include/hardware/interp.h
```diff
@@ -343,23 +343,23 @@ void interp_save(interp_hw_t *interp, interp_hw_save_t *saver);
  */
 void interp_restore(interp_hw_t *interp, interp_hw_save_t *saver);
 
-/*! \brief Sets the interpolator base register by lane
+/*! \brief Sets the interpolator base register by index
```
kilograham, line 346: i dont think you meant to change these lane -> index?

lurch: I deliberately changed this, because it was implying there's a "lane 2" which I think was confusing / misleading? Happy to accept any improved wording.

lurch: (Another approach I considered was to leave these as "Sets the interpolator base register by lane" but to restrict the `lane` parameter to just be `0` or `1`, and then have a separate function for setting `interp->base[2]`, but I dismissed that idea because it would be backwards-incompatible)

lurch: Just realised that I forgot to rename the `lane` parameters - now fixed. (The heat was making it hard for me to concentrate :hot_face: )

will-v-pi (approved):
LGTM - I agree about changing the interpolator base register docs away from `lane`, as there are only 2 lanes, and `index` does seem like the simplest option

PR #3042, opened 2026-06-29
Stop using async writes in CYW43 driver for BTStack

peterharperuk:
Evidentially, the cyw43 btstack implementation is behaving like an async implementation in that it implements hci_transport_cyw43_can_send_now which always returns true and we send the HCI_EVENT_TRANSPORT_PACKET_SENT in hci_transport_cyw43_send_packet. This can cause problems, e.g. see https://github.com/bluekitchen/btstack/issues/738


peterharperuk:
@mringwal

will-v-pi (approved):
LGTM

PR #3044, opened 2026-06-29
sort SVD peripherals alphabetically

kilograham:
non functional change in preparation for updating with the latest headers/SVD which are sorted


kilograham:
merging non-functional change

lurch:
In hindsight it's a shame that this PR didn't also do the re-ordering of `src/rp2_common/cmsis/stub/CMSIS/Device/RP2040/Include/RP2040.h` and `src/rp2_common/cmsis/stub/CMSIS/Device/RP2350/Include/RP2350.h`, but no point crying over spilt milk :wink:

Issue #3045, opened 2026-06-29
pwm_set_mask_enabled missing complementary pwm_get_mask_enabled


sprickett:
Adding a formal get function would complement the existing set function
```C
uint32_t slice_mask = pwm_hw->en;  // Where is pwm_get_mask_enabled() when you want it?
slice_mask |= 1u << pwm_gpio_to_slice_num(RED_PIN);
slice_mask |= 1u << pwm_gpio_to_slice_num(GREEN_PIN);
slice_mask |= 1u << pwm_gpio_to_slice_num(BLUE_PIN);
pwm_set_mask_enabled(slice_mask);
```

Obviously one could just write:
```C
pwm_hw->en |= slice_mask;
```
But formal get and set functions are perhaps nicer.

I notice that get functions are few and far between so there might be a reason why this is not here already

lurch:
Hmmmm... In hindsight, I wonder if it would have made more sense for `pwm_set_mask_enabled` to have an API more like [`pio_set_sm_mask_enabled`](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_pio_1gaf726f77ee21f5373cadf064a4e0b4935) i.e. a separate `mask` and `enabled` parameter, so that your example above would be:
```c
uint32_t enable_mask = 1u << pwm_gpio_to_slice_num(RED_PIN) | 1u << pwm_gpio_to_slice_num(GREEN_PIN) | 1u << pwm_gpio_to_slice_num(BLUE_PIN);
pwm_set_mask_enabled(enable_mask, true);
```

We can't change the existing function's prototype because that would break backwards-compatibility, but maybe it's worth adding a new function for this? 🤔

PR #3046, opened 2026-06-29
Implement pwm_get_mask_enabled

sprickett:
Fixes [#3045](https://github.com/raspberrypi/pico-sdk/issues/3045), pwm_set_mask_enabled missing complementary pwm_get_mask_enabled

It also addresses a small typo in pwm_set_mask_enabled comments, Bitmap --> Bitmask

### DIFF
--- a/src/rp2_common/hardware_pwm/include/hardware/pwm.h
+++ b/src/rp2_common/hardware_pwm/include/hardware/pwm.h
@@ -566,7 +566,7 @@ static inline void pwm_set_mask_enabled(uint32_t mask) {
 /** \brief Get PWM slice enabled status
  *  \ingroup hardware_pwm
  *
- * \param Bitmask of all PWM slices currently enabled
+ * \return Bitmask of all PWM slices currently enabled
  */
 static inline uint32_t pwm_get_mask_enabled(void) {
     return pwm_hw->en;
--- a/src/rp2_common/hardware_pwm/include/hardware/pwm.h
+++ b/src/rp2_common/hardware_pwm/include/hardware/pwm.h
@@ -557,10 +557,19 @@ static inline void pwm_set_enabled(uint slice_num, bool enabled) {
 /** \brief Enable/Disable multiple PWM slices simultaneously
  *  \ingroup hardware_pwm
  *
- * \param mask Bitmap of PWMs to enable/disable. Bits 0 to 7 enable slices 0-7 respectively
+ * \param mask Bitmask of PWMs to enable/disable. Bits 0 to 7 enable slices 0-7 respectively
  */
 static inline void pwm_set_mask_enabled(uint32_t mask) {
     pwm_hw->en = mask;
+}
+
+/** \brief Get PWM slice enabled status
+ *  \ingroup hardware_pwm
+ *
+ * \param Bitmask of all PWM slices currently enabled
+ */
+static inline uint32_t pwm_get_mask_enabled(void) {
+    return pwm_hw->en;
 }
 
 /**


src/rp2_common/hardware_pwm/include/hardware/pwm.h
```diff
@@ -557,12 +557,21 @@ static inline void pwm_set_enabled(uint slice_num, bool enabled) {
...
  *  \ingroup hardware_pwm
  *
- * \param mask Bitmap of PWMs to enable/disable. Bits 0 to 7 enable slices 0-7 respectively
+ * \param mask Bitmask of PWMs to enable/disable. Bits 0 to 7 enable slices 0-7 respectively
  */
 static inline void pwm_set_mask_enabled(uint32_t mask) {
     pwm_hw->en = mask;
 }
 
+/** \brief Get PWM slice enabled status
```
sprickett, line 566: or Get PWM enabled slice status?

Wren6991 (commented):
That's not a typo. Bitmap is a correct term to use here. I'd even argue bitmask is incorrect because it's not masking anything.

What's the use case for this? Any combination of `get()` + `set()` is a read-modify-write which causes a data race with concurrent writers. The correct way to modify the register is something like:

```c
hw_set_bits(&pwm_hw->en, (1 << x) | (1 << y));
```

...which is fairly straightforward and doesn't have the same failure modes.

sprickett:
@Wren6991 Thanks for commenting.

I don't have much skin in the game here. To argue my case...

Regarding `pwm_get_mask_enabled(void)`
I was addressing what i saw as an asymmetry with `pwm_set_mask_enabled(uint32_t mask)` which is implemented simply as `pwm_hw->en = mask;`

It would seem from your comment that this should maybe be implemented as `hw_set_bits(&pwm_hw->en, mask);`?

Regardless it still does not solve the conundrum I was facing when implementing logic that utilised multiple pwm pins. 
I wanted logic that was reusable and self contained. `pwm_set_mask_enabled(uint32_t mask)` offers an attractive way to toggle all the associated slices in one go. This does not work when there is some other self contained logic that is using pwm because one disables the other. The initial solution was to read  `pwm_hw->en` directly. This felt a bit wrong so I proposed this getter. This getter also would allow the logic to assert a slice conflict. 

Regarding concurrency. I might have the wrong idea of some fundamentals here. Its great that the SDK provides functions that handle concurrent writes but for a simple use case utilising one core is there any concurrency to guard against?

Ultimately I used `pwm_set_enabled(uint slice_num, bool enabled)` on each of the slices individually. I used a mask to keep track of which slices had already been enabled by other pins in the set. I have also since noted that this function enables/disables a slice using different register addresses to  `pwm_set_mask_enabled(uint32_t mask)`.

Regarding Bitmap vs Bitmask, to me they both describe an integral type with zero or more bits set.
They are both used throughout the SDK interchangeably. Bitmask heavily outweighs Bitmap.
`pwm.h` has eight instances of Bitmask and one instance of Bitmap
Overall there are roughly 144 instances of Bitmask to 13 instances of Bitmap
There is only one instance of "Bitmap" exact.
The suggested change was for more for consistency with the SDK as a whole than to say it is wrong. 
 
Happy to close if you feel this is not well reasoned.

lurch:
> What's the use case for this? Any combination of `get()` + `set()` is a read-modify-write which causes a data race with concurrent writers.

@Wren6991 See my [comment](https://github.com/raspberrypi/pico-sdk/issues/3045#issuecomment-4843075674) in the associated issue.

PR #3048, opened 2026-06-30
Add DORMANT_CLOCK_SOURCE_RTC option on RP2040

will-v-pi:
This runs RTC from XOSC, which allows going dormant without an external timer, at the expense of higher power consumption (about double `DORMANT_CLOCK_SOURCE_XOSC`, but still 5x less than sleep)

I think this is the lowest power the RP2040 can run at without an external clock source for the RTC, so is a useful option to add

### DIFF
--- a/src/rp2_common/pico_low_power/include/pico/low_power.h
+++ b/src/rp2_common/pico_low_power/include/pico/low_power.h
@@ -60,13 +60,13 @@ extern "C" {
  * Mode                  | Pico (VSYS)    | Pico 2 (VSYS)   | Pico (3V3)     | Pico 2 (3V3)
  * ----------------------|----------------|-----------------|----------------|----------------
  * Sleep                 | 7.3mA (37.9mW) | 5.9mA (30.7mW)  | 8.7mA (28.5mW) | 6.9mA (22.7mW)
- * Dormant               | 0.76mA (4.0mW) | 3.3mA (17.0mW)  | 0.75mA (2.5mW) | 3.7mA (12.0mW)
+ * Dormant               | 0.95mA (5.0mW) | 3.3mA (17.0mW)  | 1.2mA (4.0mW)  | 3.7mA (12.0mW)
  * Pstate (SRAM0 On)     | N/A            | 0.25mA (1.32mW) | N/A            | 0.14mA (0.47mW)
  * Pstate (XIP SRAM On)  | N/A            | 0.22mA (1.21mW) | N/A            | 0.10mA (0.44mW)
  * Pstate (All SRAM Off) | N/A            | 0.18mA (1.10mW) | N/A            | 0.08mA (0.40mW)
  * 
  * NOTE: The RP2350 dormant values are higher than the RP2040 ones because RP2350 continues running clk_ref from the LPOSC to run the timer,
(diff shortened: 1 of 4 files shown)


src/rp2_common/pico_low_power/include/pico/low_power.h
```diff
@@ -222,13 +224,17 @@ static inline int low_power_set_external_clock_source(__unused uint src_hz, __un
  * The clocks specified in keep_enabled will be kept enabled during dormant, but XOSC and ROSC will be stopped.
  *
  * \if rp2040_specific
- * This requires an external clock source to be set using \ref low_power_set_external_clock_source before calling this function.
- * If the external clock source is not set, or it is not running, this will return PICO_ERROR_PRECONDITION_NOT_MET.
+ * If the clock source is set to DORMANT_CLOCK_SOURCE_RTC, all clocks will be switched to the ROSC while dormant so
+ * they can be stopped, except clk_rtc which will be run from the XOSC so that it continues running for the timer.
```
lurch, line 228: Does the "except clk_rtc which will be run from the XOSC" here mean that the "but XOSC and ROSC will be stopped" wording a few lines above needs to be changed? :shrug:  (I'm afraid I never got to grips with all this clock-source stuff)

will-v-pi: I think I won't change it up there, but can add more to this note:

```suggestion
 * they can be stopped, except clk_rtc which will be run from the XOSC so that it continues running for the timer.
 * In this case the XOSC will not be stopped.
```

src/rp2_common/pico_low_power/include/pico/low_power.h
```diff
@@ -243,6 +250,12 @@ int low_power_dormant_until_aon_timer(absolute_time_t until, dormant_clock_sourc
  *
  * Go dormant until the given GPIO pin changes state.
  * The clocks specified in keep_enabled will be kept enabled during dormant, but XOSC and ROSC will be stopped.
+ * 
+ * \if rp2040_specific
+ * If the clock source is set to DORMANT_CLOCK_SOURCE_RTC, all clocks will be switched to the ROSC while dormant so
+ * they can be stopped, except clk_rtc which will be run from the XOSC. For the lowest power consumption, you should use
```
lurch: I guess the "In this case the XOSC will not be stopped." ought to be mentioned here too?

will-v-pi: Thanks, added

src/rp2_common/pico_low_power/include/pico/low_power.h
will-v-pi: Switching to this might be a better default, as it doesn’t require an external clock source for the RTC
```suggestion
#define DORMANT_CLOCK_SOURCE_DEFAULT DORMANT_CLOCK_SOURCE_RTC
```

liamfraser (dismissed):

src/rp2_common/pico_low_power/include/pico/low_power.h
will-v-pi: Done - also updated the tests and power consumption values for this change

liamfraser (approved):

PR #3049, opened 2026-06-30
Add low_power_set_pins_low_leakage function

will-v-pi:
This sets the pins to what I think is the lowest leakage state (pads_bank0 have pulls & input disabled, io_bank0 has output override set to disabled)

### DIFF
--- a/src/rp2_common/pico_low_power/include/pico/low_power.h
+++ b/src/rp2_common/pico_low_power/include/pico/low_power.h
@@ -140,21 +140,21 @@ typedef void (*low_power_pstate_resume_func)(pstate_bitset_t *pstate);
  *
  * \param exclude_mask Mask of the pins to exclude from this
  */
-void low_power_set_pins_low_leakage(uint32_t exclude_mask);
+void low_power_set_all_pins_low_leakage(uint32_t exclude_mask);
 
 /*! \brief  Set all pins to a low leakage state (64-bit mask version)
  *  \ingroup pico_low_power
  *
- * \see low_power_set_pins_low_leakage
+ * \see low_power_set_all_pins_low_leakage
  *
  * \param exclude_mask Mask of the pins to exclude from this
  */
 #if NUM_BANK0_GPIOS <= 32
-static inline void low_power_set_pins_low_leakage64(uint64_t exclude_mask) {
-    low_power_set_pins_low_leakage((uint32_t)exclude_mask);
+static inline void low_power_set_all_pins_low_leakage64(uint64_t exclude_mask) {
+    low_power_set_all_pins_low_leakage((uint32_t)exclude_mask);
 }
 #else
-void low_power_set_pins_low_leakage64(uint64_t exclude_mask);
+void low_power_set_all_pins_low_leakage64(uint64_t exclude_mask);
 #endif
 
 /*! \brief  Sleep until an interrupt occurs
--- a/src/rp2_common/pico_low_power/include/pico/low_power.h
+++ b/src/rp2_common/pico_low_power/include/pico/low_power.h
@@ -142,7 +142,7 @@ typedef void (*low_power_pstate_resume_func)(pstate_bitset_t *pstate);
  */
 void low_power_set_pins_low_leakage(uint32_t exclude_mask);
 
-/*! \brief  Set all pins to a low leakage state (64-bit version)
+/*! \brief  Set all pins to a low leakage state (64-bit mask version)
  *  \ingroup pico_low_power
  *
  * \see low_power_set_pins_low_leakage
--- a/src/rp2_common/pico_low_power/include/pico/low_power.h
+++ b/src/rp2_common/pico_low_power/include/pico/low_power.h
@@ -133,7 +133,7 @@ typedef void (*low_power_pstate_resume_func)(pstate_bitset_t *pstate);
 /*! \brief  Set all pins to a low leakage state
  *  \ingroup pico_low_power
  *
- * Disables pulls & inputs on the pads, and sets the IO function to GPIO_FUNC_SIO
+ * Disables pulls & inputs on the pads, and disables the IO output
  * with all pins set to inputs. This results in the lowest leakage current.
  * 
  * Does not change the state of pins in the exclude_mask
--- a/src/rp2_common/pico_low_power/include/pico/low_power.h
+++ b/src/rp2_common/pico_low_power/include/pico/low_power.h
@@ -130,6 +130,12 @@ typedef enum {
 typedef void (*low_power_pstate_resume_func)(pstate_bitset_t *pstate);
 #endif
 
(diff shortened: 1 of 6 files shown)


lurch:
Naive question, but do the GPIO isolation latches on RP2350 help at all?

src/rp2_common/pico_low_power/low_power.c
```diff
@@ -224,6 +224,31 @@ int low_power_set_external_clock_source(uint src_hz, uint gpio_pin) {
...
+    }
+}
+
+// this function is provided (in low_power.h) as an inlined call to the 32 bit version if we have <= 32 GPIOs
+#if NUM_BANK0_GPIOS > 32
+void low_power_set_pins_low_leakage64(uint64_t exclude_mask) {
+    uint64_t include_mask = ~exclude_mask;
+    gpio_set_dir_masked64(include_mask, 0);
+    for (uint i=0; i < NUM_BANK0_GPIOS; i++) {
+        if (exclude_mask & (1u << i)) continue;
```
peterharperuk: I wonder whether that neede to be (1ull << i)

will-v-pi: Yes, it probably does
```suggestion
        if (exclude_mask & (1ull << i)) continue;
```

will-v-pi:
> Naive question, but do the GPIO isolation latches on RP2350 help at all?

As far as I understand it, the isolation latches just isolate the pads logic from the switched core until they are configured - isolation is enabled anyway when the switched core is powered off (9.7. Pad isolation latches in the datasheet)

src/rp2_common/pico_low_power/low_power.c
```diff
@@ -224,6 +224,31 @@ int low_power_set_external_clock_source(uint src_hz, uint gpio_pin) {
 }
 #endif  // inline in header for other platforms
 
+void low_power_set_pins_low_leakage(uint32_t exclude_mask) {
+    uint32_t include_mask = ~exclude_mask;
+    gpio_set_dir_masked(include_mask, 0);
+    for (uint i=0; i < NUM_BANK0_GPIOS; i++) {
+        if (exclude_mask & (1u << i)) continue;
```
lurch, line 229: I guess you _could_ also add
```c
if (! (include_mask >> i) ) break;
```
but it's probably not worth it?

will-v-pi: I've removed `include_mask` now anyway, so not worth adding it back in - generally I'd expect this function to be operating on lots of pins anyway, as opposed to the masked functions in `hardware_gpio` which I'd expect to be operating on a smaller set of pins

src/rp2_common/pico_low_power/include/pico/low_power.h
```diff
@@ -130,6 +130,32 @@ typedef enum {
 typedef void (*low_power_pstate_resume_func)(pstate_bitset_t *pstate);
 #endif
 
+/*! \brief  Set all pins to a low leakage state
+ *  \ingroup pico_low_power
+ *
+ * Disables pulls & inputs on the pads, and sets the IO function to GPIO_FUNC_SIO
```
lurch: I guess the `GPIO_FUNC_SIO` comment here needs to be updated to reflect one of the later commits?

will-v-pi:
```suggestion
 * Disables pulls & inputs on the pads, and disables the IO output
```

src/rp2_common/pico_low_power/include/pico/low_power.h
```diff
@@ -130,6 +130,32 @@ typedef enum {
...
+ * with all pins set to inputs. This results in the lowest leakage current.
+ * 
+ * Does not change the state of pins in the exclude_mask
+ *
+ * \param exclude_mask Mask of the pins to exclude from this
+ */
+void low_power_set_pins_low_leakage(uint32_t exclude_mask);
+
+/*! \brief  Set all pins to a low leakage state (64-bit version)
```
lurch:
```suggestion
/*! \brief  Set all pins to a low leakage state (64-bit mask version)
```
(just to clarify which part of it is "64-bit"?)

src/rp2_common/pico_low_power/include/pico/low_power.h
```diff
@@ -130,6 +130,32 @@ typedef enum {
 typedef void (*low_power_pstate_resume_func)(pstate_bitset_t *pstate);
 #endif
 
+/*! \brief  Set all pins to a low leakage state
+ *  \ingroup pico_low_power
+ *
+ * Disables pulls & inputs on the pads, and disables the IO output
+ * with all pins set to inputs. This results in the lowest leakage current.
+ * 
+ * Does not change the state of pins in the exclude_mask
```
lurch:
```suggestion
 * Does not change the state of pins in the exclude_mask.
```

test/pico_low_power_test/low_power_test_gpio.c
```diff
@@ -127,6 +127,7 @@ int main() {
     init_powman_ext_ctrl();
 
     gpio_put(SLEEP_MONITOR_PIN, 0);
+    low_power_set_pins_low_leakage(USED_PIN_MASK);
```
lurch: When reading isolated code like this (i.e. without referring to the doxygen), one might naively assume that it's the pins in `USED_PIN_MASK` that are being set to low-leakage? :thinking: 
```c
low_power_set_all_pins_low_leakage_except(USED_PIN_MASK);
```
is maybe a bit too wordy though :wink:

will-v-pi: The function naming is up for debate, but I think that might be a bit too long - maybe `low_power_set_all_pins_low_leakage` instead?

lurch: Yeah, that's probably better than `low_power_set_most_pins_low_leakage` :rofl:

kilograham (dismissed):

src/rp2_common/pico_low_power/low_power.c
```diff
@@ -224,6 +224,27 @@ int low_power_set_external_clock_source(uint src_hz, uint gpio_pin) {
 }
 #endif  // inline in header for other platforms
 
+void low_power_set_all_pins_low_leakage(uint32_t exclude_mask) {
```
kilograham: i think perhaps `low_power_set_pins_low_leakage_exclude_mask` whilst a mouthful is probably the most consistent with other naming

will-v-pi: I will change it

lurch: OTOH, it might sound like "low leakage exclude mask" is a thing, and that's what this function is setting? :wink:  But I'll stop bike-shedding and just leave it as-is.

kilograham (approved):

Issue #2988, opened 2026-06-06
sau_ns_hw does not exist


kilograham:
`#define sau_ns_hw ((armv8m_sau_hw_t *)(PPB_NONSEC_BASE + M33_SAU_CTRL_OFFSET))` in `hardware/structs/sau.h`

there is actually no NS SAU, so this is confusing/wrong (i was confused)

matiasilva:
Fixed in latest export

kilograham:
merged into develop

Issue #2920, opened 2026-05-01
some perf counter enum values missing from rp2350 hardware/structs/busctrl.h


kilograham:
```
typedef enum bus_ctrl_perf_counter_rp2350 {
    arbiter_rom_perf_event_access = 19,
    arbiter_rom_perf_event_access_contested = 18,
    arbiter_xip_main_perf_event_access = 17,
    arbiter_xip_main_perf_event_access_contested = 16,
    arbiter_sram0_perf_event_access = 15,
    arbiter_sram0_perf_event_access_contested = 14,
    arbiter_sram1_perf_event_access = 13,
    arbiter_sram1_perf_event_access_contested = 12,
    arbiter_sram2_perf_event_access = 11,
    arbiter_sram2_perf_event_access_contested = 10,
    arbiter_sram3_perf_event_access = 9,
    arbiter_sram3_perf_event_access_contested = 8,
    arbiter_sram4_perf_event_access = 7,
    arbiter_sram4_perf_event_access_contested = 6,
    arbiter_sram5_perf_event_access = 5,
    arbiter_sram5_perf_event_access_contested = 4,
    arbiter_fastperi_perf_event_access = 3,
    arbiter_fastperi_perf_event_access_contested = 2,
    arbiter_apb_perf_event_access = 1,
    arbiter_apb_perf_event_access_contested = 0
} bus_ctrl_perf_counter_t;
```

there should be a lot more; this matches RP2040 I think

lurch:
Looks like https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2350/hardware_regs/include/hardware/regs/busctrl.h#L98 contains the full list of RP2350 values.

matiasilva:
The numbering on the two are completely different. Names different too. How do we reconcile this?

lurch:
It _looks_ like the current `bus_ctrl_perf_counter_rp2350` is actually completely wrong, as it's currently just the same as `bus_ctrl_perf_counter_rp2040` https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2040/hardware_structs/include/hardware/structs/busctrl.h#L28 ?

Issue #2842, opened 2026-02-27
Possible typo in rp2350/hardware/regs/powman.h


Greg-Hawkridge:
https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2350/hardware_regs/include/hardware/regs/powman.h#L496-L513

POWMAN_CHIP_RESET_HAD_WATCHDOG_RESET_RSM should possibly be POWMAN_CHIP_RESET_HAD_WATCHDOG_RESET_***P***SM to match rp2350 datasheet.

lurch:
> POWMAN_CHIP_RESET_HAD_WATCHDOG_RESET_RSM should possibly be POWMAN_CHIP_RESET_HAD_WATCHDOG_RESET__**P**_SM to match rp2350 datasheet.

Well _that's_ super-weird! The register-headers in pico-sdk and the register-tables in the RP2350 datasheet are all auto-generated from the same internal metadata files, so I've no idea how there could be a typo in one but not the other. ping @matiasilva

matiasilva:
A fix has been deployed internally, thank you!

Issue #2380, opened 2025-03-28
More IRQs missing from RP2350.svd


lurch:
Similar to #1898 , but slightly different...

I recently noticed that Table 94 in https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf (on page 83) lists IRQ numbers 40 and 41 as `PROC0_IRQ_CTI` and `PROC1_IRQ_CTI`, but these appear to be missing from the `RP2350.svd`.

matiasilva:
The CTI allows chip debug. The debug hardware is not bus addressable, so I'm not sure it belongs in the SVD? We never added the RISC-V registers to the SVD, for example.

matiasilva:
I'll rename the existing virtual peripheral and add all IRQs there

matiasilva:
This is now generating internally and ready for SDK 2.2

kilograham:
closing for tracking purposes as it will be in 2.2.0 even though it isn't in develop yet

matiasilva:
@kilograham  Before 2.2:

* PROC0_IRQ_CTI and PROC1_IRQ_CTI not in SVD
* SPARE_IRQ_<n> under a virtual (baseAddress = 0) peripheral called SPARE_IRQ

After 2.2:

* PROC0_IRQ_CTI, PROC1_IRQ_CTI and SPARE_IRQ_<n> all under one virtual peripheral called VIRTUAL

_Edit by lurch:_ These changes didn't appear in Pico-SDK 2.2.0, but they **will** be appearing in Pico-SDK 2.3.0

PR #3050, opened 2026-06-30
Update to latest versions of the hardware headers

kilograham:


kilograham:
fixes #2988
fixes #2380
fixes #1466

EDIT: also fixes #2842
EDIT: also fixes #2920

src/rp2040/hardware_regs/RP2040.svd
```diff
@@ -50915,7 +50863,7 @@ SPDX-License-Identifier: BSD-3-Clause
                     </field>
                     <field>
                         <name>FREQ_RANGE</name>
-                        <description>Frequency range. An invalid setting will retain the previous value. The actual value being used can be read from STATUS_FREQ_RANGE. This resets to 0xAA0 and cannot be changed.</description>
+                        <description>Frequency range. This resets to 0xAA0 and cannot be changed.</description>
```
lurch, line 50866: If it "cannot be changed", should it have `<access>read-only</access>` ? Or is it more "should not be changed"?

src/rp2350/hardware_regs/include/hardware/regs/m33.h
```diff
@@ -3521,23 +3521,17 @@
 #define M33_AIRCR_PRIGROUP_ACCESS "RW"
 // -----------------------------------------------------------------------------
 // Field       : M33_AIRCR_SYSRESETREQS
-// Description : System reset request, Secure state only.
-//               0	SYSRESETREQ functionality is available to both Security
-//               states.
-//               1 SYSRESETREQ functionality is only available to Secure state.
+// Description : This resets only the core on which SYSRESETREQ is asserted, and
```
lurch: This description of `M33_AIRCR_SYSRESETREQS` is now identical to the description for `M33_AIRCR_SYSRESETREQ`, whereas previously this said "Secure state only".
I guess if this **is** Secure-only, that's still useful info to know?

src/rp2350/hardware_regs/include/hardware/regs/pio.h
```diff
@@ -461,7 +461,8 @@
 // =============================================================================
 // Register    : PIO_DBG_PADOUT
 // Description : Read to sample the pad output values PIO is currently driving
-//               to the GPIOs.
+//               to the GPIOs. On RP2040 there are 30 GPIOs, so the two most
+//               significant bits are hardwired to 0.
```
lurch, line 465: We probably don't need the RP2040-specifc comment in `src/rp2350/hardware_regs/include/hardware/regs/pio.h` ? :wink: 
But if the same source-file is generating the PIO headers for both RP2040 and RP2350 then perhaps this isn't worth special-casing.

src/rp2350/hardware_regs/include/hardware/regs/sysinfo.h
```diff
@@ -17,24 +17,24 @@
 // Description : JEDEC JEP-106 compliant chip identifier.
 #define SYSINFO_CHIP_ID_OFFSET _u(0x00000000)
 #define SYSINFO_CHIP_ID_BITS   _u(0xffffffff)
-#define SYSINFO_CHIP_ID_RESET  _u(0x00000001)
+#define SYSINFO_CHIP_ID_RESET  _u(0x30004927)
```
lurch, line 20: The previous version of this PR included `SYSINFO_CHIP_ID_RESET` values for both RP2040 and RP2350, but this new PR doesn't include the `SYSINFO_CHIP_ID_RESET` value for RP2040?

src/rp2350/hardware_structs/include/hardware/structs/scb.h
```diff
@@ -159,49 +159,56 @@ typedef struct {
...
-    io_ro_32 id_mmfr[4];
- 
+    io_ro_32 id_mmfr0;
+
+    uint32_t _pad1;
+
+    // (Description copied from array index 0 register M33_ID_MMFR2 applies similarly to other array indexes)
+    _REG_(M33_ID_MMFR2_OFFSET) // M33_ID_MMFR2
+    // Provides information about the implemented memory model and memory management support
+    // 0x0f000000 [27:24] WFISTALL     (0x1) Indicates the support for Wait For Interrupt (WFI) stalling
```
lurch, line 209: The RP2350 datasheet shows that the `ID_MMFR3` register has a different field-layout to the `ID_MMFR2` register. so _maybe_ that means that the "(Description copied from array index 0 register M33_ID_MMFR2 applies similarly to other array indexes)" isn't really applicable here?

lurch:
Bit weird that the Copyright year has jumped forward in some files, and backwards in other files, but meh :wink:

src/rp2040/hardware_structs/include/hardware/structs/sysinfo.h
```diff
@@ -1,7 +1,7 @@
 // THIS HEADER FILE IS AUTOMATICALLY GENERATED -- DO NOT EDIT
```
lurch, line 1: I've no idea which version is "correct", but in #2905 the size of the dummy padding array changed in this file (and Graham commented "because it was wrong before"), but it has remained unchanged in this PR? :eyes:

src/rp2040/hardware_regs/include/hardware/regs/sysinfo.h
```diff
@@ -1,7 +1,7 @@
 // THIS HEADER FILE IS AUTOMATICALLY GENERATED -- DO NOT EDIT
```
lurch, line 1: In #2905 the `SYSINFO_GITREF_RP2040_OFFSET` value changed from `0x00000010` to `0x00000040`, but it isn't changing in this PR?

lurch:
Hmmm, I wonder if the updates to the RP2040 headers have gone wrong somehow? In both `RP2040.svd` and `RP2350.svd` the "Warning: stop the PLLs before selecting dormant mode" text has changed to "WARNING: stop the PLLs before selecting dormant mode", and that change also appears in `src/rp2350/hardware_regs/include/hardware/regs/xosc.h` but it _doesn't_ appear in `src/rp2040/hardware_regs/include/hardware/regs/xosc.h` ?  (If my supposition is correct, this also might explain why the Copyright year has gone backwards in some files :wink: )

matiasilva:
> Hmmm, I wonder if the updates to the RP2040 headers have gone wrong somehow? In both `RP2040.svd` and `RP2350.svd` the "Warning: stop the PLLs before selecting dormant mode" text has changed to "WARNING: stop the PLLs before selecting dormant mode", and that change also appears in `src/rp2350/hardware_regs/include/hardware/regs/xosc.h` but it _doesn't_ appear in `src/rp2040/hardware_regs/include/hardware/regs/xosc.h` ? (If my supposition is correct, this also might explain why the Copyright year has gone backwards in some files 😉 )

The years going back is per a decision made to preserve the original date. If anything has moved forward, that's wrong.

As you know, the text source for register descriptions is based on the hardware present in the chip. Since both chips are quite a few years apart, I'm not surprised some descriptions have been tweaked.

Edit: SVDs have the year moving forward, so I've fixed that in the latest release

src/rp2040/hardware_regs/include/hardware/regs/sysinfo.h
matiasilva: We need the change as the previous release was wrong, there are loads of internal only values in between.

src/rp2350/hardware_regs/include/hardware/regs/m33.h
matiasilva: Thanks, the regex used to patch this reg desc was too greedy. Fixed.

src/rp2350/hardware_structs/include/hardware/structs/scb.h
matiasilva: I was asked to coalesce these into a struct array, is the suggestion here to have each register as a separate member?

src/rp2040/hardware_structs/include/hardware/structs/sysinfo.h
matiasilva: Related to the below ticket, we want to be using the newest version of sysinfo.h

src/rp2350/hardware_regs/include/hardware/regs/sysinfo.h
matiasilva: Not sure what happened to this PR as the generated headers are correct

src/rp2350/hardware_structs/include/hardware/structs/scb.h
lurch: I'm not on your team so I don't know what you were asked for, but it seems _to me_ like it would only make sense to coalesce registers into a struct array, if the fields inside each of those registers has the same widths and offsets? (which isn't the case for `ID_MMFR2` and `ID_MMFR3`)
But this is only a minor issue, so I guess we can address it in the next SDK release.

src/rp2040/hardware_regs/include/hardware/regs/usb.h
```diff
@@ -1012,7 +1012,7 @@
 #define USB_SIE_STATUS_CONNECTED_BITS   _u(0x00010000)
 #define USB_SIE_STATUS_CONNECTED_MSB    _u(16)
 #define USB_SIE_STATUS_CONNECTED_LSB    _u(16)
-#define USB_SIE_STATUS_CONNECTED_ACCESS "RO"
+#define USB_SIE_STATUS_CONNECTED_ACCESS "WC"
```
lurch, line 1015: This changes `SIE_STATUS.CONNECTED` (as well as `SIE_STATUS.SPEED` and `SIE_STATUS.SUSPENDED`) from RO to WC. Is that correct @liamfraser , and if so should the same change also be happening to `src/rp2350/hardware_regs/include/hardware/regs/usb.h` ?

kilograham:
I'm going to merge though there is the one oustanding question above

kilograham:
actually someone should approve first!

will-v-pi (approved):

PR #3055, opened 2026-07-02
RISC-V: Test whether toolchain has both assembler and compiler support for Zcmp

liamfraser:
This solves the issue where GCC 14.1.0 claims it supports zcmp in the arch flags but does not actually support it when you compile it. This results in compile errors on release. See this multi-gcc exhaustive run for an example of the failure:

https://github.com/raspberrypi/pico-sdk/actions/runs/28577860556/job/84731730132

Note: multi-gcc works on all RISC-V compilers with this patch

### DIFF
--- a/cmake/preload/toolchains/pico_riscv_gcc.cmake
+++ b/cmake/preload/toolchains/pico_riscv_gcc.cmake
@@ -7,4 +7,18 @@ set(PICO_COMMON_LANG_FLAGS_LIST
         " -mcpu=hazard3-rp2350"
         " -march=rv32ima_zicsr_zifencei_zba_zbb_zbs_zbkb_zca_zcb_zcmp -mabi=ilp32"
         " -march=rv32imac_zicsr_zifencei_zba_zbb_zbs_zbkb -mabi=ilp32")
+
+# C file used to test the flags listed above. There needs to be an equal number
+# of entries in PICO_COMMON_LANG_FLAGS_LIST and PICO_COMMON_LANG_FLAGS_TEST_FILES.
+set(PICO_COMMON_LANG_FLAGS_TEST_FILES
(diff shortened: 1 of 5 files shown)


Wren6991 (dismissed):

Wren6991:
For context when we look back and wonder why this exists: there was a period where `riscv-gnu-toolchain` shipped a version of GCC with full Zcmp support, but a version of binutils which lacked support for some Zcmp instructions (I think just `cm.mvsa01` and `cm.mva01s`), so the assembler choked on the compiler output if you actually tried to compile anything significant.

https://github.com/riscv-collab/riscv-gnu-toolchain/pull/1472#issuecomment-2143520816

https://github.com/riscv-collab/riscv-gnu-toolchain/pull/1747

cmake/preload/toolchains/pico_riscv_gcc.cmake
```diff
@@ -7,4 +7,8 @@ set(PICO_COMMON_LANG_FLAGS_LIST
         " -mcpu=hazard3-rp2350"
         " -march=rv32ima_zicsr_zifencei_zba_zbb_zbs_zbkb_zca_zcb_zcmp -mabi=ilp32"
         " -march=rv32imac_zicsr_zifencei_zba_zbb_zbs_zbkb -mabi=ilp32")
+set(PICO_COMMON_LANG_FLAGS_TEST_FILES
+        "${CMAKE_CURRENT_LIST_DIR}/util/riscv_zcmp_test.c"
```
kilograham, line 14: seems to be here twice

kilograham: oh, ok, so the two lists must be in sync - needs a big comment to that effect then

kilograham (changes requested):

kilograham (approved):

Issue #2788, opened 2026-01-08
passing individual build flags to arm-none-eabi as well as to host-arch compiler


josch:
Hi,

I'm maintaining the pico-sdk package in Debian and we ship two packages: the pico-sdk source code (architecture independent) and the pioasm tool. In Debian we are passing several hardening flags and since these flags are incompatible with gcc-arm-none-eabi, I'm performing two builds: I build the main project itself with CFLAGS, CXXFLAGS, CPPFLAGS and LDFLAGS unset and I do another build of the `tools/pioasm`  directory to build the `pioasm` tool with all the default build flags set.

Is there an easy way (or could such a way be added) to, without installing it, first build `tools/pioasm` with full build flags and then let the second build of the full project re-use the produced pioasm tool without building it a second time?

By building pioasm essentially twice (once with hardened build flags and once without) it becomes tricky to analyze the build log and make sure that hardening flags were applied as desired and not accidentally skipped.

Alternatively, could there be an easy way to pass pioasm-specific build flags to the full build such that I could set empty CFLAGS to not confuse `gcc-arm-none-eabi` but then set another variable to pass the hardened CFLAGS just to build pioasm?

What do you think?

Thanks!

cheers, josch

will-v-pi:
> Is there an easy way (or could such a way be added) to, without installing it, first build tools/pioasm with full build flags and then let the second build of the full project re-use the produced pioasm tool without building it a second time?

Not without installing it, but the install can be to a specific directory rather than systemwide - if you build & install it passing `-DPIOASM_FLAT_INSTALL=1` and `-DCMAKE_INSTALL_PREFIX=/path/to/put/pioasm` to cmake (as the SDK does when it builds pioasm), you can then point future SDK builds at that installed pioasm by passing `-Dpioasm_DIR=/path/to/put/pioasm/pioasm` to the SDK cmake build

> Alternatively, could there be an easy way to pass pioasm-specific build flags to the full build such that I could set empty CFLAGS to not confuse gcc-arm-none-eabi but then set another variable to pass the hardened CFLAGS just to build pioasm?

Would something like d6e3ea84d6dd4532065f5d0b436d75d8b457d89a work for you? This would allow you to pass `-DPIOASM_C_FLAGS=...` etc to cmake, and those flags will be passed to the pioasm build (also added for picotool).

If you need that second fix then I will PR that change, but I think we'd prefer the first method if that works for you

will-v-pi:
[...]

PR #3059, opened 2026-07-03
Add ability to pass cflags straight to pioasm/picotool build

kilograham:
found this branch lurking


will-v-pi:
This was a fix for https://github.com/raspberrypi/pico-sdk/issues/2788 - but the decision there was that if you need to pass specific flags to pioasm, you should build & install it separately with those flags and then point the SDK build at it.

So this fix isn't necessary - the correct approach is to build & install pioasm and picotool separately, passing the correct flags to those builds, and then point the SDK build at those installs.

Issue #3060, opened 2026-07-04
SDK 2.3.0 - Missing or changed defines


Memotech-Bill:
````
/home/pi/pico/PicoBB/src/pico/pico_boot.c:13:36: error: 'PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE' undeclared (first use in this function); did you mean 'PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE'?
   13 |     if (p_line_coding->bit_rate == PICO_STDIO_USB_RESET_MAGIC_BAUD_RATE)
      |                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                    PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE
/home/pi/pico/PicoBB/src/pico/pico_boot.c:13:36: note: each undeclared identifier is reported only once for each function it appears in
/home/pi/pico/PicoBB/src/pico/pico_boot.c:20:35: error: 'PICO_STDIO_USB_RESET_BOOTSEL_INTERFACE_DISABLE_MASK' undeclared (first use in this function)
   20 |         reset_usb_boot(gpio_mask, PICO_STDIO_USB_RESET_BOOTSEL_INTERFACE_DISABLE_MASK);
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/pi/pico/PicoBB/src/pico/pico_boot.c:24:31: error: 'PICO_STDIO_USB_RESET_RESET_TO_FLASH_DELAY_MS' undeclared (first use in this function); did you mean 'PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS'?
   24 |         watchdog_reboot(0, 0, PICO_STDIO_USB_RESET_RESET_TO_FLASH_DELAY_MS);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                               PICO_STDIO_USB_POST_CONNECT_WAIT_DELAY_MS
````

The above used to compile OK with v2.20 of the SDK. I cannot remember which version of the SDK I first introduced this code.

The 2.3.0 release documentation references `pico-usb-reset`, but the link does not take me to any such documentation.

Memotech-Bill:
It appears that at least some of the definitions have moved into "pico/pico-usb-reset.h".

[...]

PR #3062, opened 2026-07-06
Add backwards compatible defines for the USB reset interface

will-v-pi:
Added into `stdio_usb.h`, as that is where the old defines were

Guarded behind `PICO_STDIO_USB_BACKWARDS_COMPATIBLE_RESET_DEFINES` so it can be disabled in future

Fixes https://github.com/raspberrypi/pico-sdk/issues/3060


src/rp2_common/pico_stdio_usb/include/pico/stdio_usb.h
```diff
@@ -148,4 +153,18 @@ void stdio_usb_call_chars_available_callback(void);
 }
 #endif
 
+#if PICO_STDIO_USB_BACKWARDS_COMPATIBLE_RESET_DEFINES
+#include "pico/usb_reset_config.h"
+#define PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE PICO_ENABLE_USB_RESET_VIA_BAUD_RATE
```
lurch, line 158: Do each of these need to be wrapped in an `#ifndef`? E.g.
```c
#ifndef PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE
#define PICO_STDIO_USB_ENABLE_RESET_VIA_BAUD_RATE PICO_ENABLE_USB_RESET_VIA_BAUD_RATE
#endif
```

will-v-pi: No, that sort of wrapping should already be handled in `pico/usb_reset_config.h`

will-v-pi: Oh actually, yes, they will need some sort of wrapping to prevent duplicate definitions (although the duplicate definitions would be the same value, as that is handled in `pico/usb_reset_config.h`)

will-v-pi:
It was suggested in https://github.com/raspberrypi/pico-sdk/issues/3060#issuecomment-4892485042 that this probably isn't worth it, as anyone using the old defines will need to update their project anyway to work with the new defines, so putting this in draft and can be closed if others agree

lurch:
> as anyone using the old defines will need to update their project anyway to work with the new defines

Unless people skip directly from < 2.3.0 to >= 2.3.1 ? :shrug: 
(OTOH, if they've not updated to SDK 2.3.0 yet, then perhaps it's unlikely that they'll also update to SDK 2.3.1 or later.)

Issue #3064, opened 2026-07-08
Handle GOT sections gracefully


Prabhuk:
If any of the libraries linked against is built as position independent code, weak undefined symbols and comdat symbols may have GOT relocations in ARM targets. We observed this in the LLVM runtime libraries (libc.a/libm.a) but this behavior is not specific to Clang/LLVM toolchains. Handle GOT sections gracefully to prevent possible broken behavior.

kilograham:
ok, so what is the upshot - does it fail linking, or what? I'm not keen to include the `.got` sections in the linker script - we could discard them, or assert they are empty (are they?)

does adding `-fno-pic` or `fno-pie` to compile/link help?

Prabhuk:
Even if the final link of the binary has no PIC flags set, there might be valid GOT sections coming from libraries that are linked if they were built as PIC. In this scenario, we will not see a compile/link failure but we will end up dropping valid GOT sections leading to runtime failures.

Having said that, the push back against this patch is fair. We currently build our LLVM libc distributions in PIC mode by default. A recent change upstream https://github.com/llvm/llvm-project/pull/198577 caused GOT sections being introduced in our libm.a which prompted the proposed change. We are working with upstream to resolve this particular case through this patch https://github.com/llvm/llvm-project/pull/209660 which should make our current change here to place GOT sections in linker script irrelevant. 

I am still in favor of having the GOT sections handled in one way or the other given it is not illegal for them to be present in the final binary. Assuming we will get the GOT sections removed in our toolchain distribution, the other approach is to actually assert from linker script that no GOT sections exist in the input sections list. I'd prefer that GOT sections are not quietly ignored, if present, causing runtime crashes.

kilograham:
Yup - I was leaning towards an assertion but that doesn’t help obviously if we have toolchain versions that produce them

I wonder even if present if they actually work as intended - can u point me at something that produces them

Prabhuk:
The current version of the clang toolchain in pico-sdk does not produce GOT sections. I can make sure that we dont update the toolchain version to one which emits GOT sections.  

[...]

PR #3065, opened 2026-07-08
Handle GOT sections gracefully

Prabhuk:
If any of the libraries linked against is built as position independent code, weak undefined symbols and comdat symbols may have GOT relocations. We observed this in the LLVM runtime libraries (libc.a/libm.a) but this behavior is not specific to Clang/LLVM toolchains. Including the GOT sections into the data section to handle them gracefully.

Fixes https://github.com/raspberrypi/pico-sdk/issues/3064

### DIFF
--- a/src/rp2_common/pico_standard_link/script_include/section_default_data.incl
+++ b/src/rp2_common/pico_standard_link/script_include/section_default_data.incl
@@ -30,7 +30,7 @@ SECTIONS
         *(.data*)
         *(.sdata*)
 
-	  /* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
+	    /* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
         *(.got)
         *(.got.*)
 
--- a/src/rp2_common/pico_standard_link/script_include/section_default_data.incl
+++ b/src/rp2_common/pico_standard_link/script_include/section_default_data.incl
@@ -30,6 +30,10 @@ SECTIONS
         *(.data*)
         *(.sdata*)
 
+	/* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
+	*(.got)
+        *(.got.*)
+
         . = ALIGN(4);
         *(.after_data.*)
         . = ALIGN(4);
--- a/src/rp2_common/pico_standard_link/script_include/section_default_data.incl
+++ b/src/rp2_common/pico_standard_link/script_include/section_default_data.incl
@@ -31,7 +31,7 @@ SECTIONS
         *(.sdata*)
 
 	/* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
-	*(.got)
+        *(.got)
         *(.got.*)
 
         . = ALIGN(4);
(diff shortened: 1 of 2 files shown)


Prabhuk:
CC: @petrhosek

Prabhuk:
Also tested the gcc build on device. These were the tests I built and ran. Posting the build command for reference
```
cmake --build build-gcc --target fixed_bitset_test hardware_irq_test hardware_pwm_test hardware_sync_spin_lock_test pico_async_context_test pico_divider_test pico_float_test pico_double_test pico_sem_test pico_sha256_test pico_stdlib_test pico_sync_test_sw pico_thread_local_test pico_time_test pico_stdio_test_uart cmsis_test --parallel
```

src/rp2_common/pico_standard_link/script_include/section_default_data.incl
```diff
@@ -30,6 +30,10 @@ SECTIONS
         *(.data*)
         *(.sdata*)
 
+	/* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
```
petrhosek:
```suggestion
	  /* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
```

src/rp2_common/pico_standard_link/script_include/section_default_data.incl
```diff
@@ -30,6 +30,10 @@ SECTIONS
         *(.data*)
         *(.sdata*)
 
+	  /* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
```
petrhosek:
```suggestion
	    /* if linked against non PIC archives, ARM targets may generate dynamic data relocations via got sections */
```

petrhosek (approved):

kilograham:
note, i'm not sure this is the right solution - discussion in #3064

Wren6991:
Noting this is read-write (pulled into `.data`) -- is the GOT actually being updated at runtime?

* If so, that's slightly alarming for an embedded target, but doesn't make the patch incorrect.
* If not, should this be in `.rodata` so we don't allocate RAM for it?

Wren6991:
Based on reading the upstream LLVM threads, it sounds like a GOT indirection is conservatively emitted, but then the relocation on the GOT entry itself is resolved statically during final link. Is that correct?

kilograham:
note, i think we should still fix this with read-only data probably, but i wonder if these GOT sections are related to tls, and would be fixed via  `-ftls-model=local-exec` anwyay (which we should also add)

kilograham:
note peter's claude analysis elsewhere was concerned about .got stuff being in flash when code is in RAM for flash only code... this is a possible wrinkle, but arguably no worse than jump tables (but it would be nice to know where the .got stuff is coming from)

might be a vote for leaving in RAM for now depending

Wren6991:
Meaning #3112? It feels like you can end up with either orientation of code/GOT in RAM/flash so you have that problem either way.

I looked through the relevant `libc.a` and `libm.a` for the https://github.com/arm/arm-toolchain/releases/tag/release-22.1.0-ATfE release that Peter mentioned. Disassembled with relocations, I don't see any `R_ARM_GOT*` relocations. Am I looking in the wrong place?

Wren6991 (approved):
Fixes #3112 and the sections seem to only be a few bytes, so they're fine in RAM

PR #3071, opened 2026-07-14
Pass --platform to picotool

will-v-pi:
The --platform argument was added in picotool version 2.2.0-a4, so should be fine to pass now that picotool version 2.3.0 is required by the SDK

Fixes raspberrypi/picotool#339

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -633,6 +633,15 @@ function(pico_add_uf2_output TARGET)
                 set(picotool_family ${PICO_PLATFORM})
             endif()
         endif()
+        get_target_property(picotool_platform ${TARGET} PICOTOOL_UF2_PLATFORM)
+        if (NOT picotool_platform)
+            if (PICOTOOL_DEFAULT_PLATFORM)
+                set(picotool_platform ${PICOTOOL_DEFAULT_PLATFORM})
+            else()
+                # Set picotool_platform to the part of PICO_PLATFORM before the first -
+                string(REGEX MATCH "^[^-]+" picotool_platform ${PICO_PLATFORM})
+            endif()
+        endif()
         add_custom_command(TARGET ${TARGET} POST_BUILD
             COMMAND picotool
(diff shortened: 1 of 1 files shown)


kilograham (approved):

tools/CMakeLists.txt
```diff
@@ -641,6 +650,7 @@ function(pico_add_uf2_output TARGET)
                 $<TARGET_FILE:${TARGET}>
                 ${output_path}$<IF:$<BOOL:$<TARGET_PROPERTY:${TARGET},OUTPUT_NAME>>,$<TARGET_PROPERTY:${TARGET},OUTPUT_NAME>,$<TARGET_PROPERTY:${TARGET},NAME>>.uf2
                 --family ${picotool_family}
+                --platform ${picotool_platform}
```
lurch, line 653: Asking out of curiosity: What happens if `picotool` is given a `--platform` that isn't "compatible" with the `--family`?

will-v-pi: It works fine - `--platform` is used to verify that all the code is in valid memory regions, `--family` specifies the UF2 family. So `--platform rp2040 --family rp2350-arm-s` would give you a UF2 with an RP2350 family ID that definitely fits in the RP2040 memory regions - although that would usually fail for RP2350 binaries, because they use 0x20080000-0x20082000 for the stack, which is outside the RP2040 memory address range

lurch: Ah, and I guess that's useful when creating "universal" binaries?

will-v-pi: Not really - that happens after this stage, as you take the UF2 files with different family IDs and combine them (at least for our universal binaries). Maybe it would be useful for universal binaries with proper runtime selection between chips, but that sort of thing is definitely not currently supported by the SDK (everything is compile-time selection with `#if PICO_RP2040` etc).

I don't see any particular use currently for passing "incompatible" `--family` and `--platform` flags, but `picotool` will behave as expected if you do

PR #3075, opened 2026-07-15
Don't place libc.a in RAM

### LINKED ISSUE #3070: SDK 2.3.0  default RAM memory consumption increase
After updating to the Pico SDK 2.3.0, I am unable to build my project due to over ~40KB increase in RAM consumption. Does anyone know where I can find documentation on these changes, or the best way to identify what is causing this increase?
```
 cd /home/shtirlic/projects/8086/picocalc_x86/build && /home/shtirlic/.local/bin/picotool uf2 convert --quiet /home/shtirlic/projects/8086/picocalc_x86/build/picocalc_x86.elf picocalc_x86.uf2 --family rp2350-riscv --abs-block
/opt/riscv/bin/../lib/gcc/riscv-none-elf/15.2.0/../../../../riscv-none-elf/bin/ld: picocalc_x86.elf section `.bss' will not fit in region `RAM'
/opt/riscv/bin/../lib/gcc/riscv-none-elf/15.2.0/../../../../riscv-none-elf/bin/ld: section .scratch_x VMA [20080000,2008057f] overlaps section .bss VMA [200159d8,2008a533]
/opt/riscv/bin/../lib/gcc/riscv-none-elf/15.2.0/../../../../riscv-none-elf/bin/ld: region `RAM' overflowed by 42292 bytes
Memory region         Used Size  Region Size  %age Used
           FLASH:      374352 B         4 MB      8.93%
           PSRAM:           0 B          0 B
         XIP_RAM:           0 B        16 KB      0.00%
             RAM:      566580 B       512 KB    108.07%
       SCRATCH_X:        3456 B         4 KB     84.38%
       SCRATCH_Y:          32 B         4 KB      0.78%
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
build finished with error(s).

```

will-v-pi:
As part of a bugfix to place `mem` functions in RAM as intended (https://github.com/raspberrypi/pico-sdk/commit/dc3a2e861adebdff1b7009f950580f1f719366d1), now all of `libc.a` gets placed in RAM too which takes up a lot of extra space. This fixes that to only place the `mem` functions in RAM, and leave `libc.a` in Flash.

It looks like this may have been missed because on some (possibly most) toolchains (e.g. Debian arm-none-eabi-gcc 14.2.1) if there are any `-g` flags passed, the toolchain uses `libg.a` instead of `libc.a`, and `PICO_DEBUG_INFO_IN_RELEASE` (which defaults to 1) passes `-g`. Therefore there was no `libc.a` to place in RAM, so the size increase doesn't occur.

Reproduced the issue from #3070 with `PICO_DEBUG_INFO_IN_RELEASE=0` and the `kitchen_sink_printf_compiler` test, which shows the increase in RAM usage due to placing `libc.a` in Flash

Fixes #3070

### DIFF
--- a/src/rp2_common/pico_standard_link/script_include/default_rodata_excludes.incl
+++ b/src/rp2_common/pico_standard_link/script_include/default_rodata_excludes.incl
@@ -1,1 +1,1 @@
-*(EXCLUDE_FILE(*libgcc.a: *lib*_a-mem*.o *libm.a:) .rodata*)
+*(EXCLUDE_FILE(*libgcc.a: *libm.a:) .rodata*)
--- a/src/rp2_common/pico_standard_link/script_include/default_rodata_excludes.incl
+++ b/src/rp2_common/pico_standard_link/script_include/default_rodata_excludes.incl
@@ -1,1 +1,1 @@
-*(EXCLUDE_FILE(*libgcc.a: *libc.a: *lib*_a-mem*.o *libm.a:) .rodata*)
+*(EXCLUDE_FILE(*libgcc.a: *lib*_a-mem*.o *libm.a:) .rodata*)
--- a/src/rp2_common/pico_standard_link/script_include/default_text_excludes.incl
+++ b/src/rp2_common/pico_standard_link/script_include/default_text_excludes.incl
@@ -1,1 +1,1 @@
-*(EXCLUDE_FILE(*libgcc.a: *libc.a: *lib*_a-mem*.o *libm.a:) .text*)
+*(EXCLUDE_FILE(*libgcc.a: *lib*_a-mem*.o *libm.a:) .text*)
--- a/src/rp2_common/pico_standard_link/script_include/default_text_excludes.incl
+++ b/src/rp2_common/pico_standard_link/script_include/default_text_excludes.incl
@@ -1,1 +1,1 @@
-*(EXCLUDE_FILE(*libgcc.a: *lib*_a-mem*.o *libm.a:) .text*)
+*(EXCLUDE_FILE(*libgcc.a: *libm.a:) .text*)


kilograham:
So what you are saying was that with `-g` the functions were never placed in RAM?

Even an 800 byte size difference is something not to be scoffed at - some existing projects may be very space constrained

will-v-pi:
> So what you are saying was that with `-g` the functions were never placed in RAM?
> 
> Even an 800 byte size difference is something not to be scoffed at - some existing projects may be very space constrained

I think that they were previously never placed in RAM with or without `-g` (at least not with my compiler/newlib), because the original `*libc.a:*lib_a-mem*.o` pattern did not catch `*libc.a:*libc_a-mem*.o` or `*libg.a:*libc_a-mem*.o`

When I made this change I was going off the comment which said:
```
        /* TODO revisit this now memset/memcpy/float in ROM */
        /* bit of a hack right now to exclude all floating point and time critical (e.g. memset, memcpy) code from
         * FLASH ... we will include any thing excluded here in .data below by default */
```
but the code wasn't actually including memset/memcpy as they are under `libc_a-mem*.o` rather than `lib_a-mem*.o`, so I fixed it so that memset/memcpy etc do go into RAM but accidentally included all of `libc.a` in RAM too

src/rp2_common/pico_standard_link/script_include/default_rodata_excludes.incl
```diff
@@ -1 +1 @@
-*(EXCLUDE_FILE(*libgcc.a: *libc.a: *lib*_a-mem*.o *libm.a:) .rodata*)
\ No newline at end of file
+*(EXCLUDE_FILE(*libgcc.a: *lib*_a-mem*.o *libm.a:) .rodata*)
```
will-v-pi:
```suggestion
*(EXCLUDE_FILE(*libgcc.a: *libm.a:) .rodata*)
```

src/rp2_common/pico_standard_link/script_include/default_text_excludes.incl
```diff
@@ -1 +1 @@
-*(EXCLUDE_FILE(*libgcc.a: *libc.a: *lib*_a-mem*.o *libm.a:) .text*)
\ No newline at end of file
+*(EXCLUDE_FILE(*libgcc.a: *lib*_a-mem*.o *libm.a:) .text*)
```
will-v-pi:
```suggestion
*(EXCLUDE_FILE(*libgcc.a: *libm.a:) .text*)
```

kilograham (approved):

kilograham:
@will-v-pi will you fix the comment too when u r back!

Issue #3073, opened 2026-07-15
SDK 2.3.0 risc-v extensions detection


shtirlic:
In this commit https://github.com/raspberrypi/pico-sdk/commit/b9e3417f4539949334de2fe580360f10b0f8d333
we now have tests for compressed extensions `zca_zcb_zcmp`
But commenting # set(PICO_COMPILER pico_riscv_gcc_zcb_zcmp ) produces ASM without `cm.` instructions

Using https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack/releases/tag/v15.2.0-1 as toolchain  

```
set(PICO_BOARD pico2)
set(PICO_PLATFORM rp2350-riscv)
set(PICO_USE_STACK_GUARDS ON)
# set(PICO_DEOPTIMIZED_DEBUG ON)
# set(PICO_COMPILER pico_riscv_gcc_zcb_zcmp )
# set(PICO_GCC_TRIPLE "riscv-none-elf")
set(PICO_TOOLCHAIN_PATH /opt/riscv/)
```
with 
```
set(PICO_COMMON_LANG_FLAGS " -march=rv32ima_zicsr_zifencei_zba_zbb_zbs_zbkb_zca_zcb_zcmp  -mabi=ilp32")
```

I see the the `cm.mva01s	s1,s0` in dis file.

lurch:
#3076 should fix this...

kilograham:
fix merged into develop

PR #3076, opened 2026-07-15
Don't use s0 in Risc-V zcmp test compile

will-v-pi:
This file currently fails to compile with any compilers, resulting in `pico_riscv_gcc` always falling through to the non-zcmp arch string

The fix is to change the registers used to `s1` and `s2` instead

### DIFF
--- a/cmake/preload/toolchains/util/riscv_zcmp_test.c
+++ b/cmake/preload/toolchains/util/riscv_zcmp_test.c
@@ -1,4 +1,5 @@
 /* Test that the toolchain can assemble Zcmp instructions */
 void _start(void) {
-    asm volatile ("cm.mvsa01 s0, s1" : : : "s0", "s1");
+    // Cannot use s0 as it is also fp, so use s1 & s2 instead
+    asm volatile ("cm.mvsa01 s1, s2" : : : "s1", "s2");
 }


kilograham:
ha; i was just making the same fix - but with said compiler i still didn't get any cm.push/pop etc.

Wren6991 (approved):

Wren6991:
> but with said compiler i still didn't get any cm.push/pop etc.

I think the Zcmp codegen itself has regressed in GCC. It looks like prolog/epilog scheduling can cause the push/pop pattern matching to fail (with no performance benefit from the scheduling). You should still get them, but they're rarer than they should be.

Do you still get `cm.mva01s` and `cm.mvsa01`?

lurch:
Is this also the fix for #3073 or is that a similar-but-different problem?

will-v-pi:
> Is this also the fix for #3073 or is that a similar-but-different problem?

Yes, that is this exact problem

shtirlic:
When I build with `Release` variant there are only 75  `cm.mva01s` instructions,  but with `MinSizeRel`  i got around ~600  `cm.` instructions.

will-v-pi:
The Multi GCC run (https://github.com/raspberrypi/pico-sdk/actions/runs/29430965108/job/87405548355) has passed the RISC-V GCC 14.1.0 builds, so I think that means this is tested now and can be merged

lurch:
Is it "expected" for the CI to fail on the Pico W builds for GCC 6 and GCC 7? https://github.com/raspberrypi/pico-sdk/actions/runs/29430965108/job/87405548355

will-v-pi:
> Is it "expected" for the CI to fail on the Pico W builds for GCC 6 and GCC 7? https://github.com/raspberrypi/pico-sdk/actions/runs/29430965108/job/87405548355

Yes, I think that's fixed in some way in multi-gcc-develop branch - ping @liamfraser

Wren6991:
> I think the Zcmp codegen itself has regressed in GCC. It looks like prolog/epilog scheduling can cause the push/pop pattern matching to fail (with no performance benefit from the scheduling). You should still get them, but they're rarer than they should be.

Just to close the loop on this because I was completely wrong about the cause:

* You don't get any `cm.push`/`cm.pop*` at `-O3` because GCC separately shrink-wraps each save, which inhibits the Zcmp grouping. You can opt out of this with `-fno-shrink-wrap-separate` which gives you a smaller binary at some minor perf cost.

* Separately there is a bug/limitation where even at `-Os` GCC refuses to emit Zcmp prologs/epilogs in functions with frame pointers, even when you have `-fomit-frame-pointer` (_which does not in fact always omit the frame pointer_). It's because there's conflation between the function's own `fp` used for VLAs etc, and the saved outgoing `fp` used for DWARF-less backtraces, which needs to be at a fixed offset. I think this is fixable in upstream GCC, and it's only a minority of functions.

PR #3079, opened 2026-07-19
More sleep fixes

### LINKED ISSUE #3078: `sleep_us()`/`sleep_until()` hangs forever on RP2350 (SDK 2.3.0): `sleep_until()` waits indefinitely with hardware alarm unarmed
## Environment

- pico-sdk: 2.3.0
- Target: RP2350 (Pico 2)
- Toolchain: `arm-none-eabi-gcc (Arm GNU Toolchain 15.2.Rel1 (Build arm-15.86)) 15.2.1 20251203`
- Host OS: macOS (Homebrew)
- Build system: CMake + Ninja
- Reproduces with both:
  - `-O3 -DNDEBUG` (`CMAKE_BUILD_TYPE=Release`)
  - `-Og` (`CMAKE_BUILD_TYPE=Debug`)
- Reproduces both:
  - with a debugger attached
  - standalone after flashing, with no debugger attached

Debugging setup used during investigation (not required to reproduce):

- OpenOCD: `raspberrypi/openocd`, branch `rpi-common`, commit `acff23ffd100479d6481a1155123774b79a990b9`
  (`0.12.0+dev-00004-gacff23ffd`)
- GDB: GNU gdb 17.2

Comparison test:

- RP2040 (Raspberry Pi Pico): could not reproduce with the same code and SDK version
- RP2350 (Pico 2): reproduces as described above

## Minimal repro

`CMakeLists.txt`

```cmake
cmake_minimum_required(VERSION 3.13)

include($ENV{PICO_SDK_PATH}/external/pico_sdk_import.cmake)

project(sleep-bug)

pico_sdk_init()

add_executable(main main.c)
target_link_libraries(main pico_stdlib)

pico_enable_stdio_usb(main 0)
pico_enable_stdio_uart(main 1)

pico_add_extra_outputs(main)
```

`main.c`

```c
#include <stdio.h>

#include <pico/stdio.h>
#include <pico/time.h>

int main() {
    stdio_init_all();
    printf("start\n");

    for (int i = 0; i < 10; ++i) {
        sleep_us(10);
    }

    printf("end\n");
    return 0;
}
```

## Steps to reproduce

1. Build for RP2350 (`PICO_BOARD=pico2`)
2. Flash to the board
3. Observe UART output

## Expected behavior

`start` is printed, followed shortly by `end`.

## Actual behavior

`start` is printed, but `end` never appears.

When inspected with a debugger, execution is stuck in the `sleep_us()` → `sleep_until()` path and does not return from the sleep call.

## Reproduction characteristics

The issue is sensitive to the number of consecutive `sleep_us()` calls.

Observed on my setup:

- 1 call: could not reproduce
- 3 consecutive calls: reproduces roughly 50% of runs
- 10 consecutive calls: reproduces nearly every run

The loop in the reproducer is only intended to make the failure more reproducible.

The issue is not specific to the `for` loop itself. Replacing the loop with multiple consecutive `sleep_us()` calls produces the same behavior.

## Debugging details

Backtrace obtained via OpenOCD + GDB (`-Og` build):

```text
#0  sleep_until (t=<optimized out>) at /pico-sdk/src/common/pico_time/time.c:418
#1  0x10000ed8 in sleep_us (us=<optimized out>) at /pico-sdk/src/common/pico_time/include/pico/time.h:120
#2  0x100001f8 in main () at /sleep-bug/main.c:11
```

The PC is consistently observed around the `__wfe()` wait path reached from:

```c
if (add_alarm_at(t_before, sleep_until_callback, NULL, false) >= 0) {
    // able to add alarm for just before the time
    while (!time_reached(t_before)) {
        uint32_t save = spin_lock_blocking(sleep_notifier.spin_lock);
        lock_internal_spin_unlock_with_wait(&sleep_notifier, save);
    }
}
```

With `-O3`, more code is inlined, but repeated debugger interrupts still place execution immediately after a `wfe` instruction inside the same wait path.

## Register dump while hung (TIMER0 @ 0x400B0000)

```text
(gdb) p/x ((timer_hw_t*)0x400B0000)->alarm[3]
$1 = 0x528

(gdb) p/x ((timer_hw_t*)0x400B0000)->inte
$2 = 0x8

(gdb) p/x ((timer_hw_t*)0x400B0000)->ints
$3 = 0x0

(gdb) p/x ((timer_hw_t*)0x400B0000)->armed
$4 = 0x0

(gdb) p/x ((timer_hw_t*)0x400B0000)->timerawl
$5 = 0x49ace2
```

Observed state:

- `alarm[3]` (default alarm pool hardware alarm) contains `0x528` (1,320 us).
- `timerawl` contains `0x49ace2` (4,828,386 us), far past the programmed alarm value.
- `armed == 0x0`; alarm 3 is not armed.
- `inte == 0x8`; alarm 3 interrupts are enabled.
- `ints == 0x0`; no alarm interrupt is pending.
- `timerawl` was observed to continue incrementing across multiple reads, indicating that the timer itself is still running.

At the point where execution is stuck in `sleep_until()`, the timer counter continues to advance, the programmed alarm value lies far in the past, and the corresponding hardware alarm is not armed.

I have not yet determined how this state is reached.

## Additional notes

- Replacing `sleep_us()` with `busy_wait_us()` avoids the hang.
- The issue reproduces with and without a debugger attached.
- The issue reproduces in both Debug and Release builds.

## Question

Given the observed state:

- execution blocked in the `sleep_until()` wait path,
- timer counter still advancing,
- alarm value already in the past,
- no pending interrupt,
- hardware alarm not armed,

does this match any known RP2350 or pico-sdk alarm scheduling issue?

kilograham:
fixes #3078


test/pico_sync_test/pico_sync_test.c
```diff
@@ -31,36 +35,45 @@ int64_t notify_lock_with_flag(__unused alarm_id_t id, void *user_data) {
 static lock_core_t sleep_notifier;
 
 static int64_t sleep_until_callback(__unused alarm_id_t id, __unused void *user_data) {
-    uint32_t save = spin_lock_blocking(sleep_notifier.spin_lock);
-    lock_internal_spin_unlock_with_notify(&sleep_notifier, save);
+    __sev();
```
lurch, line 38: I don't claim to understand what this code is doing, but does the callback function now _always_ need to do a `__sev()`, or is that just an "implementation detail" of this particular test?

If the `__sev()` isn't required, maybe it's worth adding a separate test that uses a different callback function (which doesn't use `__sev()`), just to rule out any "side effect" behaviour? :shrug:

test/pico_time_test/CMakeLists.txt
```diff
@@ -3,10 +3,25 @@ if (NOT PICO_TIME_NO_ALARM_SUPPORT)
...
+        add_executable(pico_time_test_sw pico_time_test.c)
+        set_target_properties(pico_time_test_sw PROPERTIES PICO_TEST_TIMEOUT "20")
+        target_compile_definitions(pico_time_test_sw PRIVATE
+                PICO_TIME_DEFAULT_ALARM_POOL_MAX_TIMERS=250
+                PICO_USE_SW_SPIN_LOCKS=1
+        )
+        target_link_libraries(pico_time_test_sw PRIVATE pico_test)
+        if (PICO_RP2040)
+            target_link_libraries(pico_time_test_sw PRIVATE pico_aon_timer)
+        endif()
```
lurch: It looks like you've got a `if (PICO_RP2040)` inside a `if (NOT PICO_RP2040)`, so presumably this code will never be reached?

test/short_sleep_test/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+if (NOT PICO_TIME_NO_ALARM_SUPPORT)
+    add_executable(short_sleep_test short_sleep_test.c)
+    set_target_properties(short_sleep_test PROPERTIES PICO_TEST_TIMEOUT "20")
+    target_compile_definitions(short_sleep_test PRIVATE
+            PICO_TIME_SLEEP_OVERHEAD_ADJUST_US=0
```
lurch, line 5: It looks like you need to add this define to `test/short_sleep_test/BUILD.bazel` too? (Because the Bazel builds are currently failing CI)

test/pico_sync_test/pico_sync_test.c
lurch: Thanks Graham :smiley:

test/pico_time_test/CMakeLists.txt
kilograham: ahha

kilograham:
Summary of changes:

1. `sleep_until()` adds a timer callback, to notify itself to wake up. This was using a `spin_lock/spin_unlock_with_wait`, with the callback doing a `spin_lock/spin_unlock_with_notify`. This was, if I recall, done (vs just WFE and SEV) because these primitives are what are overridden by FreeRTOS, and the additional overhead of the spinlock is minimal. The problem on RP2350 with software spinlocks, is that the spin_lock/spin_unlock_with_wait wakes up immediately. The fix here is to use WFE/SEV unless the macros are overridden
2. Violation of "alarm always wakes up on or before active alarm" was broken if an alarm was added, and deleted again before the IRQ handler woke up. In this case the alarm was never added to the timer. This was fixed by re-ordering code in `alarm_pool_irq_handler` and using a function callback marker constant instead of changing the timeout, so that the timeout value was not lost
3. The timeout function markers (`repeating_timer_marker` and `deleted_timer_marker` were changed to be small immediate even values, to reduce code size, and so attempting to call them will trap)
4. The recent addition of a special check in `best_effort_wfe_or_timeout`  for last_added == timeout_timestamp meant we always did a WFE in that case, which is wrong if best_effotrt_wfe_or_timeout is called repeatedly for the same timestamp that has now passed. A check is added for the WFE for this case

src/common/pico_time/time.c
```diff
@@ -13,6 +13,13 @@
 const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(nil_time, 0);
 const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(at_the_end_of_time, INT64_MAX);
 
+// not currently documented as the user isn't really expected to change this
+#ifndef PICO_TIME_USE_SLEEP_NOTIFIER
+// we don't want to circumvent use of spin_unlock_with_xxx when using for example FreeRTOS
+#if LOCK_INTERNAL_SPIN_UNLOCK_WITH_WAIT_OVERRIDDEN | LOCK_INTERNAL_SPIN_UNLOCK_WITH_NOTIFY_OVERRIDDEN | LOCK_INTERNAL_SPIN_UNLOCK_WITH_WAIT_OVERRIDDEN
```
lurch: I'm, guessing that this was supposed to be
```suggestion
#if LOCK_INTERNAL_SPIN_UNLOCK_WITH_WAIT_OVERRIDDEN | LOCK_INTERNAL_SPIN_UNLOCK_WITH_NOTIFY_OVERRIDDEN | LOCK_INTERNAL_SPIN_UNLOCK_WITH_BEST_EFFORT_WAIT_OR_TIMEOUT_OVERRIDDEN
```
?

kilograham:
Additional changes: 
5. Fixed a claude detected race conditoin
6. Added PICO_PLATFORM_NAME so tests that want to print something meaningful don't have to #Ifdef
7. Added newer even more improved claude co-written test which is also designed to test (the almost certainly broken) FreeRTOS - interop.. tests currently pass on `rp2040`,`rp2350-arm-s`,`rrp2350-riscv` with hardware/software spinlocks on RP2350

Wren6991:
Reviewed the deletion race fixes -- look correct to me. Not digested the other changes.

src/common/pico_time/time.c
```diff
@@ -13,6 +13,27 @@
 const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(nil_time, 0);
 const absolute_time_t ABSOLUTE_TIME_INITIALIZED_VAR(at_the_end_of_time, INT64_MAX);
 
+// Whether sleep_until() waits on a lock_core rather than on a bare __wfe().
```
Wren6991, line 16: If you're going to generate code, please at least write the comments yourself. This is obvious slop and I'd rather not read it every time I look in this file.

src/common/pico_time/time.c
```diff
@@ -476,9 +516,54 @@ bool best_effort_wfe_or_timeout(absolute_time_t timeout_timestamp) {
...
+        // decide that on its own - it lags an add (the IRQ that programs it runs on the pool's
+        // own core), and after our cancel it may be rearmed later or not at all - so we have to
+        // remember that we already asked.
+        //
+        // Recorded to the adapter's compare width, not to 64 bits. Where the timer compares 32
+        // bits a target is only honoured to that resolution anyway - an alarm armed for low word
+        // L fires at any time whose low word is L - so a wider record would claim precision the
+        // hardware cannot act on. Matching the width also makes this safe to share between the
+        // cores unsynchronised, which it always has been: a 32-bit aligned access is atomic on
+        // both Armv8-M and Hazard3 so it cannot tear, and every value either core can observe is
```
Wren6991: Single-copy atomicity of aligned loads and stores is a RISC-V guarantee, not a Hazard3 one.

Wren6991:
I don't think the use of LLMs here is a good direction. This code is already incredibly demanding to understand, which is why I left it unreviewed for so long. I think it's made worse by these changes.

kilograham:
yeah i plan to redact a lot of the comments which is why i haven't merged this yet

kilograham:
i disagree on whether LLM is helpful here - it has been; the upshot though is actually that a fix I'd made in June fixed A and introduced B. A separate fix i myself made as part of this PR fixed C (and unknown to me was actually a fix for A). Fixing B in this PR actually broke D. so the actual correct fix was just my change that fixed C. The tests added by this PR have been invaluable. The net code changes themselves are small, but I will actually split them out into separate commits because half of them are not even related to fixing the particular known bugs and can be applied individually to fix E & F etc

kilograham:
closed in favor of tidied up version #3127

PR #3085, opened 2026-07-20
`sem_init` `initial_permits` parameter doc comment minor correction

liamHowatt:
A semaphore initialized as below starts with one initial permit _released_. Not _acquired_, as the doc comment would leave users to believe.

```c
sem_init(&sem, 1, 1);
```

### DIFF
--- a/src/common/pico_sync/include/pico/sem.h
+++ b/src/common/pico_sync/include/pico/sem.h
@@ -45,7 +45,7 @@ typedef struct semaphore {
  *  \ingroup sem
  *
  * \param sem Pointer to semaphore structure
- * \param initial_permits How many permits are initially released
+ * \param initial_permits How many permits are initially available
  * \param max_permits  Total number of permits allowed for this semaphore
  */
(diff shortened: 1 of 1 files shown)


lurch:
Yeah, looking at https://github.com/raspberrypi/pico-sdk/blob/develop/src/common/pico_sync/sem.c `sem_init` does `sem->permits = initial_permits`, and then the "acquire" methods decrement `sem->permits` (if `sem->permits > 0`) and the "release" methods increment `sem->permits` (if `sem->permits < sem->max_permits`).
`sem_reset` also does `sem->permits = permits` but the [Doxygen for that](https://github.com/raspberrypi/pico-sdk/blob/develop/src/common/pico_sync/include/pico/sem.h#L72) says `\param permits the new number of available permits`, so perhaps this PR also ought to say "available" instead of "released" ?

liamHowatt:
Ah that's more consistent. Done.

kilograham (approved):

PR #3088, opened 2026-07-20
Add RP2350-only flash functions to Doxygen output

lurch:
Fixes https://github.com/raspberrypi/documentation/issues/4304

This is what the generated HTML docs will look like after this PR is merged:
Screenshot from 2026-07-20 18-28-27

(and the same fix will also affect the [Pico C SDK PDF](https://pip.raspberrypi.com//documents/RP-009085-KB) too :slightly_smiling_face: )

### DIFF
--- a/src/rp2_common/hardware_flash/include/hardware/flash.h
+++ b/src/rp2_common/hardware_flash/include/hardware/flash.h
@@ -169,7 +169,7 @@ static inline void flash_do_cmd(const uint8_t *txbuf, uint8_t *rxbuf, size_t cou
 
 void flash_flush_cache(void);
 
-#if !PICO_RP2040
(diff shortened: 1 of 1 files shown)


will-v-pi (approved):
Thanks! Are there any other `#if !PICO_RP2040` or `#if !PICO_RP2350` in header files that need fixing too?

lurch:
> Thanks! Are there any other `#if !PICO_RP2040` or `#if !PICO_RP2350` in header files that need fixing too?

I did a `git grep '!PICO_RP2040' "*.h"` and `git grep '!PICO_RP2350' "*.h"` and manually looked at all the matches, but (apart from the `flash.h` in this PR), none of them were excluding any Doxygen markup.

katshann-raspberry (approved):

PR #3089, opened 2026-07-21
Use the count_of macro in a couple more places

lurch:


will-v-pi (approved):

lurch:
I'm gonna fold this directly into #3094

PR #3094, opened 2026-07-22
Standardise usage of rom_get_sys_info function

lurch:
I copied the `rom_get_sys_info` usage-pattern from [`bootrom.h`](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_bootrom/include/pico/bootrom.h#L1122).
RP2350 Datasheet says "You should always check (the first word in the returned buffer) before interpreting the buffer.", so do that too.
Also some small unique_id tidy-ups.

### DIFF
--- a/src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
+++ b/src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
@@ -239,25 +239,25 @@ typedef int (*bootrom_api_callback_generic_t)(uint32_t r0, uint32_t r1, uint32_t
 // items are returned in order
 // 3 words package_id, device_id_lo, device_id_hi
 #define SYS_INFO_CHIP_INFO                      0x0001
-#define SYS_INFO_CHIP_INFO_WORDS_RETURNED       _u(3)
+#define SYS_INFO_CHIP_INFO_WORD_COUNT           _u(3)
 // 1 word: chip specific critical bits
 #define SYS_INFO_CRITICAL                       0x0002
-#define SYS_INFO_CRITICAL_WORDS_RETURNED        _u(1)
+#define SYS_INFO_CRITICAL_WORD_COUNT            _u(1)
 // 1 word: bytes: cpu_type, supported_cpu_type_bitfield
 #define SYS_INFO_CPU_INFO                       0x0004
-#define SYS_INFO_CPU_INFO_WORDS_RETURNED        _u(1)
+#define SYS_INFO_CPU_INFO_WORD_COUNT            _u(1)
 // 1 word: same as FLASH_DEVINFO row in OTP
 #define SYS_INFO_FLASH_DEV_INFO                 0x0008
-#define SYS_INFO_FLASH_DEV_INFO_WORDS_RETURNED  _u(1)
+#define SYS_INFO_FLASH_DEV_INFO_WORD_COUNT      _u(1)
 // 4 words
 #define SYS_INFO_BOOT_RANDOM                    0x0010
-#define SYS_INFO_BOOT_RANDOM_WORDS_RETURNED     _u(4)
+#define SYS_INFO_BOOT_RANDOM_WORD_COUNT         _u(4)
 // 2 words lsb first
 #define SYS_INFO_NONCE                          0x0020
-#define SYS_INFO_NONCE_WORDS_RETURNED           _u(2)
+#define SYS_INFO_NONCE_WORD_COUNT              _u(2)
 // 4 words boot_info, boot_diagnostic, boot_param0, boot_param1
 #define SYS_INFO_BOOT_INFO                      0x0040
-#define SYS_INFO_BOOT_INFO_WORDS_RETURNED       _u(4)
+#define SYS_INFO_BOOT_INFO_WORD_COUNT           _u(4)
 
 #define BOOTROM_NS_API_get_sys_info 0
 #define BOOTROM_NS_API_checked_flash_op 1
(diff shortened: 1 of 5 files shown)


src/rp2_common/pico_bootrom/bootrom.c
```diff
@@ -86,8 +86,8 @@ void __attribute__((noreturn)) rom_reset_usb_boot_extra(int usb_activity_gpio_pi
 #if !PICO_RP2040
 bool rom_get_boot_random(uint32_t out[4]) {
     uint32_t result[5];
-    rom_get_sys_info_fn func = (rom_get_sys_info_fn) rom_func_lookup_inline(ROM_FUNC_GET_SYS_INFO);
-    if (5 == func(result, count_of(result), SYS_INFO_BOOT_RANDOM)) {
+    int words_returned = rom_get_sys_info(result, 5, SYS_INFO_BOOT_RANDOM);
```
will-v-pi: I think using `count_of` might be clearer - and then the uses in `bootrom.h` could be updated to `count_of` too

```suggestion
    int words_returned = rom_get_sys_info(result, count_of(result), SYS_INFO_BOOT_RANDOM);
```

src/rp2_common/pico_unique_id/unique_id.c
```diff
@@ -38,16 +43,22 @@ static void __attribute__((PICO_UNIQUE_BOARD_ID_INIT_ATTRIBUTES)) _retrieve_uniq
...
-    assert(rc == 4);
-    for (int i = 0; i < PICO_UNIQUE_BOARD_ID_SIZE_BYTES; i++) {
-        retrieved_id.id[i] = out.bytes[PICO_UNIQUE_BOARD_ID_SIZE_BYTES - 1 + 2 * 4 - i];
-    }
+    #if PICO_UNIQUE_BOARD_ID_SIZE_BYTES <= SYS_INFO_DEVICE_ID_SIZE_BYTES
+        union {
+            uint32_t words[4];
+            uint8_t bytes[4 * 4];
+        } out;
+        int words_returned = rom_get_sys_info(out.words, 4, SYS_INFO_CHIP_INFO);
```
will-v-pi: Ditto on `count_of`

```suggestion
        int words_returned = rom_get_sys_info(out.words, count_of(out.words), SYS_INFO_CHIP_INFO);
```

src/rp2_common/pico_bootrom/bootrom.c
lurch: Okay, I can change all of these. Is it still appropriate for `count_of` to continue being used in the `words_returned == count_of(result)` check on the next line too? This would mean that the number `5` _only_ appears in the `uint32_t result[5]` line. (Which would be fine with me, I'm always happy to see the elimination of duplicated magic-numbers; I'm just double-checking that this is what's desired.)

And in that case I guess it might make sense to fold #3089 into this PR too.

will-v-pi: I think it's still appropriate - I also agree it's nice to remove the magic numbers.

Re folding the PRs, I don't mind - you could make the changes to use `count_of` in the existing `bootrom.h` functions in that PR, or fold it into this one and do it here

lurch: Taking that idea of removing magic-numbers even further, _perhaps_ it would also make sense to add e.g.
```c
#define SYS_INFO_BOOT_RANDOM_WORDS_RETURNED     _u(4)
```
to `bootrom_constants.h`, and then we could do
```c
uint32_t result[SYS_INFO_BOOT_RANDOM_WORDS_RETURNED + 1];
```
?

will-v-pi: Yeah, that sounds good too - for all the `SYS_INFO_XXX` defines having a corresponding `SYS_INFO_XXX_WORDS_RETURNED` rather than just having it in the comment above.

lurch: Finally got round to making and pushing this change :slightly_smiling_face:

src/rp2_common/boot_bootrom_headers/include/boot/bootrom_constants.h
```diff
@@ -239,18 +239,25 @@ typedef int (*bootrom_api_callback_generic_t)(uint32_t r0, uint32_t r1, uint32_t
 // items are returned in order
 // 3 words package_id, device_id_lo, device_id_hi
 #define SYS_INFO_CHIP_INFO                      0x0001
+#define SYS_INFO_CHIP_INFO_WORDS_RETURNED       _u(3)
```
kilograham: I think `_WORD_COUNT` would be better, as WORDS_RETURNED is confusing below, and who says they are returned

src/rp2_common/pico_bootrom/include/pico/bootrom.h
```diff
@@ -1109,9 +1109,9 @@ typedef struct {
 } boot_info_t;
 
 static inline int rom_get_boot_info(boot_info_t *info) {
-    uint32_t result[5];
-    int words_returned = rom_get_sys_info(result, 5, SYS_INFO_BOOT_INFO);
-    if (words_returned == (sizeof(result)/sizeof(result[0])) && result[0] == SYS_INFO_BOOT_INFO) {
+    uint32_t result[SYS_INFO_BOOT_INFO_WORDS_RETURNED + 1];
+    int words_returned = rom_get_sys_info(result, count_of(result), SYS_INFO_BOOT_INFO);
+    if ((words_returned - 1) == SYS_INFO_BOOT_INFO_WORDS_RETURNED && result[0] == SYS_INFO_BOOT_INFO) {
```
kilograham: I think i prefer `wordss_returned == SYS_INFO... + 1` as per below

Issue #3081, opened 2026-07-19
Missing crypto_random in CMakeLists.txt


fink-at-trmc-dk:
it is in mbedTLS CMake, but Pico SDK does not consume that CMake list; it uses a separate curated source list, and this file is missing there.

Created PR to fix this

peterharperuk:
Thanks. We will fix this either with your change or we'll improve how we build mbedtls to reduce the chance of mistakes like this. Here's my attempt https://github.com/raspberrypi/pico-sdk/pull/3095

kilograham:
short term fix merged into develop

PR #3095, opened 2026-07-23
Change source lists for mbedtls

peterharperuk:
Currently we list the mbedtls source files in some complicated variables which change depending on the version in use.
This makes it hard to know if you have all the source files needed.

Instead, just copy the source file lists from the mbedtls CMakeLists.txt file into our own included cmake file.
This should reduce the chance of us missing files in future.

See #3081


fink-at-trmc-dk:
Much better approach IMO. I also ran into something similar in LWIP, so maybe this approach could be used there as well?

kilograham:
#3099 is a more generic version of this, which should be used in preference if it works

src/rp2_common/pico_mbedtls/mbedtls_library_2.28.8.cmake
```diff
@@ -0,0 +1,104 @@
+set(src_crypto
```
lurch, line 1: If this PR ends up getting used in preference to #3099 , it'd probably be useful if there was a comment in here saying something like:
```cmake
# These file-lists were copied from https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-2.28.8/library/CMakeLists.txt
```
(and similarly for the other files too of course)

PR #3099, opened 2026-07-27
parse mbedtls source file names directly from cmakelists.txt

### LINKED ISSUE #3081: Missing crypto_random in CMakeLists.txt

kilograham:
fixes #3081


kilograham:
note i have only tried this with the current version of medtls

lurch:
Looks like this is a more-automated version of #3095 ?

PR #3109, opened 2026-08-03
Set HEAP_LOC when persistent data is above heap

### LINKED ISSUE #3108: pico_set_persistent_data_loc with an address in sram1 can give you an unbounded heap
When using picolibc and the SRAM1 branch of pico_set_persistent_data_loc - the .heap is placed above HEAP_LIMIT. The picoclib code (search for __fallback_sbrk) does unsigned arithmetic which gives you a very large heap.

On my PC (arm-gnu-toolchain-15.2.rel1-x86_64-arm-none-eabi which is using picolibc) If I build the low_power_pstate example for Pico 2 I get a binary with these symbols...

with pico_set_persistent_data_loc(low_power_pstate_timer 0x20040000) # SRAM1 power domain

$ nm low_power_pstate_timer.elf | grep heap
20040000 A __heap_end
20040008 B __heap_start

with pico_set_persistent_data_loc removed...

$ nm low_power_pstate_timer.elf | grep heap
20080000 A __heap_end
20000ce4 B __heap_start

I tested it with this...

    // With pico_set_persistent_data_loc(low_power_pstate_timer 0x20040000)
    // the following crashes with a isr_hardfault (bit weird)
    // you should get an out of memory panic unless PICO_MALLOC_PANIC=0 in which case it should return NULL
    volatile uint8_t *p = malloc(1024 * 1024);
    hard_assert(p == NULL);
    printf("p=%p\n", p);

Note that this is not a problem with newlib because it's using __StackLimit - although that's another story.

peterharperuk:
Fixes #3108

### DIFF
--- a/src/rp2_common/pico_low_power/CMakeLists.txt
+++ b/src/rp2_common/pico_low_power/CMakeLists.txt
@@ -55,7 +55,12 @@ function(pico_set_persistent_data_loc TARGET PERSISTENT_DATA_LOC)
                 pico_set_linker_script_var(${TARGET} HEAP_LOC __persistent_data_end__)
                 set(PERSISTENT_DATA_SECTION "RAM")
         elseif(PERSISTENT_DATA_LOC LESS 0x20080000)
-                # SRAM1, so heap should come before persistent data
+                # SRAM1, so heap should come before persistent data. This needs HEAP_LOC as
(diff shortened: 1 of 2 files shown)


will-v-pi (approved):
LGTM

Issue #3112, opened 2026-08-05
using external TLS variables with ATfE-22.1.0-Linux-x86_64 compiler fails with a link error


peterharperuk:
I got the compiler ATfE-22.1.0-Linux-x86_64 from here https://github.com/arm/arm-toolchain/releases (also tried ATfE-21.1.1-Linux-x86_64)
It is based on llvm and uses picoclib.
I have a very simple program that uses errno and when I build it I get a link error.

```
cd pico-sdk
mkdir build
cd build
cmake .. -DPICO_TOOLCHAIN_PATH=/opt/compilers/arm/llvm/ATfE-22.1.0-Linux-x86_64
cd test/pico_clib_interface
make
<blah>
[100%] Linking CXX executable tls_repro.elf
ld.lld: error: /home/peterh/source/pico-sdk1/src/rp2_common/pico_standard_link/script_include/section_default_data.incl:71: unable to move location counter (0x20000318) backward to 0x20000314 for section '.tbss_space'
ld.lld: error: section .tbss_space at 0x20000318 of size 0xfffffffffffffffc exceeds available address space
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [test/pico_clib_interface/CMakeFiles/tls_repro.dir/build.make:1092: test/pico_clib_interface/tls_repro.elf] Error 1
make[1]: *** [CMakeFiles/Makefile2:5139: test/pico_clib_interface/CMakeFiles/tls_repro.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

It seems related to the .tbss_space section. The spacer's arithmetic assumes nothing is allocated between __bss_start__ and the spacer itself, and an unplaced .got breaks that assumption on lld. (Thank claude for that)

Anyway - the fix suggested is to add the following into pico_arm_clang_common.cmake - it's what picolibc.specs does evidently.

```
foreach(LANG IN ITEMS C CXX)
    set(CMAKE_${LANG}_FLAGS_INIT "${CMAKE_${LANG}_FLAGS_INIT} -ftls-model=local-exec")
endforeach()
```

Claude seems to think this is related to the following issue - but it's a bit over my head: https://github.com/raspberrypi/pico-sdk/issues/3064

I've pushed the test to a branch. The above fix is currently commented out but it does seem to resolve the problem.

Wren6991:
Was able to reproduce this using your SDK branch and:

```
[...]

PR #3113, opened 2026-08-05
Add pico_clib_interface_test

peterharperuk:
This is a very simple test that references errno.
A TLS reference from a translation unit compiled by the SDK. With ATfE-22.1.0-Linux-x86_64 it generates a linker error

unable to move location counter (0x20000318) backward to 0x20000314 for section '.tbss_space'

See https://github.com/raspberrypi/pico-sdk/issues/3112

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -22,6 +22,7 @@ if (PICO_ON_DEVICE)
     add_subdirectory(pico_low_power_test)
     add_subdirectory(pico_async_context_test)
     add_subdirectory(pico_thread_local_test)
+    add_subdirectory(pico_clib_interface_test)
 endif()
 
 # PICO_TEST_FILE_GENERATOR: Path to the a CMake script to generate test files for ci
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ if (PICO_ON_DEVICE)
     add_subdirectory(pico_low_power_test)
     add_subdirectory(pico_async_context_test)
     add_subdirectory(pico_thread_local_test)
+    add_subdirectory(pico_clib_interface)
```
kilograham: one of these directories is not like the other ones; one of these directories is not the same!

lurch: Are you suggesting that `pico_clib_interface` ought to have `_test` at the end?

kilograham: yes

peterharperuk: Fixed. pico_clib_interface_test is probably a bad name

PR #3117, opened 2026-08-09
Extract configs speedup

kilograham:
these are painfully slow on my machine because of lots of build directories


lurch:
I wonder if parsing `.gitignore` and ignoring the patterns specified in that, would be a better solution than explicitly looking for `CMakeCache.txt` ?
EDIT: A quick search finds https://github.com/mherrmann/gitignore_parser

kilograham:
feel free, otherwise this gets the job done (edit: takes it from 76s to <1s for me locally)

PR #1865, opened 2024-08-23
Use lib instead of CMAKE_INSTALL_LIBDIR for pioasm install

will-v-pi:
The CMake GNUInstallDirs module prior to CMake version 3.23.1 used an architecture-specific CMAKE_INSTALL_LIBDIR on debian, which is not searched by `find_package` calls from the SDK

Fix this in the same way as picotool https://github.com/raspberrypi/picotool/issues/117 by hard-coding the lib directory

PR #3120, opened 2026-08-10
Install pioasm cmake files to CMAKE_INSTALL_LIBDIR

### LINKED ISSUE #3119: pioasm cmake configs are unconditionally installed into prefix/lib
https://github.com/raspberrypi/pico-sdk/blob/6a368987b7bd4627d3d6d126fabaa7495991041f/tools/pioasm/CMakeLists.txt#L72

On some systems the directory is different, e.g. `/usr/lib64/cmake`. The convention seems to be to use `CMAKE_INSTALL_LIBDIR` as the default.

ypsvlq:
Fixes #3119.

### DIFF
--- a/tools/pioasm/CMakeLists.txt
+++ b/tools/pioasm/CMakeLists.txt
@@ -69,7 +69,7 @@ if (PIOASM_FLAT_INSTALL)
     set(INSTALL_CONFIGDIR pioasm)
     set(INSTALL_BINDIR pioasm)
 else()
-    set(INSTALL_CONFIGDIR lib/cmake/pioasm)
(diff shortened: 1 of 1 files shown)


will-v-pi:
See https://github.com/raspberrypi/picotool/issues/117 for why this was changed to `lib` instead of using `CMAKE_INSTALL_LIBDIR` (tldr is that it was broken on Debian using CMake 3.23 and earlier), and https://github.com/raspberrypi/pico-sdk/pull/1865 for the SDK PR that made this fix. I don't think we will revert this fix unless there is a very good reason to.

PR #3127, opened 2026-08-13
Combined sleep fixes

kilograham:
Against usual tradition, this is worth landing as individual commits, however keeping here for review.

@Wren6991 please pile any (we expect comment only) changes in a new PR on top


src/common/pico_base_headers/include/pico/config.h
```diff
@@ -18,6 +18,11 @@
 
 #include "pico/config_autogen.h"
 
+// PICO_CONFIG: PICO_PLATFORM_STRING, Name of platform as a string; defaulted to "unknown" if the build does not set it, group=pico_base
```
lurch, line 21: The code in `tools/extract_configs.py` will assume that any `PICO_CONFIG:` without an explicit `type=`, is of type `int`. I guess I ought to add support for `type=string` ?

Hmm, looks like `PICO_CONFIG: PICO_CONFIG_RTOS_ADAPTER_HEADER` below should also be `type=string` :slightly_smiling_face:

test/pico_float_test/double_benchmark.c
```diff
@@ -1229,16 +1229,7 @@ int main() {
 #else
     printf("abi               soft(fp)\n");
 #endif
-#if PICO_RP2040
-    printf("platform          rp2040\n");
-#elif PICO_RP2350
-    printf("platform          rp2350\n");
-#endif
-#if PICO_RISCV
-    printf("arch              risc-v\n");
```
lurch, line 1238: Is it worth adding a `PICO_ARCH_STRING`, or is that not needed? (because it can be inferred from `PICO_PLATFORM_STRING`?)

Wren6991: No, we don't really want people decoding strings at runtime to infer booleans.

Wren6991: `PICO_PLATFORM_STRING` looks to me to be a human-readable convenience and we shouldn't make it (ugh) load-bearing

lurch: Yeah, I _think_ I kind of meant "inferred by the human reading the logfile output", rather than "inferred by code at runtime". Maybe.
In case it clarifies things, I was referring specifically to this part of the diff:
```patch
-#if PICO_RP2040
-    printf("platform          rp2040\n");
-#elif PICO_RP2350
-    printf("platform          rp2350\n");
-#endif
-#if PICO_RISCV
-    printf("arch              risc-v\n");
-#else
-    printf("arch              arm\n");
-#endif
+    printf("platform          " PICO_PLATFORM_STRING "\n");
```

I was wondering if somebody might want the ability to do e.g,
```c
-    printf("arch              " PICO_ARCH_STRING "\n");
```

kilograham: yeah tbc - this is PICO_PLATFORM as a string... note i thought there were already some string PICO_CONFIGS?

src/common/pico_base_headers/include/pico/config.h
Wren6991: > Hmm, looks like PICO_CONFIG: PICO_CONFIG_RTOS_ADAPTER_HEADER below should also be type=string 🙂

That one looks to me like it's a bare macro that gets expanded and stringified by the preprocessor, not a C string. So:

```
-DPICO_CONFIG_RTOS_ADAPTER_HEADER=foo.h
```

yields:

```c
#include "foo.h"
```

This isn't the same thing as a C string literal. Maybe we shouldn't infer `int` for configs with unspecified types.

lurch: > Maybe we shouldn't infer `int` for configs with unspecified types.

I'd be more than happy to modify my config-extraction scripts to insist that a `type=` is **always** present, but note that `git grep "PICO_CONFIG:" | grep -v "type=" | wc -l` reports that there are currently 88 such instances.

> This isn't the same thing as a C string literal.

Okay, so could you suggest a suitable `type=` to be used in this instance?

src/common/pico_sync/include/pico/lock_core.h
```diff
@@ -210,15 +236,90 @@ extern volatile uint8_t lock_internal_notify_count;
...
 })
 #else
+// see the comment on lock_internal_spin_unlock_with_wait above for why the event left by the
+// spin_unlock must be drained before the notify count is tested
 #define lock_internal_spin_unlock_with_best_effort_wait_or_timeout(lock, save, until) ({ \
     uint8_t _notify_count = lock_internal_notify_count;                                  \
     spin_unlock((lock)->spin_lock, save);                                                \
-    if (_notify_count == lock_internal_notify_count) __wfe();                            \
-    best_effort_wfe_or_timeout(until);                                                   \
+    /* consume the event caused by the spin_unlock; cannot block */                      \
```
Wren6991: "cannot block": is this true? IRQs are enabled inside of the `spin_unlock`.  An interrupt could be immediately taken, and consume the event before the WFE executes. Should this instead be:

```
spin_unlock_unsafe((lock)->spin_lock);
__wfe(); /* consume self-generated event without interruption */
restore_interrupts_from_disabled(save);
```

It's probably safe on Armv8-M because it sets the event register on return from any non-fault exception, but on Hazard3 it's only set by `__h3_unblock()` (including from the same core due to external loopback).

src/common/pico_base_headers/include/pico/config.h
Wren6991: My point is there is not always a suitable type, and plastering `int` on things with unspecified types doesn't help.

src/common/pico_sync/include/pico/lock_core.h
Wren6991: Existing from last set of fixes, but this comment is a bit misleading: `spin_unlock` sends an event to the current core only. The `__sev();` below is to the other core.

Wren6991, line 208: This read-modify-write updates global state, potentially under *different* locks. I think this is safe for a 2-core system, because the race only happens when both cores concurrently issue `lock_internal_spin_unlock_with_notify()` on different locks; since IRQs are disabled while holding a lock, this can't be concurrent with `lock_internal_spin_unlock_with_wait()`. However it becomes unsound with >2 cores. It's ok for an RP2350-specific workaround but just wanted to flag this.

src/common/pico_time/time.c
```diff
@@ -13,6 +13,29 @@
...
+// primitive (sleep_notifier) vs just doing a __wfe().
+//
+// It exists for RTOS integration and for nothing else. Where the lock_internal_* primitives
+// are overridden, waiting on the notifier lets the RTOS block the *task*, so other tasks keep
+// running for the duration of a sleep; a bare __wfe() would leave the sleeping task ready but
+// parked, starving every lower-priority task until it woke.
+//
+// On bare metal it buys nothing and should not be turned on. The callback's __sev() already
+// sets a per-core, sticky event latch, which reaches a waiter that has not yet reached its
+// __wfe() and cannot be consumed by anyone else - which is precisely the property the notifier
```
Wren6991: > The callback's __sev() already  sets a per-core, sticky event latch, which reaches a waiter that has not yet reached its  __wfe() and cannot be consumed by anyone else

Already pointed this out in another comment, but Claude was wrong here: it can be consumed by an IRQ on RISC-V. (I'm going to strip out all the slop comments, but pointing out the issue before I do.)

src/common/pico_base_headers/include/pico/config.h
Wren6991: `filename` or `path` would work here but it's not really general.

src/common/pico_sync/include/pico/lock_core.h
Wren6991: Fixed by [971351f](https://github.com/raspberrypi/pico-sdk/pull/3127/commits/971351f198838ad0f89ca6e5402d240015bf3c6d), which just swaps the `__wfe()` with the preceding `csrs mstatus`/`msr PRIMASK`.

To hit this in practice you'd have to have a `h3.block` inside an IRQ handler (or in a switched-to context), which would be odd, but anyway I'm pretty sure it was a bug.

src/common/pico_time/time.c
```diff
@@ -170,23 +199,23 @@ static void alarm_pool_irq_handler(void) {
...
                     if (earliest_entry->callback == repeating_timer_marker) {
+                        // special case repeating timer without making another function call which adds overhead
                         repeating_timer_t *rpt = (repeating_timer_t *)earliest_entry->user_data;
                         delta = rpt->callback(rpt) ? rpt->delay_us : 0;
                     } else {
                         alarm_id_t id = make_alarm_id(pool->ordered_head, earliest_entry->sequence);
                         delta = earliest_entry->callback(id, earliest_entry->user_data);
                     }
                 } else {
-                    // negative target means cancel alarm
```
Wren6991, line 189: Potentially a regression here: The negative target forced us to continue looping, but that is no longer the case. If multiple alarms are deleted, and there are other entries with target times earlier than the deleted alarms' target times, then the non-deleted alarms can get stranded.

Potential fix is to force the loop to continue while there are deleted entries at the head:

```patch
diff --git a/src/common/pico_time/time.c b/src/common/pico_time/time.c
index 37037e45..14bd26ce 100644
--- a/src/common/pico_time/time.c
+++ b/src/common/pico_time/time.c
@@ -333,7 +333,8 @@ static void alarm_pool_irq_handler(void) {
             }
         }
         // check we haven't now passed the target time; if not we don't want to loop again
-    } while ((earliest_target - (int64_t)ta_time_us_64(timer)) <= 0);
+    } while ((earliest_target - (int64_t)ta_time_us_64(timer)) <= 0 ||
+        (pool->ordered_head >= 0 && pool->entries[pool->ordered_head].callback == deleted_timer_marker));
     // We always want the timer IRQ to wake a WFE so that best_effort_wfe_or_timeout() will wake up. It will wake
     // a WFE on its own core by nature of having taken an IRQ, but we do an explicit SEV so it wakes the other core
     __sev();
```

Wren6991: Resolved by f2fc82b, which pushes the deleted entries straight to the free list instead of doing the weird shuffle-to-front.

test/sync_interop_test/interop_platform.c
```diff
@@ -0,0 +1,1035 @@
...
+ * Faithful copies of mutex_enter_blocking() and mutex_enter_block_until() from
+ * src/common/pico_sync/mutex.c, with the wait iterations counted.
+ *
+ * Counting is the only reliable way to tell sleeping from spinning here. On Armv8-M
+ * DWT_CYCCNT keeps counting straight through WFE - that is exactly what DWT_SLEEPCNT is
+ * for, and the RP2350 M33 does not implement it - so a cycle-based occupancy figure cannot
+ * distinguish the two. A wait that really blocks takes 1-5 iterations; a busy-poll runs to
+ * thousands.
+ *
+ * NOTE: keep these in step with mutex.c, as pico_sync_test.c already has to.
```
Wren6991, line 38: This test uses a copy/pasted version of the `mutex` primitive. It doesn't cover the version in the SDK that people actually use. The comment here in the test will be missed by anyone who changes the SDK implementation. The implementation has a comment pointing to `pico_sync_test`, a different test which *also* now has a copy/pasted `mutex`.

Maybe this instrumentation should be added to the SDK implementation (either guarded by or injected through macros) rather than hoping things will stay in sync.

test/hardware_sync_spin_lock_test/hardware_sync_spin_lock_test.c
```diff
@@ -190,6 +190,109 @@ static const test_t tests[] = {
...
 };
 
+// ---------------------------------------------------------------------------
+// PICO_SPIN_LOCK_UNLOCK_CAUSES_SEV probe
+//
+// The RP2350 lock_core wait/notify primitives (see PICO_SYNC_RP2350_SPIN_LOCK_WORKAROUND in
+// pico/lock_core.h) are built on taking and releasing a software spin lock setting the calling
+// core's own event flag, so that a following __wfe() returns immediately and a second one is
+// needed to actually wait. PICO_SPIN_LOCK_UNLOCK_CAUSES_SEV asserts that at compile time, and
+// nothing checks it against the hardware. Worth checking, because the Arm and RISC-V software
```
Wren6991: I think this should be deleted. It's "testing" a fact about the hardware that we simply know to be true. The fact Claude wanted to confirm it while debugging does not make it a meaningful test we should carry forward in the SDK. Also this part:

> different instructions -- ldaexb/strexb, which takes a reservation, against amoor.w.aq, which does not

...is simply false, and we shouldn't leave something misleading like that lying around.

test/pico_time_test/pico_time_test.c
```diff
@@ -266,9 +267,97 @@ int main() {
...
+    batch_cancel_test();
+
     PICOTEST_END_TEST();
 }
 
+// Cancelling several alarms before the pool's handler gets a chance to run. Both bugs this
+// covers need the head of the ordered list cancelled along with at least one entry behind it,
+// which is why they are cancelled as a batch:
+//
+//  - the scan left prev pointing at ordered_head while the head itself was cancelled, so moving
```
Wren6991: "the scan" here refers to the two-pass deletion code that was removed in the same batch of commits, so I don't think this is useful context to anyone reading this file in the future.

src/common/pico_sync/include/pico/lock_core.h
Wren6991: resolved by 0d6fcd2

test/hardware_sync_spin_lock_test/hardware_sync_spin_lock_test.c
Wren6991: Probably just revert or rebase out the commit that added this: 27aa1bd96

Claude's commit message for that commit is also false. AMOs *do* take reservations.

Wren6991 (commented):
I think the functional changes as of f2fc82b are all correct, and fix some severe bugs.

The copy/pasted mutex implementations in the new sync tests should ideally be linked to the actual SDK implementations, and I think it's worth cleaning up the things I flagged in the tests that will confuse us when we read them later.

Other than that LGTM.

test/hardware_sync_spin_lock_test/hardware_sync_spin_lock_test.c
kilograham: i think i'm gonna keep it because it is verifying a fact about the SDK define being set correctly, but i will fix the other issues

PR #3132, opened 2026-08-14
Pass -mstrict-align to Risc-V builds by default

will-v-pi:
GCC shouldn't generate misaligned code anyway, as it is marked as slow by the default tune, but add this to ensure __riscv_misaligned_avoid is set in builds as well

Adds new PICO_COMMON_LANG_FLAGS_EXTRA which can be used to append extra flags after pico_choose_compiler_flags chooses from PICO_COMMON_LANG_FLAGS_LIST

### DIFF
--- a/cmake/preload/toolchains/pico_riscv_gcc.cmake
+++ b/cmake/preload/toolchains/pico_riscv_gcc.cmake
@@ -4,14 +4,14 @@ set(PICO_DEFAULT_GCC_TRIPLE riscv32-pico-elf riscv32-unknown-elf riscv32-corev-e
 
 # PICO_CMAKE_CONFIG: PICO_NO_STRICT_ALIGN, Disable -mstrict-align for Risc-V, type=bool, default=0, group=build, docref=cmake-toolchain-config
 if (NOT PICO_NO_STRICT_ALIGN)
-    set(PICO_COMMON_LANG_FLAGS_EXTRA "${PICO_COMMON_LANG_FLAGS_EXTRA} -mstrict-align")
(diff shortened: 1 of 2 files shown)


will-v-pi:
@liamfraser Could you kick off a multi-gcc run with this branch which includes all the Risc-V compilers? I'm not sure where the workflow to do that has ended up, as the one on `develop`/`master` only has `RISCV GCC 14.1.0`, but I thought we were testing with more.

Wren6991:
Yeah, using `-mstrict-align` is correct. We don't handle alignment traps so this is a correctness problem not a performance problem.

So the new variable, `PICO_COMMON_LANG_FLAGS_EXTRA`, contains flags that are not chosen? That makes it tricky to override, and also means the `_EXTRA` flags are not tested during compiler testing. Wondering if it might be cleaner to just add `-mstrict-align` to all of the existing `PICO_COMMON_LANG_FLAGS` choices.

Wren6991:
Would definitely prefer a local fix in `pico_riscv_gcc.cmake` over more variable spaghetti

will-v-pi:
I've made it self-contained inside `pico_riscv_gcc.cmake`, and added the flags to the test. I'm still keeping a `PICO_RISCV_LANG_FLAGS_EXTRA` variable for it, otherwise would need duplicate `PICO_COMMON_LANG_FLAGS_LIST` for each `PICO_NO_STRICT_ALIGN` value, and that will also make it easier to add more flags in the future if we need.

Wren6991 (approved):
Nice, looks clean to me.

PR #3134, opened 2026-08-14
Test for unaligned access in newlibc functions.

peterharperuk:
This can be used to verify that the riscv compiler is behaving correctly

### DIFF
--- a/test/pico_unaligned_string_test/CMakeLists.txt
+++ b/test/pico_unaligned_string_test/CMakeLists.txt
@@ -1,21 +1,17 @@
-PROJECT(pico_unaligned_string_test)
+add_executable(pico_unaligned_string_test
+        pico_unaligned_string_test.c
+        )
 
-if (PICO_ON_DEVICE)
-    add_executable(pico_unaligned_string_test
-            pico_unaligned_string_test.c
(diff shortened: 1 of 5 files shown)


test/pico_unaligned_string_test/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+PROJECT(pico_unaligned_string_test)
+
+if (PICO_ON_DEVICE)
```
will-v-pi: This probably shouldn't call `PROJECT`, given none of the other tests do? Also, it doesn't need the `PICO_ON_DEVICE` gate because that's already gated in tests/CMakeLists.txt before including this directory.

will-v-pi (approved):
LGTM - can confirm this test fails with the [2.3.0-0 compiler](https://github.com/raspberrypi/pico-sdk-tools/releases/tag/v2.3.0-0), and passes with the [fixed compiler](https://github.com/raspberrypi/pico-sdk-tools/actions/runs/31793445982) and the [2.2.0-3 compiler](https://github.com/raspberrypi/pico-sdk-tools/releases/tag/v2.2.0-3)

Issue #3115, opened 2026-08-06
pico_low_power: timed dormant leaves the powman wakeup alarm armed, so a later GPIO dormant wakes spuriously


hasumikin:
### Description

On RP2350, after a timed dormant (`low_power_dormant_for_ms` / `low_power_dormant_until_aon_timer`), a subsequent GPIO-wake dormant (`low_power_dormant_until_gpio_pin_state`) does not wait for the pin: it returns without a GPIO event shortly after entry.

There are two layers to it:

1. `low_power_dormant_until_aon_timer` arms the wakeup with `aon_timer_enable_alarm(&ts, NULL, true)`
   ([low_power.c:609](https://github.com/raspberrypi/pico-sdk/blob/2.3.0/src/rp2_common/pico_low_power/low_power.c#L609))
   but never disarms it after waking. Its sleep-mode counterpart does: `low_power_sleep_until_aon_timer` calls `aon_timer_disable_alarm()` on the way out
   ([low_power.c:367](https://github.com/raspberrypi/pico-sdk/blob/2.3.0/src/rp2_common/pico_low_power/low_power.c#L367)).
   Once the target time has passed, the powman wake condition remains asserted until the alarm is disabled or reconfigured, so a subsequent dormant entry that observes this wake source returns without waiting for its intended event.

2. Even disarming manually with `aon_timer_disable_alarm()` is not enough on RP2350: it reaches only `powman_timer_disable_alarm()`
   ([aon_timer.c:174](https://github.com/raspberrypi/pico-sdk/blob/2.3.0/src/rp2_common/pico_aon_timer/aon_timer.c#L174)),
   which clears `INTE.TIMER` and `ALARM_ENAB` but leaves `POWMAN_TIMER_PWRUP_ON_ALARM` set
   ([powman.c:278](https://github.com/raspberrypi/pico-sdk/blob/2.3.0/src/rp2_common/hardware_powman/powman.c#L278)).
   Thus, starting from the `0x72` state below, `aon_timer_disable_alarm()` leaves `0x62` = `ALARM | PWRUP_ON_ALARM | RUN`. The API that also disables the alarm as a low-power wake source is `powman_disable_alarm_wakeup()`.

Observed register evidence (POWMAN `TIMER` register, 0x40100088):

- after a timed dormant completes: `0x72` = `ALARM | PWRUP_ON_ALARM | ALARM_ENAB | RUN`
[...]

PR #3135, opened 2026-08-14
Disable the powman alarm wakeup after waking up

will-v-pi:
Fix aon_timer_disable_alarm to undo aon_timer_enable_alarm entirely

Add aon_timer_disable_alarm after waking from dormant, and powman_disable_all_wakeups after waking from pstate

Fixes https://github.com/raspberrypi/pico-sdk/issues/3115

### DIFF
--- a/src/rp2_common/pico_aon_timer/aon_timer.c
+++ b/src/rp2_common/pico_aon_timer/aon_timer.c
@@ -176,7 +176,8 @@ void aon_timer_disable_alarm(void) {
 #if HAS_RP2040_RTC
     rtc_disable_alarm();
 #elif HAS_POWMAN_TIMER
-    powman_timer_disable_alarm();
+    // This calls powman_timer_disable_alarm()
+    powman_disable_alarm_wakeup();
 #else
     panic_unsupported();
 #endif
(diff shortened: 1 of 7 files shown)


will-v-pi:
Note that this new test fails on RP2040 - it fails to wake from DORMANT using the GPIO. I have not figured out why yet

src/rp2_common/pico_aon_timer/aon_timer.c
```diff
@@ -176,7 +176,8 @@ void aon_timer_disable_alarm(void) {
 #if HAS_RP2040_RTC
     rtc_disable_alarm();
 #elif HAS_POWMAN_TIMER
-    powman_timer_disable_alarm();
+    // This calls powman_timer_disable_alarm()
```
kilograham, line 179: this comment is a bit weird... i.e. great - but why is that important ;-)

will-v-pi: I added it because `powman_disable_alarm_wakeup` didn’t seem clear enough that it also disabled the alarm - happy to remove the comment if you don’t want it

kilograham: i more meant, i didn't know what it was trying to tell me (presumably the fact that it does is important). if that is the extent of it, then maybe fine as is

will-v-pi: Yeah, that's the extent of it - mostly to remind me if I look at the code again

kilograham (approved):

PR #3136, opened 2026-08-14
Add check_riscv_flag_pick CI check

will-v-pi:
Checks that the Risc-V flag selection chooses the correct flags for known compilers

This should catch the 2.3.0 issue that was fixed in https://github.com/raspberrypi/pico-sdk/pull/3076, where it fell through to the non-zcmp arch on a zcmp compiler


.github/workflows/check_riscv_flag_pick.yml
```diff
@@ -0,0 +1,63 @@
+name: Check RISCV Flag Choices
```
lurch:
```suggestion
name: Check RISC-V flag choices
```

.github/workflows/check_riscv_flag_pick.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Check RISCV GCC 16.1.0 gives cpu=hazard3-rp2350
+      if: ${{ !cancelled() }}
+      shell: bash
+      run: |
+        cd ${{ github.workspace }}
+        mkdir -p build && rm -rf build/* && cd build
+        cmake ../ -DPICO_NO_PICOTOOL=1 -DPICO_PLATFORM=rp2350-riscv -DPICO_TOOLCHAIN_PATH=/opt/riscv/gcc-riscv-32-pico-elf-16.1.0
+        cat CMakeCache.txt | grep "CMAKE_ASM_FLAGS"
+        cat CMakeCache.txt | grep "cpu=hazard3-rp2350"
```
lurch: Alternatively:
```suggestion
        rm -rf build && mkdir build
        cmake -S . -B build -DPICO_NO_PICOTOOL=1 -DPICO_PLATFORM=rp2350-riscv -DPICO_TOOLCHAIN_PATH=/opt/riscv/gcc-riscv-32-pico-elf-16.1.0
        grep "CMAKE_ASM_FLAGS" build/CMakeCache.txt
        grep "cpu=hazard3-rp2350" build/CMakeCache.txt
```

lurch:
Hmmm, I guess there's probably not much value in running this on every single CI run, so perhaps this is worth adding to https://github.com/raspberrypi/pico-sdk/blob/develop/.github/workflows/multi-gcc.yml instead, and _maybe_ that'd also make it more likely that these RISC-V compiler-tests get updated as and when we add new compiler-versions into the mix? :thinking:

.github/workflows/check_riscv_flag_pick.yml
will-v-pi: The `mkdir build` can go too, as cmake will create the directory - I'll apply manually

will-v-pi:
I've moved it into multi-gcc, and kicked off a run https://github.com/raspberrypi/pico-sdk/actions/runs/32718183803/job/97403804842

PR #3138, opened 2026-08-14
riscv: restore sp correctly in __unhandled_exception

### LINKED ISSUE #3137: riscv restores sp incorrectly in __unhandled_exception
The exception prologue does `addi sp, sp, -64` before saving registers, and the normal return path undoes it with `addi sp, sp, 64`. 

__unhandled_exception instead applies -64 a second time, leaving sp 64 bytes below the foreground stack pointer rather than restored.

That defeats the stated purpose of the surrounding instructions ("Restore original registers and stack pointer so debugger can backtrace") -- every unhandled exception produces a misleading backtrace, which makes an already awkward class of fault harder to diagnose. The fix is simple.

```
--- a/src/rp2_common/hardware_exception/exception_table_riscv.S
+++ b/src/rp2_common/hardware_exception/exception_table_riscv.S
@@ -117,7 +117,7 @@ __unhandled_exception:
     // Restore original registers and stack pointer so debugger can backtrace
     csrr ra, mscratch
     lw t6, 60(sp)
-    addi sp, sp, -64
+    addi sp, sp, 64
     // Second symbol here just to make it clearer in the debugger why you got
     // here; the entry point can appear labelled with the name of any one of the
     // unhandled exceptions, which is less clear.

```

Here's the difference it makes (the example coincidentally is using 128 bytes of stack)

```
BROKEN                                    FIXED
#0 __halt_on_unhandled_exception          #0 __halt_on_unhandled_exception
#1 level3_faults (<optimized out>)        #1 level3_faults (marker_a=0xAAAAAAAA, marker_b=0xBBBBBBBB)
#2 0x00000000 in ?? ()                    #2 level2 (marker_a=0xAAAAAAAA)
Backtrace stopped: frame did not          #3 level1 ()
save the PC                               #4 main ()

$sp = 0x20081f00                          $sp = 0x20081f80      (+0x80 = 128)
$mcause = 4                               $mcause = 4
```

Note this is masked at -O0, where gcc emits a frame pointer and gdb unwinds through s0 rather than sp, so the backtrace looks correct with or without the fix. It shows from -Og upwards, which includes the default Debug build.

[sp_fix_demo.zip](https://github.com/user-attachments/files/31074946/sp_fix_demo.zip)

peterharperuk:
The exception prologue does `addi sp, sp, -64` before saving registers, and the normal return path undoes it with `addi sp, sp, 64`. __unhandled_exception instead applied -64 a second time, leaving sp 128 bytes below the foreground stack pointer rather than restored.

That defeats the stated purpose of the surrounding instructions ("Restore original registers and stack pointer so debugger can backtrace"). Taking a misaligned load four frames deep and halting the core, gdb gives:

Fixes #3137

### DIFF
--- a/src/rp2_common/hardware_exception/exception_table_riscv.S
+++ b/src/rp2_common/hardware_exception/exception_table_riscv.S
@@ -117,7 +117,7 @@ __unhandled_exception:
     // Restore original registers and stack pointer so debugger can backtrace
     csrr ra, mscratch
     lw t6, 60(sp)
-    addi sp, sp, -64
+    addi sp, sp, 64
     // Second symbol here just to make it clearer in the debugger why you got
     // here; the entry point can appear labelled with the name of any one of the
     // unhandled exceptions, which is less clear.


Wren6991 (approved):
Fix is minimal and correct. Please try to describe it in your own words next time instead of pasting LLM output.

Wren6991:
bazel build failure has nothing to do with this 1-line change to RISC-V asm. Other checks are fine.

peterharperuk:
I couldn't have written it any better than that.

PR #3139, opened 2026-08-14
pico_ubsan: report undefined behaviour

peterharperuk:
The recent RISC-V alignment issue made me wonder if we should have a way to check code for this sort of problem. -fsanitize achives this.

RP2350's RISC-V cores fault on a misaligned load or store where the Arm cores handle it in hardware, so code that works on Arm can hang on RISC-V.

Also reading from address zero is bad as it does not fault, it returns bootrom contents and execution continues.

Add UBSan handlers and two CMake flags to enable the corresponding checks across a build:

PICO_UBSAN_ALIGNMENT_CHECKS
PICO_UBSAN_NULL_CHECKS
PICO_UBSAN_RECOVER - report and continue rather than panicking

Debug only. Costs roughly 10% of code size, so it is a development aid, off by default.

### DIFF
--- a/pico_sdk_init.cmake
+++ b/pico_sdk_init.cmake
@@ -46,6 +46,41 @@ if (NOT TARGET _pico_sdk_pre_init_marker)
             message(WARNING "pico_sdk_init() should be called after the project is created (and languages added)")
         endif()
         add_subdirectory(${PICO_SDK_PATH} pico-sdk)
+
+        pico_apply_ubsan_options()
+    endmacro()
+
+    # PICO_CMAKE_CONFIG: PICO_UBSAN_ALIGNMENT_CHECKS, Check every load and store for correct alignment at runtime, type=bool, default=0, group=build
+    # PICO_CMAKE_CONFIG: PICO_UBSAN_NULL_CHECKS, Check every load and store for a null pointer at runtime, type=bool, default=0, group=build
+    # PICO_CMAKE_CONFIG: PICO_UBSAN_RECOVER, Report and continue rather than panicking on the first failure, type=bool, default=0, group=build
+    macro(pico_apply_ubsan_options)
+        set(_pico_ubsan_checks "")
+        if (PICO_UBSAN_ALIGNMENT_CHECKS)
+            list(APPEND _pico_ubsan_checks alignment)
+        endif()
+        if (PICO_UBSAN_NULL_CHECKS)
+            list(APPEND _pico_ubsan_checks null)
+        endif()
+        if (_pico_ubsan_checks)
+            # Refuse to apply checks in release builds
+            if (CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
+                message(FATAL_ERROR "UBSan checks are not supported with"
+                        " CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}: Use CMAKE_BUILD_TYPE=Debug, or"
+                        " drop PICO_UBSAN_ALIGNMENT_CHECKS/PICO_UBSAN_NULL_CHECKS.")
+            endif()
+            list(JOIN _pico_ubsan_checks "," _pico_ubsan_list)
+            get_property(_pico_ubsan_announced GLOBAL PROPERTY PICO_UBSAN_ANNOUNCED)
+            if (NOT _pico_ubsan_announced)
+                message("Enabling UBSan checks: ${_pico_ubsan_list}")
+                set_property(GLOBAL PROPERTY PICO_UBSAN_ANNOUNCED 1)
+            endif()
+            add_compile_options(-fsanitize=${_pico_ubsan_list})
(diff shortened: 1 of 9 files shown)


peterharperuk:
@kilograham What do you think of this? I'm thinking it would be useful to enable for pico-examples testing

test/pico_ubsan_test/pico_ubsan_align_test.c
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    printf("pico_ubsan_align_test\n");
+
+    // A correctly aligned access through the same code path must not trip the
+    // check
+    offset = 4;
+    void *aligned = &buf[offset];
+    sink = *(uint32_t *)aligned;
+    printf("aligned load at offset 4 returned 0x%08x\n", (unsigned)sink);
```
lurch, line 38: In both here and line 26 of `pico_ubsan_null_test.c` is it also worth displaying what the _expected_ value is? (I've not run this code myself, but if my maths is right I think it'd be `0x07060504`?)

test/pico_ubsan_test/pico_ubsan_null_test.c
```diff
@@ -0,0 +1,43 @@
...
+    printf("pico_ubsan_null_test\n");
+
+    // A non-null load through the same code path must not trip the check.
+    ptr = (uint32_t *)&value;
+    uint32_t ok = *ptr;
+    printf("non-null load returned 0x%08x\n", (unsigned)ok);
+
+    ptr = NULL;
+    printf("loading through a null pointer -- expecting a ubsan panic now\n");
+    stdio_flush();
```
lurch, line 30: Does `pico_ubsan_align_test.c` also need a `stdio_flush()` ?

pico_sdk_init.cmake
```diff
@@ -46,6 +46,41 @@ if (NOT TARGET _pico_sdk_pre_init_marker)
             message(WARNING "pico_sdk_init() should be called after the project is created (and languages added)")
         endif()
         add_subdirectory(${PICO_SDK_PATH} pico-sdk)
+
+        pico_apply_ubsan_options()
+    endmacro()
+
+    # PICO_CMAKE_CONFIG: PICO_UBSAN_ALIGNMENT_CHECKS, Check every load and store for correct alignment at runtime, type=bool, default=0, group=build
+    # PICO_CMAKE_CONFIG: PICO_UBSAN_NULL_CHECKS, Check every load and store for a null pointer at runtime, type=bool, default=0, group=build
+    # PICO_CMAKE_CONFIG: PICO_UBSAN_RECOVER, Report and continue rather than panicking on the first failure, type=bool, default=0, group=build
```
lurch, line 55:
```suggestion
    # PICO_CMAKE_CONFIG: PICO_UBSAN_RECOVER, Report and continue rather than panicking on the first ubsan failure, type=bool, default=0, group=build
```
Perhaps this wording-tweak makes it clearer that this setting is ignored if both `PICO_UBSAN_ALIGNMENT_CHECKS` and `PICO_UBSAN_NULL_CHECKS` are false?

src/rp2_common/pico_ubsan/ubsan.c
```diff
@@ -0,0 +1,117 @@
...
+#define TYPE_CHECK_KIND_STORE   1
+
+// UBSan calls one of two entry points depending on whether the check was
+// compiled as recoverable.
+//
+//   __ubsan_handle_type_mismatch_v1        must return; the program then goes
+//                                          ahead and performs the access
+//   __ubsan_handle_type_mismatch_v1_abort  must not return
+
+// PICO_CONFIG: PICO_UBSAN_MAX_REPORTS, Maximum number of reports the recoverable UBSan handler will print before falling silent, min=1, default=20, group=pico_ubsan
```
lurch, line 51: I know nothing about ubsan, but I wonder if it's worth having a special value (e.g. `-1`) to indicate "never fall silent"?

src/rp2_common/pico_ubsan/ubsan.c
```diff
@@ -0,0 +1,117 @@
...
+struct mismatch {
+    const char *file;
+    unsigned line;
+    const char *type_name;
+    const char *op;
+    uintptr_t ptr;
+    unsigned alignment;
+};
+
+PICO_UBSAN_HANDLER static struct mismatch describe(void *_data, void *_ptr) {
```
lurch, line 67: I guess that only one of `__ubsan_handle_type_mismatch_v1` or `__ubsan_handle_type_mismatch_v1_abort` will get used, so is it worth declaring `describe` as `inline`?

Wren6991 (changes requested):
Putting aside that this appears to be 100% LLM-generated: Armv8-M already has a hardware feature for this.

Image

I don't think we should add a library to support compiler instrumentation for something that hardware can already check if you set a register bit.

peterharperuk:
Interesting. M33_CCR_UNALIGN_TRP_BITS. That's a lot simpler.

PR #3140, opened 2026-08-16
Correct description for doorbell functions

aaronmdjones:
The first couple of doorbell functions were using copy-pasted documentation from the alarm functions.

### DIFF
--- a/src/rp2_common/pico_multicore/include/pico/multicore.h
+++ b/src/rp2_common/pico_multicore/include/pico/multicore.h
@@ -314,10 +314,10 @@ static inline void check_doorbell_num_param(__unused uint doorbell_num) {
     invalid_params_if(PICO_MULTICORE, doorbell_num >= NUM_DOORBELLS);
 }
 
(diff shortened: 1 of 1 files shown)


lurch:
The doxygen for `multicore_doorbell_claim_unused` also says "\return the doorbell number claimed or -1 if required was false, and none are available", but looking at the [implementation](https://github.com/raspberrypi/pico-sdk/blob/develop/src/rp2_common/pico_multicore/multicore.c#L427) it actually returns `PICO_ERROR_INSUFFICIENT_RESOURCES` in the failure mode (which has the [value](https://github.com/raspberrypi/pico-sdk/blob/develop/src/common/pico_base_headers/include/pico/error.h#L33) `-9`), so IMHO the doxygen ought to be updated to say "\return the doorbell number claimed or PICO_ERROR_INSUFFICIENT_RESOURCES if required was false, and none are available".

aaronmdjones:
Good catch!  Updated.

kilograham (approved):

Issue #3106, opened 2026-07-31
Cannot complete a bt connection after last update to v2.3.0


elrir01:
Hello,

I am building a device that streams Bluetooth audio using the A2DP protocol over I2S to a DAC, in order to play music from my phone to a set of speakers. I am using the **Pimoroni Pico 2 W Plus**.

For this project, I based my code on the btstack example a2dp_sink_demo.c found in pico-examples. I am using FreeRTOS to manage the various tasks, and I have a dedicated task for the btstack loop.

Anyway, the issue I am experiencing can also be reproduced with just the aforementioned example; in fact, the Bluetooth handling code in my project is almost identical.

The problem is that I cannot connect to the phone after updating to the latest version of the pico-sdk (v2.3.0), and consequently to btstack version 1.8.2. (*The btstack changelog mentions an increased security level in this release.*) Let me explain further: the pairing between the device and the phone, when initiating the request from the phone, is successful and I am able to connect; however, after a reboot using the following function
```c
a2dp_sink_establish_stream(device_addr, &a2dp_sink_demo_a2dp_connection.a2dp_cid); 
```
where `device_addr` is the MAC address of my phone a Pixel 8a,
as you can see in the log **AudioPico_Pixel_connection_failed.pklg**, it returns:
```
357	-31622388.284000    0.000000	PKTLOG	59	A2DP  Sink : Streaming connection failed, status 0x6a
```
After several tests and attempts, my guess is that the process might be failing after authentication when requesting to raise the security level at the following line:
```
306	-31622392.284000	0.009000			PKTLOG	91	hci.c.9154: gap_request_security_level requested level 2, planned level 0, current level 0

```
Looking at the log from the phone's perspective in the **Pixel_AudioPico_connection_failed.pcapng** file, after the line below, the Pico sends the disconnect command. I assume because the security levels do not match, meaning level 2 is requested but the current level is 0:
```
57	0.970892000	0.049205000	controller	host	HCI_EVT	12	Rcvd Encryption Change [v2]
```
The same code works with pico-sdk 2.2.0.
[...]

PR #3143, opened 2026-08-17
Add some btstack tests

peterharperuk:
This adds some firmware based on spp_and_gatt_counter and a python script using bumble to talk to a pico device. It checks bonding works over BR/EDR bond, BR/EDR bond + CTKD over BR/EDR and LE SMP bond and re-encryption.

See #3106


peterharperuk:
This is more of a smoke test for https://github.com/raspberrypi/pico-sdk/issues/3106
We're still waiting for the btstack fix for this - but this script reproduces the problem.

test/pico_btstack/bond/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+endif()
+
+# pico_btstack sets PICO_BTSTACK_PATH in its own directory scope, so work it out
+# again here rather than relying on it leaking.
+if (DEFINED ENV{PICO_BTSTACK_PATH} AND (NOT PICO_BTSTACK_PATH))
+    set(PICO_BTSTACK_PATH $ENV{PICO_BTSTACK_PATH})
+endif()
+if (NOT PICO_BTSTACK_PATH)
+    set(PICO_BTSTACK_PATH ${PICO_SDK_PATH}/lib/btstack)
+endif()
```
will-v-pi: `PICO_BTSTACK_PATH` is explicitly registered as a common scope variable, so it should be available here, so this code can be removed (it's not "relying on it leaking", there are calls to `pico_register_common_scope_var` and `pico_promote_common_scope_vars` to ensure it is propogated)

peterharperuk: ah - you're right. I did have a look and couldn't spot it calling register.

test/pico_btstack/bond/bond_dut.c
```diff
@@ -0,0 +1,351 @@
...
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holders nor the names of
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ * 4. Any redistribution, use, or modification is done solely for
+ *    personal benefit and not for any commercial purpose or for
+ *    monetary gain.
```
will-v-pi: I don't think this file can be included directly in the SDK with this licensing? Even with our custom license in src/rp2_common/pico_btstack/LICENSE.RP which allows commercial use with some of our products, the list there does not include non-W boards, so purchasers of non-W boards (or bare chips) may not be permitted to redistribute, use, or modify this file?

test/pico_btstack/bond/CMakeLists.txt
peterharperuk: And we don't need it anyway. An earlier version included the source from btstack.

test/pico_btstack/bond/bond_dut.c
peterharperuk: Yes, annoying

peterharperuk:
I'm going to put this in another repo

PR #3145, opened 2026-08-18
Add source file missing from btstack.BUILD

peterharperuk:
These files are in the cmake but missing from bazel.


kilograham (approved):

kilograham:
if you want this in 2.3.1 then put a milestone and merge it

peterharperuk:
Well no one has noticed the issue, but I guess it can't do any harm.

lurch:
@armandomontanez Could the bazel-checking scripts be modified to also detect whichever inconsistency is being fixed in this PR?

PR #3149, opened 2026-08-19
Add -ftls-model=local-exec to embedded builds

kilograham:


kilograham:
fixes #3112

test/pico_thread_local_test/pico_thread_local_test_extern.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _PICO_THREAD_LOCAL_TEST_EXTERN_H
+#define _PICO_THREAD_LOCAL_TEST_EXTERN_H
+
+// A thread local defined in a second translation unit. The compiler can see that a thread local
+// defined in the same file is local to the executable and reaches it straight off the thread
+// pointer; one it only has a declaration for, it cannot, so it goes via a GOT entry instead
```
Wren6991: Trying to parse the LLM babble: this seems to insinuate you still get a GOT indirection for TLS in a different TU. I thought the whole point of forcing this TLS model was to *not* do that, and always emit `tp`-relative accesses.

bazel/toolchain/BUILD.bazel
```diff
@@ -59,6 +59,15 @@ cc_args(
 # compiler include directories from resolving to absolute paths. Prefer to use
 # :bazel_no_absolute_paths, since it correctly guides based on the current
 # compiler type.
+# Without this, a thread local that a translation unit only has a declaration for is reached
```
lurch: It looks like this has been inserted in the wrong place? Line 58 (before this insertion) describes the `:no_canonical_system_headers` and `:no_canonical_prefixes` settings, and line 72 (after this insertion) is adding the `no_canonical_system_headers` setting and line 78 is adding the `no_canonical_prefixes` setting.

lurch:
What does "embedded builds" mean in this context?

test/pico_thread_local_test/pico_thread_local_test_extern.h
kilograham: it did mention the fact that this only happens with the old tls-model ... but it is something worth testing works anyway - i have modified the comment

bazel/toolchain/BUILD.bazel
kilograham: yup i have moved it up

Wren6991:
> What does "embedded builds" mean in this context?

I would guess non-`host` builds, aka `PICO_ON_DEVICE=1`

test/pico_thread_local_test/pico_thread_local_test_extern.h
Wren6991: Thanks, the one you wrote is much better.

Wren6991 (approved):

test/pico_thread_local_test/pico_thread_local_test.c
```diff
@@ -13,29 +13,35 @@
...
 int core1_count;
+int core1_extern;
 
 semaphore_t sem;
 
 int do_count(int delta) {
     for (int i=0;i<COUNT;i++) {
         counter += delta;
+        // the same thing again, but on a thread local this TU only has a declaration for
+        extern_bump(delta);
```
lurch, line 31: Was this supposed to be 
```suggestion
        counter += extern_bump(delta);
```
(Or if not, should `extern_bump` be declared as returning `void` instead of `int`?)

kilograham: the test is fine - could return void but i don't care enough ;-)

PR #3150, opened 2026-08-20
some comment cleanup

kilograham:


Wren6991 (approved):

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
```diff
@@ -17,7 +17,19 @@
...
+//
+// On RP2350, writes to SIO register offsets +0x180..0x1fc spuriously release the spinlock addressed
+// 0x80 bytes lower (erratum RP2350-E2). By default we use software spinlocks to work around this.
+// Otherwise (PICO_USE_SW_SPIN_LOCKS=0) we change the default numbering to avoid locks which alias
+// with writable registers: 5-7, 10-11 and 26-31 have no register at that offset, and 18-25 alias
+// the read-only TMDS PEEK/POP registers. The rest alias a writable register: 0-4
+// DOORBELL_*/PERI_NONSEC, 8-9 RISCV_SOFTIRQ/MTIME_CTRL, 12-15 MTIME*, 16-17 TMDS_CTRL/WDATA.
+//
+// On RP2040 we preserve the original default lock numbers existing since SDK v1.0, since they are a
+// compatibility detail that out-of-tree code might rely on.
```
lurch, line 30: The above paragraph explains why/when spinliock numbers are different on RP2350. I'm not sure we need this "On RP2040 we preserve the original default lock numbers" sentence?

src/rp2_common/pico_runtime_init/include/pico/runtime_init.h
```diff
@@ -259,7 +259,7 @@ void runtime_init_post_clock_resets(void);
 // RP2040 IE disable for GPIO 26-29
 // ----------------------------------------
 
-// PICO_CONFIG: PICO_RUNTIME_SKIP_INIT_RP2040_GPIO_IE_DISABLE, Skip calling of `runtime_init_rp2040_gpio_ie_disable` function during runtime init, type=bool, default=0 on RP2040, group=pico_runtime_init
+// PICO_CONFIG: PICO_RUNTIME_SKIP_INIT_RP2040_GPIO_IE_DISABLE, Skip calling of `runtime_init_rp2040_gpio_ie_disable` function during runtime init, type=bool, default=0 on RP2040 unless PICO_IE_26_29_UNCHANGED_ON_RESET is 1, group=pico_runtime_init
```
lurch, line 262: This "0 on RP2040 unless..." wording gives no hint as to what the default value might be on RP2350, so perhaps it's better to explicitly state exactly when it **is** 0 (with the implication then being that it's 1 in any other situation)? I.e.
```suggestion
// PICO_CONFIG: PICO_RUNTIME_SKIP_INIT_RP2040_GPIO_IE_DISABLE, Skip calling of `runtime_init_rp2040_gpio_ie_disable` function during runtime init, type=bool, default=0 on RP2040 when PICO_IE_26_29_UNCHANGED_ON_RESET is unset, group=pico_runtime_init
```

P.S. Looks like we might also need a `PICO_CONFIG:` entry for `PICO_IE_26_29_UNCHANGED_ON_RESET` ?
```console
$ git grep PICO_IE_26_29_UNCHANGED_ON_RESET
src/rp2_common/pico_runtime_init/include/pico/runtime_init.h:#if !PICO_RP2040 || PICO_IE_26_29_UNCHANGED_ON_RESET
src/rp2_common/pico_runtime_init/runtime_init.c:#if PICO_RP2040 && !PICO_IE_26_29_UNCHANGED_ON_RESET
```

src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h
Wren6991: otherwise we would have just changed the default for all platforms

src/rp2_common/pico_runtime_init/include/pico/runtime_init.h
kilograham: unset is unclear since it is unset or 0... feel free to make a PR/issue for adding `PICO_IE_26_20_UNCHANGED_ON_RESET`

lurch: LOL. I was originally going to write "...is 0" since that's more technically accurate, but I thought you'd prefer the "is unset" wording :laughing:

lurch: I created an issue ( #3153 ) rather than a PR, because I've no idea what the `description` for `PICO_IE_26_20_UNCHANGED_ON_RESET` should be (nor where would be a sensible place to add it).

PR #3152, opened 2026-08-21
Add support for configuring XOSC FREQ_RANGE

### LINKED ISSUE #3074: hardware_xosc: Hardcoded frequency range in xosc_init() prevents using >15MHz crystals on RP2350 without modifying the SDK
When trying to configure pico-sdk to use a 32MHz crystal as the clock source for XOSC I found that `xosc_init()` in [xosc.c](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_xosc/xosc.c#L29-L43)  hardcodes the XOSC CTRL register to XOSC_CTRL_FREQ_RANGE_VALUE_1_15MHZ. This cannot be overridden without modifying the SDK code.

It would be incredibly helpful if the frequency range in `xosc_init()` could be made configurable via a preprocessor definition, similar to other clock configurations in the SDK.

For example, introducing a define like:
```c
#ifndef PICO_XOSC_FREQ_RANGE
#define PICO_XOSC_FREQ_RANGE XOSC_CTRL_FREQ_RANGE_VALUE_1_15MHZ
#endif
```
so developers can override this behavior in their board.h files.

will-v-pi:
Add PICO_XOSC_FREQ_RANGE_MAX config for non-RP2040 chips - defaults to XOSC_HZ / 1000000, but can be configured higher (e.g. to select the 40-100MHz range for a 50MHz crystal)

RP2040 only has one range, so this config is not defined for RP2040, as CMOS clock inputs >15MHz are still supported despite being out of the XOSC FREQ_RANGE value

Fixes #3074

### DIFF
--- a/src/rp2_common/hardware_xosc/include/hardware/xosc.h
+++ b/src/rp2_common/hardware_xosc/include/hardware/xosc.h
@@ -10,6 +10,9 @@
 #include "pico.h"
 #include "hardware/structs/xosc.h"
 
+// For frequency related definitions etc
+#include "hardware/clocks.h"
+
 
 // Allow lengthening startup delay to accommodate slow-starting oscillators
 
@@ -18,12 +21,12 @@
 #define PICO_XOSC_STARTUP_DELAY_MULTIPLIER 6
 #endif
 
-// PICO_CONFIG: PICO_XOSC_FREQ_RANGE_MAX, The maximum frequency in MHz for the XOSC to support (not required when using CMOS clock input instead of XOSC) - defining this variable is not supported on RP2040 as it only has one range, type=int, min=1, max=100, default=XOSC_HZ/1000000, group=hardware_xosc
+// PICO_CONFIG: PICO_XOSC_FREQ_RANGE_MAX, The maximum frequency in MHz for the XOSC to support (not required when using CMOS clock input instead of XOSC) - defining this variable is not supported on RP2040 as it only has one range, type=int, min=1, max=100, default=XOSC_HZ/MHZ, group=hardware_xosc
 #if PICO_RP2040
 #define PICO_XOSC_FREQ_RANGE_MAX 15
 #else
 #ifndef PICO_XOSC_FREQ_RANGE_MAX
-#define PICO_XOSC_FREQ_RANGE_MAX (XOSC_HZ / 1000000)
+#define PICO_XOSC_FREQ_RANGE_MAX (XOSC_HZ / MHZ)
 #endif
 #endif
 
--- a/src/rp2_common/hardware_xosc/include/hardware/xosc.h
+++ b/src/rp2_common/hardware_xosc/include/hardware/xosc.h
@@ -18,6 +18,15 @@
 #define PICO_XOSC_STARTUP_DELAY_MULTIPLIER 6
 #endif
 
+// PICO_CONFIG: PICO_XOSC_FREQ_RANGE_MAX, The maximum frequency in MHz for the XOSC to support (not required when using CMOS clock input instead of XOSC) - defining this variable is not supported on RP2040 as it only has one range, type=int, min=1, max=100, default=XOSC_HZ/1000000, group=hardware_xosc
+#if PICO_RP2040
+#define PICO_XOSC_FREQ_RANGE_MAX 15
+#else
+#ifndef PICO_XOSC_FREQ_RANGE_MAX
+#define PICO_XOSC_FREQ_RANGE_MAX (XOSC_HZ / 1000000)
+#endif
+#endif
+
 
 #ifdef __cplusplus
 extern "C" {
--- a/src/rp2_common/hardware_xosc/xosc.c
+++ b/src/rp2_common/hardware_xosc/xosc.c
@@ -27,8 +27,17 @@
 #endif
 
 void xosc_init(void) {
-    // Assumes 1-15 MHz input, checked above.
+#if PICO_RP2040 || PICO_XOSC_FREQ_RANGE_MAX <= 15 // RP2040 only has this range
     xosc_hw->ctrl = XOSC_CTRL_FREQ_RANGE_VALUE_1_15MHZ;
+#elif PICO_XOSC_FREQ_RANGE_MAX <= 30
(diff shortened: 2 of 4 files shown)


src/rp2_common/hardware_xosc/xosc.c
```diff
@@ -27,8 +27,17 @@
...
+#if PICO_RP2040 || PICO_XOSC_FREQ_RANGE_MAX <= 15 // RP2040 only has this range
     xosc_hw->ctrl = XOSC_CTRL_FREQ_RANGE_VALUE_1_15MHZ;
+#elif PICO_XOSC_FREQ_RANGE_MAX <= 30
+    xosc_hw->ctrl = XOSC_CTRL_FREQ_RANGE_VALUE_10_30MHZ;
+#elif PICO_XOSC_FREQ_RANGE_MAX <= 60
+    xosc_hw->ctrl = XOSC_CTRL_FREQ_RANGE_VALUE_25_60MHZ;
+#elif PICO_XOSC_FREQ_RANGE_MAX <= 100
+    xosc_hw->ctrl = XOSC_CTRL_FREQ_RANGE_VALUE_40_100MHZ;
+#else
+    #error PICO_XOSC_FREQ_RANGE_MAX is too large, maximum is 100
```
lurch, line 39: This might be a stupid question, but why does this allow `PICO_XOSC_FREQ_RANGE_MAX` values up to 100 MHz, when the check on line 16 of this same file caps `XOSC_HZ` at 50 MHz?

lurch:
It's _very slightly_ OffTopic, but maybe this PR would also be a good place to edit `src/rp2040/hardware_regs/include/hardware/platform_defs.h` and `src/rp2350/hardware_regs/include/hardware/platform_defs.h` to add `min=1000000, max=50000000,` to the `PICO_CONFIG: XOSC_HZ` lines? :slightly_smiling_face:

src/rp2_common/hardware_xosc/include/hardware/xosc.h
```diff
@@ -18,6 +18,15 @@
 #define PICO_XOSC_STARTUP_DELAY_MULTIPLIER 6
 #endif
 
+// PICO_CONFIG: PICO_XOSC_FREQ_RANGE_MAX, The maximum frequency in MHz for the XOSC to support (not required when using CMOS clock input instead of XOSC) - defining this variable is not supported on RP2040 as it only has one range, type=int, min=1, max=100, default=XOSC_HZ/1000000, group=hardware_xosc
+#if PICO_RP2040
+#define PICO_XOSC_FREQ_RANGE_MAX 15
+#else
+#ifndef PICO_XOSC_FREQ_RANGE_MAX
+#define PICO_XOSC_FREQ_RANGE_MAX (XOSC_HZ / 1000000)
```
lurch:
```suggestion
#define PICO_XOSC_FREQ_RANGE_MAX (XOSC_HZ / MHZ)
```
?

src/rp2_common/hardware_xosc/xosc.c
will-v-pi: Because the ranges are:
- 1-15
- 10-30
- 25-60
- 40-100

so you need to be able to specify values above 50 to pick the final range, if you want that range instead of the 25-60 one

src/rp2_common/hardware_xosc/include/hardware/xosc.h
will-v-pi: Yes, this already links `hardware_clocks`, so I can include `hardware/clocks.h` and use `MHZ`

will-v-pi:
> It's _very slightly_ OffTopic, but maybe this PR would also be a good place to edit `src/rp2040/hardware_regs/include/hardware/platform_defs.h` and `src/rp2350/hardware_regs/include/hardware/platform_defs.h` to add `min=1000000, max=50000000,` to the `PICO_CONFIG: XOSC_HZ` lines? 🙂

Thanks, added - I didn't add a note on RP2040 that values above 15MHz are only supported when using CMOS clock inputs, as that is very much a board design thing rather than an SDK thing

src/rp2_common/hardware_xosc/xosc.c
lurch: Okay, next stupid question: If using a 50 MHz frequency, is there any reason for choosing a range of 40-100 vs. choosing a range of 25-60?

EDIT: The reason I'm asking, is because I wonder if this could instead be calculated from the existing `XOSC_HZ` define, rather than requiring the addition of a new `PICO_XOSC_FREQ_RANGE_MAX` define?

EDIT2: Although if we _do_ need both a `XOSC_HZ` and `PICO_XOSC_FREQ_RANGE_MAX`, maybe there could / should also be a check that `XOSC_HZ  <= (PICO_XOSC_FREQ_RANGE_MAX * MHZ)` ? :thinking:

will-v-pi: I'm not sure, you would have to ask the ASIC team (@davidb990 might know?)

davidb990: I can't spot anything in the datasheet to say you should pick one range over the other. 

FWIW, the frequency ranges are used as a proxy for drive strength, so the choice ought to depend on how difficult the crystal is to drive - there could be a case where one 50 MHz crystal is better served by the 40-100 range, and another is better served by the 25-60 range (so perhaps leaving the choice up to the user to select based on their implementation is marginally preferable?)

lurch: > there could be a case where one 50 MHz crystal is better served by the 40-100 range, and another is better served by the 25-60 range (so perhaps leaving the choice up to the user to select based on their implementation is marginally preferable?)

Huh, fair enough :+1: 

But that still leaves my EDIT2 question above un-answered :wink:

will-v-pi: I think it’s fine to not check it, because if you’re explicitly setting it lower than your XOSC_HZ you may have a reason (eg wanting lower drive strengths because your crystal works with them, or you’re using a CMOS clock input and want a lower setting, or something)

kilograham (approved):

PR #3154, opened 2026-08-22
add cytron motion 2350 pro board 

ItsJupiter21:
Adds the  cytron motion 2350 board.

pinout:
image

datasheet:
https://docs.google.com/document/d/1qYD0RdZG1tiklflnt8Z7jBy-wUJ3LbNBJ4aT1EgnjqQ/edit?usp=sharing

`tools/check_board_header.py src/boards/include/boards/cytron_motion_2350_pro.h` runs without errors.

### DIFF
--- a/src/boards/include/boards/cytron_motion_2350_pro.h
+++ b/src/boards/include/boards/cytron_motion_2350_pro.h
@@ -94,6 +94,9 @@ pico_board_cmake_set(PICO_PLATFORM, rp2350)
 #define PICO_DEFAULT_UART_RX_PIN 1
 #endif
 
+
+#define MOTION_2350_PRO_BUZZER_PIN 22
+
 // --- LED ---
 #ifndef PICO_DEFAULT_WS2812_PIN
 #define PICO_DEFAULT_WS2812_PIN 23
--- a/src/boards/include/boards/cytron_motion_2350_pro.h
+++ b/src/boards/include/boards/cytron_motion_2350_pro.h
@@ -1,0 +1,163 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// -----------------------------------------------------
+// NOTE: THIS HEADER IS ALSO INCLUDED BY ASSEMBLER SO
+//       SHOULD ONLY CONSIST OF PREPROCESSOR DIRECTIVES
+// -----------------------------------------------------
+
+
+#ifndef _BOARDS_CYTRON_MOTION_2350_PRO_H
+#define _BOARDS_CYTRON_MOTION_2350_PRO_H
+
+
+pico_board_cmake_set(PICO_PLATFORM, rp2350)
+
+// For board detection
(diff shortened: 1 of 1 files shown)


src/boards/include/boards/cytron_motion_2350_pro.h
```diff
@@ -0,0 +1,163 @@
...
+#define CYTRON_MOTION_2350_PRO
+
+// --- RP2350 VARIANT ---
+#define PICO_RP2350A 1
+
+
+// --- BOARD SPECIFIC ---
+
+// Motor drivers
+#ifndef MOTION_2350_PRO_M1A_PIN
```
lurch, line 29: I'm not sure if these board-specific pins need to be inside `#ifndef` guards, because it seems pretty unlikely that a user would reassign e.g. `MOTION_2350_PRO_M1A_PIN` to a different pin-number? :thinking:

src/boards/include/boards/cytron_motion_2350_pro.h
```diff
@@ -0,0 +1,163 @@
...
+
+// --- LED ---
+#ifndef PICO_DEFAULT_WS2812_PIN
+#define PICO_DEFAULT_WS2812_PIN 23
+#endif
+
+// This board has two WS2812 leds in series
+#ifndef PICO_DEFAULT_WS2812_NUM_PIXELS
+#define PICO_DEFAULT_WS2812_NUM_PIXELS 2
+#endif
```
lurch, line 108: Interesting. @peterharperuk How does your status_led code behave if there's two WS2812 LEDs connected to `PICO_DEFAULT_WS2812_PIN` ? (Does only the first LED change colour, do both LEDs change colour, or do any colour-changes end up getting cascaded down the chain?)

src/boards/include/boards/cytron_motion_2350_pro.h
ItsJupiter21: I copied this behaviour from the maker pi: https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/cytron_maker_pi_rp2040.h , i can remove the guards if thatd be better

lurch:
Looking at your pinout diagram, is it worth also adding some `#define` values for the buzzer and the "maker ports"?

ItsJupiter21:
Added the buzzer pin. Dont think there's any value in adding the maker ports since they're only labelled with the pin numbers.

src/boards/include/boards/cytron_motion_2350_pro.h
lurch: I _personally_ don't think that they're needed, but they obviously don't do any harm, so I'm happy to leave them in there.

lurch (approved):

src/boards/include/boards/cytron_motion_2350_pro.h
peterharperuk: No sure - I'd have to test it. This might be the first board that doesn't define a separate LED pin?

peterharperuk: Seems to behave fine

PR #3156, opened 2026-08-24
Fix race in RP2040 low_power_test_timers_gpio

will-v-pi:
Fix race in `external_sleep_timer` which would put the pin low after waking up, so there would be no edge when it gets put low again - this race was hit in the `low_power_test_timers_gpio` test on RP2040


lurch:
I guess the "// Can take longer to wake up that Arm" comment on line 17 ought to say "// Can take longer to wake up than Arm" ?

kilograham (approved):

PR #3157, opened 2026-08-24
Make sure CFG_TUSB_OS is defined

peterharperuk:
When building the USB examples with a newer version of tinyusb it's possible for CFG_TUSB_OS to be set wrongly.

### DIFF
--- a/src/rp2_common/pico_stdio_usb/include/tusb_config.h
+++ b/src/rp2_common/pico_stdio_usb/include/tusb_config.h
@@ -29,6 +29,15 @@
 
 #include "pico/stdio_usb.h"
 
+// CFG_TUSB_OS=OPT_OS_PICO is set by tinyusb_common_base
+// Newer versions of tinyusb remove it again in family_configure_example
+// Presumably this is to avoid issues when setting CFG_TUSB_OS=OPT_OS_FREERTOS
+// Because of this CFG_TUSB_OS can end up being set to OPT_OS_NONE
+// Make sure CFG_TUSB_OS=OPT_OS_PICO if it's undefined
+#ifndef CFG_TUSB_OS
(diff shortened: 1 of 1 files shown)


kilograham:
i somewhat dislike this as it works around a slightly dubious change in TinyUSB (which makes a global change in an example scoped configuration).

That said, in my testing i haven't needed this at least with #3163 (can you try again with that and my latest PR from pico-examples - https://github.com/raspberrypi/pico-examples/pull/787)

kilograham:
That said, it seems like it ought to be necessary, given that the configuration of a freertos example should then remove this flag, so it may just be luck (or rather that the order of pico-example building has no stdio_usb examples after the tinyusb freertos examples)... so i'm leaning back towards merging this (without the hand-wavey speculation in the comments)

Note i did say dubious above, but it has always been slightly dubious that SDK tinyusb_base or whatever sets this unequivocally. I'm still a bit in the dark as to what it means to use TinyUSB w/FreeRTOS on Pico (I assume setting the RTOS away from OS_PICO is fine and likely required in this case, in which case we _should_ keep **this** change, but also look for a better, more configurable way to apply the default of OS_PICO for the common SDK tinyusb case - doing it via CMake might be vestigial, and could be potentially fixed by an upstream TinyUSB change. failing that we should do it in such away that it can be overriden per TARGET at least)

Side note, it'd be helpful to have a version number define coming from TinyUSB in CMake

cc @hathach for my musings

kilograham:
moved this back into 2.3.2 since

1. it seems like the change ought to be within the existing guard for whether TinyUSB is being used directly
2. i'm a bit surprised that we always add our own `tusb_config.h` to the include path here. this likely predates `pico_stdio_usb` being used in conjunction with user use of TinyUSB. I guess we just happen to be fiurther down the include path!?

PR #3161, opened 2026-08-27
Add Pimoroni board headers

Gadgetoid:
Add all the Pimoroni boards we've neglected to upstream for the last forever. Draft because there might be more, or I might have bungled this.

@ZodiusInfuser I did a full sweep of everything we're currently missing, and might have got a bit carried away. Let me know if any of these should be dropped.


src/boards/include/boards/pimoroni_presto.h
```diff
@@ -0,0 +1,155 @@
...
+
+pico_board_cmake_set(PICO_PLATFORM, rp2350)
+pico_board_cmake_set(PICO_CYW43_SUPPORTED, 1)
+
+// For board detection
+#define PIMORONI_PRESTO
+
+// --- BOARD SPECIFIC ---
+// 480x480 display on a 16 bit parallel interface, data on GPIO 1 through 16
+#define PRESTO_LCD_D0_PIN 1
```
lurch: Just to be completely unambiguous, it might be worth specifying _all_ of these GPIOs? E.g.
```c
#define PRESTO_LCD_D0_PIN 1
#define PRESTO_LCD_D1_PIN 2
#define PRESTO_LCD_D2_PIN 3
#define PRESTO_LCD_D3_PIN 4
// etc.
```
(This might also help indicate "Don't use these GPIOs for anything else".)

src/boards/include/boards/pimoroni_presto.h
```diff
@@ -0,0 +1,155 @@
...
+
+// --- UART ---
+#ifndef PICO_DEFAULT_UART
+#define PICO_DEFAULT_UART 0
+#endif
+#ifndef PICO_DEFAULT_UART_TX_PIN
+#define PICO_DEFAULT_UART_TX_PIN 0
+#endif
+#ifndef PICO_DEFAULT_UART_RX_PIN
+#define PICO_DEFAULT_UART_RX_PIN 1
```
lurch: The same as `PRESTO_LCD_D0_PIN`? Is that right?

src/boards/include/boards/pimoroni_tinyfx.h
```diff
@@ -0,0 +1,138 @@
...
+// For board detection
+#define PIMORONI_TINYFX
+
+// --- BOARD SPECIFIC ---
+#define TINYFX_OUT2_PIN 2
+#define TINYFX_OUT1_PIN 3
+#define TINYFX_OUT3_PIN 4
+#define TINYFX_OUT4_PIN 5
+#define TINYFX_OUT5_PIN 8
+#define TINYFX_OUT6_PIN 9
```
lurch: Probably makes more sense to list the defines in name-order, rather than in value-order? I.e.
```suggestion
#define TINYFX_OUT1_PIN 3
#define TINYFX_OUT2_PIN 2
#define TINYFX_OUT3_PIN 4
#define TINYFX_OUT4_PIN 5
#define TINYFX_OUT5_PIN 8
#define TINYFX_OUT6_PIN 9
```

src/boards/include/boards/pimoroni_tufty2040.h
```diff
@@ -0,0 +1,96 @@
...
+#define TUFTY2040_SW_DOWN_PIN 6
+#define TUFTY2040_SW_A_PIN 7
+#define TUFTY2040_SW_B_PIN 8
+#define TUFTY2040_SW_C_PIN 9
+#define TUFTY2040_SW_UP_PIN 22
+
+#define TUFTY2040_LED_PIN 25
+#define TUFTY2040_BACKLIGHT_PIN 2
+
+// 8 bit parallel display interface, data on GPIO 14 through 21
```
lurch, line 39: See earlier comment :wink:

src/boards/include/boards/pimoroni_presto.h
Gadgetoid: Agreed. Explicit is better than implicit. Thank you! (specially when it's not obvious how many pins there are)

src/boards/include/boards/pimoroni_tinyfx.h
Gadgetoid: My head hurts either way but I think you're right!

src/boards/include/boards/pimoroni_tufty2040.h
Gadgetoid: Thank you! Agreed.

Gadgetoid:
@ZodiusInfuser do you have any thoughts on these? (Yukon and Tiny in particular, since those are your babies)

src/boards/include/boards/pimoroni_presto.h
```diff
@@ -0,0 +1,190 @@
...
+
+pico_board_cmake_set(PICO_PLATFORM, rp2350)
+pico_board_cmake_set(PICO_CYW43_SUPPORTED, 1)
+
+// For board detection
+#define PIMORONI_PRESTO
+
+// --- BOARD SPECIFIC ---
+// 480x480 display on a 16 bit parallel interface, data on GPIO 1 through 16
+// GPIO 17 and 18 carry the two extra bits when the driver runs 18 bit palette mode
```
lurch: ...in which case it's probably worth adding defines for GPIOs 17 and 18 too? These board-header files describe the **hardware** and don't care what the software-driver does or doesn't do :wink:

ZodiusInfuser:
I haven't had a chance to look at this yet, will do so tomorrow.

ZodiusInfuser:
@Gadgetoid I've made some edits across the range if you wish to double check.

kilograham:
FYI; hurry, hurry as 2.3.1 is coming probably tomorrow (Friday)... so if you want this to make that and not just land in develop, then get your changes in, and hope @lurch blesses them!!

Gadgetoid:
*presses the button*

I won’t be able to make any changes today but hopefully Chris’ll pick up any last minute fixes! 👀

src/boards/include/boards/pimoroni_badger2350.h
```diff
@@ -0,0 +1,179 @@
...
+pico_board_cmake_set(PICO_PLATFORM, rp2350)
+pico_board_cmake_set(PICO_CYW43_SUPPORTED, 1)
+
+// For board detection
+#define PIMORONI_BADGER2350
+
+// --- BOARD SPECIFIC ---
+// PCF85063a real time clock
+#define BADGER2350_RTC_I2C 0
+#define BADGER2350_RTC_I2C_ADDR 0x51
```
lurch, line 26: For the other boards that have on-board devices connected to their I2C pins, maybe it'd be useful if their headers included the I2C addresses too? :thinking: 
(I won't let that stop the merging of this PR though)

lurch:
Not _directly_ related to this PR, but whilst adding some additional checks to the `check_board_header.py` script (which still need to be tidied up before PRing), I spotted that https://github.com/raspberrypi/pico-sdk/blob/develop/src/boards/include/boards/pimoroni_pico_plus2_w_rp2350.h is missing a define for `CYW43_PIN_WL_DYNAMIC`, so maybe you'd like to squeeze that fix into this PR too? :wink:

lurch (approved):
Blessed :+1:

Issue #3160, opened 2026-08-25
Use of `PICO_HARD_FLOAT_ABI`


dpgeorge:
In MicroPython we want to enable `-mfloat-abi=hard` for all RP2350 boards when in ARM mode.

But simply adding `set(PICO_HARD_FLOAT_ABI ON)` to our `CMakeLists.txt` forces it to be on for RP2040 as well which leads to a compile error.

On the other hand, we don't know that we are on RP2350 until we have included `${PICO_SDK_PATH}/pico_sdk_init.cmake` and done `pico_sdk_init()`, but by then it's too late to set `PICO_HARD_FLOAT_ABI`.

Maybe I'm missing something simple here... how to enable hard float ABI for only the RP2350 ARM builds?

dpgeorge:
I did find somewhat of a workaround for this, which requires each board to set `PICO_PLATFORM` (and all rp2350 boards already did that):
```
# include board-specific config
# ...

if(PICO_PLATFORM)
    if(${PICO_PLATFORM} STREQUAL "rp2350")
        set(PICO_HARD_FLOAT_ABI ON)
    endif()
endif()

# include and set up pico-sdk
# ...
```

kilograham:
sorted in develop

dpgeorge:
Thank you!

PR #3162, opened 2026-08-29
Ignore PICO_HARD_FLOAT_ABI=1 on RP2040 with warning; add PICO_NO_HARD_FLOAT_ABI_WARNING to suppress

kilograham:


dpgeorge:
Thanks!

cmake/preload/toolchains/util/pico_arm_gcc_common.cmake
```diff
@@ -1,12 +1,18 @@
 # PICO_CMAKE_CONFIG: PICO_HARD_FLOAT_ABI, use hard floating point ABI, type=bool, default=0, group=build, docref=cmake-toolchain-config
+# PICO_CMAKE_CONFIG: PICO_NO_HARD_FLOAT_ABI_WARNING, disable warning when PICO_HARD_FLOAT_ABI is set when not supported, type=bool, default=0, group=build, docref=cmake-toolchain-config
```
will-v-pi: Is it worth adding `(e.g. on RP2040)`?


```suggestion
# PICO_CMAKE_CONFIG: PICO_NO_HARD_FLOAT_ABI_WARNING, disable warning when PICO_HARD_FLOAT_ABI is set when not supported (e.g. on RP2040), type=bool, default=0, group=build, docref=cmake-toolchain-config
```

lurch:
Does this fix #3160 ?

kilograham:
fixes #3160

cmake/preload/toolchains/util/pico_arm_gcc_common.cmake
kilograham: i had it in there but took it out - i guess i'll put it back ;=)

will-v-pi (approved):
Thanks, LGTM

PR #3165, opened 2026-09-01
Fix \param names in interp_get_base and riscv_timer_set_mtimecmp

darkdi:
Was checking doxygen output on the interp and riscv timer headers and hit two \param names that don't match the argument below them. interp_get_base documents `lane`, which looks copied from interp_set_base just above it, and riscv_timer_set_mtimecmp documents `mtime` from its neighbour. Doxygen drops unmatched names quietly so the build stays green either way. I left the mtimecmp wording alone, it still reads as if the call sets mtime and that seemed like your call.

### DIFF
--- a/src/rp2_common/hardware_interp/include/hardware/interp.h
+++ b/src/rp2_common/hardware_interp/include/hardware/interp.h
@@ -358,7 +358,7 @@ static inline void interp_set_base(interp_hw_t *interp, uint index, uint32_t val
  *  \ingroup hardware_interp
  *
  * \param interp Interpolator instance, interp0 or interp1.
- * \param lane The base register index, 0 or 1 or 2
+ * \param index The base register index, 0 or 1 or 2
  * \return  The current content of the base register
  */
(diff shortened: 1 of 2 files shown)


lurch (approved):

lurch:
> Doxygen drops unmatched names quietly

Well that's a bit annoying :slightly_frowning_face:

darkdi:
I was wrong about that. It does warn, I just checked with 1.18 against the defaults `docs/Doxyfile.in` inherits:

```
t.h:2: warning: argument 'lane' of command @param is not found in the argument list of interp_get_base(void *interp, unsigned index)
t.h:2: warning: The following parameter of interp_get_base(void *interp, unsigned index) is not documented:
  parameter 'index'
```

Both of those, and then it exits 0. So the docs job stays green and the two lines scroll past with everything else. `WARN_AS_ERROR = FAIL_ON_WARNINGS` in `docs/Doxyfile.in` would catch the next one without changing how much output you get, it runs the whole build and only sets the exit code at the end.

PR #3168, opened 2026-09-03
disable pico_unaligned_string_test on google llvm < 21 since it doesn…

kilograham:
this test does not build (missing c library funcs prior to 21.0.0)


will-v-pi:
Does it need disabling in Bazel too, given that's where this google llvm was used previously?

will-v-pi (approved):
Approving for now, but might be good to come back later for the Bazel fix in another PR

kilograham:
yeah, though bazel should already be using a newer version

PR #3171, opened 2026-09-03
temp workaround for failed build with no PICO_DEFAULT_UART

kilograham:
Fix broken build when there is no PICO_DEFAULT_UART... note this is a band-aid fix; opening .. #3172 to track

### DIFF
--- a/src/rp2_common/pico_low_power/low_power.c
+++ b/src/rp2_common/pico_low_power/low_power.c
@@ -172,7 +172,7 @@ static void replace_null_enable_values(const clock_dest_bitset_t *keep_enabled,
     }
 }
 
-static void add_library_clocks(clock_dest_bitset_t *local_keep_enabled) {
+static void add_library_clocks(__unused clock_dest_bitset_t *local_keep_enabled) {
 #if LIB_PICO_STDIO_USB || LIB_TINYUSB_HOST || LIB_TINYUSB_DEVICE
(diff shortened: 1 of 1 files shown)


will-v-pi:
A better fix might be to replace the `PICO_DEFAULT_UART` references with `UART_NUM(uart_default)`, as other `LIB_PICO_STDIO_UART` sections rely on `uart_default` always being defined

kilograham:
Well not uart_default is derived from PICO_DEFAULT_UART, but if you are referring to the better fix in the other issue then we can discuss there

src/rp2_common/pico_low_power/low_power.c
will-v-pi: Needs to be marked `__unused` too, otherwise `kitchen_sink` with `PICO_BOARD=none` fails with `error: unused parameter 'local_keep_enabled'`

```suggestion
static void add_library_clocks(__unused clock_dest_bitset_t *local_keep_enabled) {
```

will-v-pi:
> A better fix might be to replace the `PICO_DEFAULT_UART` references with `UART_NUM(uart_default)`, as other `LIB_PICO_STDIO_UART` sections rely on `uart_default` always being defined

I got confused - I thought `uart_default` was being used without `#ifdef uart_default` in `stdio_uart_init`, but had just misread it

will-v-pi (approved):

PR #3187, opened 2026-09-07
Improve docs around AON timer

### LINKED ISSUE #3180: aon_timer_is_running() is true on boot in SDK 2.3.1! Previously the timer would have to be started
Pico SDK 2.3.1 on Pico 2 W
On boot the aon_timer_is_running() is reported as started, yet does not have a "current" calendar date set. Previous versions of the SDK did not have the aon timer running at boot.

Test case:

int main( void )
{
    stdio_init_all();
    sleep_ms(2000);

    if (aon_timer_is_running())
    	printf("AON Timer is running\n");

...
}

Use Case:
I check the aon_timer to see if the SNTP has sync-ed time to the Pico 2 W. The can pull calendar and time stamps from aon.

will-v-pi:
Improve `low_power_start_aon_timer` docs, and refer to them from `low_power_dormant_for_ms` and `low_power_pstate_for_ms`

Also mention that AON timer persists across most warm reboots on RP2350

Fixes #3180


src/rp2_common/pico_low_power/include/pico/low_power.h
```diff
@@ -414,6 +415,9 @@ static inline int low_power_sleep_for_ms(uint32_t ms, const clock_dest_bitset_t
  *  \ingroup pico_low_power
  *
  * See \ref low_power_dormant_until_aon_timer for more information.
+ * 
+ * NOTE: This calls \ref low_power_start_aon_timer() to ensure the
```
lurch, line 419: Based on the comments in the linked issue, should this also mention that the AON timer will also _continue_ to run after wake-up? (i.e. even if it wasn't running before)

will-v-pi: I don't think that's necessary - the `low_power_start_aon_timer` docs should make it clear that the timer is started, and this makes no mention of stopping it after

PR #3192, opened 2026-09-11
Persistent pstate improvements

will-v-pi:
This now determines the persistent pstate at link time and uses that in `low_power_persistent_pstate_get`, rather than calculating it at runtime (it cannot be done at compile time, because it depends on the size and location of the persistent data section).

I've kept the option for runtime calculation by setting `PICO_LOW_POWER_PERSISTENT_PSTATE_STATIC=0` (e.g. custom linker scripts that don't provide the new `__persistent_data_pstate__`), with the logic copied from the linker script to avoid differing logic in the two places. Also added tests to make sure they give the same value.

Add `pstate_bitset_to/from_uint32` for use with scratch registers and with `__persistent_data_pstate__`, rather than `powman_power_state` directly

Also add `pstate_resume_func_common` for common `pstate_resume_func` stuff in tests, rather than duplicating it everywhere


peterharperuk:
Sorry, dumb questions... what's the advantage of doing this at link time rather than runtime? Is PICO_LOW_POWER_PERSISTENT_PSTATE_STATIC a good name. I guess PICO_LOW_POWER_PERSISTENT_PSTATE_RUN_TIME=1 would be more obvious to me (to get back the old behaviour?)

will-v-pi:
> Sorry, dumb questions... what's the advantage of doing this at link time rather than runtime? Is PICO_LOW_POWER_PERSISTENT_PSTATE_STATIC a good name. I guess PICO_LOW_POWER_PERSISTENT_PSTATE_RUN_TIME=1 would be more obvious to me (to get back the old behaviour?)

It saves about 100 bytes of code, and about half that in CPU cycles

Yeah, the variable name probably needs improvement - `PICO_LOW_POWER_PERSISTENT_PSTATE_RUN_TIME` is probably better

src/rp2350/pico_platform/script_include/section_platform_end.incl
```diff
@@ -1,5 +1,47 @@
...
+    /* Convert into bits */
+    POWMAN_POWER_DOMAIN_XIP_CACHE_BITS  = 1 << POWMAN_POWER_DOMAIN_XIP_CACHE;
+    POWMAN_POWER_DOMAIN_SRAM_BANK0_BITS = 1 << POWMAN_POWER_DOMAIN_SRAM_BANK0;
+    POWMAN_POWER_DOMAIN_SRAM_BANK1_BITS = 1 << POWMAN_POWER_DOMAIN_SRAM_BANK1;
+
+    /* Assert the maximum size, to ensure it doesn't cross from xip cache into main sram */
+    ASSERT(__persistent_data_end__ - __persistent_data_start__ <= SRAM_STRIPED_END - SRAM_BASE, "Persistent data is too large")
+
+    /* Assert end is in striped sram */
+    ASSERT(__persistent_data_end__ <= SRAM_STRIPED_END, "Persistent data ends outside striped sram")
```
peterharperuk, line 31: Is <= right?

src/rp2_common/pico_low_power/low_power.c
```diff
@@ -680,30 +680,43 @@ static pstate_bitset_t *low_power_pstate_get(pstate_bitset_t *pstate) {
...
+    /* Assert the maximum size, to ensure it doesn't cross from xip cache into main sram */
+    assert((uint32_t)__persistent_data_end__ - (uint32_t)__persistent_data_start__ <= SRAM_STRIPED_END - SRAM_BASE);
 
-    // Keep __persistent_data_start__ on
-    if ((uint32_t)__persistent_data_start__ < SRAM_BASE) {
-        pstate_bitset_add(pstate, POWMAN_POWER_DOMAIN_XIP_CACHE);
-    } else if ((uint32_t)__persistent_data_start__ < SRAM4_BASE) {
-        pstate_bitset_add(pstate, POWMAN_POWER_DOMAIN_SRAM_BANK0);
+    /* Assert end is in striped sram */
+    assert((uint32_t)__persistent_data_end__ <= SRAM_STRIPED_END);
```
peterharperuk, line 695: ditto

src/rp2350/pico_platform/script_include/section_platform_end.incl
will-v-pi: Yes, for the end checks, as the end address is set to the next address you can place data

PR #3194, opened 2026-09-11
Fixes #3193: Add pio_sm_get_enabled

### LINKED ISSUE #3193: No function for checking if PIO SM is enabled
You can do this by checking `pio->ctrl` obviously, but it would be a bit cleaner if there was a little helper function for this.

badcf00d:
Adds a little helper function to check if a PIO state machine is enabled


src/rp2_common/hardware_pio/include/hardware/pio.h
```diff
@@ -1098,6 +1098,19 @@ static inline void pio_sm_set_enabled(PIO pio, uint sm, bool enabled) {
     pio->ctrl = (pio->ctrl & ~(1u << sm)) | (bool_to_bit(enabled) << sm);
 }
 
+/*! \brief Check if a PIO state machine is enabled or disabled
+ *  \ingroup hardware_pio
+ *
+ * \param pio The PIO instance; e.g. \ref pio0, \ref pio1 etc.
+ * \param sm State machine index (0..3)
+ * \return true if the SM is enabled
```
```suggestion
 * \return true if the state machine is enabled
```

PR #3197, opened 2026-09-16
Set execute bit on scripts with shebangs

### LINKED ISSUE #3196: Some scripts are missing the execute bit
Some scripts have a shebang but not execute permission, so they cannot be run directly. This is flagged in Fedora packaging.

ypsvlq:
Fixes #3196.


lurch:
Nice catch. However it looks like `tools/bazel_common.py` should probably have the shebang line removed, because it only consists of functions (which are presumably called from other scripts) and doesn't have any "executable" section of it's own?

EDIT: Yep :slightly_smiling_face: 
```
$ git grep bazel_common
tools/bazel_build.py:from bazel_common import (
tools/bazel_buildifier_lint_check.py:from bazel_common import (
tools/check_source_files_in_bazel_build.py:from bazel_common import (
tools/compare_build_systems.py:from bazel_common import SDK_ROOT, setup_logging
tools/run_all_bazel_checks.py:from bazel_common import (
```

lurch (approved):

PR #3199, opened 2026-09-17
Add picotool combine commands

will-v-pi:
Adds `pico_combine_uf2_outputs`, `pico_create_partition_table_uf2` and `pico_create_erase_partition_uf2`

These add support for creating a UF2 containing a partition table, and data for all partitions, including overwriting the first sectors of partitions to deal with pre-existing contents

For example, the uart and encrypted bootloader pico-examples could use these to create a single UF2 containing the bootloader and the file to load:
```cmake
# Create combined UF2 with bootloader and binary in a partition
pico_combine_uf2_outputs(uart_boot_combined uart_boot uart_binary PARTITIONS 0)
```

The `pico_create_..._uf2` functions are provided for when there isn't a bootloader


will-v-pi:
Closing as this is now part of #3202

PR #3202, opened 2026-09-19
Rasperry Pi Connect library preview

kilograham:
APIs and integrations are subject to change


src/common/pico_rpi_connect/rpi_connect.c
```diff
@@ -0,0 +1,1603 @@
...
+        free(json_str);
+        rpi_connect_memory_struct_free(&chunk);
+        return NULL;
+    }
+
+    struct curl_slist *additional_headers = NULL;
+    char sig_header[512];
+    snprintf(sig_header, sizeof(sig_header),
+             "X-Connect-Identity-Signature: %s", signature);
+    additional_headers = curl_slist_append(additional_headers, sig_header);
```
will-v-pi, line 987: I think this needs the `Accept: */*` header added, for on-device use?

```suggestion
    additional_headers = curl_slist_append(additional_headers, sig_header);
    additional_headers = curl_slist_append(additional_headers, "Accept: */*");
```

src/common/pico_rpi_connect_ota/rpi_connect_ota.c
```diff
@@ -0,0 +1,890 @@
...
+
+    for (unsigned int i = 0; i < rows && key_offset < 32; i++) {
+        uint32_t row = otp_raw[start_row + i] & 0x00ffffff;
+        for (int b = 0; b < 3 && key_offset < 32; b++) {
+            out_key[key_offset++] = (unsigned char)(row & 0xff);
+            row >>= 8;
+        }
+    }
+    return 0;
+}
```
will-v-pi, line 878: Keys should be ECC, not RAW - I've added `picotool` support in https://github.com/raspberrypi/picotool/pull/362 for `picotool otp load -s 0xc0 device-priv-key.pem` to go with this change

```suggestion
int rpi_connect_ota_read_identity_key_otp(unsigned int start_row, unsigned char out_key[32]) {
    volatile uint16_t *otp_data = (volatile uint16_t *)OTP_DATA_GUARDED_BASE;
    size_t key_offset = 0;
    unsigned int rows = 32 / 2; // 16 rows of 2 bytes

    for (unsigned int i = 0; i < rows && key_offset < 32; i++) {
        uint16_t row = otp_data[start_row + i];
        for (int b = 0; b < 2 && key_offset < 32; b++) {
            out_key[key_offset++] = (unsigned char)(row & 0xff);
            row >>= 8;
        }
    }
    return 0;
}
```

src/common/pico_rpi_connect_ota/rpi_connect_ota.c
```diff
@@ -0,0 +1,890 @@
...
+
+    rc = ffs_write(file_id, data, data_len + 1); // Include the null terminator
+    if (rc != 0) {
+        RPI_CONNECT_OTA_ERROR("FFS: Failed to write file (%d) length %d\n", rc, data_len);
+        return -1;
+    }
+    return 0;
+}
+
+static char *ffs_get_string(uint8_t file_id) {
```
will-v-pi, line 159: This and `ffs_update_string` seem like useful functions (and is duplicated in the pico-example too) - could they be made public in the `pico_ffs` library instead?

Preferably with the change I suggested in pico-examples to use `strndup(data, rc)` instead, so it supports non-NUL-terminated strings, as those will be easier to provision

tools/CMakeLists.txt
```diff
@@ -784,3 +822,257 @@ function(picotool_postprocess_binary TARGET)
...
+
+# pico_create_erase_partition_uf2(TARGET [FAMILY <family>])
+# \brief_nodesc\ Create a UF2 file which can erase the start of a partition
+#
+# Create a UF2 containing 4096 0xff bytes, which will erase the start of a given partition,
+# and therefore mark the partition as invalid
+#
+# \param\ TARGET The name of the target to create
+# \param\ FAMILY Specify the family for the output UF2 (default is absolute)
+function(pico_create_erase_partition_uf2 TARGET)
```
will-v-pi, line 1056: @lurch suggested there might be a better name for this function - given it just creates a UF2 that overwrites the first 4k of wherever it ends up (start of flash, a partition, or an offset) maybe `pico_create_4k_overwrite_uf2`?

```suggestion
function(pico_create_4k_overwrite_uf2 TARGET)
```

src/rp2_common/pico_flash_image/flash_image.c
```diff
@@ -0,0 +1,559 @@
...
+    if (0 > target_part_idx) {
+        FIMG_DEBUG_PRINTF("error - %s() target_part_idx = %i\n", __FUNCTION__, target_part_idx);
+        return target_part_idx;  // -ve means error ...
+    }
+
+    uint32_t first_sector_number = FIRST_SECTOR_NUMBER(uf2_target_partition.permissions_and_location);
+    uint32_t last_sector_number = LAST_SECTOR_NUMBER(uf2_target_partition.permissions_and_location);
+    uint32_t total_sector_count = 1 + last_sector_number - first_sector_number;
+    uint32_t maximum_code_size = FLASH_SECTOR_SIZE * total_sector_count;
+    uint32_t update_size = UF2_BLOCK_SIZE * fimg_state->total_num_blocks;
```
will-v-pi: This should be `UF2_DATA_PAGE_SIZE`, as that's the size of the actual data being written

```suggestion
    uint32_t update_size = UF2_DATA_PAGE_SIZE * fimg_state->total_num_blocks;
```

PR #3203, opened 2026-09-21
Add ffsgen CMake support

will-v-pi:
This adds CMake support for `ffsgen`, similar to how `pioasm` is used

Also includes https://github.com/raspberrypi/pico-sdk/pull/3199, and adds `pico_create_ffs_blob_uf2` function to create FFS blobs

Adds `kitchen_sink_ffs` test for this functionality, using `kitchen_sink_ffs_combined` for the combined partition table, binary, and FFS blob


tools/Findffsgen.cmake
```diff
@@ -0,0 +1,53 @@
+# Finds (or builds) the ffsgen executable
+#
+# This will define the following imported targets
+#
+#     ffsgen
+#
+
+if (NOT TARGET ffsgen)
+    # todo we would like to use pckgconfig to look for it first
```
lurch:
```suggestion
    # todo we would like to use pkg-config to look for it first
```
?

tools/CMakeLists.txt
```diff
@@ -784,3 +822,257 @@ function(picotool_postprocess_binary TARGET)
...
+
+    if (NOT UF2_FAMILY)
+        set(UF2_FAMILY "absolute")
+    endif()
+
+    pico_init_picotool()
+    if (picotool_FOUND)
+        add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.uf2
+            COMMAND picotool uf2 convert --quiet --family ${UF2_FAMILY}
+                ${PICO_SDK_PATH}/tools/ff.bin ${CMAKE_CURRENT_BINARY_DIR}/${TARGET}.uf2
```
lurch, line 1068: Would there be an (easy) way to generate this programmatically, rather than adding a file full of `0xff` to git? :shrug: If not, perhaps it would make sense to just add the `erase_partition.uf2` file to git directly?

will-v-pi: Unfortunately CMake doesn't have a way to generate binary files, hence having to include this file

I've used the `.bin` file, because you might want to use a different UF2 family

timg236 (approved):

timg236 (approved):

